PR #8, opened 2021-01-28
Feature normalize RLP encodings

ferranbt:
This PR normalizes how the RLP marshal/unmarshal methods. Before, the same function signature would have different inputs for different objects. Now, all the structs have the same function signatures with the same behaviour.

For marshaling, each object has three functions
```
// Encode an object in RLP format
MarshalRLP() []byte
// Encode an object in RLP format and put the result in dst
MarshalRLPTo(dst []byte) []byte
// Encode an object in RLP format and return a fastrlp.Value object
MarshalRLPWith(ar *fastrlp.Arena) *fastrlp.Value
```

For unmarshaling, each object has two functions:
```
// Unmarshal an object in RLP from a fastrlp.Parser and fastrlp.Value objects
UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error
// Unmarshal an object in RLP from a bytes representation
UnmarshalRLP(input []byte) error
```


zivkovicmilos (approved):
Looks good 👍

PR #9, opened 2021-02-05
feature/rpc-calls

zivkovicmilos:
# Description

This PR brings the implementation for the following JSON RPC endpoints:
- eth_getStorageAt
- eth_gasPrice
- eth_call
- eth_estimateGas
- eth_getLogs

# Changes include

# Checklist

- [ ] I have tested this code -> Waiting on the test framework overhaul to write concrete tests

# Additional comments

There was some generic code cleanup, as well as addition of several utility functions (mostly concerning hex manipulation).


api/jsonrpc/eth_endpoint.go
```diff
@@ -117,12 +119,13 @@ func (e *Eth) GetTransactionReceipt(hash string) (interface{}, error) {
 	return nil, fmt.Errorf("transaction not found")
 }
 
-// CurrentBlock returns current block number
+// GetBalance returns the account's balance
 func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
```
ferranbt: GetBalance takes a BlockNumber reference (number or "latest").

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
 	return new(types.Big), nil
 }
 
+// GetStorageAt returns the contract storage at the index position
+func (e *Eth) GetStorageAt(address string, index []byte, number string) (interface{}, error) {
```
ferranbt: - index []byte should be index types.Hash because the types.Hash object already has functions to UnmarshalJSON, otherwise, this function would fail.
- Number is a BlockNumber object

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
...
+func (e *Eth) GasPrice() (interface{}, error) {
+
+	// Grab the average gas price and convert it to a hex value
+	avgGasPrice := hex.EncodeBig(e.d.minimal.Blockchain.GetAvgGasPrice())
+
+	return avgGasPrice, nil
+}
+
+// Call executes a smart contract call using the transaction object data
+func (e *Eth) Call(transaction *types.Transaction, number string) (interface{}, error) {
```
ferranbt: number is a BlockNumber object

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return transition.ReturnValue(), nil
+}
+
+// GetBlockHeader returns the specific header of the requested block
+// latest, earliest, pending, or a specific block number
+func (e *Eth) GetBlockHeader(number string) (*types.Header, error) {
```
ferranbt: Use the BlockNumber object

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
...
+}
+
+// EstimateGasParams - Optional params used for the EstimateGas call
+type EstimateGasParams struct {
+	transaction *types.Transaction
+	number      string
+}
+
+// EstimateGas estimates the gas needed to execute a transaction
+func (e *Eth) EstimateGas(params EstimateGasParams) (interface{}, error) {
```
ferranbt: EstimateGas does not take a block number. Only the transaction object.

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
...
+		if failed {
+			return 0, fmt.Errorf("gas required exceeds allowance (%d)", gasCap)
+		}
+	}
+
+	return hex.EncodeUint64(highEnd), nil
+}
+
+// GetLogs returns an array of logs matching the filter options
+func (e *Eth) GetLogs(filterOptions LogFilter) ([]*types.Log, error) {
```
ferranbt: Use a pointer like (filterOptions *LogFilter).

api/jsonrpc/eth_endpoint.go
```diff
@@ -138,12 +141,319 @@ func (e *Eth) GetBalance(address string, number string) (interface{}, error) {
...
+				// }
+			}
+		}
+	}
+
+	return result, nil
+}
+
 // GetTransactionCount returns account nonce
 func (e *Eth) GetTransactionCount(address string, number string) (interface{}, error) {
```
ferranbt: Use BlockNumber

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Clarified it in the comment 👍

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍

api/jsonrpc/eth_endpoint.go
zivkovicmilos: According to the Ethereum wiki, estimateGas should take in the same params as eth_call, except that they're all optional:
[eth_call](https://eth.wiki/json-rpc/API#eth_calll)

> See eth_call parameters, expect that all properties are optional. If no gas limit is specified geth uses the block gas limit from the pending block as an upper bound. As a result the returned estimate might not be enough to executed the call/transaction when the amount of gas is higher than the pending block gas limit.

> eth_call

Executes a new message call immediately without creating a transaction on the block chain.
Parameters

    Object - The transaction call object

    from: DATA, 20 Bytes - (optional) The address the transaction is sent from.
    to: DATA, 20 Bytes - The address the transaction is directed to.
    gas: QUANTITY - (optional) Integer of the gas provided for the transaction execution. eth_call consumes zero gas, but this parameter may be needed by some executions.
    gasPrice: QUANTITY - (optional) Integer of the gasPrice used for each paid gas
    value: QUANTITY - (optional) Integer of the value sent with this transaction
    data: DATA - (optional) Hash of the method signature and encoded parameters. For details see Ethereum Contract ABI in the Solidity documentation

    QUANTITY|TAG - integer block number, or the string "latest", "earliest" or "pending", see the default block parameter

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍

api/jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍

api/jsonrpc/eth_endpoint.go
ferranbt: That refers to the transaction object not to the whole set of function params.

ferranbt: You are right, estimateGas takes two function params https://github.com/ethereum/go-ethereum/blob/fba5a63afe0993da70896c763b4fdfa953e066ff/internal/ethapi/api.go#L941

PR #14, opened 2021-03-18
Feature/dht

Kourin1996:
# Description

This PR integrates `libp2p/go-libp2p-kad-dht` into libp2p function in order to connect new peer automatically. 

# Changes include

# Checklist

- [x] I have tested this code


minimal/libp2p.go
```diff
@@ -30,19 +31,26 @@ func (s *Server) setupLibP2P() error {
 	}
 	s.addrs = []ma.Multiaddr{addr}
 
+	bwc := metrics.NewBandwidthCounter()
```
ferranbt: What does this do?

Kourin1996: Sorry, it's caused by my misunderstanding. Used this as an indicator to get good peer from closest peers

PR #33, opened 2021-04-14
Add limit of inbound connection request

Kourin1996:
# Description

Relevant: https://github.com/0xPolygon/polygon-sdk/issues/28
This PR adds limitations for the inbound connection requests. Node rejects connection request if it doesn't have available slots. (MaxPeers - NumOfConnectingPeers - NumOfDialingPeers)

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


network/server_test.go
```diff
@@ -14,26 +15,174 @@ func TestDialLifecycle(t *testing.T) {
...
+	tests := []struct {
+		name     string
+		cfgs     []func(*Config)
+		requests []Request
+		expected []Expected
+	}{
+		{
+			name: "server 1 should reject incoming connection request from server 3 due to limit",
+			cfgs: []func(*Config){
+				func(c *Config) {
```
ferranbt: Since this config is repeated a lot, could you create something like
```
defConfig := func(c *Config) {
   c.MaxPeers = 1
   c.NoDiscover = true
}
```
and then use it instead

ferranbt (changes requested):
Just some minor stuff, otherwise is good to go

PR #36, opened 2021-04-21
Add Broadcast e2e test

Kourin1996:
# Description

This PR adds e2e test for transaction broadcast.
And it adds `dummy` consensus to do e2e test for discovery and broadcast.
(Dev Mode always forece to enable sealing, disable discovery and gossip)

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


consensus/dummy/dummy.go
```diff
@@ -0,0 +1,126 @@
...
+	// All blocks are valid
+	return nil
+}
+
+func (d *Dummy) Close() error {
+	close(d.closeCh)
+	return nil
+}
+
+func (d *Dummy) run() {
```
ferranbt, line 59: Is there any point in sealing blocks, the goal of the dummy consensus is to enable everything on the network side of things (txpool and discovery) but without the hassle of syncing logic. I think the run function can be removed.

command/genesis.go
```diff
@@ -54,13 +54,17 @@ func (c *GenesisCommand) Run(args []string) int {
...
+	var dummyConsensus bool
+
 	flags.StringVar(&dataDir, "data-dir", "", "")
 	flags.StringVar(&name, "name", "example", "")
 	flags.Var(&premine, "premine", "")
 	flags.Uint64Var(&chainID, "chainid", 100, "")
 	flags.BoolVar(&ibftConsensus, "ibft", false, "")
 	flags.Var(&ibftValidators, "ibft-validator", "list of ibft validators")
 	flags.StringVar(&ibftValidatorsPrefixPath, "ibft-validators-prefix-path", "", "")
+	flags.BoolVar(&dummyConsensus, "dummy", false, "")
```
ferranbt: Could you do instead a flag called "consensus" which you use to select either ibft or dummy? and then
```
ibftConsensus := consensus == "ibft"
```

PR #43, opened 2021-05-27
Minor improvements and fix a goroutine leak

nicovogelaar:
# Description

# Changes include

# Breaking changes

# Checklist

- [ ] I have tested this code

# Additional comments

### DIFF
--- a/network/server.go
+++ b/network/server.go
@@ -520,12 +520,37 @@ func (s *Server) SubscribeFn(handler func(evnt *PeerEvent)) error {
 }
 
 // SubscribeCh returns an event of of subscription events
-func (s *Server) SubscribeCh() (chan *PeerEvent, error) {
+func (s *Server) SubscribeCh() (<-chan *PeerEvent, error) {
 	ch := make(chan *PeerEvent)
+
+	var closed bool
+	var mutex sync.Mutex
+
+	isClosed := func() bool {
+		mutex.Lock()
+		defer mutex.Unlock()
+		return closed
+	}
+
 	err := s.SubscribeFn(func(evnt *PeerEvent) {
-		ch <- evnt
+		if !isClosed() {
+			ch <- evnt
+		}
 	})
-	return ch, err
+	if err != nil {
+		close(ch)
+		return nil, err
+	}
+
+	go func() {
+		<-s.closeCh
+		mutex.Lock()
+		closed = true
+		mutex.Unlock()
+		close(ch)
+	}()
+
+	return ch, nil
 }
 
 func StringToAddrInfo(addr string) (*peer.AddrInfo, error) {
--- a/protocol/service_v1.go
+++ b/protocol/service_v1.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/hashicorp/go-hclog"
+	"github.com/libp2p/go-libp2p-core/peer"
 )
 
 // serviceV1 is the GRPC server implementation for the v1 protocol
@@ -28,7 +29,13 @@ type rlpObject interface {
 }
 
 func (s *serviceV1) Notify(ctx context.Context, req *proto.NotifyReq) (*empty.Empty, error) {
-	id := ctx.(*grpc.Context).PeerID
+	var id peer.ID
+
+	if ctx, ok := ctx.(*grpc.Context); ok {
+		id = ctx.PeerID
+	} else {
+		return &empty.Empty{}, nil
+	}
 
 	b := new(types.Block)
 	if err := b.UnmarshalRLP(req.Raw.Value); err != nil {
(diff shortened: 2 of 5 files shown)


protocol/syncer.go
```diff
@@ -265,16 +266,24 @@ func (s *Syncer) Start() {
...
+		s.logger.Error("failed to subscribe", "err", err)
+		return
+	}
 
 	go func() {
 		for {
-			evnt := <-updateCh
+			evnt, ok := <-updateCh
+			if !ok {
+				return
```
nicovogelaar, line 284: This stops the go-routine once the channel is closed.

protocol/service_v1.go
```diff
@@ -28,7 +29,13 @@ type rlpObject interface {
 }
 
 func (s *serviceV1) Notify(ctx context.Context, req *proto.NotifyReq) (*empty.Empty, error) {
-	id := ctx.(*grpc.Context).PeerID
```
nicovogelaar, line 31: This can cause a panic

lazartravica (approved):
Good catch and thank you for your quick contribution!

👍

nicovogelaar:
Note: I didn't manage to update the generated proto files: *.pb.go, due to having a different version of protoc-gen-go.

protocol/syncer.go
```diff
@@ -379,21 +387,26 @@ func (s *Syncer) findCommonAncestor(clt proto.V1Client, status *Status) (*types.
...
-					return nil, nil, fmt.Errorf("genesis does not match?")
+					return nil, nil, errors.New("genesis does not match?")
 				}
 				max = m - 1
 			}
 		}
 	}
 
+	var num uint64
+	if header != nil {
```
nicovogelaar: `header` can be nil. How to handle this?

network/server.go
```diff
@@ -520,12 +520,37 @@ func (s *Server) SubscribeFn(handler func(evnt *PeerEvent)) error {
 }
 
 // SubscribeCh returns an event of of subscription events
-func (s *Server) SubscribeCh() (chan *PeerEvent, error) {
+func (s *Server) SubscribeCh() (<-chan *PeerEvent, error) {
```
nicovogelaar, line 523: `<-chan *PeerEvent` only allow reading from the channel

nicovogelaar:
I managed to update the generated .pb.go file and added a shell script to install the same version of protoc to have the same output.

zivkovicmilos (changes requested):
Everything looks good 👍
Thank you for the fast contribution!
 
I'm not sure whether we should be including the .sh script in the root directory for the repo.

@lazartravica @Vuksan @Kourin1996 what do you guys think?

Vuksan (approved):
Thank you very much for your contribution 🙏

Vuksan:
> Everything looks good 👍
> Thank you for the fast contribution!
> 
> I'm not sure whether we should be including the .sh script in the root directory for the repo.
> 
> @lazartravica @Vuksan @Kourin1996 what do you guys think?

I agree, but want to hear more from @lazartravica

install_protoc_osx_x86_64.sh
```diff
@@ -0,0 +1,10 @@
+# http://google.github.io/proto-lens/installing-protoc.html
+PROTOC_ZIP=protoc-3.12.0-osx-x86_64.zip
+curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v3.12.0/$PROTOC_ZIP
+sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc
+sudo unzip -o $PROTOC_ZIP -d /usr/local 'include/*'
+rm -f $PROTOC_ZIP
+
+# https://grpc.io/docs/languages/go/quickstart/
+GO111MODULE=on go get -v -u google.golang.org/protobuf/cmd/protoc-gen-go@v1.25
+GO111MODULE=on go get -v -u google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1
```
lazartravica: @nicovogelaar Thank you for this effort, but we would rather not  include shell scripts for installing protoc like this on OSX.

We should have a platform agnostic way of installing all the dependencies for local builds (and for CI builds).

E.g. a Docker image for local, and possibly to reuse it for the CI.

protocol/syncer.go
lazartravica: As it seems like the same type of problem as with being unable to find a certain header by number or with being unable to read the local genesis, I'd advise: 

```
		return nil, nil, errors.New("header is nil")
```

Also, can you please clarify why did you keep `errors.New()` in this case, while keeping the`fmt.Errorf()` above?  Thanks
@nicovogelaar

lazartravica (changes requested):
Added a couple of points to the discussion. Great stuff 👌

install_protoc_osx_x86_64.sh
nicovogelaar: Sure, I will remove it. It may be good to document it somewhere to make it easier for other developers to contribute. Keep in mind that different versions of protoc will have different outputs of generated code.

nicovogelaar:
> Also, can you please clarify why did you keep errors.New() in this case, while keeping the fmt.Errorf() above? Thanks

`fmt.Errorf` is to format errors, for example: `fmt.Errorf("something went wrong %d: %w", number, err)`. Providing parameters to format is optional. However using `fmt.Errorf` without any formatting, can be confusing and is unnecessary.

I suggest to use a simple `errors.New("something went wrong")`, when there is nothing to format.

@lazartravica

nicovogelaar:
The shell script is removed.

zivkovicmilos (approved):
Looks good 💯

PR #45, opened 2021-05-28
Fix failed tests

Kourin1996:
# Description

This PR fixes current failed tests without change of implementation.

- [x] TestEth_State_GetBalance in jsonrpc/eth_endpoint_test.go
- [x] TestConnLimit_Inbound in network/server_test.go
- [x] TestTxnQueue_Promotion in txpool/txpool_test.go
- [x] TestEVM in tests/evm_test.go

Skipped TestEncoding in e2e/encoding_test.go because we need to add some codes into implementation.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

I summarized the reasons why tests failed and directions how to fix tests for now.
https://drive.google.com/file/d/13leoGsTlAVHTdUVsV28g9M-ksqg9SqFj/view?usp=sharing

### DIFF
--- a/tests/evm_test.go
+++ b/tests/evm_test.go
@@ -2,11 +2,14 @@ package tests
 
 import (
 	"encoding/json"
-	"github.com/umbracle/fastrlp"
 	"io/ioutil"
 	"math/big"
+	"reflect"
 	"strings"
 	"testing"
+	"unsafe"
+
+	"github.com/umbracle/fastrlp"
 
 	"github.com/0xPolygon/minimal/chain"
 	"github.com/0xPolygon/minimal/helper/hex"
@@ -58,6 +61,16 @@ func testVMCase(t *testing.T, name string, c *VMCase) {
 	}
 
 	e, _ := executor.BeginTxn(root, c.Env.ToHeader(t))
+	{
+		// todo: workaround fix to pass the test
+		// we need to consider best way to give GasPrice and Origin as context value into evm in test
+		t := reflect.ValueOf(e).Elem()
+		ctx := t.FieldByName("ctx")
+		gasPrice := (*types.Hash)(unsafe.Pointer(ctx.FieldByName("GasPrice").UnsafeAddr()))
+		origin := (*types.Address)(unsafe.Pointer(ctx.FieldByName("Origin").UnsafeAddr()))
+		*gasPrice = types.BytesToHash(env.GasPrice.Bytes())
+		*origin = env.Origin
+	}
 
 	evmR := evm.NewEVM()
 
--- a/tests/evm_test.go
+++ b/tests/evm_test.go
@@ -4,10 +4,8 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"math/big"
-	"reflect"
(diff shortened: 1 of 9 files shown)


tests/evm_test.go
```diff
@@ -58,6 +61,16 @@ func testVMCase(t *testing.T, name string, c *VMCase) {
...
+	{
+		// todo: workaround fix to pass the test
+		// we need to consider best way to give GasPrice and Origin as context value into evm in test
+		t := reflect.ValueOf(e).Elem()
+		ctx := t.FieldByName("ctx")
+		gasPrice := (*types.Hash)(unsafe.Pointer(ctx.FieldByName("GasPrice").UnsafeAddr()))
+		origin := (*types.Address)(unsafe.Pointer(ctx.FieldByName("Origin").UnsafeAddr()))
+		*gasPrice = types.BytesToHash(env.GasPrice.Bytes())
+		*origin = env.Origin
+	}
```
Vuksan: I just noticed that the `Context()` method of the `Transition` struct isn't used anywhere, so this being the only place using it, you could just change the signature of it to return the pointer instead of the value and then do:
```
e.Context().GasPrice = types.BytesToHash(env.GasPrice.Bytes())
e.Context().Origin = env.Origin
```

Again, this might not be the proper solution either - I would love to hear an opinion from @lazartravica @zivkovicmilos

Vuksan (approved):
@Kourin1996 If you can please attach the doc that you assembled for this task to the PR 🙏

tests/evm_test.go
Kourin1996: @Vuksan Yes, you're correct. Thank you for your discovery and suggestion. I checked it on my side. Transition.Context() might not be called from anywhere. And if I change this function to return pointer instead of value, we can pass the test without using reflect. But strange thing is that Transition has Context() and GetTxContext(). Both method return same value

Kourin1996:
I added the link of additional documents in the description

tests/evm_test.go
zivkovicmilos: Yep, this is probably just an oversight. We can convert the Transition.Context() method to return a pointer, probably change the name as well, to something like Transition.GetContextPtr() - as @Vuksan suggested.

zivkovicmilos (approved):
Great catches, and fixed quickly, as always 👍  💯

tests/evm_test.go
Kourin1996: I'd like to get feedback from @lazartravica, if it's okay.

lazartravica: > Yep, this is probably just an oversight. We can convert the Transition.Context() method to return a pointer, probably change the name as well, to something like Transition.GetContextPtr() 

I agree with this 👍

lazartravica (approved):

Kourin1996:
@Vuksan @zivkovicmilos I pushed new change. Please let me know if there is any issue. Thanks

Vuksan (approved):
LGTM

zivkovicmilos (approved):
💯

PR #47, opened 2021-05-31
Pass bootnodes through CLI when generating genesis.json

Vuksan:
# Description

- Enabled passing of bootnodes through the CLI.
- Added default values to some flags for the `genesis` command.
- Made `ibft init` command print the `node ID` for libp2p multiaddr format.

### How to test:

Follow [this how-to](https://github.com/0xPolygon/polygon-sdk-docs/blob/b62d0ba46c15cfcd69bbce8d901c57c9915eac27/docs/how-tos/howto-set-ibft.md).

# Changes include

# Checklist

- [x] I have tested this code


command/genesis.go
```diff
@@ -13,15 +13,16 @@ import (
...
 	"github.com/0xPolygon/minimal/types"
 	"github.com/mitchellh/cli"
 )
 
 const (
 	genesisFileName       = "./genesis.json"
+	defaultChainName	  = "example"
 	defaultChainID        = 100
 	defaultPremineBalance = "0x3635C9ADC5DEA00000" // 1000 ETH
+	defaultConsensus	  = "pow"
```
zivkovicmilos, line 25: Are these default values leftover?
Love the standardized naming for flag descriptions now, it's more concise 👍

zivkovicmilos (approved):
It's a subtle change, but I think it will definitely improve the UX

Kourin1996:
Just informing. If I don't give bootnodes on genesis command, `bootnodes` will be `null` in genesis.json. 

```
{
   ...
    "bootnodes": null
}
```

Please set empty slice to `bootnodes` as initial value to avoid this.

Kourin1996 (approved):
LGTM!

command/genesis.go
Vuksan: The default consensus definitely is, but I left it since I wasn't sure what to do in case it isn't provided through the CLI.

lazartravica (approved):
LGTM

PR #50, opened 2021-06-03
Add help output for grpc-address flag when using operator commands.

Vuksan:
# Description

Multiple operator commands had the flag for gRPC address, it just wasn't added to the `--help` output.
It had the default value of `127.0.0.1:8545` if not set. Since users didn't know they can set it, they would call the commands without it, resulting in the commands failing, since their gRPC was running on a different port if they followed ibft guide.

This PR enables the use of "global flags" - for now only `--grpc-address`.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

PR to modify SDK docs: [link](https://github.com/0xPolygon/polygon-sdk-docs/pull/5)

### DIFF
--- a/command/commands.go
+++ b/command/commands.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/0xPolygon/minimal/minimal"
 	"os"
+	"strings"
 
 	"github.com/0xPolygon/minimal/command/server"
 	"github.com/mitchellh/cli"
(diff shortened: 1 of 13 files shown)


zivkovicmilos (approved):
Everything seems good to me 👍

Kourin1996 (approved):
LGTM. This additional help message is very kind.

lazartravica (approved):
LGTM!

PR #52, opened 2021-06-04
Refactor e2e test

Kourin1996:
# Description

This PR aims to remove some dependencies from e2e test. It eliminates pre-requirements for e2e and enables to run them with a clean environment and same conditions.
- Stop using solc (solidity compiler) in e2e test
- Run new clean IBFT nodes internally in every test instead of using IBFT nodes running locally
  + Fix e2e/framework package and separate processes to generate genesis.json and start node into two methods.
  + Find available ports automatically
  + Add Manager for IBFT nodes

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


e2e/README.md
```diff
@@ -30,4 +30,16 @@ To clean up these folders, simply run:
...
+
+### constants
+
+Constant values that used in some e2e tests are defined in `e2e/const.go`.
+Mock contract are pre-compiled and the result is stored in `const.go` in order to avoid dependencies of solc command (solidity compiler).
+`e2e/sample.go` is original source code, and you can get byte code by following command.
+
+```shell
+$ solc --bin e2e/sample.sol
+```
```
lazartravica: Thank you for the docs! @Kourin1996  🙏

lazartravica (changes requested):
Excellent job!

All tests are passing except `TestBroadcast:tx_should_not_reach_to_last_node`.

@Kourin1996 is it the same on your end?

e2e/README.md
```diff
@@ -30,4 +30,16 @@ To clean up these folders, simply run:
...
+## Note
+
+### constants
+
+Constant values that used in some e2e tests are defined in `e2e/const.go`.
+Mock contract are pre-compiled and the result is stored in `const.go` in order to avoid dependencies of solc command (solidity compiler).
+`e2e/sample.go` is original source code, and you can get byte code by following command.
+
+```shell
+$ solc --bin e2e/sample.sol
```
Vuksan, line 45: Can you please add the version of the compiler you used as well.
I have tried the command with this compiler and it fails:
```
0.7.1+commit.f4a555be.Darwin.appleclang
```

e2e/framework/config.go
```diff
@@ -0,0 +1,78 @@
...
+func (t *TestServerConfig) SetConsensus(c ConsensusType) {
+	t.Consensus = c
+}
+
+// SetIBFTDirPrefix callback sets prefix of IBFT directories
+func (t *TestServerConfig) SetIBFTDirPrefix(ibftDirPrefix string) {
+	t.IBFTDirPrefix = ibftDirPrefix
+}
+
+// SetIBFTDirPrefix callback sets prefix of IBFT directories
```
Vuksan: Wrong comment, probably c/p from previous method

e2e/framework/config.go
```diff
@@ -0,0 +1,78 @@
...
+func (t *TestServerConfig) SetSeal(state bool) {
+	t.Seal = state
+}
+
+// SetBootnodes sets bootnodes
+func (t *TestServerConfig) SetBootnodes(bootnodes []string) {
+	t.Bootnodes = bootnodes
+}
+
+// SetBootnodes sets bootnodes
```
Vuksan: Wrong comment, probably c/p from previous method

e2e/framework/helper.go
```diff
@@ -36,36 +52,168 @@ func MultiJoin(t *testing.T, srvs ...*TestServer) {
...
+				resCh <- result{nil, ErrTimeout}
+				return
+			default:
+				res, retry := f()
+				if !retry {
+					resCh <- result{res, nil}
+					return
+				}
+			}
+			time.Sleep(time.Second)
```
Vuksan, line 146: Why the Sleep()?

e2e/discovery_test.go
```diff
@@ -30,67 +32,103 @@ func TestDiscovery(t *testing.T) {
...
 	}
 
-	conf := func(config *framework.TestServerConfig) {
-		config.SetConsensus(framework.ConsensusDummy)
-	}
-
 	for _, tt := range tests {
+		conf := func(config *framework.TestServerConfig) {
+			config.SetConsensus(framework.ConsensusDummy)
+		}
```
Vuksan: Shouldn't this `conf` initialization be outside the for loop?

e2e/discovery_test.go
```diff
@@ -30,67 +32,103 @@ func TestDiscovery(t *testing.T) {
...
-				res, err := s.Operator().PeersList(context.Background(), &empty.Empty{})
-				if err != nil {
-					t.Fatal(err)
+			for i, srv := range srvs {
+				shouldKnowPeers := false
+				subTestName := fmt.Sprintf("node %d should know other peers", i)
+				if i < tt.numInitConnectNodes {
+					shouldKnowPeers = true
+					subTestName = fmt.Sprintf("node %d shouldn't know other peers", i)
 				}
```
Vuksan: Shouldn't the message in the if block be "should know", and before the if block "shouldn't know"?
If I understand the tests, first 4 nodes should know each other, and not the other way around.

e2e/broadcast_test.go
```diff
@@ -37,30 +39,67 @@ func TestBroadcast(t *testing.T) {
...
 				}
-			}()
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
+				if err := srv.Start(ctx); err != nil {
+					t.Fatal(err)
+				}
+
+				srvs = append(srvs, srv)
+			}
```
Vuksan: I see that this is repeated through some tests... what do you think about extracting it to a helper function for tests?

[from line 40 to 75, in case this github selector bugs out]

e2e/logs_test.go
```diff
@@ -1,73 +1,123 @@
...
+		defer cancel()
+		srv.TxnTo(ctx, contractAddr, "setA1")
 	}
-	time.Sleep(10 * time.Second)
 
 	res, err := client.Eth().GetFilterChanges(id)
 	assert.NoError(t, err)
-	assert.Len(t, res, 10)
+	// todo: need to check implementation because there is a possibility of losing some logs
+	assert.GreaterOrEqual(t, len(res), numCalls/2)
```
Vuksan, line 44: Just to check, you didn't forget about this todo?
I noticed that test doesn't receive all filter changes, but most of them.

e2e/logs_test.go
```diff
@@ -1,73 +1,123 @@
...
+	})
+
+	if err := srv.GenerateGenesis(); err != nil {
+		t.Fatal(err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := srv.Start(ctx); err != nil {
+		t.Fatal(err)
+	}
```
Vuksan: The code (ln74-98) seems to be the same as in the previous test. Is it possible to extract it in a helper function?

e2e/logs_test.go
```diff
@@ -1,73 +1,123 @@
...
-		time.Sleep(1 * time.Second)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		_, err = srv.WaitForReceipt(ctx, hash)
+		assert.NoError(t, err)
 	}
 
-	// wait for the last txn to be sealed
-	time.Sleep(2 * time.Second)
-
```
Vuksan, line 70: I just noticed that at the end of this test we have `fmt.Println()` instead of `assert`.
Maybe just assert if 3 changes occurred?

e2e/ibft_test.go
```diff
@@ -1,52 +1,68 @@
...
+	})
 
-	signer := crypto.NewEIP155Signer(100)
-	target := types.StringToAddress("0x1010101010101010101010101010101010101010")
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	ibftManager.StartServers(ctx)
 
+	srv := ibftManager.GetServer(0)
 	for i := 0; i < 3; i++ {
```
Vuksan:
```suggestion
	for i := 0; i < IBFTMinNodes - 1; i++ {
```

e2e/ibft_test.go
```diff
@@ -1,52 +1,68 @@
...
+		assert.NoError(t, err)
+		assert.NotNil(t, hash)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		receipt, err := srv.WaitForReceipt(ctx, hash)
+
 		assert.NoError(t, err)
-		fmt.Println(hash)
+		assert.NotNil(t, receipt)
```
Vuksan, line 66: Maybe add additional check, to make sure it is the same transaction:
```suggestion
		assert.NotNil(t, receipt)
		assert.Equal(t, receipt.TransactionHash, hash)
```

e2e/transaction_test.go
```diff
@@ -1,34 +1,48 @@
...
-)
-
-var (
-	addr0 = types.Address{}
-	addr1 = types.Address{0x1}
 )
 
 func TestSignedTransaction(t *testing.T) {
-	fr := framework.NewTestServerFromGenesis(t)
+	key, err := wallet.GenerateKey()
```
Vuksan: Just curious - why not `framework.GenerateKeyAndAddr(t)` as in other tests?

e2e/transaction_test.go
```diff
@@ -65,7 +79,9 @@ func TestSignedTransaction(t *testing.T) {
 		hash, err := clt.Eth().SendRawTransaction(data)
 		assert.NoError(t, err)
 
-		fr.WaitForReceipt(hash)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		srv.WaitForReceipt(ctx, hash)
```
Vuksan: To add some checks

```suggestion
		srv.WaitForReceipt(ctx, hash)
		
		assert.NoError(t, err)
		assert.NotNil(t, receipt)
		assert.Equal(t, receipt.TransactionHash, hash)
```

e2e/transaction_test.go
```diff
@@ -104,38 +120,49 @@ func TestPreminedBalance(t *testing.T) {
...
 
 			if err != nil && !testCase.shouldFail {
 				assert.Failf(t, "Uncaught error", err.Error())
 			}
-
 			if !testCase.shouldFail {
 				assert.Equalf(t, testCase.balance, balance, "Balances don't match")
 			}
 		})
 	}
```
Vuksan: I have an issue with this test, since we aren't actually asserting the third case, since there `testCase.shouldFail` is `true`.

- If the expected behavior should be that for the nonexisting account we should get an error, we should append the latest if with the following code and then the test will fail, since it doesn't return the error, but we expect it to (or at least that seems to me).
```
 else {
     assert.NotNil(t, err)
}
```

- However, if the expected behaviour should be that for "nonexistent" account we should get a 0 balance, then we should change the test to reflect  that.

I am in favour of the second approach, since there are no "nonexistent" accounts if you ask me - any valid address is an existing account.

Vuksan: We can open another task for this, it doesn't have to be solved in this PR, just wanted to discuss.

e2e/transaction_test.go
```diff
@@ -169,36 +196,54 @@ func TestEthTransfer(t *testing.T) {
...
 			false,
 		},
 		{
 			// ACC #2 -> ACC #3
 			"Invalid ETH transfer",
 			validAccounts[1].address,
 			validAccounts[2].address,
-			ethToWei(100),
+			framework.EthToWei(100),
 			true,
```
Vuksan: This case is never asserted in tests. We can add a task to cover that.

e2e/transaction_test.go
```diff
@@ -169,36 +196,54 @@ func TestEthTransfer(t *testing.T) {
...
 		},
 		{
 			// ACC #2 -> ACC #1
```
Vuksan:
```suggestion
			// ACC #3 -> ACC #2
```

Vuksan (changes requested):
Great work @Kourin1996 ! 🚀 
I have several observations - nothing major, mostly small things.

_EDIT:_ I see now that one e2e test is failing:
```
--- FAIL: TestEncoding (3.09s)
    encoding_test.go:57:
        	Error Trace:	encoding_test.go:57
        	Error:      	Received unexpected error:
        	            	field 'blockHash' not found
        	Test:       	TestEncoding
panic: field 'blockHash' not found [recovered]
	panic: field 'blockHash' not found
```

Kourin1996:
@Vuksan Thank you for your feedback! About `field 'blockHash' not found` issue, we need to merge https://github.com/0xPolygon/polygon-sdk/pull/59 first. If you don't have any issues except for this, should be ok.
I will fix codes based on your feedback. Thank you :)

e2e/framework/helper.go
Kourin1996: It's for suppressing many calles in a moment.

e2e/README.md
Kourin1996: I modified document

e2e/framework/config.go
Kourin1996: I updated comment

e2e/discovery_test.go
Kourin1996: I fixed wrong sub test name

e2e/broadcast_test.go
Kourin1996: I created `NewTestServers` helper function in e2e/framework/helper.go

e2e/logs_test.go
Kourin1996: I need to check implementation for log and fix if need. I will create another PR for this.

e2e/logs_test.go
Kourin1996: I created NewTestServers helper function in e2e/framework/helper.go

e2e/logs_test.go
Kourin1996: I added new assert

e2e/ibft_test.go
Kourin1996: I added this assertion

e2e/transaction_test.go
Kourin1996: I fixed this and could test without `wallet.GenerateKey()`

e2e/transaction_test.go
Kourin1996: I fixed the test and now it checks balance change for every case.

Kourin1996:
@Vuksan I fixed codes based on your feedback. If you noticed something, please let me know. Thank you.

e2e/framework/helper.go
```diff
@@ -36,36 +53,202 @@ func MultiJoin(t *testing.T, srvs ...*TestServer) {
...
+			}
+			return nil, errors.New("couldn't reserve required number of ports")
+		}
+		ports = append(ports, *newPort)
+		nextFrom = newPort.Port() + 1
+	}
+	return ports, nil
+}
+
+func NewTestServers(t *testing.T, num int, setConfig func(int, *TestServerConfig)) []*TestServer {
```
Vuksan: Why does `setConfig` function need to have the index parameter?
I don't see it used anywhere...

e2e/logs_test.go
```diff
@@ -1,75 +1,107 @@
...
+	})
+
+	if err := srv.GenerateGenesis(); err != nil {
+		t.Fatal(err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := srv.Start(ctx); err != nil {
+		t.Fatal(err)
+	}
```
Vuksan: I think there is no need for this `newLogsTestServer ` function. You should be able to call helper's `NewTestServers()` instead of this if I'm not mistaken...

e2e/encoding_test.go
```diff
@@ -1,34 +1,59 @@
...
+	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	contractAddr, err := srv.DeployContract(ctx, sampleByteCode)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	receipt := srv.TxnTo(ctx, contractAddr, "setA1")
```
Vuksan: You can use helper's `NewTestServers()` here.

e2e/transaction_test.go
```diff
@@ -144,167 +154,168 @@ func TestEthTransfer(t *testing.T) {
...
 		}
+	})
+
+	if err := srv.GenerateGenesis(); err != nil {
+		t.Fatal(err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := srv.Start(ctx); err != nil {
+		t.Fatal(err)
```
Vuksan: You can use helper's `NewTestServers()` here.

e2e/transaction_test.go
```diff
@@ -79,63 +88,64 @@ func TestPreminedBalance(t *testing.T) {
...
 
-	rpcClient := srv.JSONRPC()
+	if err := srv.GenerateGenesis(); err != nil {
+		t.Fatal(err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := srv.Start(ctx); err != nil {
+		t.Fatal(err)
+	}
```
Vuksan: You can use helper's `NewTestServers()` here.

e2e/transaction_test.go
```diff
@@ -1,76 +1,85 @@
...
-	key, err := wallet.NewWalletFromPrivKey(hex.MustDecodeHex(privKeyRaw))
-	assert.NoError(t, err)
+	dataDir, err := framework.TempDir()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	preminedAmount := framework.EthToWei(10)
+	ibftManager := framework.NewIBFTServersManager(t, IBFTMinNodes, dataDir, IBFTDirPrefix, func(i int, config *framework.TestServerConfig) {
+		config.Premine(types.Address(senderAddr), preminedAmount)
```
Vuksan: Redundant type conversion
```suggestion
		config.Premine(senderAddr, preminedAmount)
```

Vuksan (changes requested):
👏  for applying suggestions. 
Let's just call `NewTestServers()` in several more places

Vuksan (approved):
Great job! 🚀

lazartravica (approved):
Recompiled, reran the tests and it all works excellent!

zivkovicmilos (approved):
[Test Results — Run all tests.pdf](https://github.com/0xPolygon/polygon-sdk/files/6654870/Test.Results.Run.all.tests.pdf)
All tests are passing, looks good to me.
Clean and efficient as always 💯

PR #61, opened 2021-06-07
Pass command line arguments as flags

Vuksan:
# Description

Modified these commands to have mandatory flags passed as flags:

- peers add
- peers status
- ibft init

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Will update the docs as well.
_EDIT:_ Docs update PR: https://github.com/0xPolygon/polygon-sdk-docs/pull/6

### DIFF
--- a/command/peers_status.go
+++ b/command/peers_status.go
@@ -19,7 +19,7 @@ func (p *PeersStatus) DefineFlags() {
 	}
 
 	p.flagMap["libp2p-node-id"] = FlagDescriptor{
-		description: fmt.Sprintf("A unique reference to a specific peer within p2p network"),
+		description: "A unique reference to a specific peer within p2p network",
 		arguments: []string{
 			"PEER_ID",
 		},
--- a/command/peers_status.go
+++ b/command/peers_status.go
@@ -12,6 +12,21 @@ type PeersStatus struct {
 	Meta
 }
 
+func (p *PeersStatus) DefineFlags() {
+	if p.flagMap == nil {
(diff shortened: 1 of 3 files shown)


lazartravica (approved):
LGTM

command/peers_status.go
```diff
@@ -20,8 +35,9 @@ func (p *PeersStatus) GetHelperText() string {
 // Help implements the cli.PeersStatus interface
 func (p *PeersStatus) Help() string {
 	p.Meta.DefineFlags()
+	p.DefineFlags()
 
-	usage := "peers status PEER_ID"
+	usage := "peers status --libp2p-node-id PEER_ID"
```
zivkovicmilos: Kind of a nitpick, but I'm trying to think of a different way we can name this flag.
**libp2p-node-id** just seems a bit long, especially for a command that just returns basic peer data.

What you you think about just **peer-id**?
@Kourin1996 @lazartravica

zivkovicmilos (approved):
Great and clean, as always 💯

Kourin1996 (approved):
LGTM. It works locally.

command/peers_status.go
Kourin1996: I think **peer-id** sounds ok. Can understand fully enough. It's better to change help message also if you change to **peer-id**.

Vuksan: My intention was to add more context to the flag itself, but I agree, the command is often used, so the flag should be shorter. Will modify.

PR #62, opened 2021-06-09
feature/cli-commands-touchup

zivkovicmilos:
# Description

This PR fixes some cli command outputs, as well as standardizes the way some commands are being called.

# Changes include

# Checklist

- [x] I have tested this code


lazartravica (changes requested):
Excellent stuff!

Just one question, I would appreciate to hear your thoughts on this: Why did we move away from the subcommands, e.g. `ibft init` towards `ibft-init`?

Unless we have a very good argument for that I would advise against doing this as we would need to update the docs and communicate with the users that they need to change the way they do everything with polygon-sdk.

command/commands.go
```diff
@@ -124,11 +126,28 @@ func Commands() map[string]cli.CommandFactory {
...
+func (m MetaFlagDescriptor) GetArgumentsList() []string {
+	return m.arguments
+}
+
+func (m MetaFlagDescriptor) GetArgumentsOptional() bool {
+	return m.argumentsOptional
+}
+
+func (m MetaFlagDescriptor) GetFlagOptional() bool {
+	return m.flagOptional
```
Vuksan: Getter methods for bool flags shouldn't start with a `Get` keyword, but with `is`/`has`. - `IsFlagOptional()`
Ref: https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter

As for plural arguments `are` is used very often, but it's still better than `Get` IMO. `AreArgumentsOptional()`
Btw not sure why we have this `argumentsOptional` flag when none of the arguments are actually optional, and we only pass false to it.

zivkovicmilos:
@lazartravica 
We moved away from the subcommands model because it's easier to keep everything uniform, [same as with geth](https://geth.ethereum.org/docs/interface/command-line-options).
The commands also now follow the [UCB utility guidelines](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

There is already [a PR waiting](https://github.com/0xPolygon/polygon-sdk-docs/pull/7) on the documentation repo, with the changes from this PR.

command/ibft_propose.go
```diff
@@ -17,39 +18,43 @@ type IbftPropose struct {
 func (p *IbftPropose) DefineFlags() {
 	if p.flagMap == nil {
 		// Flag map not initialized
-		p.flagMap = make(map[string]FlagDescriptor)
+		p.flagMap = make(map[string]types.FlagDescriptor)
 	}
 
-	p.flagMap["add"] = FlagDescriptor{
-		description: "Proposes a new validator to be added to the validator set",
+	p.flagMap["a"] = MetaFlagDescriptor{
```
Vuksan: let's use `address` as a flag, or at least `addr` 🙏

command/ibft_propose.go
```diff
@@ -59,32 +64,27 @@ func (p *IbftPropose) Synopsis() string {
...
-	flags := p.FlagSet("ibft propose")
+	flags := p.FlagSet(p.GetBaseCommand())
 
-	var add, del bool
-	flags.BoolVar(&add, "add", false, "add")
-	flags.BoolVar(&del, "del", false, "del")
+	var vote bool
+	var ethAddress string
+
+	flags.BoolVar(&vote, "vote", false, "")
```
Vuksan: I think it isn't wise to allow `vote` flag to be optional, implicitly meaning "Remove validator"

command/ibft_snapshot.go
```diff
@@ -87,27 +91,46 @@ func (p *IbftSnapshot) Run(args []string) int {
...
-		votes[i+1] = fmt.Sprintf("%s|%s|%v", d.Validator, d.Proposed, d.Auth)
+	output += "\n"
+
+	var votes []string = nil
+	if len(s.Votes) == 0 {
+		votes = make([]string, 1)
+		votes[0] = "No votes found"
+	} else {
+		votes = make([]string, len(s.Votes)+1)
+		votes[0] = "Proposer|Address|Authorize"
```
Vuksan:
```suggestion
	votes := make([]string, len(s.Votes)+1)
	if len(s.Votes) == 0 {
		votes[0] = "No votes found"
	} else {
		votes[0] = "Proposer|Address|Authorize"
```

command/ibft_snapshot.go
```diff
@@ -87,27 +91,46 @@ func (p *IbftSnapshot) Run(args []string) int {
...
-		validators[i+1] = d.Address
+	output += "\n"
+
+	var validators []string = nil
+	if len(s.Validators) == 0 {
+		validators = make([]string, 1)
+		validators[0] = "No validators found"
+	} else {
+		validators = make([]string, len(s.Validators)+1)
+		validators[0] = "Address"
```
Vuksan:
```suggestion
	validators := make([]string, len(s.Validators)+1)
	if len(s.Validators) == 0 {
		validators[0] = "No validators found"
	} else {
		validators[0] = "Address"
```

command/monitor.go
```diff
@@ -18,16 +19,14 @@ type MonitorCommand struct {
...
 }
 
 // Help implements the cli.Command interface
 func (m *MonitorCommand) Help() string {
 	m.Meta.DefineFlags()
 
-	usage := "monitor"
-
-	return m.GenerateHelp(m.Synopsis(), usage)
+	return types.GenerateHelp(m.Synopsis(), types.GenerateUsage("monitor", m.flagMap), m.flagMap)
```
Vuksan: Add `GetBaseCommand()` method here as well?

command/peers_add.go
```diff
@@ -47,33 +54,63 @@ func (p *PeersAdd) Synopsis() string {
...
 
-	p.UI.Info("Peer added")
+	var output = "\n[PEERS ADDED]\n"
+	output += formatKV([]string{
+		fmt.Sprintf("Peers listed|%d", len(passedInAddresses)), // The number of peers the user wanted to add
+		fmt.Sprintf("Peers added|%d", peersAdded),              // The number of peers that have been added
+	})
+
+	if len(addedPeers) > 0 {
+		output += "\n\n[LIST OF DIALED PEERS]\n"
```
Vuksan: you meant `LIST OF ADDED PEERS`?

command/peers_add.go
```diff
@@ -13,31 +17,34 @@ type PeersAdd struct {
 func (p *PeersAdd) DefineFlags() {
 	if p.flagMap == nil {
 		// Flag map not initialized
-		p.flagMap = make(map[string]FlagDescriptor)
+		p.flagMap = make(map[string]types.FlagDescriptor)
 	}
 
-	p.flagMap["libp2p-address"] = FlagDescriptor{
+	p.flagMap["addr"] = MetaFlagDescriptor{
```
Vuksan: Can you explain a bit more in the description how multiple addresses should be passed as arguments?

command/status.go
```diff
@@ -18,13 +19,15 @@ func (c *StatusCommand) GetHelperText() string {
...
+}
+
 // Help implements the cli.Command interface
 func (c *StatusCommand) Help() string {
 	c.Meta.DefineFlags()
 
-	usage := "status"
-
-	return c.GenerateHelp(c.Synopsis(), usage)
+	return types.GenerateHelp(c.Synopsis(), c.GetBaseCommand(), c.flagMap)
```
Vuksan: You need to call `GenerateUsage` because of `--grpc-address` global flag.
```suggestion
	return types.GenerateHelp(c.Synopsis(), types.GenerateUsage(c.GetBaseCommand(), c.flagMap), c.flagMap)
```

command/server/command.go
```diff
@@ -9,28 +9,193 @@ import (
...
+	return s.arguments
+}
+
+func (s ServerFlagDescriptor) GetArgumentsOptional() bool {
+	return s.argumentsOptional
+}
+
+func (s ServerFlagDescriptor) GetFlagOptional() bool {
+	return s.flagOptional
+}
```
Vuksan: We shouldn't have to create duplicate `structs` to resolve cyclic imports. 😐 
What do you think of creating a new `helper` package within `command` and move `MetaFlagDescriptor` there?

command/server/command.go
```diff
@@ -9,28 +9,193 @@ import (
...
+	c.flagMap["nat"] = ServerFlagDescriptor{
+		description: "Sets the the external IP address without the port, as it can be seen by peers",
+		arguments: []string{
+			"NAT_ADDRESS",
+		},
+		flagOptional: true,
+	}
+
+	c.flagMap["no-discover"] = ServerFlagDescriptor{
+		description: "Sets the client's discoverability for other peers. Default false",
```
Vuksan: By this description, I would suppose that `true` would set the client as discoverable by other peers.
```suggestion
		description: "Prevents the client from discovering other peers. Default false",
```

Vuksan (changes requested):
Nice job!

Take a look at the suggestions I left - feel free to discuss anything you don't agree with.

command/commands.go
zivkovicmilos: Completely agree. Changing it now 👍

command/ibft_propose.go
zivkovicmilos: Missed this, should definitely be `addr`, so we're consistent

command/ibft_propose.go
zivkovicmilos: If we want to keep it within a bool, it needs to be either `true` or `false`, and in Golang it's false by default.
I'll change it to `true`, since it feels more natural.
@lazartravica thoughts?

command/monitor.go
zivkovicmilos: Yep, fixed 👍

command/peers_add.go
zivkovicmilos: The reasoning behind adding the ability to add more peers is just better ux - not having to type out the same command a lot of times. 

How it would work after this PR, if you wanted to add multiple peers:
`go run main.go peers-add -addr <ADDR1> -addr <ADDR2> -addr <ADDR3>` and so on.

command/status.go
zivkovicmilos: You're right - I missed this when adding in the generateUsage

command/server/command.go
zivkovicmilos: You're completely right. I've moved it to a separate package, and now we have a single FlagDescriptor 👍

command/ibft_propose.go
zivkovicmilos: Updated the vote so it takes in a string (`add` or `remove`) 👍

lazartravica: @zivkovicmilos Looks good, only comment is that in the eip the vote types are named "auth" and "drop" https://github.com/ethereum/EIPs/issues/650, not add or remove

lazartravica: As for the argument being mandatory I agree completely 👍

zivkovicmilos: This is even better. We should try to stick to the EIP naming conventions as much as possible

e2e/framework/testserver.go
```diff
@@ -125,33 +131,49 @@ type InitIBFTResult struct {
...
+		return nil, err
+	}
+
+	// Get the node ID from the private key
+	nodeId, err := peer.IDFromPrivateKey(libp2pKey)
+	if err != nil {
+		return nil, err
 	}
 
+	res.PublicKey = crypto.PubKeyToAddress(&key.PublicKey).String()
```
Vuksan: I noticed it now, but our naming is way off here. We call it a `PublicKey` but it is in fact the address.
Can you maybe rename it to `res.Address` or something similar?

Vuksan (approved):
Kudos for refactoring testserver as well 👏 

I left a comment to rename a parameter, which isn't a showstopper. If you agree, you can do that and merge this one.

e2e/framework/testserver.go
zivkovicmilos: Agreed and fixed 👍

command/helper/helper.go
```diff
@@ -0,0 +1,145 @@
...
+
+// FlagDescriptor contains the description elements for a command flag
+type FlagDescriptor struct {
+	Description       string   // Flag description
+	Arguments         []string // Arguments list
+	ArgumentsOptional bool     // Flag indicating if flag arguments are optional
+	FlagOptional      bool
+}
+
+// GetDescription ets the flag description
```
Kourin1996: It may be a typo?

```suggestion
// GetDescription gets the flag description
```

Kourin1996 (approved):
Thank you for your improvements and fixing e2e codes accordingly.
I noticed a small thing, it'd be great if you could fix if necessary.
Thanks.

lazartravica (approved):
you did an awesome job!

everything works and it's much clearer now.

PR #65, opened 2021-06-11
Fix crash issue on moving to new epoch due to no snapshot in memory

Kourin1996:
# Description

This PR fixes the issue to crash process due to no snapshot in memory.

## Issue Description

The issue happens when the sequence reaches `3*EpochSize` if validators and votes are not changed from beginning. It causes panic with following message. (I changed EpochSize to 10 for debug)

```shell
2021-06-11T15:33:46.024+0900 [INFO]  polygon.consensus.ibft.acceptState: Accept state: sequence=31
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x18b38f2]

goroutine 152 [running]:
github.com/0xPolygon/minimal/consensus/ibft.(*Ibft).runAcceptState(0xc0008a4750)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-sdk/consensus/ibft/ibft.go:449 +0x332
github.com/0xPolygon/minimal/consensus/ibft.(*Ibft).runCycle(0xc0008a4750)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-sdk/consensus/ibft/ibft.go:247 +0x4d
github.com/0xPolygon/minimal/consensus/ibft.(*Ibft).start(0xc0008a4750)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-sdk/consensus/ibft/ibft.go:233 +0x11d
created by github.com/0xPolygon/minimal/consensus/ibft.(*Ibft).Start
        /Users/kourin/go/src/github.com/0xPolygon/polygon-sdk/consensus/ibft/ibft.go:128 +0x89
exit status 2
make: *** [run] Error 1
```

The reason is that node deletes old snapshots from memory every epoch but will reuse same snapshot if snapshot doesn't change.

Snapshot is the data of validators and votes at a certain `sequence`. IBFT consensus has a list of snapshot, and votes are reset when move to new epoch. (So epoch size means the period when votes are available) 

Snapshot is appended to the list when snapshot change (validators or votes change) or new epoch. But snapshot is not appended to the list if it equals to previous snapshot. Snapshots with lower number than `(current epoch - 2)*(epoch size)` are removed from the list on new epoch.

Let's say validators and votes are not changed from beginning. In this case, snapshot list has only a snapshot with 0 number (genesis one). No new snapshot is added to the list because new snapshot equals to previous one. When the sequence reaches (epoch size)*3, all snapshots with lower number than (epoch size) is removed from the list. As a result, the list doesn't have any snapshot and crashes from next sequence.

## Change of this PR

This PR fixes this issue and makes sure snapshot is created on new epoch. For that, I moved conditional branch for equality of snapshot.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/consensus/ibft/snapshot.go
+++ b/consensus/ibft/snapshot.go
@@ -128,10 +128,6 @@ func (i *Ibft) processHeaders(headers []*types.Header) error {
 
 	// saveSnap is a callback function for saving the passed in header to the snapshot store
 	saveSnap := func(h *types.Header) error {
-		if snap.Equal(parentSnap) {
-			return nil
-		}
-
 		snap.Number = h.Number
 		snap.Hash = h.Hash.String()
 
@@ -242,8 +238,10 @@ func (i *Ibft) processHeaders(headers []*types.Header) error {
 			})
 		}
 
-		if err := saveSnap(h); err != nil {
-			return nil
+		if !snap.Equal(parentSnap) {
+			if err := saveSnap(h); err != nil {
+				return nil
+			}
 		}
 	}
 
--- a/consensus/ibft/snapshot_test.go
+++ b/consensus/ibft/snapshot_test.go
@@ -123,12 +123,13 @@ func TestSnapshot_ProcessHeaders(t *testing.T) {
 	}
 
 	var cases = []struct {
+		name       string
 		epoch      uint64
 		validators []string
 		headers    []mockHeader
 	}{
 		{
-			// single validator casts no vote
+			name: "single validator casts no vote",
 			validators: []string{
 				"A",
 			},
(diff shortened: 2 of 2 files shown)


lazartravica (approved):
Tested, this fix works.

zivkovicmilos (approved):

lazartravica (changes requested):
LGTM! Tested, but we are lacking tests for this case.

Kourin1996:
@lazartravica I added new unit test. This should be failed on current develop but passed on this PR.

lazartravica (approved):
🤘 Works excellent

Vuksan (approved):
Great job on such a detailed description! 👏

Kourin1996:
@lazartravica Thanks for your improvements :)

PR #67, opened 2021-06-15
Restore snapshot from blocks if can't find snapshot from data on initialize

Kourin1996:
# Description

Relevant issue: https://github.com/0xPolygon/polygon-sdk/issues/55
This issue happens when node can't load snapshot from file in `setupSnapshot` and can't get no snapshot to produce next snapshot.

This PR adds code to restore snapshot from block header at beginning of current epoch if list doesn't have the snapshot to calculate next snapshot. 

Snapshot has two kinds of data, validator set and votes. Validator set is in extra field of header but votes aren't, so it needs to process headers from beginning of epoch to current block to generate snapshot at specific block number.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


consensus/ibft/snapshot.go
```diff
@@ -46,6 +46,27 @@ func (i *Ibft) setupSnapshot() error {
 		}
 	}
 
+	currentEpoch, metaEpoch := (header.Number / i.epochSize), (meta.LastBlock / i.epochSize)
+	snapshot, _ := i.getSnapshot(header.Number)
+	if snapshot == nil || metaEpoch < currentEpoch {
+		// If there is no snapshot for next snapshot, restore snapshot at beginning of current epoch and update last block number
```
zivkovicmilos: This was probably a typo: 
"If there is no snapshot for the next `snapshot`..."

zivkovicmilos (approved):
Good fix, even better tests 💯

consensus/ibft/snapshot.go
```diff
@@ -46,6 +46,28 @@ func (i *Ibft) setupSnapshot() error {
 		}
 	}
 
+	currentEpoch, metaEpoch := (header.Number / i.epochSize), (meta.LastBlock / i.epochSize)
```
lazartravica: Please add comment to what we are doing here.

For example:
```
If the snapshot is not found, or the latest snapshot belongs to a previous epoch, we need to start rebuilding the snapshot from the beginning of the current epoch in order to have all the votes and validators correctly set in the snapshot, since they reset every epoch.
```

lazartravica: Just a comment on the code style, open for discussion:

`x, y := 1, 2`
seems to be harder to read than
```
x := 1
y := 2
```

lazartravica (changes requested):
I really like what the IBFT snapshot tests look like now, excellent work!

Please just add a bit of comments on what we are doing when snapshot is not found and the latest processed block belongs to the previous epoch.

consensus/ibft/snapshot.go
```diff
@@ -46,6 +46,35 @@ func (i *Ibft) setupSnapshot() error {
...
+			return err
+		}
+		i.store.updateLastBlock(beginHeight)
+
+		if meta, err = i.getSnapshotMetadata(); err != nil {
+			return err
+		}
+	}
+
 	// Some of the data might get lost due to ungrateful disconnections
```
Vuksan: Better description in light of these changes
```suggestion
	// Process headers if we missed some blocks in the current epoch
```

consensus/ibft/snapshot.go
```diff
@@ -46,6 +46,35 @@ func (i *Ibft) setupSnapshot() error {
 		}
 	}
 
+	// If the snapshot is not found, or the latest snapshot belongs to a previous epoch,
+	// we need to start rebuilding the snapshot from the beginning of the current epoch
+	// in order to have all the votes and validators correctly set in the snapshot,
+	// since they reset every epoch.
```
Vuksan, line 52: Just to confirm - we are fine with not having some snapshots in the store?
If the difference between `currentEpoch` and `metaEpoch` is more than 1 we will disregard those epochs in between

cc @lazartravica @zivkovicmilos

consensus/ibft/snapshot_test.go
```diff
@@ -96,32 +109,254 @@ func (ap *testerAccountPool) ValidatorSet() ValidatorSet {
...
 }
 
-func TestSnapshot_ProcessHeaders(t *testing.T) {
-	type mockVote struct {
-		validator string
-		voted     string
-		auth      bool
+type mockVote struct {
+	validator string
+	voted     string
```
Vuksan: It's a bit confusing what do `voted` and `validator` represent here?
a) Is `validator` - validator who voted for `voted` candidate?
b) Is `voted` - validator who voted for `validator` candidate?

If `a` is the case, can you rename `voted` to `candidate` to make it less ambiguous?

consensus/ibft/snapshot_test.go
```diff
@@ -96,32 +109,254 @@ func (ap *testerAccountPool) ValidatorSet() ValidatorSet {
...
+	auth      bool
+}
+
+func mine(validator string) mockVote {
+	return mockVote{validator: validator}
+}
```
Vuksan: I fail to grasp why is this function called `mine` 😕 
Can you pls explain?

Vuksan: Did you mean it in a sense of "mine a regular block"?

consensus/ibft/snapshot_test.go
```diff
@@ -96,32 +109,254 @@ func (ap *testerAccountPool) ValidatorSet() ValidatorSet {
...
+		}
+	}
+
+	type snapshotData struct {
+		LastBlock uint64
+		Snapshots []*Snapshot
+	}
+	var cases = []struct {
+		name           string
+		epoch          uint64
```
Vuksan: Can you please rename this to `epochSize`?
My mind kept thinking it was a specific epoch 🙂

consensus/ibft/snapshot_test.go
```diff
@@ -96,32 +109,254 @@ func (ap *testerAccountPool) ValidatorSet() ValidatorSet {
...
+				Snapshots: []*Snapshot{
+					newSnapshot(3, validatorSet, []*Vote{}),
+					newSnapshot(4, validatorSet, []*Vote{{
+						Validator: pool.get("D").Address(),
+						Address:   pool.get("E").Address(),
+						Authorize: true,
+					}}),
+				},
+			},
+		},
```
Vuksan: I'm struggling to understand this last case.
If epoch size is 3, shouldn't first 3 blocks be in the first epoch, and only the last one with vote for `D` and `E` in second epoch, so the expected snapshots should contain just the last one, and not `newSnapshot(3, validatorSet, []*Vote{})`?

consensus/ibft/snapshot_test.go
```diff
@@ -96,32 +109,254 @@ func (ap *testerAccountPool) ValidatorSet() ValidatorSet {
...
 	}
```
Vuksan, line 345: Add this test case if you can:
```suggestion
		{
			name:  "should not count votes from the beginning of current epoch as there cannot be any proposals during the checkpoint block",
			epoch: 3,
			headers: []mockHeader{
				newMockHeader(validators, mine("A")),
				newMockHeader(validators, vote("B", "F", true)),
				newMockHeader(validators, mine("C")),
				newMockHeader(validators, vote("D", "E", true)),
				newMockHeader(validators, mine("D")),
				newMockHeader(validators, vote("C", "F", true)),
				newMockHeader(validators, vote("A", "F", true)),
			},
			savedSnapshots: nil,
			expectedResult: snapshotData{
				LastBlock: 7,
				Snapshots: []*Snapshot{
					newSnapshot(6, validatorSet, []*Vote{}),
					newSnapshot(7, validatorSet, []*Vote{{
						Validator: pool.get("A").Address(),
						Address:   pool.get("F").Address(),
						Authorize: true,
					}}),
				},
			},
		},
	}
```

Vuksan (changes requested):
Great job on the tests! 👏 

I left some comments just to understand the implementation fully.
Probably nothing major needs to be changed, but better safe than sorry. 🙂

consensus/ibft/snapshot.go
Vuksan: I understand that we dump old epochs anyway, just wanted to make sure if it's ok to have a gap of 1 epoch in the store, for example.

Kourin1996: It'll work as long as there is a snapshot to be used to calculate next snapshot. Restored snapshot must be at the beginning of the epoch. Subsequent codes advances snapshot from the beginning of the epoch to the current number.

The stored snapshot doesn't have to be in same epoch to the current. So that, second condition `metaEpoch < currentEpoch` is just for jumping to current epoch.

consensus/ibft/snapshot_test.go
Kourin1996: (a) is correct, I will rename `voted` to `candidate `.

consensus/ibft/snapshot_test.go
Kourin1996: This is the function to propose header without any vote. I'm renaming `mine` to `skipVote` for similar name

consensus/ibft/snapshot_test.go
Kourin1996: We have a genesis header, so `Number` of normal header begins from 1.
So in this case, blockchain has following headers.
0: genesis header
1: header produced by `A`
2: header produced by `B` with vote to `F`
3: header produced by `C`
4: header produced by `D` with vote to `E`

Currently epoch size is 3, epoch advances every 3 header. So 0, 1, 2 are in first epoch while 3, 4 are in second epoch.

consensus/ibft/snapshot_test.go
Kourin1996: I added this case and it works without any change. Thanks!

consensus/ibft/snapshot_test.go
Kourin1996: I renamed it. Thank you :)

consensus/ibft/snapshot.go
Kourin1996: I updated this comment. Thanks :)

consensus/ibft/snapshot_test.go
Vuksan: Thanks, it wasn't obvious from looking at the test 🙂

Vuksan (approved):

lazartravica (approved):
LGTM

PR #69, opened 2021-06-16
IBFT POS Staking

zivkovicmilos:
# Description

This PR modifies the Account state, and adds support for the staking mechanism in PoS.

# Changes include

# Breaking changes

The account state is modified to support the `StakedBalance`, as such, it causes a chain reaction in where this field is updated.

# Checklist

- [x] I have tested this code


lazartravica (commented):
Just a comment, since this is still a draft.

Try to separate all the staking logic to different files / packages so that it can be easier to refactor into a plugin later.

If it seems like it's not possible, that's okay, then we should be aware of that, and somehow "mark" which functions are used for staking only

e2e/framework/config.go
```diff
@@ -42,8 +43,9 @@ func (t *TestServerConfig) Premine(addr types.Address, amount *big.Int) {
 		t.PremineAccts = []*SrvAccount{}
 	}
 	t.PremineAccts = append(t.PremineAccts, &SrvAccount{
-		Addr:    addr,
-		Balance: amount,
+		Addr:          addr,
+		Balance:       amount,
+		StakedBalance: big.NewInt(0),
```
zivkovicmilos, line 48: @lazartravica @Vuksan @Kourin1996 
What do you guys think about adding the ability to set premined staked balance for accounts?

chain/chain.go
```diff
@@ -254,6 +256,11 @@ func (g *GenesisAccount) MarshalJSON() ([]byte, error) {
 	if g.Balance != nil {
 		obj.Balance = types.EncodeBigInt(g.Balance)
 	}
+	// Staking implementation
+	if g.StakedBalance != nil {
+		obj.StakedBalance = types.EncodeBigInt(g.StakedBalance)
+	}
+	//
```
Vuksan:
```suggestion

```

chain/chain.go
```diff
@@ -226,19 +226,21 @@ func (g *Genesis) UnmarshalJSON(data []byte) error {
...
 type GenesisAccount struct {
-	Code       []byte                    `json:"code,omitempty"`
-	Storage    map[types.Hash]types.Hash `json:"storage,omitempty"`
-	Balance    *big.Int                  `json:"balance,omitempty"`
-	Nonce      uint64                    `json:"nonce,omitempty"`
-	PrivateKey []byte                    `json:"secretKey,omitempty"` // for tests
+	Code          []byte                    `json:"code,omitempty"`
+	Storage       map[types.Hash]types.Hash `json:"storage,omitempty"`
+	Balance       *big.Int                  `json:"balance,omitempty"`
+	StakedBalance *big.Int                  `json:"stakedBalance,omitempty"` // Staking implementation
```
Vuksan, line 232: Why is it `omitempty` only here, and not in the Encoder as well?

jsonrpc/stake_endpoint.go
```diff
@@ -0,0 +1,29 @@
...
+// GetStakedBalance returns the account's staked balance at the referenced block
+func (s *Stake) GetStakedBalance(address types.Address, number BlockNumber) (interface{}, error) {
+	header, err := s.d.getBlockHeaderImpl(number)
+	if err != nil {
+		return nil, err
+	}
+
+	acc, err := s.d.store.GetAccount(header.StateRoot, address)
+	// TODO Consider tackling this: `error getting storage snapshot`,
+	// when the account is not initialized
```
Vuksan: TODO left?

state/executor.go
```diff
@@ -591,6 +591,26 @@ func (t *Transition) GetBalance(addr types.Address) *big.Int {
...
+	return t.state.GetStakedBalance(addr)
+}
+
+func (t *Transition) AddStakedBalance(addr types.Address, balance *big.Int) {
+	t.state.AddStakedBalance(addr, balance)
+}
+
+func (t *Transition) SubStakedBalance(addr types.Address, balance *big.Int) {
+	t.state.SubStakedBalance(addr, balance)
+}
```
Vuksan: Out of these 5 I only see `AddStakedBalance` called.
I guess you added them for future usage, but I would advise against it.
Do not try to think "What will we need in the future?". Only implement what's needed at this very moment. 🙂

state/state.go
```diff
@@ -35,6 +35,26 @@ type Account struct {
...
+}
+
+func (a *AccountLegacy) MarshalWith(ar *fastrlp.Arena) *fastrlp.Value {
+	v := ar.NewArray()
+	v.Set(ar.NewUint(a.Nonce))
+	v.Set(ar.NewBigInt(a.Balance))
+	v.Set(ar.NewBytes(a.Root.Bytes()))
+	v.Set(ar.NewBytes(a.CodeHash))
+	return v
 }
```
Vuksan: This `AccountLegacy` isn't used as well?

state/state.go
```diff
@@ -163,14 +197,29 @@ func (s *StateObject) Copy() *StateObject {
...
+	Deleted       bool
+
+	// TODO: Move this to executor
+	DirtyCode bool
+	Code      []byte
+
+	Storage []*StorageObject
+}
+
+type ObjectLegacy struct {
```
Vuksan: ObjectLegacy is unused?

state/txn.go
```diff
@@ -187,14 +188,30 @@ func (txn *Txn) SubBalance(addr types.Address, balance *big.Int) {
 	})
 }
 
+// SubStakedBalance reduces the staked balance
+func (txn *Txn) SubStakedBalance(addr types.Address, balance *big.Int) {
+	if balance.Sign() == 0 {
+		return
+	}
+	txn.upsertAccount(addr, true, func(object *StateObject) {
+		object.Account.StakedBalance.Sub(object.Account.StakedBalance, balance)
```
Vuksan: Won't this move `StakedBalance` to negative if `balance` > `Account.StakedBalance`?

state/txn_test.go
```diff
@@ -100,7 +100,8 @@ func buildMockPreState(p *PreState) (*Account, *mockSnapshot) {
 	account := &Account{
 		Nonce:   p.Nonce,
 		Balance: big.NewInt(int64(p.Balance)),
-		Root:    root,
+		// TODO Mock the Staked balance
```
Vuksan: TODO left

tests/testing.go
```diff
@@ -204,6 +204,7 @@ func buildState(t *testing.T, allocs map[types.Address]*chain.GenesisAccount) (s
 		txn.CreateAccount(addr)
 		txn.SetNonce(addr, alloc.Nonce)
 		txn.SetBalance(addr, alloc.Balance)
+		//txn.SetStakedBalance(addr, big.NewInt(0))
```
Vuksan: forgot to remove this?

state/runtime/system/system_test.go
```diff
@@ -0,0 +1,80 @@
...
+		},
+	}
+
+	systemRuntime := NewSystem()
+
+	for _, testCase := range testTable {
+		canRun := systemRuntime.CanRun(testCase.contract, nil, nil)
+		if !canRun && testCase.canRun {
+			t.Fatal(fmt.Sprintf("[%s] Runtime doesn't recognize address %s", testCase.name, testCase.contract.CodeAddress))
+		}
```
Vuksan: To take advantage of test table fully:
```suggestion
		t.Run(testCase.name, func(t *testing.T) {
			canRun := systemRuntime.CanRun(testCase.contract, nil, nil)
			assert.Equalf(t, testCase.canRun, canRun, "Runtime doesn't recognize address %s", testCase.contract.CodeAddress)
		})
```

tests/state_test.go
```diff
@@ -31,6 +35,105 @@ type stateCase struct {
...
+
+		// Add the legacy account to the local trie
+		trieTransaction.Insert(hashIt(accountAddress.Bytes()), data)
+		arena.Reset()
+	}
+
+	newRoot, _ := trieTransaction.Hash()
+
+	return newRoot
+}
```
Vuksan: Maybe remove these changes as well, since `pruneStakedAccounts` isn't used in this PR?

e2e/framework/config.go
Vuksan: That seems reasonable to me.

e2e/system_test.go
```diff
@@ -0,0 +1,201 @@
...
+			)
+			assert.NoError(t, err)
+
+			var out string
+			if callErr := rpcClient.Call(
+				"stake_getStakedBalance",
+				&out,
+				web3.Address(testCase.staker),
+				web3.Latest.String(),
+			); callErr != nil && testCase.shouldSucceed {
```
Vuksan: It should return the result (0) even if the account doesn't exist, so no need for the second part of the condition, right?
```suggestion
			); callErr != nil {
```

e2e/system_test.go
```diff
@@ -0,0 +1,201 @@
...
+				web3.Latest,
+			)
+			assert.NoError(t, err)
+
+			if callErr := rpcClient.Call(
+				"stake_getStakedBalance",
+				&out,
+				web3.Address(testCase.staker),
+				web3.Latest.String(),
+			); callErr != nil && testCase.shouldSucceed {
```
Vuksan: It should return the result (0) even if the account doesn't exist, so no need for the second part of the condition, right?
```suggestion
			); callErr != nil {
```

Vuksan (changes requested):

chain/chain.go
zivkovicmilos: I was just following the same json tag structure we used for `Balance`

state/executor.go
zivkovicmilos: I believe we use the rest of the unused StakedBalance functions in the unstaking PR (#76)

state/state.go
zivkovicmilos: Yup, leftover from an implementation that is on the Eth State tests fix branch now 🙂
Removed 👍

state/state.go
zivkovicmilos: Same as with AccountLegacy.

Removed 👍

state/txn.go
zivkovicmilos: The only place this is called from is from the [unstaking handler](https://github.com/0xPolygon/polygon-sdk/pull/76/files#diff-d4e1bf18e1d1050f24fa6a0d47dc44b3ac67e2866a634b72342739164372dac9R45), where we previously check what the staked account balance was and deduct it

state/txn_test.go
zivkovicmilos: Removed 👍

tests/testing.go
zivkovicmilos: Removed 👍

state/runtime/system/system_test.go
zivkovicmilos: You're right, missed this as I was testing
Fixed 👍

tests/state_test.go
zivkovicmilos: Agreed. This code will still be present on the fix branch

Removed 👍

e2e/system_test.go
zivkovicmilos: Yes. The test should fail if this call fails, regardless of the test case, since getStakedBalance returns 0 for unknown accounts, as you said.

Fixed 👍

e2e/system_test.go
zivkovicmilos: Fixed 👍

chain/chain.go
Vuksan: Fair enough 🙂

state/executor.go
Vuksan: Then they should be a part of that PR. 🙂

state/txn.go
Vuksan: True, but I would argue that subBalance itself should return an error if this occurs. However I can see that for regular balance this isn't done as well. Not sure what to think regarding that.

state/executor.go
Vuksan: @zivkovicmilos Are they all used there? Also getters and subBalance?

state/txn.go
zivkovicmilos: We can always go along the route - the caller has the responsibility to ensure correct state execution

state/executor.go
zivkovicmilos: I think so. 
I'll go ahead and remove them from this branch, since they are present and used on the other one. You're right - it shouldn't be a part of this PR 💯

Vuksan: @zivkovicmilos Be careful there!
If you remove them here and later on merge this branch into PoS one, when the time comes to merge the Unstaking branch with the PoS you will pull changes from the PoS branch which will remove these functions from the unstaking branch as well, so take note to add them back there.

zivkovicmilos: Got it. I'll make sure they're included on the unstaking PR 👍

zivkovicmilos: Removed 👍

Vuksan: @zivkovicmilos You can remove `AddBalance` and `SubBalance` that you've added to the executor. They're unused as well.

state/txn.go
Vuksan: Ok, let's go like that for now 🙂

state/executor.go
zivkovicmilos: Removed 👍

Vuksan (approved):
LGTM

Kourin1996 (approved):
LGTM, it works properly 👍

PR #71, opened 2021-06-18
Fix GetFilterChanges doesn't return every log from NewFilter

Kourin1996:
# Description

This PR fixes the issue of missing some logs in response of `eth_getFilterChanges`.
Number of logs doesn't equal to the number of transactions which are sent after call `eth_newFilter` before call `eth_getFilterChanges`

This issue caused by wrong order of some codes.
Following codes are a part of codes in `WriteBlocks` of `Blockchain` in blockchain/blockchain.go
https://github.com/0xPolygon/polygon-sdk/blob/4df9810cb85fcf69b5f777a33eb5fd1f8c6c9224/blockchain/blockchain.go#L577-L584

`dispatchEvent` dispatches an event to notify that new header is included and `WriteReceipts` writes receipts of transactions in block into storage (leveldb or memory).
`FilterManager` defined in jsonrpc/filter_manager.go subscribes to the event and get receipts for transaction by `GetReceiptsByHash` and append to each filter. So there's a possibility that FilterManager loads receipts before saving receipts in current code.

This PR swaps the order of these two call. `WriteReceipts` is used only in this method and `GetReceiptsByHash` is called from FilterManager, jsonrpc, and grpc. So perhaps there is no side effect without issue fix.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -574,7 +574,6 @@ func (b *Blockchain) WriteBlocks(blocks []*types.Block) error {
 		if err := b.writeHeaderImpl(evnt, header); err != nil {
 			return err
 		}
-		b.dispatchEvent(evnt)
 
 		// write the receipts, do it only after the header has been written.
 		// Otherwise, a client might ask for a header once the receipt is valid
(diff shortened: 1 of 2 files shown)


Vuksan (approved):
Excellent PR description 🚀

zivkovicmilos (approved):
Good catch, as always 👍 
The description is 💯

lazartravica (approved):
LGTM

PR #72, opened 2021-06-22
IBFT proposer reward not assigned to a block proposer

Vuksan:
# Description

Currently, block reward is assigned to `Miner` field in the block header.
For regular PoW this is fine, but in IBFT `Miner` field is not used as an address of the block proposer. Instead, it holds the candidate's address during the voting process. If there is no voting process, this field is set to zero address.

You can read more about it in the [Repurposing header fields for signing and voting section](https://github.com/ethereum/EIPs/issues/225) of Clique's EIP

I decided to let consensus be responsible for defining how block creator is fetched.
Most consensuses would simply return `Miner` field, but IBFT ecrecovers it from the `Seal` in the `extradata`

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -47,10 +47,11 @@ type Blockchain struct {
 
 type Verifier interface {
 	VerifyHeader(parent, header *types.Header) error
+	GetBlockCreator(header *types.Header) (types.Address, error)
 }
 
 type Executor interface {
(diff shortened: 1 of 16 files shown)


zivkovicmilos (approved):
Looks good to me 💯

Kourin1996 (approved):
LGTM 👍

PR #73, opened 2021-06-23
Fix concurrent issues in network

Kourin1996:
# Description

This PR fixes following concurrent issues in network.
In addition, it fixes some unit tests to make sure that a node connects to peer as pre-process.

## Deadlock issue in runDial

`runDial` in network/server.go runs an infinite loop to connect new peer. In each loop, runDial gets number of open slots , pick as many peers as the number up from dial queue, and sends request. Number of open slots is calculated by `MaxPeers - (number of connected peers) - (number of pending peers)`. The loop begins to sleep after it sends requests. It wakes up once it receives one of the events and goes to next loop. (PeerEventConnected, PeerEventConnectedFailed, PeerEventDisconnected)

On the other hand, `ConnectedF` callback in `setup` of `network/identity.go` tries handshake after connection established. During this time, identity increase number of pending and decrease it after identity get result from peer.

The problem happens if DisconnectedEvent is emitted before decreasing number of pending. In this case, node gets wrong number of open slots in loop. So sometimes deadlock happens and loop continues to sleep.

This PR adds new event `PeerDialCompleted` which is emitted after identify decreases number of pending.

## handshake should be executed only once in discovery

Nodes try to connect to each other at the same moment if nodes found the peer by discovery protocol and the peer has highest priority. In this case, the handshake called later will be rejected sometimes because the status is already pending and all of open slots has been consumed. 

The handshake called later should be ignored.

## gRPC service should be registered before gRPC server started

grpc server needs to register grpc service before `serve` is called, but `NewGrpcStream` in network/grpc/grpc.go creates new server and calls `serve` inside. If a service is registered after called `serve`, server throws error like `[core] grpc: Server.RegisterService after Server.Serve for "v1.Discovery`.
This PR removes call of `serve` from `NewGrpcStream` and add calls of `Serve` method after a service is registered.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


zivkovicmilos (approved):
I can't decide what's better, the PR itself or just the description, because they're both 💯 
Great work 👍

Vuksan (approved):
Good job, great description 👏

Issue #64, opened 2021-06-10
Error getting storage snapshot when deploying a smart contract to private blockchain via hardhat


e74000:
### Description

When using hardhat to deploy a copy of the sample Greeter.sol script to a private polygon blockchain, I get the error: `error getting storage snapshot` in the server log. I am a bit of a noob at this so please correct me if I'm doing something stupid.

### Your environment

OS: Manjaro linux 5.10.41
Ver: 0.1.0-dev

### Steps to reproduce

- To reproduce start a polygon blockchain, exactly copying this guide: https://sdk-docs.polygon.technology/docs/how-tos/howto-set-ibft,
- Start a JsonRPC server with `go run main.go --chain genesis.json --log-level debug`,
- Compile the sample script in hardhat with `npx hardhat compile` in the sample repository,
- Your JsonRPC servers ip and port, + an account in the `hardhat.config.js` config file
- Run the provided script to deploy the compiled smart contract with `npx hardhat run scripts/sample-script.js --network polygon`

### Expected behaviour

The simple hello world script should be deployed to the blockchain.

### Actual behaviour

In hardhat you get the error `HardhatError: HH110: Invalid JSON-RPC response received: {"code":-32603,"message":"internal error"}`
In the polygon server you get the error `[ERROR] polygon.dispatcher: failed to dispatch: method=eth_estimateGas err="error getting storage snapshot"`

### Logs

```
2021-06-10T13:17:22.133+0100 [DEBUG] polygon.dispatcher: request: method=web3_clientVersion id=1
2021-06-10T13:17:22.144+0100 [DEBUG] polygon.dispatcher: request: method=eth_chainId id=2
2021-06-10T13:17:22.154+0100 [DEBUG] polygon.dispatcher: request: method=eth_chainId id=3
2021-06-10T13:17:22.163+0100 [DEBUG] polygon.dispatcher: request: method=eth_estimateGas id=4
2021-06-10T13:17:22.163+0100 [ERROR] polygon.dispatcher: failed to dispatch: method=eth_estimateGas err="error getting storage snapshot"
```

### Proposed solution

No idea how to fix it, but I found that the file giving the error was minimal/server.go

fastepic:
[...]

PR #74, opened 2021-06-25
Fix/error getting storage snapshot

lazartravica:
# Description

This PR encompasses fixing 3 bugs that I found while deploying Polygon smart contracts (as per the conversation on #64).

## Limit registered libp2p server addresses to 1

While implementing the `--nat` flag a while ago, I appended the NAT address to the list of all registered addresses (which if you are listening on 0.0.0.0 are all network interfaces). I also changed the `AddrInfoToString` as it needed to serialize more than 1 registered address.

This caused an issue since while gossiping, the servers would share multiple multiaddrs when in fact they only need to share their NAT address.

The sharing of the known nodes while gossiping is done with the `AddrInfoToString` fn.

## Not printing an error if the account is uninitialized

Currently on every lookup in the storage using the root hash and the slot, if we are unable to find anything at that location, we will print "error getting storage snapshot".

The error handling for different server calls which may fail gracefully is done outside of the `getState()` method, making the `fmt.Errorf("error getting storage snapshot")` a false positive for an error, when in fact caller functions will handle gracefully if they can (such as `GetBalance` and `getNextNonce`).

## When storing a block from another validator the `From` address should be properly set 

Previously, the `From` was set to ZeroAddress.

The reason behind this is that we are writing the block body (with txs) first into the storage, and only then recovering the `From` address from the signature of the transaction.

I have changed the order of these operations so that we first process the block (effectively recovering the `From` address for each transaction) and only then write it in the TxLookup & block databases.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/minimal/server.go
+++ b/minimal/server.go
@@ -239,16 +239,13 @@ func (j *jsonRPCHub) getState(root types.Hash, slot []byte) ([]byte, error) {
 }
 
 func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
-	obj, err := j.getState(root, addr.Bytes())
+	obj, _ := j.getState(root, addr.Bytes())
 	var account state.Account
-
-	if err != nil {
-		return &account, nil
-	}
 
 	if err := account.UnmarshalRlp(obj); err != nil {
 		return nil, err
 	}
+
 	return &account, nil
 }
 
--- a/minimal/server.go
+++ b/minimal/server.go
@@ -2,6 +2,7 @@ package minimal
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -233,19 +234,20 @@ func (j *jsonRPCHub) getState(root types.Hash, slot []byte) ([]byte, error) {
 	}
 	result, ok := snap.Get(key)
 	if !ok {
-		return nil, fmt.Errorf("error getting storage snapshot")
+		return nil, errors.New("given root and slot not found in storage")
 	}
 	return result, nil
 }
 
 func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
-	obj, _ := j.getState(root, addr.Bytes())
+	obj, err := j.getState(root, addr.Bytes())
+	if err != nil {
+		return nil, err
+	}
 	var account state.Account
-
 	if err := account.UnmarshalRlp(obj); err != nil {
 		return nil, err
 	}
-
 	return &account, nil
 }
 
--- a/minimal/server.go
+++ b/minimal/server.go
@@ -240,10 +240,12 @@ func (j *jsonRPCHub) getState(root types.Hash, slot []byte) ([]byte, error) {
 
 func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
 	obj, err := j.getState(root, addr.Bytes())
-	if err != nil {
-		return nil, err
-	}
 	var account state.Account
+
+	if err != nil {
+		return &account, nil
+	}
+
 	if err := account.UnmarshalRlp(obj); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


Kourin1996 (approved):
LGTM 👏

network/server_test.go
```diff
@@ -173,6 +173,19 @@ func TestNat(t *testing.T) {
 	})
 	defer srv.Close()
 
+	t.Run("there should be multiple listening addresses", func(t *testing.T) {
+		listenAddresses := srv.host.Network().ListenAddresses()
+
+		assert.GreaterOrEqual(t, len(listenAddresses), 1)
```
Vuksan: Since this test states that there should be multiple listening addresses, I guess we should expect strictly "higher than 1":
```suggestion
		assert.Greater(t, len(listenAddresses), 1)
```

minimal/server.go
```diff
@@ -239,14 +239,13 @@ func (j *jsonRPCHub) getState(root types.Hash, slot []byte) ([]byte, error) {
...
-	obj, err := j.getState(root, addr.Bytes())
-	if err != nil {
-		return nil, err
-	}
+	obj, _ := j.getState(root, addr.Bytes())
 	var account state.Account
+
 	if err := account.UnmarshalRlp(obj); err != nil {
 		return nil, err
 	}
```
Vuksan: I am confused by the latest commit 😕 
Won't `getState()` return nil for an address that doesn't exist, and then that nil will be passed to `UnmarshalRlp()` which would result in an error? In the end, we would still get an error instead of a 0 for a nonexisting account.

Vuksan: It would be good to have a test case covering this, to make sure we don't repeat the same mistake again. 🤔

Vuksan (commented):
Excellent commit messages! 👏 
Check out the comments I left. Probably nothing needs to change, I am just unsure if we haven't reverted one of the issues in your last commit.

minimal/server.go
lazartravica: @Vuksan I've made wrong assumptions on how and what needs to be fixed in this case.

I have updated this PR's description accordingly.

Basically, I've just got rid of the `fmt.Errorf()` and replaced it with the simplest `errors.New()` because the severity of the slot & root not being found in the storage is determined from the caller funcs, so we shouldn't keep screaming `unable to get storage snapshot` when we are correctly returning the 0 for the next nonce or balance of an uninitialized address.

Vuksan: @lazartravica I thought that the issue was that we are returning an error for the uninitialized account that we don't have in memory, instead of returning an account with zero balance, as geth does?

lazartravica: @Vuksan 
We are  returning an error which in certain caller functions indicates that the account is uninitialized, allowing them to return some default values for uninitialized accounts such as having 0 balances or having the next nonce set to 0.

The only change is that the error is now created using `errors.New()`, and not `fmt.Errorf()` which will create an error and also print a message, because printing `error getting storage snapshot` when you're correctly getting back the balance of an uninitialized account as 0 is something we don't need.

You can take a look at the examples of how these errors returned from getState() are used to return gracefully even with an error here:
- https://github.com/0xPolygon/polygon-sdk/blob/97a06c1aa7a36758d7dae9efccf78935bfb620f7/jsonrpc/eth_endpoint.go#L438-L442
- https://github.com/0xPolygon/polygon-sdk/blob/97a06c1aa7a36758d7dae9efccf78935bfb620f7/jsonrpc/dispatcher.go#L396-L399

Vuksan: @lazartravica `fmt.Errorf()` doesn't print a message AFAIK... it only formats an error - This is what got me confused. It would mean that the error is printed somewhere else

lazartravica: @Vuksan you're absolutely right 🤦 

I probably got confused as I was unable to reproduce this exact issue with `error getting storage snapshot` when deploying the `pos-portal` smartcontracts with truffle, which was occurring 2 weeks ago.

Sorry for the mess, I'll squash & merge after your approval not to keep confusing commit mesages!

Vuksan: And I was wondering what am I not getting 😄 
Cheers! 🍻

Vuksan (approved):

PR #75, opened 2021-06-28
Add validator after account staked in IBFT POS

Kourin1996:
# Description

This PR adds StakingEvent, UnstakingEvent and codes to modifies validator set after the account stake balance.

This PR adds following changes:
* Adds StakingEvent and UnstakingEvent in executor and enable Transition to emit this event when account staked/unstaked balance. In addition, I add subscription function for these events so that consensus layer can subscribe the events.
* Calculate next validator set by current validator set and the events and update snapshot

~~This PR needs the change of #69 and #76, so the status will change to ready for review after they are merged.~~

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


Kourin1996:
Sorry, to trouble you. I just added an e2e test to check if validator leaves after it sends unstake transaction.

state/executor.go
```diff
@@ -27,6 +28,31 @@ type GetHashByNumber = func(i uint64) types.Hash
...
+
+type StakingEvent struct {
+	Number  int64
+	Type    StakingEventType
+	Address types.Address
+	Amount  *big.Int
+}
+
+type StakingEventSubscription struct {
+	EventCh chan *StakingEvent
```
Vuksan: This can be one-directional channel if I'm not mistaken:
```suggestion
	EventCh chan<- *StakingEvent
```

state/executor.go
```diff
@@ -120,6 +151,44 @@ func (e *Executor) StateAt(root types.Hash) (Snapshot, error) {
...
+	for _, s := range e.stakingEventSubscriptions {
+		if s != target {
+			newSubs = append(newSubs, s)
+		}
+	}
+	e.stakingEventSubscriptions = newSubs
+}
+
+// PublishStakingEvent publishes a staking event to subscriptions
+func (e *Executor) PublishStakingEvent(event *StakingEvent) {
```
Vuksan: What do you think of making this method private?

e2e/pos_test.go
```diff
@@ -0,0 +1,144 @@
...
+			return v
+		}
+	}
+	return nil
+}
+
+func TestPoS_Stake(t *testing.T) {
+	signer := &crypto.FrontierSigner{}
+	stakerKey, stakerAddr := framework.GenerateKeyAndAddr(t)
+	stakingContractAddr := types.StringToAddress("1001")
```
zivkovicmilos: We should not be hardcoding these address values anywhere, you can grab them as constants from the system runtime

e2e/pos_test.go
```diff
@@ -0,0 +1,144 @@
...
+	assert.NotNil(t, snapshot)
+
+	validator := findValidatorByAddress(snapshot.Validators, stakerAddr.String())
+	assert.Len(t, snapshot.Validators, numGenesisValidators+1)
+	assert.NotNil(t, validator, "account should have joined to validator set, but doesn't belong to")
+}
+
+func TestPoS_Unstake(t *testing.T) {
+	signer := &crypto.FrontierSigner{}
+	unstakingContractAddr := types.StringToAddress("1002")
```
zivkovicmilos: Same comment as the previous test - Try not to hardcode these address values

e2e/pos_test.go
Kourin1996: Fixed

e2e/pos_test.go
Kourin1996: Fixed

zivkovicmilos (approved):
Clean and detailed as always 💯 
I especially liked the event mechanism, it's an elegant solution.

consensus/ibft/pos.go
```diff
@@ -0,0 +1,160 @@
...
+		i.executor.UnsubscribeStakingEvent(subscription)
+		subscription.WaitForDone()
+		close(subscription.EventCh)
+		return <-resCh
+	}
+}
+
+// getNextValidatorSet returns the validator set for the next
+func (i *Ibft) getNextValidatorSet(header *types.Header, stakingEvents []*state.StakingEvent) (ValidatorSet, error) {
+	transition, err := i.executor.BeginTxn(header.StateRoot, header, header.Miner)
```
Vuksan: Not sure if the coinbase receiver will be in the `header.Miner` field.
If I understood the code correctly, here it isn't important who the coinbase receiver is, since we don't write the transition later? In that case, what do you think of using the empty `address` instead?

consensus/ibft/pos.go
```diff
@@ -0,0 +1,160 @@
...
+
+		header, ok := i.blockchain.GetHeaderByNumber(n)
+		if !ok {
+			return fmt.Errorf("cannot find header at %d", n)
+		}
+		validators, err := i.getNextValidatorSet(header, events)
+		if err != nil {
+			return err
+		}
+		i.updateSnapshotValidators(n, validators)
```
Vuksan: Did you forgot to handle errors that `updateSnapshotValidators` can return?

consensus/ibft/pos.go
```diff
@@ -0,0 +1,160 @@
...
+			i.store.add(newSnap)
+		} else {
+			i.store.replace(newSnap)
+		}
+	}
+	return nil
+}
+
+// bulkUpdateSnapshots updates validator set in multiple snapshots
+// ignore events with greater number than end
```
Vuksan: Just a bit more clarification:
```suggestion
// ignore events with a greater block number than the latest header's block number in memory
```

consensus/ibft/ibft.go
```diff
@@ -331,6 +343,8 @@ func (i *Ibft) runSyncState() {
 			return !isValidator
 		})
 
+		i.bulkUpdateSnapshots(stopSubscription())
```
Vuksan: Unhandled error

consensus/ibft/ibft.go
```diff
@@ -311,10 +311,18 @@ func (i *Ibft) runSyncState() {
...
+		})
+
 		if err := i.syncer.BulkSyncWithPeer(p); err != nil {
 			i.logger.Error("failed to bulk sync", "err", err)
+
+			// Update by the number of the header that has been merged successful
+			i.bulkUpdateSnapshots(stopSubscription())
 			continue
 		}
+		i.bulkUpdateSnapshots(stopSubscription())
```
Vuksan: Unhandled error

consensus/ibft/ibft.go
```diff
@@ -311,10 +311,18 @@ func (i *Ibft) runSyncState() {
...
 
+		stopSubscription := i.SubscribeStakingEvent(func(e *state.StakingEvent) bool {
+			return true
+		})
+
 		if err := i.syncer.BulkSyncWithPeer(p); err != nil {
 			i.logger.Error("failed to bulk sync", "err", err)
+
+			// Update by the number of the header that has been merged successful
+			i.bulkUpdateSnapshots(stopSubscription())
```
Vuksan: Unhandled error

e2e/framework/config.go
```diff
@@ -20,20 +20,26 @@ type SrvAccount struct {
 	StakedBalance *big.Int
 }
 
+type GenesisValidatorBalance struct {
+	Balance       *big.Int
+	StakedBalance *big.Int
+}
```
Vuksan, line 32: Why did you need this new struct?
Can you help me understand why didn't you use `SrvAccount` instead?
It has both balance and staked balance as well...

e2e/framework/testserver.go
```diff
@@ -267,6 +274,19 @@ func (t *TestServer) Start(ctx context.Context) error {
...
+	case ConsensusIBFT:
+		return filepath.Join(t.Config.RootDir, t.Config.IBFTDir)
+	default:
+		return t.Config.RootDir
+	}
+}
+
+func (t *TestServer) PrivateKey() (*ecdsa.PrivateKey, error) {
+	return crypto.ReadPrivKey(filepath.Join(t.DataDir(), "consensus", ibft.IbftKeyName))
+}
```
Vuksan: Can you call this method in other places as well, instead of `crypto.ReadPrivKey(...)`?

Vuksan: What do you think of moving these 2 methods (`DataDir` and `PrivateKey`) to `TestServerConfig`?

e2e/framework/testserver.go
```diff
@@ -388,6 +408,27 @@ func (t *TestServer) WaitForReady(ctx context.Context) error {
 	return err
 }
 
+func (t *TestServer) WaitForIBFTSnapshot(ctx context.Context, number uint64) (*ibftProto.Snapshot, error) {
+	client := t.IBFTOperator()
+	res, err := RetryUntilTimeout(ctx, func() (interface{}, bool) {
+		receipt, _ := client.GetSnapshot(context.Background(), &ibftProto.SnapshotReq{
```
Vuksan: Any advantage of using `context.Background()` instead of `context.WithTimeout`?

e2e/pos_test.go
```diff
@@ -0,0 +1,145 @@
...
+	// Check validator set
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	snapshot, err := srv.WaitForIBFTSnapshot(ctx, receipt.BlockNumber)
+	assert.NoError(t, err)
+	assert.NotNil(t, snapshot)
+
+	validator := findValidatorByAddress(snapshot.Validators, stakerAddr.String())
+	assert.Len(t, snapshot.Validators, numGenesisValidators+1)
+	assert.NotNil(t, validator, "account should have joined to validator set, but doesn't belong to")
```
Vuksan: Wording suggestion:
```suggestion
	assert.NotNil(t, validator, "expected staker to join the validator set")
```

e2e/pos_test.go
```diff
@@ -0,0 +1,145 @@
...
+	data := txn.MarshalRLP()
+	hash, err := srv.JSONRPC().Eth().SendRawTransaction(data)
+	assert.NoError(t, err)
+	assert.NotNil(t, hash)
+
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	receipt, err := srv.WaitForReceipt(ctx, hash)
+	assert.NoError(t, err)
+	assert.NotNil(t, receipt)
```
Vuksan: You can make this simpler by using `testserver.go` `SendRawTx()`

e2e/pos_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Check validator set
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	snapshot, err = srv.WaitForIBFTSnapshot(ctx, receipt.BlockNumber)
+	assert.NoError(t, err)
+	assert.NotNil(t, snapshot)
+
+	validator = findValidatorByAddress(snapshot.Validators, unstakerAddr.String())
+	assert.Nil(t, validator, "account should have left from validator set, but still belongs to")
```
Vuksan:
```suggestion
	assert.Nil(t, validator, "account should have left from validator set, but still belongs to it")
```

e2e/pos_test.go
```diff
@@ -0,0 +1,145 @@
...
+	data := txn.MarshalRLP()
+	hash, err := srv.JSONRPC().Eth().SendRawTransaction(data)
+	assert.NoError(t, err)
+	assert.NotNil(t, hash)
+
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	receipt, err := srv.WaitForReceipt(ctx, hash)
+	assert.NoError(t, err)
+	assert.NotNil(t, receipt)
```
Vuksan: You can make this simpler by using `testserver.go` `SendRawTx()`

Vuksan (changes requested):
Everything looks good overall, gj on the event subscription mechanism. 👏 
My comments are mostly minor suggestions.

e2e/framework/testserver.go
Kourin1996: I changed it to context with timeout. It's also canceled by parent context given in args.

e2e/framework/config.go
Kourin1996: The main reason is we don't know the validator address server runs until it executes `ibft init` command. (after config callback is called) Just can change type of GenesisValidatorBalance to `*SrvAccount` but the `Addr` won't be used for this.

consensus/ibft/pos.go
Kourin1996: Yes, you're correct. No transaction will be not applied to this transition here and just use for reading state. So you can give any address as coinbase receiver. I changed it to zero address

e2e/pos_test.go
Kourin1996: Thanks for good suggestion!

state/executor.go
Kourin1996: Can't change the type to one-directional channel here.
Instead of this, I made this field private and add method `EventCh` which returns receive only channel. Other package won't be able to send events into this channel. Accordingly, I change `WaitForDone` method to `Close`. It closes own channel after all transmissions finish.

Kourin1996:
@Vuksan I've fixed the codes based on your suggestions. I confirmed all your comments, I wrote a reply in the comment about GenesisValidatorBalance. Could you check this?

state/executor.go
Vuksan: Oh, I see. You didn't have to do that, I thought that it could be made one-directional, but as it seems I was mistaken. 🙂

Vuksan (approved):
LGTM

lazartravica (approved):
LGTM 👍

PR #76, opened 2021-06-28
IBFT PoS Unstaking

zivkovicmilos:
# Description

This PR introduces the unstaking handler for the PoS implementation, while modifying Account state.

# Changes include

# Breaking changes

As stated in the PR #69, the current implementation modifies parts of the codebase that pertain to how Account state is kept and marshaled in the world trie.

# Checklist

- [x] I have tested this code


zivkovicmilos:
@lazartravica @Vuksan @Kourin1996 
The only difference that exists between this and the ibft-pos-staking branch is in [this commit](https://github.com/0xPolygon/polygon-sdk/pull/76/commits/2a9140f4240e6c3e3cc81eb6f6a9bc626a6c6752).

There is no need to check other files again, just the linked commit is fine.

command/genesis.go
```diff
@@ -252,6 +264,40 @@ func (c *GenesisCommand) Run(args []string) int {
 		}
 	}
 
+	if len(prestake) != 0 {
```
Kourin1996: We need to check if prestake is not empty here? because you just run for-loop inside if block and for-loop will never run any loop if prestake is nil or empty slice.

zivkovicmilos: You're actually right. I'll remove it, along with the same code block for `premine` 👍

state/runtime/system/system_unstaking.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Sanity check
+	// Can't unstake if the balance isn't previously present on the unstaking account
+	// Can't unstake if the value is different from 0
+	if unstakingAccountBalance.Cmp(stakedBalance) < 0 || state.contract.Value.Cmp(big.NewInt(0)) != 0 {
+		return nil, nil
+	}
+
+	// Decrease the staked balance on the unstaking address
+	state.host.SubBalance(unstakingAddress, stakedBalance)
```
Kourin1996: Just curious question, I remember you didn't change the balance in system contract when an account stake balance. 
1. Is it correct behaviour that you reduce the total staked balance from system contract?
2. Is unstakingAddress the correct address to reduce balance?

zivkovicmilos: The reason we modified the balance for the `unstaking address`, as well as the staked balance in the system contract here, is because this event (`unstaking`) is triggered by sending a transaction to the `unstaking address` with a value of **`0`**.

We didn't need to modify it directly like this in the staking handler, because the flow that previously executes in `executor.go` (specifically the `transfer` method) would deduce the staker's balance, and increase the staking address balance.
This just doesn't happen with unstaking, because the value of the transaction is 0.

Kourin1996: Understood. Sorry, it's my misunderstanding about first question. Yes, transfer is processed by original flow. When a user wants to stake, the user sends to "1001" (and staking event happens on process). So staked amout is saved to "1001" (, which is not unstaking address). It it correct?

zivkovicmilos: I understand what you mean now. I subtracted the balance from the wrong address. It should subtract the balance from `staking address` (1001) which holds the funds, not `unstaking address` (1002).
Fixing now 👍

command/genesis.go
```diff
@@ -228,26 +241,55 @@ func (c *GenesisCommand) Run(args []string) int {
...
+			c.UI.Error(fmt.Sprintf("failed to parse stake amount %s: %v", val, err))
+			return 1
+		}
+
+		previousAccount := cc.Genesis.Alloc[addr]
+
+		if previousAccount != nil {
+			// Account already has a premined balance
+			previousAccount.StakedBalance = stakeAmount
+			cc.Genesis.Alloc[addr] = previousAccount
```
Vuksan: Maybe this line is not needed since `previousAccount` is a pointer to the `GenesisAccount`?
```suggestion
			previousAccount.StakedBalance = stakeAmount
```

state/runtime/system/system.go
```diff
@@ -22,9 +22,22 @@ type systemContract interface {
 	run(state *systemState) ([]byte, error)
 }
 
+// operationsAddressMap defines an operation -> address (string) mapping
+var operationsMap = map[string]string{
+	"staking":   "1001",
+	"unstaking": "1002",
```
Vuksan: What do you think of adding constants for `"staking"` and `"unstaking"`?

Kourin1996 (approved):
LGTM 👍

state/executor.go
```diff
@@ -49,10 +50,17 @@ func (e *Executor) WriteGenesis(alloc map[types.Address]*chain.GenesisAccount) t
...
+	unstakingAddress := types.StringToAddress(system.GetOperationsMap()["unstaking"])
+
 	for addr, account := range alloc {
 		if account.Balance != nil {
 			txn.AddBalance(addr, account.Balance)
 		}
+		if account.StakedBalance != nil {
+			txn.AddStakedBalance(addr, account.StakedBalance)
+
+			txn.AddBalance(unstakingAddress, account.StakedBalance)
```
Vuksan: I can't understand why staked balance is added to the unstaking address? 😕

state/runtime/system/system_unstaking.go
```diff
@@ -0,0 +1,53 @@
...
+	// Grab the staking address
+	stakingAddress := types.StringToAddress(GetOperationsMap()["staking"])
+
+	// Grab the address calling the staking method
+	staker := state.contract.Caller
+
+	// Grab the staked balance for the account
+	stakedBalance := state.host.GetStakedBalance(staker)
+
+	// Grab the balance on the unstaking address
```
Vuksan:
```suggestion
	// Grab the balance on the staking address
```

state/runtime/system/system_unstaking.go
```diff
@@ -0,0 +1,53 @@
...
+	stakingAccountBalance := state.host.GetBalance(stakingAddress)
+
+	// Sanity check
+	// Can't unstake if the balance isn't previously present on the unstaking account
```
Vuksan:
```suggestion
	// Can't unstake if the balance isn't previously present on the staking account
```

state/runtime/system/system_unstaking.go
```diff
@@ -0,0 +1,53 @@
...
+	// Can't unstake if the value is different from 0
+	if stakingAccountBalance.Cmp(stakedBalance) < 0 || state.contract.Value.Cmp(big.NewInt(0)) != 0 {
+		return nil, nil
```
Vuksan: Shouldn't an error be returned in this case?

state/runtime/system/system_unstaking.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	// Decrease the staked balance on the staking address
+	state.host.SubBalance(stakingAddress, stakedBalance)
+
+	// Decrease the staked amount from the account's staked balance
+	state.host.SubStakedBalance(staker, stakedBalance)
```
Vuksan: Just revert these 2 calls. Probably won't make a difference but in SC development it is a good practice to reduce the caller's balance first.
```suggestion
	// Decrease the staked amount from the account's staked balance
	state.host.SubStakedBalance(staker, stakedBalance)

	// Decrease the staked balance on the staking address
	state.host.SubBalance(stakingAddress, stakedBalance)
```

state/txn_test.go
```diff
@@ -100,7 +100,8 @@ func buildMockPreState(p *PreState) (*Account, *mockSnapshot) {
 	account := &Account{
 		Nonce:   p.Nonce,
 		Balance: big.NewInt(int64(p.Balance)),
-		Root:    root,
+		// TODO Mock the Staked balance
+		Root: root,
```
Vuksan: This `TODO` is very persistent 😄 
```suggestion
		Root: root,
```

tests/testing.go
```diff
@@ -204,6 +204,7 @@ func buildState(t *testing.T, allocs map[types.Address]*chain.GenesisAccount) (s
 		txn.CreateAccount(addr)
 		txn.SetNonce(addr, alloc.Nonce)
 		txn.SetBalance(addr, alloc.Balance)
+		//txn.SetStakedBalance(addr, big.NewInt(0))
```
Vuksan: Leftover from the merge I guess
```suggestion

```

tests/state_test.go
```diff
@@ -31,6 +35,105 @@ type stateCase struct {
...
+
+		// Add the legacy account to the local trie
+		trieTransaction.Insert(hashIt(accountAddress.Bytes()), data)
+		arena.Reset()
+	}
+
+	newRoot, _ := trieTransaction.Hash()
+
+	return newRoot
+}
```
Vuksan: Leftover from the merge?

e2e/system_test.go
```diff
@@ -199,3 +200,182 @@ func TestSystem_StakeAmount(t *testing.T) {
...
+		// Valid account with stake
+		{
+			addressKeyPairs[0].address,
+			framework.EthToWei(50), // 50 ETH
+			framework.EthToWei(10), // 10 ETH
+		},
+		// Valid account without stake
+		{
+			addressKeyPairs[1].address,
+			big.NewInt(0),
```
Vuksan: For consistency 🙂 
```suggestion
			framework.EthToWei(0)
```

e2e/system_test.go
```diff
@@ -199,3 +200,182 @@ func TestSystem_StakeAmount(t *testing.T) {
...
+			)
+			assert.NoError(t, err)
+
+			var out string
+			if callErr := rpcClient.Call(
+				"stake_getStakedBalance",
+				&out,
+				web3.Address(testCase.staker),
+				web3.Latest.String(),
+			); callErr != nil && testCase.shouldSucceed {
```
Vuksan: I guess this isn't needed?
```suggestion
			); callErr != nil {
```

e2e/system_test.go
```diff
@@ -199,3 +200,182 @@ func TestSystem_StakeAmount(t *testing.T) {
...
+				web3.Latest,
+			)
+			assert.NoError(t, err)
+
+			if callErr := rpcClient.Call(
+				"stake_getStakedBalance",
+				&out,
+				web3.Address(testCase.staker),
+				web3.Latest.String(),
+			); callErr != nil && testCase.shouldSucceed {
```
Vuksan:
```suggestion
			); callErr != nil {
```

e2e/system_test.go
```diff
@@ -199,3 +200,182 @@ func TestSystem_StakeAmount(t *testing.T) {
 		})
 	}
 }
+
+func TestSystem_UnstakeAmount(t *testing.T) {
```
Vuksan, line 198: It seems that you can extract a helper method for this test and `TestSystem_StakeAmount`. Looks like some code is duplicated.

Vuksan (changes requested):

command/genesis.go
zivkovicmilos: Agreed. Removing now 👍

state/runtime/system/system.go
zivkovicmilos: I agree it looks a lot cleaner.
Changing it now 👍

state/executor.go
zivkovicmilos: This was done just to simulate a transaction execution.
When a user is staking something, they include the staked amount in the value field of the transaction.
The account at the staking address is incremented by this amount.

We can completely ignore the balances, both on the staking and unstaking address.
What do you think?

state/runtime/system/system_unstaking.go
zivkovicmilos: Fixed 👍

state/runtime/system/system_unstaking.go
zivkovicmilos: Fixed 👍

state/runtime/system/system_unstaking.go
zivkovicmilos: Fixed 👍

state/runtime/system/system_unstaking.go
zivkovicmilos: Fixed 👍

state/txn_test.go
zivkovicmilos: Fixed 👍

tests/testing.go
zivkovicmilos: Yup, removing now 👍

tests/state_test.go
zivkovicmilos: Yes.
Removing it now 👍

e2e/system_test.go
zivkovicmilos: Agreed. Fixed 👍

e2e/system_test.go
zivkovicmilos: Leftover from the merge.
Removed 👍

e2e/system_test.go
zivkovicmilos: Removed 👍

e2e/system_test.go
zivkovicmilos: Agreed 👍 
I extracted out two useful methods:
`getAccountBalance` and `getStakedBalance` to avoid code duplication.
Also, I've extracted out the transaction object construction as it can be parametrized

state/executor.go
Vuksan: `The account at the staking address is incremented by this amount.`

This makes sense, but on this line you have incremented the account at the **unstaking address**

e2e/system_test.go
```diff
@@ -32,11 +33,81 @@ func generateAddressKeyPairs(num int, t *testing.T) []*addressKeyPair {
...
+	}
+
+	signedTxn, err := signer.SignTx(txnObject, params.signingKey)
+	assert.NoError(params.t, err)
+
+	data := signedTxn.MarshalRLP()
+
+	// Do the transfer
+	return params.rpcClient.Eth().SendRawTransaction(data)
+}
```
Vuksan: This method already exists in testserver.go L321. You can probably reuse that one? 🙂

Vuksan (changes requested):

state/executor.go
zivkovicmilos: You're completely right. I put in the wrong address.
Sorry about that. Fixed now 👍

e2e/system_test.go
zivkovicmilos: Got it, didn't catch it earlier.
Refactored to use the existing method 👍

Vuksan (approved):

PR #78, opened 2021-06-30
Dev command

zivkovicmilos:
# Description

This PR restructures the commands module a bit, and implements the base dev command

# Changes include

# Checklist

- [x] I have tested this code


command/dev/dev_command.go
```diff
@@ -0,0 +1,102 @@
...
+	config, err := conf.BuildConfig()
+	if err != nil {
+		d.UI.Error(err.Error())
+
+		return 1
+	}
+
+	logger := hclog.New(&hclog.LoggerOptions{
+		Name:  "polygon-dev",
+		Level: hclog.LevelFromString("debug"),
```
Kourin1996: I believe you should pass log level from flag value.

zivkovicmilos: Good catch.
Fixed 👍

e2e/framework/testserver.go
```diff
@@ -16,7 +15,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/0xPolygon/minimal/command"
+	"github.com/0xPolygon/minimal/command/genesis"
+	ibft2 "github.com/0xPolygon/minimal/command/ibft"
```
lazartravica: Please call it something among the lines of ibftCommand, not ibft2

main.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/0xPolygon/minimal/command"
+	"github.com/0xPolygon/minimal/command/util"
```
lazartravica, line 7: Feel free to remove the comments at the bottom of this file

command/helper/helper.go
```diff
@@ -143,3 +171,310 @@ func GenerateUsage(baseCommand string, flagMap map[string]FlagDescriptor) string
...
+	DefineFlags()
+}
+
+// Meta is a helper utility for the commands
+type Meta struct {
+	UI   cli.Ui
+	Addr string
+
+	FlagMap        map[string]FlagDescriptor
+	HasGlobalFlags bool
```
lazartravica: HasGlobalFlags is only assigned to true in the below method and that's it. We can remove it.

lazartravica (changes requested):
Great job! All is good in your implementation, just a few  minor comments so that we can clean up some more around the comments.

command/helper/helper.go
```diff
@@ -143,3 +171,310 @@ func GenerateUsage(baseCommand string, flagMap map[string]FlagDescriptor) string
...
+		}
+	}
+	if !os.IsNotExist(err) {
+		return &GenesisGenError{
+			message:   fmt.Sprintf("genesis file at path (%s) already exists", genesisPath),
+			errorType: ExistsError,
+		}
+	}
+
+	return nil
```
Vuksan: What's specific about Genesis here? Why not `DoesFileExist`?

command/peers/peers_add.go
```diff
@@ -57,7 +57,7 @@ func (p *PeersAdd) Run(args []string) int {
 	flags := p.FlagSet(p.GetBaseCommand())
 
 	var passedInAddresses = make(helperFlags.ArrayFlags, 0)
-	flags.Var(&passedInAddresses, "addr", "")
+	flags.Var(&passedInAddresses, "Addr", "")
```
Vuksan: Did you make this parameter capital by accident?
```suggestion
	flags.Var(&passedInAddresses, "addr", "")
```

Vuksan (approved):
LGTM 👍 
I left two comments, but nothing major really.

e2e/framework/testserver.go
zivkovicmilos: Got it 👍 
Fixed now

main.go
zivkovicmilos: Removed 👍

command/helper/helper.go
zivkovicmilos: I agree. It was part of an earlier implementation that required a different structure for `Meta`, but it's useless now 👍

command/helper/helper.go
zivkovicmilos: It throws genesis specific errors and is meant to be used only when handling genesis generation.
We have other helper methods for checking if a file at a path exists.

I've renamed it to something more appropriate now 👍

command/peers/peers_add.go
zivkovicmilos: I do all my find-and-replaces manually, and IntelliJ still somehow manages to sneak in a rogue replace.
Fixed here, and in other places it changed 👍

command/genesis/genesis.go
```diff
@@ -18,78 +17,70 @@ import (
...
-	if len(c.flagMap) > 0 {
+	if len(c.FlagMap) > 0 {
 		// No need to redefine the flags again
 		return
 	}
 
-	c.flagMap["data-dir"] = helper.FlagDescriptor{
-		Description: fmt.Sprintf("Sets the directory for the Polygon SDK data. Default: %s", genesisFileName),
+	c.FlagMap["data-dir"] = helper.FlagDescriptor{
+		Description: fmt.Sprintf("Sets the directory for the Polygon SDK genesis data. Default: %s", helper.GenesisFileName),
```
Kourin1996, line 39: Expects user to give directory path for `data-dir`, but description says its default is `./genesis.json`. It's not directory.

Kourin1996 (commented):
I added a comment about small point. Please check it.
And I realized there's still a fixed log level
https://github.com/0xPolygon/polygon-sdk/blob/74ab26731fef271be4982072208cdfe61d8c7b67/command/server/server_command.go#L184

Looks good to me, apart from them.

command/genesis/genesis.go
zivkovicmilos: Got it. I'll change the flag to something more appropriate since this does actually point just to the genesis file, and not any directory 👍

zivkovicmilos: Fixed 👍

Kourin1996 (approved):
LGTM 👍 Thanks for the fixing.

command/genesis/genesis.go
```diff
@@ -18,78 +17,70 @@ import (
...
-	if len(c.flagMap) > 0 {
+	if len(c.FlagMap) > 0 {
 		// No need to redefine the flags again
 		return
 	}
 
-	c.flagMap["data-dir"] = helper.FlagDescriptor{
-		Description: fmt.Sprintf("Sets the directory for the Polygon SDK data. Default: %s", genesisFileName),
+	c.FlagMap["dir"] = helper.FlagDescriptor{
+		Description: fmt.Sprintf("Sets the directory for the Polygon SDK genesis data. Default: ./%s", helper.GenesisFileName),
```
Vuksan: You don't need `./` since `GenesisFileName` is already `./genesis.json`, so it outputs `Default: ././genesis.json`
```suggestion
		Description: fmt.Sprintf("Sets the directory for the Polygon SDK genesis data. Default: %s", helper.GenesisFileName),
```

Vuksan (approved):
GJ
I left a minor, cosmetic suggestion

command/genesis/genesis.go
zivkovicmilos: Fixed 👍

main.go
RauleHernandeza: where can i find the library github.com/0xPolygon/minimal/command/util? i want to check the functions it has so i would appreciate if you could pass the link as i can't find it in go.dev or in the old github repository.

Vuksan: > where can i find the library github.com/0xPolygon/minimal/command/util? i want to check the functions it has so i would appreciate if you could pass the link as i can't find it in go.dev or in the old github repository.

Here it is: https://github.com/0xPolygon/polygon-sdk/blob/develop/command/util/commands.go

PR #79, opened 2021-07-02
Disable vote function in IBFT PoS

Kourin1996:
# Description

This PR disables vote function to invite/kick a validator in PoS

# Changes include

# Breaking changes

Disable `ibft propose` and `ibft candidates` commands

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/consensus/ibft/snapshot.go
+++ b/consensus/ibft/snapshot.go
@@ -193,12 +193,12 @@ func (i *Ibft) processHeaders(headers []*types.Header) error {
 
 		// Disable voting temporarily for PoS
 		// if err := i.applyVote(h, proposer, snap); err != nil {
-		// 	return nil
+		// 	return err
 		// }
 
 		if !snap.Equal(parentSnap) {
 			if err := saveSnap(h); err != nil {
-				return nil
+				return err
 			}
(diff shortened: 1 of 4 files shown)


zivkovicmilos (approved):
Looks good 💯

consensus/ibft/snapshot.go
```diff
@@ -191,87 +191,98 @@ func (i *Ibft) processHeaders(headers []*types.Header) error {
...
-			if snap.Set.Includes(h.Miner) {
-				continue
-			}
-		} else {
-			// we can only remove if they are part of the validators list
-			if !snap.Set.Includes(h.Miner) {
-				continue
+		if !snap.Equal(parentSnap) {
+			if err := saveSnap(h); err != nil {
+				return nil
```
Vuksan: Isn't it strange that we return `nil` here if there is error, and not `err`?

Vuksan (approved):
LGTM

consensus/ibft/snapshot.go
Kourin1996: I agree with you. I fixed the code

PR #84, opened 2021-07-12
IBFT PoS v2

zivkovicmilos:
# Description

This PR reverts the Account state back to its original form, while still maintaining the staking / unstaking functionality.

# Changes include

# Breaking changes

Stated in the initial paragraph.

# Checklist

- [x] I have tested this code

# Additional comments

## Staking Hub

The Staking Hub is the central place from which different modules from Polygon SDK can interact with the staking tally.
Currently, the staking tally (map) is kept in memory, and is written back to disk in JSON format for a period specified in `WriteBackPeriod` (default is 60s).

## Earlier staking / unstaking flow
Initial

## Current staking / unstaking flow
Modified


Kourin1996:
Just quick comment. We don't need event subscription in consensus layer if consensus can trust data in staking hub 100% and staking hub can return list of stakers fast (O(1) or O(n)). In PoS version1, there was no way to get current stakers. Traversing all accounts to check staking amount was unrealistic. But now, we can have the list and consensus can get stakers directly. I'm not sure subscription will be used in the future but it's not necessary for now.

What do you think about?

staking/staking_hub_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+	for indx, testCase := range testTable {
+		t.Run(testCase.name, func(t *testing.T) {
+			instances := make([]*StakingHub, testCase.numInstances)
+
+			for i := 0; i < testCase.numInstances; i++ {
+				instances[i] = GetStakingHub()
+			}
+
+			for i := 0; i < testCase.numInstances; i++ {
```
lazartravica: Why the additional for loop?

jsonrpc/stake_endpoint.go
```diff
@@ -17,11 +18,13 @@ func (s *Stake) GetStakedBalance(address types.Address, number BlockNumber) (int
 		return nil, err
 	}
 
-	acc, err := s.d.store.GetAccount(header.StateRoot, address)
+	_, err = s.d.store.GetAccount(header.StateRoot, address)
```
lazartravica, line 21: How am I able to interface with the GetStakedBalance endpoint?

lazartravica (changes requested):
Please take a look at @Kourin1996 's comment.

It also seems to me like we only need the event publishing / subscription between the Staking Hub and the Executor and not between Consensus and Staking Hub.

staking/staking_hub_test.go
zivkovicmilos: Good catch. It probably contained additional logic that got swapped out later.
Fixing now 👍

jsonrpc/stake_endpoint.go
zivkovicmilos: What do you mean?
All of the stake endpoints are registered with the keyword `stake`, so if you wanted to shoot an http request to get staked balance, you would curl:
`stake_getStakedBalance` with parameters that match the same ones as for `eth_getBalance`

zivkovicmilos:
@Kourin1996 @lazartravica 
I've tried to figure out if we can ditch the staking / unstaking event mechanism entirely now that we have the staking hub, but the only thing that's bugging me is that if we do that - we completely lose the ability to have a valid snapshot store state sync on a per block basis:
https://github.com/0xPolygon/polygon-sdk/blob/3ddf382c4c80cfcb6c02c1682b87fc754c427e76/consensus/ibft/pos.go#L111

https://github.com/0xPolygon/polygon-sdk/blob/3ddf382c4c80cfcb6c02c1682b87fc754c427e76/consensus/ibft/ibft.go#L319

These events are not just used to update the current validator set when the chain is synced up, but also to update the snapshot store for each individual block that we sync (block A has validators B, C, block D has validators E, F...). Moreover, they depend on the staking / unstaking events that occurred _during that specific block_.

This is a problem because the Staking Hub has no notion of _past_ staking / unstaking events. The pending events that we do have are cleared out as soon as the transaction is executed correctly - with the only thing left being the staking map, which doesn't have a notion of block numbers. We use `PendingEvent`s to keep track of possible rollbacks and pending stakes / unstakes in case a block fails or a transaction is faulty.

I'm probably missing something since you've both come to the same conclusion, so please let me know if the above mentioned is a valid concern.

jsonrpc/stake_endpoint.go
lazartravica: I meant exactly that, thanks Milos! 🙏

state/executor.go
```diff
@@ -123,6 +129,82 @@ type BlockResult struct {
...
+				discardedTxn.From,
+				discardedTxn.Value,
+				blockNumber,
+			)
+
+			hub := staking.GetStakingHub()
+			if hub.ContainsPendingEvent(pendingEvent) {
+				// Remove the event from the queue
+				hub.RemovePendingEvent(pendingEvent)
+			}
```
Kourin1996: We traverse EventQueue twice here. I think we can call only RemovePendingEvent here, or RemovePendingEvent can return `deleted` flag for other code (it's useful the code in commitApprovedEvents)

Kourin1996:
@lazartravica @zivkovicmilos's opinion makes sense for me. There are 2 facts (1) StakingHub holds only staked balances at latest height (2) IBFT processes blocks in bulk in sync-state, we can't do it one by one (unless we fix the code in BulkSyncWithPeer). 

So we still need the event subscription in order to generate past snapshots (between the heights before and after sync) But we need to fix bulkUpdateSnapshots because it's referring staked balances in staking hub.

lazartravica:
> @Kourin1996 @lazartravica
> I've tried to figure out if we can ditch the staking / unstaking event mechanism entirely now that we have the staking hub, but the only thing that's bugging me is that if we do that - we completely lose the ability to have a valid snapshot store state sync on a per block basis:
> 
> https://github.com/0xPolygon/polygon-sdk/blob/3ddf382c4c80cfcb6c02c1682b87fc754c427e76/consensus/ibft/pos.go#L111
> 
> https://github.com/0xPolygon/polygon-sdk/blob/3ddf382c4c80cfcb6c02c1682b87fc754c427e76/consensus/ibft/ibft.go#L319
> 
> These events are not just used to update the current validator set when the chain is synced up, but also to update the snapshot store for each individual block that we sync (block A has validators B, C, block D has validators E, F...). Moreover, they depend on the staking / unstaking events that occurred _during that specific block_.
> 
> This is a problem because the Staking Hub has no notion of _past_ staking / unstaking events. The pending events that we do have are cleared out as soon as the transaction is executed correctly - with the only thing left being the staking map, which doesn't have a notion of block numbers. We use `PendingEvent`s to keep track of possible rollbacks and pending stakes / unstakes in case a block fails or a transaction is faulty.
> 
> I'm probably missing something since you've both come to the same conclusion, so please let me know if the above mentioned is a valid concern.

I have overlooked the implications given by syncing numerous blocks from peers, you're absolutely right, I don't think a different solution to this issue would be more elegant.

Approving the pull request 💯

lazartravica (approved):

state/executor.go
zivkovicmilos: You're right, there is no need to traverse it twice. Removing the double traversal 👍

state/executor.go
```diff
@@ -677,15 +764,18 @@ func (t *Transition) SubBalance(addr types.Address, balance *big.Int) {
...
-	return t.state.GetStakedBalance(addr)
+	hub := staking.GetStakingHub()
+	return hub.GetStakedBalance(addr)
 }
 
 func (t *Transition) AddStakedBalance(addr types.Address, balance *big.Int) {
-	t.state.AddStakedBalance(addr, balance)
+	hub := staking.GetStakingHub()
+	hub.IncreaseStake(addr, balance)
 }
```
Vuksan: You can remove this method since it's unused.

state/executor.go
```diff
@@ -677,15 +764,18 @@ func (t *Transition) SubBalance(addr types.Address, balance *big.Int) {
...
 
 func (t *Transition) SubStakedBalance(addr types.Address, balance *big.Int) {
-	t.state.SubStakedBalance(addr, balance)
+	hub := staking.GetStakingHub()
+	hub.DecreaseStake(addr, balance)
 }
```
Vuksan: You can remove this method since it's unused.

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+
+	// Alert the closing channel
+	sh.CloseCh <- struct{}{}
+
+	close(sh.CloseCh)
+}
+
+// PendingEvent contains useful information about a staking / unstaking event
+type PendingEvent struct {
+	Number    int64
```
Vuksan: Can you please rename this to `BlockNumber`?
It took me a while to understand what we mean by number in transaction context. It is misnamed there as well, let's not repeat the same mistake here 🙂

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+
+// readFromDisk reads the staking map from a previously saved disk copy, if it exists
+func (sh *StakingHub) readFromDisk() error {
+	// Check if the file exists
+	mapFile, err := os.Open(filepath.Join(sh.WorkingDirectory, defaultFileName))
+	if err != nil {
+		return fmt.Errorf("no exiting map file is present")
+	}
+
+	byteValue, _ := ioutil.ReadAll(mapFile)
```
Vuksan: Why don't you handle the error here?

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// StakerMapping is a representation of a staked account balance
+type StakerMapping struct {
+	Address types.Address `json:"address"`
+	Stake   *big.Int      `json:"stake"`
+}
+
+// GetStakerMappings returns the staking addresses and their staking balances
+func (sh *StakingHub) GetStakerMappings() []StakerMapping {
```
Vuksan: This can be a private method since `StakerMapping` is only really used in this file?
I think that you can make `StakerMapping` private as well.

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+		sh.StakingMap[val.Address] = val.Stake
+	}
+
+	return nil
+}
+
+// marshalJSON generates the json object for staker mappings
+func (sh *StakingHub) marshalJSON(mappings []StakerMapping) (io.Reader, error) {
+	sh.StakingMutex.Lock()
+	defer sh.StakingMutex.Unlock()
```
Vuksan: Do we really need a mutex lock on this method?
It seems like it's only working on the local `mappings` list.

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+
+		// Save the current staking map to disk, in JSON
+		mappings := sh.GetStakerMappings()
+
+		reader, err := sh.marshalJSON(mappings)
+		if err != nil {
+			continue
+		}
+
+		// Save the json to workingDirectory/StakingMap.json
```
Vuksan:
```suggestion
		// Save the json to workingDirectory/staking-map.json
```

chain/chain.go
```diff
@@ -226,21 +250,27 @@ func (g *Genesis) UnmarshalJSON(data []byte) error {
...
-	Storage       map[types.Hash]types.Hash `json:"storage,omitempty"`
-	Balance       *big.Int                  `json:"balance,omitempty"`
-	StakedBalance *big.Int                  `json:"stakedBalance,omitempty"` // Staking implementation
-	Nonce         uint64                    `json:"nonce,omitempty"`
-	PrivateKey    []byte                    `json:"secretKey,omitempty"` // for tests
+	Code       []byte                    `json:"code,omitempty"`
+	Storage    map[types.Hash]types.Hash `json:"storage,omitempty"`
+	Balance    *big.Int                  `json:"balance,omitempty"`
+	Nonce      uint64                    `json:"nonce,omitempty"`
+	PrivateKey []byte                    `json:"secretKey,omitempty"` // for tests
```
Vuksan, line 257: Why didn't you keep `StakedBalance` in the `GenesisAccount`?
I understand GenesisAccount is only there to load the data from genesis.json, which means It doesn't have to comply 100% with the Account structure, right?

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+		return nil, err
+	}
+	return bytes.NewReader(b), nil
+}
+
+// isStaker is a helper method to check whether or not an address has a staked balance
+func (sh *StakingHub) isStaker(address types.Address) bool {
+	if _, ok := sh.StakingMap[address]; ok {
+		return true
+	}
```
Vuksan: Isn't it dangerous to only check if address is in the map like this?
As far as I see, increasing stake and decreasing it by the same balance will still leave the address in the staking map resulting in this method returning true

state/executor.go
Vuksan: You can also remove `hub.DecreaseStake()` as well since it's only used here.

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+
+	if sh.isStaker(address) {
+		return sh.StakingMap[address]
+	}
+
+	return big.NewInt(0)
+}
+
+// GetStakerAddresses returns a list of all addresses that have a stake > 0
+func (sh *StakingHub) GetStakerAddresses() []types.Address {
```
Vuksan: You can remove this method since it's unused.

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+		if el.Compare(event) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// ContainsPendingEvent checks if an identical event is present in the queue
+func (sh *StakingHub) ContainsPendingEvent(event PendingEvent) bool {
```
Vuksan: You can remove this method since it's unused.

state/executor.go
```diff
@@ -123,6 +129,79 @@ type BlockResult struct {
 	TotalGas uint64
 }
 
+// getStakingEventType is a helper method for getting the type of system event
+func getStakingEventType(toAddress types.Address) staking.StakingEventType {
+	if toAddress == types.StringToAddress(system.StakingAddress) {
+		return staking.StakingEvent
+	}
+
+	return staking.UnstakingEvent
```
Vuksan: I think that this method (by itself) shouldn't assume that anything that's passed into it is either staking or unstaking address. It should check if address is unstaking one, and if not return either error or something like "unknown event"

staking/staking_hub.go
```diff
@@ -0,0 +1,394 @@
...
+			foundIndx = indx
+		}
+	}
+
+	if foundIndx >= 0 {
+		sh.EventQueue = append(sh.EventQueue[:foundIndx], sh.EventQueue[foundIndx+1:]...)
+		foundFlag = true
+	}
+
+	return foundFlag
```
Vuksan: I think you can do without `foundFlag`
```suggestion
	if foundIndx >= 0 {
		sh.EventQueue = append(sh.EventQueue[:foundIndx], sh.EventQueue[foundIndx+1:]...)
		return true
	}

	return false
```

state/executor.go
```diff
@@ -123,6 +129,79 @@ type BlockResult struct {
...
+				*t.To,
+				t.From,
+				t.Value,
+				blockNumber,
+			)
+
+			hub := staking.GetStakingHub()
+
+			// Remove the event from the queue if it is present
+			if hub.RemovePendingEvent(pendingEvent) {
```
Vuksan: Maybe log an error or warning if this returns false here?

state/executor.go
```diff
@@ -133,11 +212,19 @@ func (e *Executor) ProcessBlock(parentRoot types.Hash, block *types.Block, block
...
+			// since every transaction in this block is discarded
+			cleanDiscardedEvents(block.Transactions, block.Number())
+
 			return nil, err
 		}
 	}
 	_, root := txn.Commit()
 
+	// Checks are passed, do staking logic if possible
+	commitApprovedEvents(block.Transactions, block.Number())
```
Vuksan: If `commitApprovedEvents` is called after `txn.Commit()` then I'm afraid this attack might be possible:
1. I send n unstaking transactions to be mined in a single block
2. I make sure that the sum of my transactions isn't greater than the staked amount of the whole network
3. I receive n * myStakedBalance to my account.

In the end `commitApprovedEvents` will just have n calls to `ResetStake` which will pass.

consensus/ibft/ibft.go
```diff
@@ -67,6 +68,9 @@ type Ibft struct {
 
 	// aux test methods
 	forceTimeoutCh bool
+
+	// Staking hub used for interacting with other modules
+	stakingHub *staking.StakingHub
```
Vuksan: I don't see this used anywhere, but maybe I'm mistaken...

staking/staking_hub_test.go
```diff
@@ -0,0 +1,262 @@
...
+				dummyDir := "dummyDir"
+				instances[0].SetWorkingDirectory(dummyDir)
+
+				for i := 1; i < testCase.numInstances; i++ {
+					assert.Equalf(t,
+						instances[0].WorkingDirectory,
+						instances[i].WorkingDirectory,
+						"Staking hub instances not equal",
+					)
+				}
```
Vuksan, line 63: Neat. This one made me smile 🙂 
A clever way to test this out.

staking/staking_hub_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+	"github.com/0xPolygon/minimal/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func EthToWei(ethValue int64) *big.Int {
+	return new(big.Int).Mul(
+		big.NewInt(ethValue),
+		new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil))
+}
```
Vuksan: I guess you couldn't import the one from `e2e/framework/helper.go`?

Vuksan: Maybe pull that into some outer directory, so that we don't have two methods with the same code?

staking/staking_hub_test.go
```diff
@@ -0,0 +1,262 @@
...
+	validStakes := []struct {
+		address        types.Address
+		initialStake   *big.Int
+		increaseAmount *big.Int
+	}{
+		// Valid account #1
+		{
+			types.StringToAddress("123"),
+			EthToWei(50), // 50 ETH
+			EthToWei(5),  // 50 ETH
```
Vuksan:
```suggestion
			EthToWei(5),  // 5 ETH
```

staking/staking_hub_test.go
```diff
@@ -0,0 +1,262 @@
...
+	validStakes := []struct {
+		address        types.Address
+		initialStake   *big.Int
+		decreaseAmount *big.Int
+	}{
+		// Valid account #1
+		{
+			types.StringToAddress("123"),
+			EthToWei(50), // 50 ETH
+			EthToWei(5),  // 50 ETH
```
Vuksan:
```suggestion
			EthToWei(5),  // 5 ETH
```

state/executor.go
Vuksan: If this turns out to be true, it would be amazing to write a test that covers this before fixing it.

Vuksan (changes requested):
Great job on the diagrams in the description 🙏 
Kudos for removing those println calls from eth_endpoint 🙇 

I think we have an issue that opens up a possible exploit. I tried to explain it in [this comment](https://github.com/0xPolygon/polygon-sdk/pull/84#discussion_r674233109)

state/executor.go
zivkovicmilos: Good catch. Removed 👍

state/executor.go
zivkovicmilos: Removing now 👍

staking/staking_hub.go
zivkovicmilos: Agreed, renaming now 👍

staking/staking_hub.go
zivkovicmilos: Added the error check now here as well 👍

staking/staking_hub.go
zivkovicmilos: Making it private 👍

staking/staking_hub.go
zivkovicmilos: Good catch. This is definitely leftover, removing now 👍

staking/staking_hub.go
zivkovicmilos: The joys of IntelliJ :)
Renaming 👍

chain/chain.go
zivkovicmilos: It's just to separate it out from the staking implementation and keep it uniform, since it's not present in any other _legacy_ objects.
We added an analog way to `Alloc` -> `AllocStake` to keep track of prestaked balances

staking/staking_hub.go
zivkovicmilos: The only places where we modify the staking map directly are through methods `IncreaseStake`, which adds the address to the map, `DecreaseStake`(used only in tests) and `ResetStake`, which deletes that entry in the staking map:
https://github.com/0xPolygon/polygon-sdk/blob/c2e920b30aa55937496816e098d90b57adafc5ca/staking/staking_hub.go#L332-L334

GIven this, we can be sure that if the stake was reset, the address will not be in the map

staking/staking_hub.go
zivkovicmilos: Removing now 👍

staking/staking_hub.go
zivkovicmilos: Good catch, it's leftover.
Removing now 👍

state/executor.go
zivkovicmilos: Got it. Added a new event type for unknown events, and now handle them accordingly 👍

staking/staking_hub.go
zivkovicmilos: You're right.

Removing now 👍

state/executor.go
zivkovicmilos: Got it. Logging an error if the event was not found 👍

consensus/ibft/ibft.go
zivkovicmilos: You're right, this was leftover from a previous implementation.
Removing now 👍

staking/staking_hub_test.go
zivkovicmilos: Pulling this out to a specific helper directory, so it's visible wherever we need it 👍

staking/staking_hub_test.go
zivkovicmilos: Fixed 👍

staking/staking_hub_test.go
zivkovicmilos: Fixed 👍

Kourin1996 (approved):

helper/currency/currency.go
```diff
@@ -0,0 +1,9 @@
+package currency
+
+import "math/big"
+
+func EthToWei(ethValue int64) *big.Int {
```
Vuksan: Since this is only used in tests, can you maybe put it in `helper/tests/testing.go`?
I did something similar in [this PR](https://github.com/0xPolygon/polygon-sdk/pull/104).

staking/staking_hub.go
Vuksan: What happens if some address increases its stake by 10, then decreases it by 10?
It will be in the`StakingMap`, and `isStaker` will return `true` for that address, but `false` for an address that never staked anything.

I find that a bit misleading for anyone using this method, especially given it's description:
```
// isStaker is a helper method to check whether or not an address has a staked balance
```

staking/staking_hub.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	sh.log(
+		fmt.Sprintf("Stake increase:\t%s %s", address.String(), stakeBalance.String()),
+		logInfo,
+	)
+}
+
+// DecreaseStake decreases the account's staked balance if the account is present
+func (sh *StakingHub) DecreaseStake(address types.Address, unstakeBalance *big.Int) {
```
Vuksan, line 321: Shouldn't have implemented this yet, but since you wrote a neat test as well it would be a shame to remove it now 🙂

state/executor.go
```diff
@@ -123,6 +128,85 @@ type BlockResult struct {
...
+				discardedTxn.From,
+				discardedTxn.Value,
+				blockNumber,
+			)
+
+			if pendingEvent.EventType != staking.UnknownEvent {
+				hub := staking.GetStakingHub()
+				// Remove the event from the queue if it is present
+				hub.RemovePendingEvent(*pendingEvent)
+			}
```
Vuksan: What do you think of leaving this as it was, discarding any event produced by this `discardedTxn` - that way we would discard `UnknownEvents` as well, right?
This way any `UnknownEvent` ends up in the `EventQueue` for good

staking/staking_hub.go
```diff
@@ -0,0 +1,409 @@
...
+func (pe *PendingEvent) Compare(event PendingEvent) bool {
+	if pe.EventType == event.EventType &&
+		pe.Address.String() == event.Address.String() &&
+		pe.Value.Cmp(event.Value) == 0 &&
+		pe.BlockNumber == event.BlockNumber {
+		return true
+	}
+
+	return false
+}
```
Vuksan: I don't like this check, because based on it we assume that user cannot do the following in the same block:
`Send two transactions to stake the same amount of tokens.`
If he does that, we would see these events as one, which they are not.

My suggestion is to add some sort of `ID` to `PendingEvent` struct and compare events based on that

Vuksan: Furthermore, the user would only get his staked balance increased once, although he sent two transactions and paid for their execution.
Please, write a test that shows this is possible before implementing the solution.

state/runtime/system/system_unstaking.go
```diff
@@ -32,23 +33,43 @@ func (uh *unstakingHandler) run(state *systemState) ([]byte, error) {
...
 
 	// Decrease the staked balance on the staking address
 	state.host.SubBalance(stakingAddress, stakedBalance)
 
 	// Increase the account's actual balance
 	state.host.AddBalance(staker, stakedBalance)
 
-	// TODO Remove the staker from the validator set after this point + checks
+	// Emit an unstaked event for the consensus layer
 	state.host.EmitUnstakedEvent(staker, stakedBalance)
```
Vuksan: I think we've opened a new exploit:
```
Context: user has 10 ETH staked prior to this block.
He fits following transactions in the same block:

- unstake (user gains 10 ETH)
- stake 1 ETH
- unstake (user gains 10 ETH)
- stake 1 ETH
- repeat until he either reaches the block gas limit or depletes the network's staking address of funds.

```
Try to write an e2e test that does this to confirm before fixing it, please

e2e/pos_test.go
```diff
@@ -121,3 +139,108 @@ func TestPoS_Unstake(t *testing.T) {
...
+	expectedBalance := big.NewInt(0).Add(defaultBalance, defaultBalance)
+
+	// Subtract tx fees
+	expectedBalance = big.NewInt(0).Sub(
+		expectedBalance,
+		fee,
+	)
+
+	actualStakedBalance := getStakedBalance(referenceAddr, srv.JSONRPC(), t)
+	actualBalance := getAccountBalance(referenceAddr, srv.JSONRPC(), t)
```
Vuksan: Nitpick warning:
You can do this, to prevent instantiation of multiple JSONRPC clients:
```suggestion
	client := srv.JSONRPC()
	actualStakedBalance := getStakedBalance(referenceAddr, client, t)
	actualBalance := getAccountBalance(referenceAddr, client, t)
```

state/executor.go
Vuksan: @zivkovicmilos GJ on the test, exactly what I was hoping for 💯

staking/staking_hub.go
```diff
@@ -0,0 +1,409 @@
...
+			Stake:   stake,
+		}
+		indx++
+	}
+
+	return mappings
+}
+
+// ClearEvents resets the event queue. Called from the consensus layer
+// to discard any events that might have been created during initial block creation
```
Vuksan: Let callers decide on why they are calling this 🙂 
```suggestion
// ClearEvents resets the event queue
```

consensus/ibft/ibft.go
```diff
@@ -422,6 +423,10 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 	}
 
 	_, root := transition.Commit()
+	// If we are the proposer, these events will be added in Validate state again
+	// No need to double add them here and have complicated logic in unstaking
+	staking.GetStakingHub().ClearEvents()
```
Vuksan: Doesn't this mean that we might build a block with invalid (or exploiting) staking transactions, send `PREPREPARE` and `PREPARE` messages to other nodes, and only after that discover the invalid transactions in the `VALIDATE` state?

Can you elaborate a bit on this, how would that play out?

consensus/dev/dev.go
```diff
@@ -150,6 +151,7 @@ func (d *Dev) writeNewBlock(parent *types.Header) error {
 
 	// Commit the changes
 	_, root := transition.Commit()
+	staking.GetStakingHub().ClearEvents()
```
Vuksan: Please just add a comment that we're clearing events because they are added in the `transaction.Write()`, and need to be reset before we call `WriteBlocks()`

Vuksan (changes requested):
I think we might have a new exploit, a bit different this time. I outlined it in [this comment](https://github.com/0xPolygon/polygon-sdk/pull/84#discussion_r679172229).

e2e/pos_test.go
```diff
@@ -121,3 +139,108 @@ func TestPoS_Unstake(t *testing.T) {
...
+	for i := 0; i < numTransactions; i++ {
+		go sendFunc(&wg, t, i)
+	}
+
+	wg.Wait()
+
+	fee := new(big.Int).Mul(
+		big.NewInt(int64(receiptArr[0].GasUsed)),
+		big.NewInt(gasPrice.Int64()),
+	)
```
Vuksan: Shouldn't user pay the fee for failed transactions as well?

helper/currency/currency.go
zivkovicmilos: Got it, we'll keep it uniform on this branch as well 👍

e2e/pos_test.go
zivkovicmilos: Good catch.

Fixed 👍

staking/staking_hub.go
zivkovicmilos: Fixed 👍

consensus/dev/dev.go
zivkovicmilos: Fixed 👍

e2e/pos_test.go
zivkovicmilos: I'm not sure what you mean, this is the fee that will get charged to the user for executing the transaction, regardless if the transaction failed - unless we have some faulty logic in the executor.

consensus/ibft/ibft.go
zivkovicmilos: As a proposer we go through the exact same transaction execution / block validation process as any other node, but we do it seemingly twice.

If the transaction is faulty / invalid, it should get handled accordingly and this is out of the scope of the staking hub.
We disregard all events regardless when initially building the block.

staking/staking_hub.go
zivkovicmilos: The implementation assumed that the user cannot / should not send 2+ staking / unstaking transactions within a single block. I believe our IBFT validator event logic is based on this as well.

The only thing that is missing from the solution for it to respect the point stated above is to deny the user staking if he has tried to stake in the same block in `system_staking.go`.

@lazartravica @Kourin1996 correct me if I'm wrong on this assumption

state/executor.go
zivkovicmilos: You're right 👍 
Not only here, but in `commitApprovedEvents` there is a chance for `UnknownEvent`s to be leftover.
I've added a function that tidies up the queue.

staking/staking_hub.go
zivkovicmilos: @Vuksan For some reason I cannot reply to [this comment](https://github.com/0xPolygon/polygon-sdk/pull/84#discussion_r679095525) directly, so I'll write the reply here.

I'll modify the description a bit, so it's more clear as to what the method returns and at which point in time 👍

state/runtime/system/system_unstaking.go
zivkovicmilos: I have to test this out to confirm it, and I'll get back to you 👍

staking/staking_hub.go
zivkovicmilos: > The implementation assumed that the user cannot / should not send 2+ staking / unstaking transactions within a single block. I believe our IBFT validator event logic is based on this as well.
> 
> The only thing that is missing from the solution for it to respect the point stated above is to deny the user staking if he has tried to stake in the same block in `system_staking.go`.
> 
> @lazartravica @Kourin1996 correct me if I'm wrong on this assumption

Ignore this comment completely, we do have support for multiple staking / unstaking transactions within a single block. The IBFT validator event logic is executed after things have been committed 👍 

@Vuksan Adding the ID to differentiate the events further. Good catch 👍

state/runtime/system/system_unstaking.go
zivkovicmilos: Confirming that this exploit is present 👍 

I've patched it up, and added a test covering it: c801bac

staking/staking_hub.go
Vuksan: @zivkovicmilos Is this still unresolved?

state/executor.go
Vuksan: Isn't that function obsolete if you just remove this condition here and in the `commitApprovedEvents `.
This way it looks like we add the check to skip unknown events and then clean them up afterward using more CPU than necessary.
```suggestion
			hub := staking.GetStakingHub()
			// Remove the event from the queue if it is present
			hub.RemovePendingEvent(*pendingEvent)
```

staking/staking_hub.go
Vuksan: Is this still unresolved?

staking/staking_hub.go
```diff
@@ -0,0 +1,419 @@
...
+
+	if len(foundIndxs) > 0 {
+		prevSize := len(foundIndxs)
+
+		for _, foundIndx := range foundIndxs {
+			sh.EventQueue = append(sh.EventQueue[:foundIndx], sh.EventQueue[foundIndx+1:]...)
+		}
+
+		sh.log(fmt.Sprintf("Cleared [%d] unknown pending events", prevSize), logWarning)
+	}
```
Vuksan: This might be more performant:
```suggestion
	knownEvents := make([]PendingEvent, 0)
	for _, el := range sh.EventQueue {
		if el.EventType != UnknownEvent {
			knownEvents = append(knownEvents, el)
		}
	}
	numUnknownEvents := len(sh.EventQueue) - len(knownEvents)
	if numUnknownEvents > 0 {
		sh.log(fmt.Sprintf("Cleared [%d] unknown pending events", numUnknownEvents), logWarning)
		sh.EventQueue = knownEvents
	}
```

Vuksan: If you decide to keep this function, that is. Right now I believe it's not needed.

e2e/pos_test.go
```diff
@@ -121,3 +124,249 @@ func TestPoS_Unstake(t *testing.T) {
...
+		}
+
+		_, addErr := clt.AddTxn(context.Background(), msg)
+		if addErr != nil {
+			t.Fatalf("Unable to add txn, %v", addErr)
+		}
+	}
+
+	// Mandatory sleep for the dev consensus and executor to go through the txns
+	time.Sleep(time.Duration(devInterval+5) * time.Second)
```
Vuksan: Could you possibly check the tx receipts (from `dev` consensus's block) or some signal that dev consensus has produced a block other than sleeping like this?

e2e/pos_test.go
```diff
@@ -121,3 +124,249 @@ func TestPoS_Unstake(t *testing.T) {
...
+	actualStakedBalance := getStakedBalance(senderAddr, client, t)
+	actualAccountBalance := getAccountBalance(senderAddr, client, t)
+	actualStakingAddrBalance := getAccountBalance(stakingContractAddr, client, t)
+
+	// Make sure the balances match up
+
+	// The account balance should be in the range of
+	// [previousBalance + stake refund - fees, previousBalance + stake refund]
+	prevAccountBalanceWithStake := big.NewInt(0).Add(previousAccountBalance, defaultBalance)
+	ballparkFeeValue := new(big.Int).Exp(big.NewInt(10), big.NewInt(15), nil) // 0.001 ETH
```
Vuksan: Don't like the `ballparkFeeValue`, if it is possible to retrieve the receipts from the block that should help with calculating the exact fee

e2e/pos_test.go
```diff
@@ -121,3 +124,249 @@ func TestPoS_Unstake(t *testing.T) {
...
+		}
+
+		_, addErr := clt.AddTxn(context.Background(), msg)
+		if addErr != nil {
+			t.Fatalf("Unable to add txn, %v", addErr)
+		}
+	}
+
+	// Mandatory sleep for the dev consensus and executor to go through the txns
+	time.Sleep(time.Duration(devInterval+5) * time.Second)
```
Vuksan: Could you possibly check the tx receipts (from `dev` consensus's block) or some signal that dev consensus has produced a block other than sleeping like this?

e2e/pos_test.go
```diff
@@ -121,3 +124,249 @@ func TestPoS_Unstake(t *testing.T) {
...
+	)
+
+	assert.Equalf(t,
+		expStake.String(),
+		actualStakingAddrBalance.String(),
+		"Staked address balance mismatch after stake / unstake exploit",
+	)
+
+	// Make sure the account balances match up
+	ballparkFeeValue := new(big.Int).Exp(big.NewInt(10), big.NewInt(15), nil) // 0.001 ETH
```
Vuksan: Don't like the `ballparkFeeValue`, if it is possible to retrieve the receipts from the block that should help with calculating the exact fee

Vuksan (changes requested):
I think that the test `TestPoS_StakeUnstakeExploit` doesn't check this exploit correctly.

state/runtime/system/system_unstaking.go
Vuksan: I reverted the fix and run the test, this is the error I got:

Screen Shot 2021-08-05 at 20 01 41 PM

What should happen is that the malicious user should have his actual account balance increased quite a bit (by each unstake), but what actually occurs is that his actual account balance is even smaller than what we expect.

Try to have the staking address with a lot more ETH (1000 for example) than what this user stakes.

staking/staking_hub.go
zivkovicmilos: I decided it didn't make sense to add the ID, as these events are constructed on the fly both in the staking / unstaking handles as well as in the executor when going over transactions.

The fix was simply to remove the `hasEvent` check when adding the pending event, because that's what prevented double adds, and it was leftover code to patch the proposer double event add - and has now been resolved through other code flows 👍

state/executor.go
zivkovicmilos: I actually believe we don't even need to check for unknown events thoroughly.

If you take a look at the commit / clear logic in the executor, we actually check if the event is a valid system event - meaning that it can never be an `Unknown` event type:
https://github.com/0xPolygon/polygon-sdk/blob/ddb93e332143a0e7c3aa92faa56a420e63a578eb/state/executor.go#L164

and

https://github.com/0xPolygon/polygon-sdk/blob/ddb93e332143a0e7c3aa92faa56a420e63a578eb/state/executor.go#L187

Removed the useless clearing now 👍 
50f5bac

staking/staking_hub.go
zivkovicmilos: I've changed up the description so it is more clear as to what the function actually does 👍 :
https://github.com/0xPolygon/polygon-sdk/blob/50f5bac2b7acc2e78ca446f429d6c47b0330e4a3/staking/staking_hub.go#L266-L275

staking/staking_hub.go
zivkovicmilos: Removed it as part of another comment 👍 
50f5bac

e2e/pos_test.go
zivkovicmilos: You're right. We should be exact when possible.
I've overlooked that we have an entry point to the block generation through the system operator interface.

Fixed 👍 
04c25da

e2e/pos_test.go

04c25da

e2e/pos_test.go

04c25da

e2e/pos_test.go

04c25da

state/runtime/system/system_unstaking.go
zivkovicmilos: You're right.

By increasing the initial staking address balance to something large, I've been able to get the account balance to significantly increase before the fix in `system_unstaking.go`. 👍 

5632957

staking/staking_hub.go
Vuksan: I think the description isn't enough given that this function changes behavior depending on what txs were executed in the block prior to the call. Can we make it a bit more unambiguous, like this:
```suggestion
	if _, ok := sh.StakingMap[address]; ok {
		return sh.StakingMap[address].Cmp(big.NewInt(0)) > 0
	}
```

staking/staking_hub.go
Vuksan: Alright, but now we would have to instruct the caller of the `RemovePendingEvent` to call it multiple times if he wishes to remove a specific event, because multiple instances of the same event may be present.

I think that having that ID would clear any doubts for the caller, which will prove beneficial in the long run.

state/runtime/system/system_unstaking.go
```diff
@@ -26,30 +27,50 @@ func (uh *unstakingHandler) run(state *systemState) ([]byte, error) {
...
 	// Sanity check
 	// Can't unstake if the balance isn't previously present on the staking account
 	// Can't unstake if the value is different from 0
-	if stakingAccountBalance.Cmp(stakedBalance) < 0 || state.contract.Value.Cmp(big.NewInt(0)) != 0 {
+	// Can't unstake if the account doesn't have stake
+	// Can't unstake if the account has pending events which cause it to not have stake
+	if stakingAccountBalance.Cmp(afterEventsStake) < 0 ||
+		state.contract.Value.Cmp(zeroValue) != 0 ||
+		stakedBalance.Cmp(zeroValue) == 0 ||
+		afterEventsStake.Cmp(zeroValue) == 0 {
```
Vuksan: I think that you need to omit this check (`stakedBalance.Cmp(zeroValue) == 0`):
```suggestion
	if stakingAccountBalance.Cmp(afterEventsStake) < 0 ||
		state.contract.Value.Cmp(zeroValue) != 0 ||
		afterEventsStake.Cmp(zeroValue) == 0 {
```

If you do not remove it, the following scenario becomes impossible:
```
Prior to the block user has a balance of 5 ETH and staked balance of 0 ETH.
In the same block he does the following:

1. Stake 2 ETH
2. Unstake
```

If you agree, please add this test scenario to `pos_test.go`. It isn't an exploit, so maybe something like `TestPoS_StakeUnstakeWithinSameBlock`

Vuksan (commented):
I can't see any major exploits this time. 🔒 
Looking at the code, and the whole `Event` concept that we've come up with, I feel as if we've made the code more complex than it needs to be for the basic staking functionality.
I would love to try and make it more simple since PoS requirements can (and probably will) become much more complex in the future (delegation, slashing...).

staking/staking_hub.go
zivkovicmilos: I'm not sure I understand why you would need to call `RemovePendingEvent` manually.
All events will eventually be removed / added if they are in the queue. 

The problem with adding an ID is that you have no idea what kind of ID was assigned to a specific event in the `system_staking` or `system_unstaking` handlers, while you're in the `executor` and constructing events for searches.

Could you please clarify this?

state/runtime/system/system_unstaking.go
zivkovicmilos: You're right 👍 
I changed the logic below, but forgot to remove the stale `stakedBalance` from the check. We already have `afterEventsStake` which is the most up to date staked balance.

Changing now, and adding a test 👍 

73cf867

staking/staking_hub.go
zivkovicmilos: Got it.
Fixed 👍 

37e054c

state/runtime/system/system_unstaking.go
Vuksan: Good job on the test, it's on point 🎯

staking/staking_hub.go
Vuksan: > I'm not sure I understand why you would need to call RemovePendingEvent manually.
All events will eventually be removed / added if they are in the queue.

I don't think at this moment, but this is a public method. We may have numerous bugs tomorrow because we may rely on this method to tell us if the two events are the same event, while in reality it can lie to us stating that two events are the same one, although they are not.

> The problem with adding an ID is that you have no idea what kind of ID was assigned to a specific event in the system_staking or system_unstaking handlers, while you're in the executor and constructing events for searches.

I meant to have an integer ID that increases whenever a new event is created.
How come you don't know the ID? Wouldn't it be emitted as well, since it would be a part of the event structure?

zivkovicmilos: It's not a problem to add an ID in the form of a UUID, the issue arises in `executor.go` when you're parsing transactions and need to construct an event for the search:
https://github.com/0xPolygon/polygon-sdk/blob/37e054cef0a49d6669db45c59f1778473318e7c9/state/executor.go#L184-L189

and

https://github.com/0xPolygon/polygon-sdk/blob/37e054cef0a49d6669db45c59f1778473318e7c9/state/executor.go#L165-L170

At this point in time, you have no idea what the ID of the pending event that corresponds to that transaction that you're currently parsing is - which means if you don't know this ID, you won't be able to do a search on the event queue if the ID is part of that search.

The only solution I see to this is to add some kind of tracking mechanism in `executor.go` to know which event corresponds to which ID, but it would only complicate the code even more.

minimal/server.go
```diff
@@ -78,6 +79,11 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 
 	m.logger.Info("Data dir", "path", config.DataDir)
 
+	// Spin up the staking hub if the consensus engine is ibft-pos
+	if config.Chain.Params.GetEngine() == chain.IBFTEngine {
+		staking.GetStakingHub().SetLogger(logger)
+		staking.GetStakingHub().SetWorkingDirectory(config.DataDir)
+	}
```
Kourin1996: Just a small comment. Will it work if we move this code to `Factory` of IBFT? I think minimal shouldn't care about which consensus use.

minimal/server.go
```diff
@@ -376,6 +396,9 @@ func (s *Server) Close() {
 	if err := s.consensus.Close(); err != nil {
 		s.logger.Error("failed to close consensus", "err", err.Error())
 	}
+
+	// Close the staking hub if present
+	staking.GetStakingHub().CloseStakingHub()
```
Kourin1996: If we can move init process of StakingHub to consensus, then this can be moved to Close() in consensus layer also.

staking/staking_hub.go
```diff
@@ -0,0 +1,396 @@
...
+// RemovePendingEvent removes the pending event from the event queue if it exists.
+// Returns a boolean flag indicating if an event was found and removed
+func (sh *StakingHub) RemovePendingEvent(event PendingEvent) bool {
+	sh.EventQueueMutex.Lock()
+	defer sh.EventQueueMutex.Unlock()
+
+	foundIndx := -1
+	for indx, el := range sh.EventQueue {
+		if el.Compare(event) {
+			foundIndx = indx
```
Kourin1996: If you expect to remove the first element that equals to given event. You might want to add break in if block. Or you can integrate with removing from slice.

e2e/pos_test.go
```diff
@@ -121,3 +127,400 @@ func TestPoS_Unstake(t *testing.T) {
...
+	expBalance = big.NewInt(0).Sub(expBalance, paidFee)
+
+	assert.Equalf(t,
+		expBalance.String(),
+		actualAccountBalance.String(),
+		"Account balance mismatch after unstake exploit",
+	)
+}
+
+func TestPoS_StakeUnstakeWithinSameBlock(t *testing.T) {
```
lazartravica, line 402: @zivkovicmilos  Isn't the previous tests testing the staking/unstaking in the same block?

Also is there any reason the devInterval is 5 just in this test?

helper/tests/testing.go
```diff
@@ -0,0 +1,9 @@
+package tests
+
+import "math/big"
+
+func EthToWei(ethValue int64) *big.Int {
```
lazartravica, line 13: Excellent catch that EthToWei is only used in tests.

staking/staking_hub.go
```diff
@@ -0,0 +1,396 @@
...
+// PendingEvent contains useful information about a staking / unstaking event
+type PendingEvent struct {
+	BlockNumber int64
+	Address     types.Address
+	Value       *big.Int
+	EventType   StakingEventType
+}
+
+// Compare checks if the two events match
+func (pe *PendingEvent) Compare(event PendingEvent) bool {
```
lazartravica: `Compare` should usually return `-1`, `0`, `1` since it is checking how 2 values compare to each other.
`Equal` should usually return true or false.

I think you should rename this method to `Equal`

staking/staking_hub.go
lazartravica: @zivkovicmilos @Vuksan What do you think about using the transaction nonce instead of the ID?

state/executor.go
```diff
@@ -123,6 +128,81 @@ type BlockResult struct {
...
+	}
+}
+
+// constructPendingEvent is a helper method for constructing a pending event
+func constructPendingEvent(
+	toAddress types.Address,
+	fromAddress types.Address,
+	value *big.Int,
+	blockNumber uint64,
+) *staking.PendingEvent {
```
lazartravica: perhaps pass the transaction?

lazartravica (changes requested):
We are polishing this to perfection, and we definitely should.

One thing I didn't comment in the code, but would like to ask of you:

Could you please add as large as needed comment blocks in the code for testing the different exploits we've detected just for any reader to be able to read from the test case documentation how the exploit works?

Thank you @zivkovicmilos ! 💪

Kourin1996:
Sorry, I have one question. I noticed it just now.
In `ProcessBlock`, you revert staking events if `txn.Write(t)` returns error. This looks looks right process. Please note `txn.Write(t)` will not return any error if transaction failed by like insufficient balance. (Actually according to code of `txn.Write(t)`, I see it always returns nil.)

Kourin1996:
Sorry, I have one question. I noticed it just now.
In `ProcessBlock`, you revert staking events if `txn.Write(t)` returns error. This looks looks right process. Please note `txn.Write(t)` will not return any error if transaction failed. (Actually according to code of `txn.Write(t)`, I see it always returns nil.)
~~If transaction comes from EOA, perhaps it should work. But I'm not sure it works if it comes from contract account.~~

Updated:
I checked current `develop` branch, transaction error is in result now. So we need to handle both error and transaction error.
* When error happens, we can't process block so revert all events which happen within block
* When transaction error happens, transaction is put into block with failed status so revert the events for only a txn.

But in any cases, please keep in mind that ProcessBlock cannot detect transaction failed in current implementations.

lazartravica:
@zivkovicmilos please merge develop into this branch, as we have discussed 🙏

staking/staking_hub.go
Vuksan: Oh, I get it now.
What do you think of using the `nonce` as @lazartravica suggested, of course it would have to be in combination with `From` I guess?

minimal/server.go
zivkovicmilos: You're completely right.

I remember I was looking for a good place to store this that's not minimal, but later on forgot to move it out.

Moved 👍 

e96d66e 
52eebf8

minimal/server.go
zivkovicmilos: https://github.com/0xPolygon/polygon-sdk/pull/84#discussion_r688033118

staking/staking_hub.go
zivkovicmilos: Good catch 👍 

This is indeed the intended behavior. Fixed 👍 

f2088bc

e2e/pos_test.go
zivkovicmilos: The test scenario is different for this one, where it starts with a stake, then an unstake, whereas other start with an unstake then a stake. Initial params are also different.

No specific reason for lowering the dev interval to 5s, I was just wondering if it's enough time to wait, and forgot to make it uniform. Seems like 5s is enough for the dev interval, do you want me to change it for the other tests as well?

helper/tests/testing.go
zivkovicmilos: @Vuksan Suggested putting everything common for tests in this package, so I just followed his advice :)

staking/staking_hub.go
zivkovicmilos: Completely agree 👍 

Fixed 👍 

ebc42eb

zivkovicmilos:
@lazartravica @Kourin1996 @Vuksan 

Commit that removes the event queue and simplifies our staking logic is up:
e57444f

Would love to hear thoughts / what can be improved 👍

staking/staking_hub.go
```diff
@@ -0,0 +1,416 @@
...
+	WorkingDirectory string
+
+	// Write-back period (in s) for backup staking data
+	WritebackPeriod time.Duration
+
+	// Logger for logging errors and information
+	Logger hclog.Logger
+
+	// Event list mutex
+	EventQueueMutex sync.Mutex
```
Kourin1996, line 50: Is this (EventQueueMutex) no longer used?

staking/staking_hub.go
```diff
@@ -0,0 +1,416 @@
...
+	sh.StakingMutex.Lock()
+	defer sh.StakingMutex.Unlock()
+
+	// Alert the closing channel
+	select {
+	case sh.CloseCh <- struct{}{}:
+		sh.log("Closing staking hub...", logInfo)
+	default:
+		sh.log("SH Writer not set up. Closing staking hub...", logInfo)
+	}
```
Kourin1996, line 132: I didn't understand what you want to do here 100%, but there is a possibility that nothing is sent to channel if you add default case. To avoid this, you can set channel capacity to 1 or remove default case.

consensus/ibft/ibft.go
```diff
@@ -423,6 +428,10 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 	i.logger.Info("picked out txns from pool", "num", len(txns), "remaining", i.txpool.Length())
 
 	_, root := transition.Commit()
+	// If we are the proposer, these events will be added in Validate state again
+	// No need to double add them here and have complicated logic in unstaking
+	staking.GetStakingHub().ClearDirtyStakes()
```
Vuksan, line 433: Comment talking about events.

state/executor.go
```diff
@@ -123,6 +129,18 @@ type BlockResult struct {
 	TotalGas uint64
 }
 
+// cleanDirtyStakes is a helper method for clearing discarded staked events
+func cleanDirtyStakes() {
+	hub := staking.GetStakingHub()
+	hub.ClearDirtyStakes()
+}
```
Vuksan, line 136: Comment talking about events.

consensus/dev/dev.go
```diff
@@ -147,9 +148,14 @@ func (d *Dev) writeNewBlock(parent *types.Header) error {
...
 		txns = append(txns, txn)
 	}
+	d.logger.Debug(fmt.Sprintf("Writing %d transactions to a block", len(txns)))
 
 	// Commit the changes
 	_, root := transition.Commit()
+	// Since the block transactions are firstly written, then the block is built (causing the txns to be written again),
+	// no need to double add staking / unstaking events that might have been added in the initial txn write,
+	// and have complicated logic in unstaking
+	staking.GetStakingHub().ClearDirtyStakes()
```
Vuksan, line 158: Comment talking about events.

staking/staking_hub.go
```diff
@@ -0,0 +1,416 @@
...
+func (sh *StakingHub) getCleanStake(address types.Address) *big.Int {
+	sh.StakingMutex.Lock()
+	defer sh.StakingMutex.Unlock()
+
+	if !sh.isStaker(address) {
+		return big.NewInt(0)
+	}
+
+	return sh.StakingMap[address]
+}
```
Vuksan, line 286: I think you can remove this method since it is the same as `GetStakedBalance()`

Vuksan (changes requested):
Minor nitpicks
GJ on the dirty structure approach 👍

Mum2541:
gh pr checkout 84

zivkovicmilos:
Closed and archived with tag `archive/ibft--pos-v2`

https://github.com/0xPolygon/polygon-sdk/releases/tag/archive%2Fibft-pos-v2

PR #85, opened 2021-07-15
State Pruning

Kourin1996:
# Description

Note: This PR is just a plan to introduce state pruning currently.

This PR adds state pruning to Polygon SDK. There are 2 approaches that have been introduced in go-ethereum in v1.8.x and v.1.10.x.

## Intermediate mempool between trie and database (added in Geth v1.8.x)

This approach suppresses writing trie nodes into disk rather than pruning nodes. It adds in-memory trie db between transition and disk trie db. It's based on the fact the most part of a state update tend to become obsolete within limited time by overwriting.

When adding trie node, it's stored with reference counter into in-memory db first. When a new trie references something that already exists, reference counter in the node is increased. When deleting old trie, dereference its root, decrementing the refcounts of all children.

> some data must be pushed to disk when restarting Geth (or periodically to avoid a full re-sync after a crash). The in-memory garbage collector in incapable of pruning this data away, so this still adds up.

This approach doesn't do state pruning completely, but we just need to extend trie tree and add in-memory layer. So it's easier to introduce this approach than offline pruning approach described next.

## Offline pruning by bloom filter (added in Geth v1.10.x)

As describes above, first approach isn't perfect. There is the possibility of saving unnecessary nodes into storage. This function cleans trie by snapshots and bloom filter.

Commits all the state entries(trie node + contract code) into the bloom filter. After that, traverse and delete nodes if the key doesn't contain in bloom filter. 

**Need to check the snapshot described here is same one in PolygonSDK**
`Snapshot` described here is a secondary data structure for storing the Ethereum state. It archives to reduce number of disk reads. (O(log N) -> O(1))

First PR for snapshot in go-ethereum
+ `https://github.com/ethereum/go-ethereum/pull/22331`

> Be advised, that pruning is a new and dangerous feature, a failure of which can cause bad blocks. We’re confident that it’s reliable, but if something goes wrong, there’s likely no way to salvage the database. Our recommendation - at least until the feature gets battle tested - is to back up your database prior to pruning, and try with testnet nodes first before going all in on mainnet.

To introduce this approach, firstly we need to snapshot feature which is a big change. In addition, we need to keep in mind that this function seems to be under test. (It's been merged to master already, but user decide if it uses this function) We need to consider deeply.

## Relevant

Article of state pruning by reference counter:
https://blog.ethereum.org/2015/06/26/state-tree-pruning/

Explanation of Geth v1.10.x including one about offline pruning:
https://blog.ethereum.org/2021/03/03/geth-v1-10-0/

Explanation of snapshot
https://blog.ethereum.org/2020/07/17/ask-about-geth-snapshot-acceleration/

Discussions about state pruning in go-ethereum:
+ `https://github.com/ethereum/go-ethereum/pull/2111`
+ `https://github.com/ethereum/go-ethereum/pull/18418`

PRs of intermediate mempool approach
+ `https://github.com/ethereum/go-ethereum/pull/15857`
+ `https://github.com/ethereum/go-ethereum/pull/15864`
+ `https://github.com/ethereum/go-ethereum/pull/15883`

PRs of offline pruning
+ `https://github.com/ethereum/go-ethereum/pull/21724`
+ `https://github.com/ethereum/go-ethereum/pull/22386`
+ `https://github.com/ethereum/go-ethereum/pull/22291`
+ `https://github.com/ethereum/go-ethereum/pull/22294`

# Changes include

# Breaking changes

# Checklist

- [ ] I have tested this code

# Additional comments


Vuksan (approved):
Good job on the research! 👏 
I agree that offline pruning requires quite a bit of work, since for it to function we first need to implement snapshots, and that is a big task.

Btw can you share a link to the PR that implemented the intermediate mempool (1st solution)?
I've found that [this one](https://github.com/ethereum/go-ethereum/pull/18418) is the most recent one, but it isn't merged.

zivkovicmilos (approved):
Very informative read 💯

Kourin1996:
@Vuksan intermediate mempool should be added in `https://github.com/ethereum/go-ethereum/pull/15857`

lazartravica:
Thanks Yoshiki for the excellent read. I will close this pull request and we can find it later by searching for state pruning.

🙏

PR #87, opened 2021-07-16
Not returning an error when trying running eth_getCode against an uni…

lazartravica:
# Description

The eth_getCode JSON RPC call returned an error if the address supplied in the parameter was uninitialized. 
Replicating the behaviour of geth, I have changed it so that it returns a `0x` string.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


Kourin1996 (approved):
LGTM 👏

jsonrpc/eth_endpoint.go
```diff
@@ -462,7 +462,7 @@ func (e *Eth) GetCode(address types.Address, number BlockNumber) (interface{}, e
 
 	acc, err := e.d.store.GetAccount(header.StateRoot, address)
 	if err != nil {
-		return nil, err
+		return "0x", nil
```
zivkovicmilos, line 465: We should consider wrapping this in the `arg` types, as we do with other endpoints:
`argUintPtr`, `argBigPtr`...

zivkovicmilos (approved):
Looks good 💯

Vuksan (approved):
LGTM.
The only issue I see is that we don't really comply with [the Ethereum specification for the getCode](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getcode), but we don't do that for other endpoints as well, so I guess fixing that is a different task.

EDIT: Meaning we don't return result with metadata fields: `id` and `jsonrpc`

PR #89, opened 2021-07-19
Fix the popped order for priced transactions

zivkovicmilos:
# Description

This PR fixes a minor bug with the ordering of popped transactions from the price-sorted transaction heap (txpool).
Now, the heap pops transactions which have a **greater** price first.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -56,8 +56,14 @@ type TxPool struct {
 	proto.UnimplementedTxnPoolOperatorServer
 }
 
-// NewTxPool creates a new pool of transactios
-func NewTxPool(logger hclog.Logger, sealing bool, store store, grpcServer *grpc.Server, network *network.Server) (*TxPool, error) {
+// NewTxPool creates a new pool for transactions
+func NewTxPool(
+	logger hclog.Logger,
+	sealing bool,
+	store store,
+	grpcServer *grpc.Server,
+	network *network.Server,
+) (*TxPool, error) {
 	txPool := &TxPool{
 		logger:     logger.Named("txpool"),
 		store:      store,
@@ -485,7 +491,8 @@ func (t txPriceHeapImpl) Less(i, j int) bool {
 	if t[i].from == t[j].from {
 		return t[i].tx.Nonce < t[j].tx.Nonce
 	}
-	return t[i].price.Cmp((t[j].price)) < 0
+
+	return t[i].price.Cmp(t[j].price) >= 0
 }
 
 func (t txPriceHeapImpl) Swap(i, j int) {
--- a/txpool/txpool_test.go
+++ b/txpool/txpool_test.go
@@ -3,6 +3,7 @@ package txpool
 import (
 	"fmt"
 	"math/big"
+	"strconv"
 	"testing"
 
 	"github.com/0xPolygon/minimal/crypto"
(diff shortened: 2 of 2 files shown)


txpool/txpool.go
```diff
@@ -485,7 +491,8 @@ func (t txPriceHeapImpl) Less(i, j int) bool {
 	if t[i].from == t[j].from {
 		return t[i].tx.Nonce < t[j].tx.Nonce
 	}
-	return t[i].price.Cmp((t[j].price)) < 0
+
+	return t[i].price.Cmp(t[j].price) >= 0
```
lazartravica, line 495: I like the `>=` optimization instead of `>` as new items would start their journey comparing with nodes that are probably smaller than them

lazartravica (approved):
LGTM, good job on improving the tests.

I will close #86

txpool/txpool_test.go
```diff
@@ -121,3 +122,100 @@ func TestTxnQueue_Promotion(t *testing.T) {
...
+			{
+				From:     addr1,
+				GasPrice: big.NewInt(2),
+				Nonce:    0,
+			},
+			{
+				From:     addr1,
+				GasPrice: big.NewInt(1),
+				Nonce:    1,
+			},
```
Vuksan: What do you think of changing the GasPrice so that it is unambiguous that here we actually look at `Nonce` instead of the `GasPrice`?
You could do that by making a second transaction have GasPrice of 3, for example.

txpool/txpool_test.go
```diff
@@ -121,3 +122,100 @@ func TestTxnQueue_Promotion(t *testing.T) {
...
+	addr1 := types.Address{0x1}
+	addr2 := types.Address{0x2}
+
+	test := func(t *testing.T, testTable []TestCase) {
+		pool, err := NewTxPool(hclog.NewNullLogger(), false, &mockStore{}, nil, nil)
+		assert.NoError(t, err)
+		pool.EnableDev()
+
+		for _, testCase := range testTable {
+			err := pool.addImpl("", &types.Transaction{
```
Vuksan, line 143: Why didn't you test the public method `AddTx` instead of the private one?

Vuksan (changes requested):
GJ on the tests. Please check out the minor suggestions that I had.

txpool/txpool_test.go
zivkovicmilos: The rest of the tests use the internal `addImpl`, because `AddTx` is just sugar coat:
- it calls `addImpl` internally
- it broadcasts the transaction
---
https://github.com/0xPolygon/polygon-sdk/blob/acfa430342569c1c74486e2b427a96e9ab908f02/txpool/txpool.go#L123-L148

Since these are unit tests and we don't have networking or broadcasts enabled for this specific test, calling `AddTx` is basically the same as calling `addImpl`.

txpool/txpool_test.go
zivkovicmilos: If I understood correctly, you propose we change the values for the GasPrice so the point of this test case (testing the nonce) is in full focus?

I'll change up the values for the second gas price as you suggested 👍

Kourin1996 (approved):
LGTM 👍

txpool/txpool_test.go
Vuksan: It was only a comment in the sense that it's more important to test public than private methods. I understand that currently it is the same, but tomorrow if someone is to change the "wrapper" method (`AddTx`) we wouldn't catch that.

No need to focus on this right now, just to keep it in mind for future testing - testing public methods should take precedence over private ones since the client (or the rest of the system) will use these, and not private ones 🙂

Vuksan (approved):

txpool/txpool_test.go
zivkovicmilos: I completely agree. We need to overhaul and standardize the way we write / run most tests in the SDK. 
We'll probably open up several tasks further down the line to refactor these tests :+1:

PR #93, opened 2021-07-22
Add comments and rename method name to improve code readability

Kourin1996:
# Description

This PR adds some comments, fixes the codes without any impact, and renames some method names to improve code readability

## Explanation about some changes

Keeps a record of the reason of some changes.

### Message code used in committed seal

> Committed seal is calculated by each of the validator signing the hash along with COMMIT_MSG_CODE message code of its private key

`COMMIT_MSG_CODE` is 1 according to EIP-650, but it's 2 in Quorum implementation because they added PREPREPARE_MSG_CODE. PolygonSDK is using same set of message codes to Quorum implementation. So we can calculate committed seal by 2.

https://docs.goquorum.consensys.net/en/stable/Concepts/Consensus/IBFT/

### Remaining gas check in apply of Transition

This PR removes following code:
https://github.com/0xPolygon/polygon-sdk/blob/2f50a194f21d972b542ccd1e1571ccba40c5c810/state/executor.go#L373-L376

The reason is that `leftoverGas` is calculated by `msg.gas - (constant value) - (size of Input with non-zero value)`, which means leftoverGas is always less than msg.gas. So will never enter the branch.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/consensus/ibft/sign.go
+++ b/consensus/ibft/sign.go
@@ -14,8 +14,8 @@ import (
 
 func commitMsg(b []byte) []byte {
 	// message that the nodes need to sign to commit to a block
-	// TODO: EIP-650 says the COMMIT_MSG_CODE is 1, not 2
-	return crypto.Keccak256(b, []byte{byte(2)})
+	// hash with COMMIT_MSG_CODE which is the same value used in quorum
+	return crypto.Keccak256(b, []byte{byte(proto.MessageReq_Commit)})
 }
 
 func ecrecoverImpl(sig, msg []byte) (types.Address, error) {
@@ -34,7 +34,7 @@ func ecrecoverFromHeader(h *types.Header) (types.Address, error) {
 		return types.Address{}, err
 	}
 	// get the sig
-	msg, err := signHash(h)
+	msg, err := getHeaderHash(h)
 	if err != nil {
 		return types.Address{}, err
 	}
@@ -43,7 +43,7 @@ func ecrecoverFromHeader(h *types.Header) (types.Address, error) {
 }
 
 func signSealImpl(prv *ecdsa.PrivateKey, h *types.Header, committed bool) ([]byte, error) {
-	sig, err := signHash(h)
+	sig, err := getHeaderHash(h)
 	if err != nil {
 		return nil, err
 	}
@@ -110,8 +110,7 @@ func writeCommittedSeals(h *types.Header, seals [][]byte) (*types.Header, error)
 	return h, nil
 }
 
-// TODO: This function is called signHash, we are doing no signing
-func signHash(h *types.Header) ([]byte, error) {
+func getHeaderHash(h *types.Header) ([]byte, error) {
 	//hash := istanbulHeaderHash(h)
 	//return hash.Bytes(), nil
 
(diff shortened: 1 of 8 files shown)


consensus/ibft/sign.go
```diff
@@ -43,7 +43,7 @@ func ecrecoverFromHeader(h *types.Header) (types.Address, error) {
 }
 
 func signSealImpl(prv *ecdsa.PrivateKey, h *types.Header, committed bool) ([]byte, error) {
-	sig, err := signHash(h)
+	sig, err := getHeaderHash(h)
```
Vuksan: Please change variable sig to a more proper name, since it's not a signature:
```suggestion
	hash, err := getHeaderHash(h)
```

consensus/ibft/sign.go
```diff
@@ -110,8 +110,7 @@ func writeCommittedSeals(h *types.Header, seals [][]byte) (*types.Header, error)
 	return h, nil
 }
 
-// TODO: This function is called signHash, we are doing no signing
-func signHash(h *types.Header) ([]byte, error) {
+func getHeaderHash(h *types.Header) ([]byte, error) {
```
Vuksan: What do you think if we rename this to `calculateHeaderHash()`?
As it is now, I would expect it to only do `return h.Hash`

state/txn.go
```diff
@@ -333,7 +334,7 @@ func (txn *Txn) GetState(addr types.Address, key types.Hash) types.Hash {
 		}
 	}
 
-	// Under which condition is this called?
+	// Get account state from trie tree
```
Vuksan: Can you please elaborate a bit here (in code comment) on what's the difference between `trie tree` and `radix tree` up on the line 327?

state/executor.go
```diff
@@ -370,10 +365,6 @@ func (t *Transition) apply(msg *types.Transaction) (
 	if err != nil {
 		return nil, 0, false, err
 	}
-	// TODO: Check if this is even possible
-	if leftoverGas > msg.Gas {
-		return nil, 0, false, errorVMOutOfGas
-	}
```
Vuksan: I'm afraid we cannot do this, it is the only check that we have if intrinsic gas cost goes above `msg.Gas`.
If you take a look at `preCheck()` method on line 349 it deducts the `msg.Gas` from `intrinsicGas` cost of the transaction it calculated.

If the intrinsic gas is greater than the gas limit that sender has provided through `msg.Gas`, then `gasAvailable` will go through underflow, and end up being some big number.
That's why we still need this weird check on the outside to check if the `leftoverGas` is greater than `msg.Gas` because then we know that what has occurred is that transaction has run out of gas during intrinsic gas cost check.

Please leave this as it is, I have a fix for this as a part of my task 🙂

Vuksan (changes requested):
Mostly minor issues

state/executor.go
Vuksan: What saddens me is that our tests still pass, even with this issue. 🙃

consensus/ibft/sign.go
lazartravica: Agree with Vuksan on this one, `get` should only be reserved for getter methods and simple helper funcs.

state/txn.go
lazartravica: Also, try to include a comment like: `If the object was not found in the radix trie (due to reasons XXX), we fetch it from the trie tre...`

In simpler terms, we should try to comment what is the logic behind these if branches.

lazartravica (changes requested):

lazartravica (approved):
💯

Vuksan (approved):
👍

zivkovicmilos (approved):
Thank you for clarifying some tricky parts of the code, it is much more understandable now what is happening 💯

PR #94, opened 2021-07-22
Fix up the JSON WebSocket interface

zivkovicmilos:
# Description

This PR solves the pending issues with WebSocket connections.
Specifically, it expands on the error handling mechanisms and cleans up the code a bit.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/e2e/websocket_test.go
+++ b/e2e/websocket_test.go
@@ -101,8 +101,12 @@ func TestWS_Response(t *testing.T) {
 			t.Fatalf("Unable to unmarshal WS result: %v", wsError)
 		}
 
-		foundBalance, wsError := types.ParseUint256orHex(&balanceHex)
-		if wsError = json.Unmarshal(res.Result, &balanceHex); wsError != nil {
+		foundBalance, parseError := types.ParseUint256orHex(&balanceHex)
+		if parseError != nil {
+			t.Fatalf(fmt.Sprintf("Unable to parse WS result balance: %v", parseError))
+		}
+
+		if wsError := json.Unmarshal(res.Result, &balanceHex); wsError != nil {
 			t.Fatalf("Unable to parse WS result balance: %v", wsError)
 		}
 
@@ -144,8 +148,11 @@ func TestWS_Response(t *testing.T) {
 			t.Fatalf("Unable to unmarshal WS result: %v", wsError)
 		}
 
-		blockNumInt, wsError := types.ParseUint256orHex(&blockNum)
-		if wsError = json.Unmarshal(res.Result, &blockNum); wsError != nil {
+		blockNumInt, parseError := types.ParseUint256orHex(&blockNum)
+		if parseError != nil {
+			t.Fatalf(fmt.Sprintf("Unable to parse WS result balance: %v", parseError))
+		}
+		if wsError := json.Unmarshal(res.Result, &blockNum); wsError != nil {
 			t.Fatalf("Unable to parse WS result balance: %v", wsError)
 		}
 
--- a/e2e/websocket_test.go
+++ b/e2e/websocket_test.go
@@ -1,0 +1,154 @@
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math/big"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/0xPolygon/minimal/e2e/framework"
+	"github.com/0xPolygon/minimal/jsonrpc"
+	"github.com/0xPolygon/minimal/types"
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/assert"
+	"github.com/umbracle/go-web3"
+)
+
+type testWSRequest struct {
+	JSONRPC string   `json:"jsonrpc"`
+	Params  []string `json:"params"`
+	Method  string   `json:"method"`
+	ID      int      `json:"id"`
+}
+
+func constructWSRequest(id int, method string, params []string) ([]byte, error) {
+	request := testWSRequest{
+		JSONRPC: "2.0",
+		Method:  method,
+		ID:      id,
+		Params:  params,
+	}
+
+	return json.Marshal(request)
+}
+
+func getWSResponse(t *testing.T, ws *websocket.Conn, request []byte) (jsonrpc.Response, error) {
+	if wsError := ws.WriteMessage(websocket.TextMessage, request); wsError != nil {
+		t.Fatalf("Unable to write message to WS connection: %v", wsError)
+	}
+	_, response, wsError := ws.ReadMessage()
+	if wsError != nil {
+		t.Fatalf("Unable to read message from WS connection: %v", wsError)
+	}
+
+	var res jsonrpc.Response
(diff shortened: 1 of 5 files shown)


lazartravica (approved):
LGTM, ws endpoints are working 👍

Kourin1996 (approved):
LGTM 👏

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	t.Run("Valid account balance", func(t *testing.T) {
+		requestID := 1
+		request, constructErr := constructWSRequest(
+			requestID,
+			"eth_getBalance",
+			[]string{preminedAccounts[0].address.String(), "latest"},
+		)
+
+		assert.Nilf(t, constructErr, fmt.Sprintf("Unable to construct request: %v", constructErr))
```
Vuksan: Might be better to do `t.Fatalf` here, since you might want the test to stop if this happens
```suggestion
		if constructErr != nil {
			t.Fatalf("Unable to construct request: %v", constructErr)
		}
```

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+		assert.Equalf(t, res.ID, float64(requestID), "Invalid response ID")
+
+		var balanceHex string
+		if wsError := json.Unmarshal(res.Result, &balanceHex); wsError != nil {
+			t.Fatalf("Unable to unmarshal WS result: %v", wsError)
+		}
+
+		foundBalance, parseError := types.ParseUint256orHex(&balanceHex)
+		if parseError != nil {
+			t.Fatalf(fmt.Sprintf("Unable to parse WS result balance: %v", parseError))
```
Vuksan: No need for `Sprintf`, `Fatalf` calls it internally:
```suggestion
			t.Fatalf("Unable to parse WS result balance: %v", parseError)
```

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		if wsError := json.Unmarshal(res.Result, &balanceHex); wsError != nil {
+			t.Fatalf("Unable to parse WS result balance: %v", wsError)
+		}
```
Vuksan: Unmarshalling again seems redundant?

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+		assert.NoError(t, err)
+
+		requestID := 2
+		request, constructErr := constructWSRequest(
+			requestID,
+			"eth_blockNumber",
+			[]string{},
+		)
+
+		assert.Nilf(t, constructErr, fmt.Sprintf("Unable to construct request: %v", constructErr))
```
Vuksan: Might be better to do `t.Fatalf` here, since you might want the test to stop if this happens
```suggestion
		if constructErr != nil {
			t.Fatalf("Unable to construct request: %v", constructErr)
		}
```

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+		JSONRPC: "2.0",
+		Method:  method,
+		ID:      id,
+		Params:  params,
+	}
+
+	return json.Marshal(request)
+}
+
+func getWSResponse(t *testing.T, ws *websocket.Conn, request []byte) (jsonrpc.Response, error) {
```
Vuksan: No need to return `error` here since you fail the test on each error in this function.

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+			t.Fatalf("Unable to unmarshal WS result: %v", wsError)
+		}
+
+		blockNumInt, parseError := types.ParseUint256orHex(&blockNum)
+		if parseError != nil {
+			t.Fatalf(fmt.Sprintf("Unable to parse WS result balance: %v", parseError))
+		}
+		if wsError := json.Unmarshal(res.Result, &blockNum); wsError != nil {
+			t.Fatalf("Unable to parse WS result balance: %v", wsError)
+		}
```
Vuksan: Unmarshalling into `blockNum` again seems redundant?

e2e/websocket_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+		assert.Greaterf(t, blockNumInt.Cmp(big.NewInt(0)), 0, "Invalid block number")
```
Vuksan: Can we do this instead, please?
```suggestion
		assert.Equalf(t, blockNumInt.Cmp(big.NewInt(0)), 1, "Invalid block number")
```

This way it takes me less time to understand we actually check if block number is greater than 0 😅

Vuksan (approved):
GJ on the tests.
Left nitpick comments, LGTM

e2e/websocket_test.go
zivkovicmilos: Fixed 👍

e2e/websocket_test.go
zivkovicmilos: Forgot about this. Good catch 👍

e2e/websocket_test.go
zivkovicmilos: We need to unmarshal `json.RawMessage`. We do this unmarshalling everywhere where we want to grab the result from the JSON RPC endpoints.

e2e/websocket_test.go
zivkovicmilos: Fixed 👍

e2e/websocket_test.go
zivkovicmilos: You're right, it's redundant.
Removing now 👍

e2e/websocket_test.go
zivkovicmilos: We need to unmarshal `json.RawMessage`. We do this unmarshalling everywhere where we want to grab the result from the JSON RPC endpoints.

e2e/websocket_test.go
zivkovicmilos: Changing now 👍

e2e/websocket_test.go
Vuksan: Don't you do that on lines 102-104 already?

e2e/websocket_test.go
Vuksan: Don't you do that on lines 151-153 already?

e2e/websocket_test.go
zivkovicmilos: When we unmarshal the initial response, we get a `json.RawMessage` in the results field. I don't think there is any way we can bypass this second unmarshalling, because it's still not in the correct form (it's []byte)

e2e/websocket_test.go
zivkovicmilos: Replied in the [other comment](https://github.com/0xPolygon/polygon-sdk/pull/94#discussion_r677755793).

e2e/websocket_test.go
zivkovicmilos: I'm obviously blind. I didn't figure out you meant the duplicated section
Removed 👍

e2e/websocket_test.go
zivkovicmilos: Removed 👍

Vuksan (approved):
🚀

PR #101, opened 2021-07-27
added txpool methods

dankostiuk:
# Description

Added the 3 geth RPC endpoints for `txpool` to support `txpool_content`, `txpool_inspect`, and `txpool_status` (see https://geth.ethereum.org/docs/rpc/ns-txpool).

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/jsonrpc/txpool_endpoint.go
+++ b/jsonrpc/txpool_endpoint.go
@@ -35,30 +35,20 @@ func (t *Txpool) Content() (interface{}, error) {
 	for address, nonces := range pendingTxs {
 		pendingRpcTxns[address] = make(map[uint64]*transaction)
 		for nonce, tx := range nonces {
-			// placeholder block to pass toTransaction
-			mockBlock := &types.Block{
-				Header: &types.Header{
-					Hash: types.Hash{0x0},
-					Number: uint64(0),
-				},
-			}
+			blockHash, _ := t.d.store.ReadTxLookup(tx.Hash)
+			block, _ := t.d.store.GetBlockByHash(blockHash, false)
 			// using 0 as txIndex
-			pendingRpcTxns[address][nonce] = toTransaction(tx, mockBlock, 0)
+			pendingRpcTxns[address][nonce] = toTransaction(tx, block, 0)
 		}
   }
 	queuedRpcTxns := make(map[types.Address]map[uint64]*transaction)
 	for address, nonces := range queuedTxs {
 		queuedRpcTxns[address] = make(map[uint64]*transaction)
 		for nonce, tx := range nonces {
-			// placeholder block to pass toTransaction
-			mockBlock := &types.Block{
-				Header: &types.Header{
-					Hash: types.Hash{0x0},
-					Number: uint64(0),
-				},
-			}
+			blockHash, _ := t.d.store.ReadTxLookup(tx.Hash)
+			block, _ := t.d.store.GetBlockByHash(blockHash, false)
 			// using 0 as txIndex
-			queuedRpcTxns[address][nonce] = toTransaction(tx, mockBlock, 0)
+			queuedRpcTxns[address][nonce] = toTransaction(tx, block, 0)
 		}
   }
 
--- a/jsonrpc/txpool_endpoint.go
+++ b/jsonrpc/txpool_endpoint.go
@@ -2,62 +2,100 @@ package jsonrpc
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/0xPolygon/minimal/types"
 )
 
 // Txpool is the txpool jsonrpc endpoint
 type Txpool struct {
-	d *Dispatcher
+  d *Dispatcher
 }
 
 type ContentResponse struct {
-	Pending   map[types.Address]map[uint64]*transaction		`json:"pending"`
-	Queued 		map[types.Address]map[uint64]*transaction		`json:"queued"`
+  Pending   map[types.Address]map[uint64]*gethTransaction   `json:"pending"`
+  Queued    map[types.Address]map[uint64]*gethTransaction   `json:"queued"`
 }
 
 type InspectResponse struct {
-	Pending   map[string]map[string]string	`json:"pending"`
-	Queued		map[string]map[string]string	`json:"queued"`
+  Pending   map[string]map[string]string  `json:"pending"`
+  Queued    map[string]map[string]string  `json:"queued"`
 }
 
 type StatusResponse struct {
-	Pending   uint64	`json:"pending"`
-	Queued 		uint64	`json:"queued"`
+  Pending   uint64  `json:"pending"`
+  Queued    uint64  `json:"queued"`
+}
+
+type gethTransaction struct {
+  Nonce       argUint64      `json:"nonce"`
+  GasPrice    argBig         `json:"gasPrice"`
+  Gas         argUint64      `json:"gas"`
+  To          *types.Address `json:"to"`
+  Value       argBig         `json:"value"`
+  Input       argBytes       `json:"input"`
+  V           argByte        `json:"v"`
+  R           argBytes       `json:"r"`
+  S           argBytes       `json:"s"`
+  Hash        types.Hash     `json:"hash"`
+  From        types.Address  `json:"from"`
+  BlockHash   types.Hash     `json:"blockHash"`
+  BlockNumber interface{}    `json:"blockNumber"`
+  TxIndex     interface{}    `json:"transactionIndex"`
+}
+
+func toGethTransaction(t *types.Transaction) *gethTransaction {
+  if t.R == nil {
+    t.R = []byte{0}
+  }
+  if t.S == nil {
+    t.S = []byte{0}
+  }
+
+  return &gethTransaction{
+    Nonce:        argUint64(t.Nonce),
+    GasPrice:     argBig(*t.GasPrice),
+    Gas:          argUint64(t.Gas),
+    To:           t.To,
+    Value:        argBig(*t.Value),
+    Input:        argBytes(t.Input),
+    V:            argByte(t.V),
+    R:            argBytes(t.R),
+    S:            argBytes(t.S),
+    Hash:         t.Hash,
+    From:         t.From,
+    BlockHash:    types.ZeroHash,
+    BlockNumber:  nil,
+    TxIndex:      nil,
+  }
 }
 
 /** 
  * Content - implemented according to https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content
  */
 func (t *Txpool) Content() (interface{}, error) {
-	pendingTxs, queuedTxs := t.d.store.GetTxs()
-	pendingRpcTxns := make(map[types.Address]map[uint64]*transaction)
-	for address, nonces := range pendingTxs {
-		pendingRpcTxns[address] = make(map[uint64]*transaction)
-		for nonce, tx := range nonces {
-			blockHash, _ := t.d.store.ReadTxLookup(tx.Hash)
-			block, _ := t.d.store.GetBlockByHash(blockHash, false)
-			// using 0 as txIndex
-			pendingRpcTxns[address][nonce] = toTransaction(tx, block, 0)
-		}
+  pendingTxs, queuedTxs := t.d.store.GetTxs()
+  pendingRpcTxns := make(map[types.Address]map[uint64]*gethTransaction)
+  for address, nonces := range pendingTxs {
+    pendingRpcTxns[address] = make(map[uint64]*gethTransaction)
+    for nonce, tx := range nonces {
+      pendingRpcTxns[address][nonce] = toGethTransaction(tx)
+    }
   }
-	queuedRpcTxns := make(map[types.Address]map[uint64]*transaction)
-	for address, nonces := range queuedTxs {
-		queuedRpcTxns[address] = make(map[uint64]*transaction)
-		for nonce, tx := range nonces {
-			blockHash, _ := t.d.store.ReadTxLookup(tx.Hash)
-			block, _ := t.d.store.GetBlockByHash(blockHash, false)
-			// using 0 as txIndex
-			queuedRpcTxns[address][nonce] = toTransaction(tx, block, 0)
-		}
+  queuedRpcTxns := make(map[types.Address]map[uint64]*gethTransaction)
+  for address, nonces := range queuedTxs {
+    queuedRpcTxns[address] = make(map[uint64]*gethTransaction)
+    for nonce, tx := range nonces {
+      queuedRpcTxns[address][nonce] = toGethTransaction(tx)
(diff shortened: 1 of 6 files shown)


jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,115 @@
...
+func (t *Txpool) Content() (interface{}, error) {
+	pendingTxs, queuedTxs := t.d.store.GetTxs()
+	pendingRpcTxns := make(map[types.Address]map[uint64]*transaction)
+	for address, nonces := range pendingTxs {
+		pendingRpcTxns[address] = make(map[uint64]*transaction)
+		for nonce, tx := range nonces {
+			blockHash, _ := t.d.store.ReadTxLookup(tx.Hash)
+			block, _ := t.d.store.GetBlockByHash(blockHash, false)
+			// using 0 as txIndex
+			pendingRpcTxns[address][nonce] = toTransaction(tx, block, 0)
```
Kourin1996: I suppose we can't use `toTransaction` here because tx is not in any block yet and `toTransaction` probably will throw panic. We need to use different function and set "0x00...." to block hash and nil to block number in the response according to docs.

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,115 @@
...
+			pendingRpcTxns[address.String()][fmt.Sprint(nonce)] = msg
+		}
+  }
+
+	queuedRpcTxns := make(map[string]map[string]string)
+	for address, nonces := range queuedTxs {
+		queuedRpcTxns[address.String()] = make(map[string]string)
+		for nonce, tx := range nonces {
+			msg := fmt.Sprintf("%d wei + %d gas x %d wei", tx.Value, tx.Gas, tx.GasPrice)
+			queuedRpcTxns[address.String()][fmt.Sprint(nonce)] = msg
```
Kourin1996: To convert uint64 to string, you can use `strconv.FormatUint` which is more efficient function.

```
key := strconv.FormatUint(nonce, 10)
```

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+    To:           t.To,
+    Value:        argBig(*t.Value),
+    Input:        argBytes(t.Input),
+    V:            argByte(t.V),
+    R:            argBytes(t.R),
+    S:            argBytes(t.S),
+    Hash:         t.Hash,
+    From:         t.From,
+    BlockHash:    types.ZeroHash,
+    BlockNumber:  nil,
```
dankostiuk: blockNumber and transactionIndex must be included and nil according to the geth spec.

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+  S           argBytes       `json:"s"`
+  Hash        types.Hash     `json:"hash"`
+  From        types.Address  `json:"from"`
+  BlockHash   types.Hash     `json:"blockHash"`
+  BlockNumber interface{}    `json:"blockNumber"`
+  TxIndex     interface{}    `json:"transactionIndex"`
+}
+
+func toGethTransaction(t *types.Transaction) *gethTransaction {
+  if t.R == nil {
```
dankostiuk: Here if R or S is `nil`, argBytes will format the response json as `0x` however this isn't an acceptable value. Therefore, I initialized a 0 byte array if either of these fields are `nil` so that the response has values to the right of x, for example `0x0`.

jsonrpc/txpool_endpoint.go
dankostiuk: Thanks @Kourin1996 , I added that and will keep that in mind for any future changes.

txpool/txpool.go
```diff
@@ -203,6 +203,27 @@ func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
 	return nil
 }
 
+// GetTxs gets both queued and pending transactions
+func (t *TxPool) GetTxs() (map[types.Address]map[uint64]*types.Transaction, map[types.Address]map[uint64]*types.Transaction) {
+	queuedTxs := make(map[types.Address]map[uint64]*types.Transaction)
+	queue := t.queue
+	for addr, queuedTxn := range queue {
+		for _, tx := range queuedTxn.txs {
+			queuedTxs[addr] = make(map[uint64]*types.Transaction)
```
Kourin1996: You might want to create new map outside second loop (between line 210 and 211). In current implementation, result shows only one transaction per an account. (According to doc, you can see transactions by multiple nonce if exist)

txpool/txpool.go
```diff
@@ -203,6 +203,27 @@ func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
...
+			queuedTxs[addr][tx.Nonce] = tx
+		}
+	}
+
+	pendingTxs := make(map[types.Address]map[uint64]*types.Transaction)
+	sortedPricedTxs := t.sorted.index
+	for _, sortedPricedTx := range sortedPricedTxs {
+		pendingTxs[sortedPricedTx.from] = make(map[uint64]*types.Transaction)
```
Kourin1996: This is the similar one. It's better to create new map only if not exist.

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+type gethTransaction struct {
+  Nonce       argUint64      `json:"nonce"`
+  GasPrice    argBig         `json:"gasPrice"`
+  Gas         argUint64      `json:"gas"`
+  To          *types.Address `json:"to"`
+  Value       argBig         `json:"value"`
+  Input       argBytes       `json:"input"`
+  V           argByte        `json:"v"`
+  R           argBytes       `json:"r"`
+  S           argBytes       `json:"s"`
```
Kourin1996: You don't need to put V & R & S in the response according to doc https://geth.ethereum.org/docs/rpc/ns-txpool

Kourin1996 (changes requested):
Hello. Thank you for your PRs. I noticed some points and put comment. It'd be great if you would check it.

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+  Pending   map[string]map[string]string  `json:"pending"`
+  Queued    map[string]map[string]string  `json:"queued"`
+}
+
+type StatusResponse struct {
+  Pending   uint64  `json:"pending"`
+  Queued    uint64  `json:"queued"`
+}
+
+type gethTransaction struct {
```
zivkovicmilos: I'm not sure about the naming on this. `gethTransaction` implies that the transaction is conforming to specific geth services, when in fact it is just emulating the same functionality.

jsonrpc/txpool_endpoint.go
zivkovicmilos: Thank you for the explanation. Could you add it to the code as well, since these comments will not be visible in the future when someone is going over it?

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+    From:         t.From,
+    BlockHash:    types.ZeroHash,
+    BlockNumber:  nil,
+    TxIndex:      nil,
+  }
+}
+
+/** 
+ * Content - implemented according to https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content
+ */
```
zivkovicmilos: This is probably a nitpick, but throughout the entire project we've never used comments like this. Please place multiline comments in // instead of /* */. Function comments should have a short description as to what they do, and you can additionaly link to the spec after the explanation

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+    Pending: pendingRpcTxns,
+    Queued: queuedRpcTxns,
+  }
+  
+  return resp, nil
+}
+
+/** 
+ * Inspect - implemented according to https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_inspect
+ */
```
zivkovicmilos: Same as with the previous comment, please conform it to what is present in the codebase

jsonrpc/txpool_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+    Pending: pendingRpcTxns,
+    Queued: queuedRpcTxns,
+  }
+
+  return resp, nil
+}
+
+/** 
+ * Status - implemented according to https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content
+ */
```
zivkovicmilos: Same as with the previous 2 comments, please conform this

zivkovicmilos (changes requested):
Overall it looks good.

Please review the comments I've left, and it will be 💯

txpool/txpool.go
dankostiuk: Thanks @Kourin1996 , for both these comments, I created a new map only if one at the particular address doesn't exist. I found out that the scenario exists where a particular address might not have any txs, so this approach ensures there's always at least 1 tx per address.

Kourin1996 (approved):
Looks good to me 👍 Thank you for new features.

lazartravica (approved):
LGTM! I especially liked how you wrote tests both for the txpool and txpool_endpoint in different places 💯

Kourin1996 (approved):

txpool/txpool.go
```diff
@@ -203,6 +204,31 @@ func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
...
+	pendingTxs := make(map[types.Address]map[uint64]*types.Transaction)
+	sortedPricedTxs := t.sorted.index
+	for _, sortedPricedTx := range sortedPricedTxs {
+		if _, ok := pendingTxs[sortedPricedTx.from]; !ok {
+			pendingTxs[sortedPricedTx.from] = make(map[uint64]*types.Transaction)
+		}
+		pendingTxs[sortedPricedTx.from][sortedPricedTx.tx.Nonce] = sortedPricedTx.tx
+	}
+
+	return queuedTxs, pendingTxs
```
Vuksan: You are returning `queuedTxs, pendingTxs` but in `txpool_endpoint.go` when you call this method you are assigning them in a different order: 
```
pendingTxs, queuedTxs := t.d.store.GetTxs()
```

txpool/txpool_test.go
```diff
@@ -116,6 +117,55 @@ func TestMultipleTransactions(t *testing.T) {
...
+		From:     from3,
+		Nonce:    3,
+		Gas:      validGasLimit,
+		GasPrice: big.NewInt(1),
+	}
+	assert.NoError(t, pool.addImpl("", txn3))
+
+	queuedTxs, pendingTxs := pool.GetTxs()
+
+	assert.Len(t, queuedTxs, 2)
```
Vuksan: Shouldn't there be 3 queued transactions (`txn1, txn2 and txn3`)?

Vuksan (changes requested):
Thank you for adding these methods 🙇 
Can you please take a look at the comments I left, I think you made a small mistake?

txpool/txpool.go
dankostiuk: Thanks @Vuksan! Fixed this. The return is now pending followed by queued txs in all occurrences exactly how is done with geth (https://github.com/ethereum/go-ethereum/blob/d3e3a460ec947c9e1e963d1a35f887d95f23f99d/core/tx_pool.go#L482)

txpool/txpool_test.go
dankostiuk: Yes, however in this test 2 txs were sent from 1 account. I now separated the `from` of each tx so that the length of the queued map is 3 which makes it easier to follow.

ferranbt:
I do not think these methods should be here. Polygon-sdk has two main interfaces: JsonRPC and GRPC. The first one is a generic interface to send Ethereum transactions, think of the minimum methods required to be usable with a client like etherjs. Thus, it was not intended to implement all the Geth namespace.

On the other side, the GRPC interface was meant to be the place to add these utility methods, besides, there is already a TxPool namespace there.

mrwillis:
@ferranbt I see your point of wanting to make the sdk RPC interface as generic as possible but I think these client-specific methods (also this one https://github.com/0xPolygon/polygon-sdk/issues/118) add a lot of value to the project.

We've created this PR to make polygon-sdk compatible with many block explorers as I don't believe the support of gRPC on block explorers in the ethereum ecosystem is strong. If the SDK can conform to a geth or parity interface, polygon-sdk can out of the box work with many explorers and node providers and not require teams to implement them themselves.

I'd also note that clients like ethers.js use these methods in many cases to offer additional functionality. 

Maybe these can be exposed via a flag similar to how geth offers you options to expose different namespaces?

lazartravica:
> I do not think these methods should be here. Polygon-sdk has two main interfaces: JsonRPC and GRPC. The first one is a generic interface to send Ethereum transactions, think of the minimum methods required to be usable with a client like etherjs. Thus, it was not intended to implement all the Geth namespace.
> 
> On the other side, the GRPC interface was meant to be the place to add these utility methods, besides, there is already a TxPool namespace there.

Thank you @ferranbt for keeping track of the pull requests 🔥 

I must say I agree with @mrwillis on this one, as we want to support people connecting polygon-sdk to blockexplorers. On the other side I agree with you that indiscriminately adding endpoints exposed in other clients to the polygon-sdk is not something that we should do, but rather keep the set of endpoints to a functional minimum.

I guess the only point where we have a conflict regarding this is whether querying the txpool should be a part of the minimal set of JSON-RPC methods or not, or is there a deeper technical issue that you see here?

Vuksan (approved):
From a technical perspective, you have my approval.
Thank you for the contribution and for making the requested changes.

As for the strategic decision, let's wait for @ferranbt to answer @lazartravica's comment.

ferranbt:
The only technical issue is to maintain the same sets of methods in two places (JSON-RPC and GRPC) and how much will grow the interface between the JSON-RPC layer.

For me, the minimal set of JSON-RPC methods should be the ones implemented in Infura since that is the main gateway used by Ethereum compatible end users. I understand that having block exporters is important but not sure at which point you draw the line.

Vuksan:
This is how Celo did it: https://github.com/celo-org/celo-blockchain/blob/master/internal/web3ext/web3ext.go#L664-L688

ferranbt:
There is something else I forgot. If you include debug or txpool namespace directly on JSON-RPC you will likely have to add support to enable different namespaces for each transport (i.e. rpcapi and wsapi in Geth) since you are mixing endpoints for two different stakeholders.

A different approach would be to implement these methods on the GRPC layer only and then have a separate microservice/repo that translates JSON-RPC calls to the GRPC endpoints.

lazartravica:
@ferranbt Thank you for the comments! Since this is a community opened PR, I will close the pull request and merge, so that we can redesign the way we expose non-essential APIs  internally.

PR #102, opened 2021-07-28
Fix private key reads from .key files

zivkovicmilos:
# Description

This PR fixes a bug with how private keys are read from .key files.

When storing keys, the polygon SDK converts them to a human readable format - base64, and saves those bytes to the .key file. However, during key read from the file, the opposite conversion never occurred, meaning the SDK never converted back from the read base64 to a base byte value.
This caused the parsed private key to be wrong (the base64 byte representation was read, instead of the serialized byte value). Likewise, the addresses retrieved from this kind of key did not match the actual private key / address pair in all `.key` files.

This PR also includes support for [EIP-55 ](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md)(address checksums)

# Changes include

# Breaking changes

Since all addresses that were read earlier are in fact invalid, this fix could cause earlier versions of the SDK to not be backwards compatible in places where addresses are stored long term.

# Checklist

- [x] I have tested this code


types/types.go
```diff
@@ -56,13 +58,41 @@ func (h *Hash) Scan(src interface{}) error {
 	return nil
 }
 
-func (a Address) EIP55() string {
-	// TODO
-	return hex.EncodeToHex(a[:])
+// eip55 returns the checksummed address with 0x prefix
+// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md
+func (a Address) eip55() string {
```
zivkovicmilos: @Kourin1996 
There has got to be a way to make this parsing more efficient. Do you have any ideas?
I haven't benched it, but maybe we can use some golang tricks and slice optimizations.

Curious to hear your thoughts as well @lazartravica @Vuksan

Kourin1996: @zivkovicmilos 
If there are following premises, it's possible to optimize it
+ the length of given address and result address are same each other
+ the address consists only hex characters (0-9, A-F)

In this case, I think we can use a slice with fixed instead of dynamically concat. And we can compare character by comparison operator.

I wrote simple benchmarks and ran them. We may be able to optimize.

```go
package main

import (
	"strconv"
	"strings"
	"testing"
	"unicode"
)

const (
	OriginalAddr = "001d3f1ef827552ae1114027bd3ecf1f086ba0f9"
	HashedAddr   = "23a69c1653e4ebbb619b0b2cb8a9bad49892a8b9695d9a19d8f673ca991deae1"
	ExpectedAddr = "001d3F1ef827552Ae1114027BD3ECF1f086bA0F9"
)

// Based on original code
func ComvertBySimpleConcat() string {
	result := ""
	for idx, ch := range OriginalAddr {
		if ch >= '0' && ch <= '9' {
			result += string(ch)
		} else {
			hashedNibble, _ := strconv.ParseInt(string(HashedAddr[idx]), 16, 64)
			if hashedNibble > 7 {
				result += strings.ToUpper(string(ch))
			} else {
				result += string(ch)
			}
		}
	}
	return result
}

// Use fixed slice of rune
func ConvertWithFixedByteSlice() string {
	result := make([]rune, len(OriginalAddr))
	for idx, ch := range OriginalAddr {
		if ch >= '0' && ch <= '9' {
			result[idx] = ch
		} else {
			hashedNibble, _ := strconv.ParseInt(string(HashedAddr[idx]), 16, 64)
			if hashedNibble > 7 {
				result[idx] = unicode.ToUpper(ch)
			} else {
				result[idx] = ch
			}
		}
	}
	return string(result)
}

// Use fixed slice + optimized compare
func ComvertWithOptimizedCompare() string {
	result := make([]rune, len(OriginalAddr))
	for idx, ch := range OriginalAddr {
                // we expect ch and HashedAddr[idx] are any of hex character
                // Invert condition and check if hashed ch is less than or equal to '7' 
                // instead of checking if hashed ch is greater than or equal to '8'
		if ch >= '0' && ch <= '9' || HashedAddr[idx] >= '0' && HashedAddr[idx] <= '7' {
			result[idx] = ch
		} else {
			result[idx] = unicode.ToUpper(ch)
		}
	}
	return string(result)
}

func BenchmarkComvertBySimpleConcat(b *testing.B) {
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		result := ComvertBySimpleConcat()
		if result != ExpectedAddr {
			panic("wrong result")
		}
	}
}

func BenchmarkConvertWithFixedByteSlice(b *testing.B) {
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		result := ConvertWithFixedByteSlice()
		if result != ExpectedAddr {
			panic("wrong result")
		}
	}
}

func BenchmarkComvertWithOptimizedCompare(b *testing.B) {
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		result := ComvertWithOptimizedCompare()
		if result != ExpectedAddr {
			panic("wrong result")
		}
	}
}
```

You can run this benchmark by this command:
```shell
$ go test -bench . -benchmem
```

Result in my environment is as follows:
```shell
goos: darwin
goarch: amd64
pkg: github.com/Kourin1996/benchmark1
BenchmarkComvertBySimpleConcat-8                  342128              3104 ns/op            1312 B/op        103 allocs/op
BenchmarkConvertWithFixedByteSlice-8             1281477               899 ns/op             108 B/op         16 allocs/op
BenchmarkComvertWithOptimizedCompare-8           2979254               409 ns/op              48 B/op          1 allocs/op
PASS
ok      github.com/Kourin1996/benchmark1        5.171s
```

crypto/crypto.go
```diff
@@ -220,35 +219,47 @@ func toECDSA(d []byte, strict bool) (*ecdsa.PrivateKey, error) {
...
 
-// ReadPrivKey reads a private key from the file path
-func ReadPrivKey(path string) (*ecdsa.PrivateKey, error) {
-	createFn := func() ([]byte, error) {
-		key, err := GenerateKey()
-		if err != nil {
-			return nil, err
-		}
+// generateKeyAndMarshall generates a new private key and serializes it to a byte array
+func generateKeyAndMarshall() ([]byte, error) {
```
lazartravica: `Marshall` is the military title, `Marshal` is a verb meaning putting something in order.

This is me giving back the favour for `shouldSuccess` :D

lazartravica: Just kidding ofcourse I just know the https://pkg.go.dev/encoding/json package has Marshal methods with single `l`

helper/keystore/keystore.go
```diff
@@ -11,30 +11,44 @@ type createFn func() ([]byte, error)
...
-	raw, err := ioutil.ReadFile(path)
-	if err != nil {
-		return nil, err
+		// Read the Base64 encoded key from the disk.
+		// Base64 decoding is done in the read method
+		keyBuff, err = ioutil.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read private key from disk (%s), %v", path, err)
+		}
 	}
```
lazartravica: You could get rid of if/else and just do an early return of the key read from disk if it exists. In my opinion it's more readable that way than having to go back and to see what the initial `if` was for in whose `else` branch we are in.

lazartravica (changes requested):
At first I was thinking that we can get rid of the `keystore.go` entirely and thus reduce the complexity but no need to do that since I think the separation of concern between disk reads/writes (in `keystore.go`) and cryptographic functions in `crypto.go` is good.

Let's discuss if it makes sense to move the `GenerateOrReadPrivateKey` entirely to the `keystore.go`, and  reduce the complexity by not receiving callback functions as arguments but by calling the `crypto.generateKeyAndMarshall()` and `crypto.bytesToPrivateKey()` since we don't have, and don't see having any other implementations any time soon.

Perha

crypto/crypto.go
zivkovicmilos: Fixed the typo for other functions as well in the `crypto.go` file 👍

helper/keystore/keystore.go
zivkovicmilos: You're right. It's probably unnecessarily complicated. 

I've simplified it a bit, and moved out the common read part out to `crypto.go` 👍

types/types.go
zivkovicmilos: Thank you for the detailed answer and for taking the time to benchmark different solutions 💯 

I was expecting a 1-2 sentence reply but you went above and beyond. Both of your assumptions are correct.
I've replaced the bulk of the eip55 method with the code you suggested with `ConvertWithOptimizedCompare` 👍

Kourin1996 (approved):
LGTM 👌

helper/keystore/keystore.go
```diff
@@ -9,32 +9,39 @@ import (
 
 type createFn func() ([]byte, error)
 
-type readFn func(b []byte) (interface{}, error)
-
-func CreateIfNotExists(path string, create createFn, read readFn) (interface{}, error) {
+// CreateIfNotExists generates a private key at the specified path,
+// or reads it if a key file is present
+func CreateIfNotExists(path string, create createFn) ([]byte, error) {
```
Vuksan: More true:
```suggestion
// CreateIfNotExists generates a private key at the specified path,
// or reads the file on that path if it is present
func CreateIfNotExists(path string, create createFn) ([]byte, error) {
```

helper/keystore/keystore.go
```diff
@@ -9,32 +9,39 @@ import (
...
-		// create the file
-		buf, err := create()
+
+	var keyBuff []byte
+	if !os.IsNotExist(err) {
+		// Key exists
+
+		// Read the Base64 encoded key from the disk.
+		// Base64 decoding is done later on
+		keyBuff, err = ioutil.ReadFile(path)
```
Vuksan: I would remove this comment, since it lies a bit 🙂 
it reads **anything** on the `path`. Doesn't care if it's a key or not.
```suggestion
		keyBuff, err = ioutil.ReadFile(path)
```

types/types.go
```diff
@@ -56,13 +58,33 @@ func (h *Hash) Scan(src interface{}) error {
...
+	addrBytes := a.Bytes() // 20 bytes
+
+	// Encode to hex without the 0x prefix
+	lowercaseHex := hex.EncodeToHex(addrBytes)[2:]
+	hashedAddress := hex.EncodeToHex(keccak.Keccak256(nil, []byte(lowercaseHex)))[2:]
+
+	result := make([]rune, len(lowercaseHex))
+	// Iterate over each character in the hashed address
+	for idx, ch := range lowercaseHex {
```
Vuksan: either the comment or code is wrong. It says `iterate over **hashed** address` but does `range lowercaseHex`?

types/types.go
Vuksan: Please rename this function to what it does.
Comment above it is enough for EIP origin IMO
```suggestion
func (a Address) checksumEncode() string {
```

Vuksan (changes requested):
It would be great if you could add tests for `eip55`.
TestTable with as many addresses as you can find (both valid and invalid, you can start from EIP55 test addresses) would be awesome.

helper/keystore/keystore.go
zivkovicmilos: I'll try to make sure the comments are more aligned to the general purpose in the future, and not just to the context we use them in 👍 

Fixed 👍

helper/keystore/keystore.go
zivkovicmilos: Fixed 👍

types/types.go
zivkovicmilos: Comments lie 🥸

Fixed 👍

zivkovicmilos:
> It would be great if you could add tests for `eip55`.
> TestTable with as many addresses as you can find (both valid and invalid, you can start from EIP55 test addresses) would be awesome.

@Vuksan 
Looks like we have the EIP55 tests ready in the `types_test.go`, written by Ferran 3.6.2019. 🙂🇪🇸
I'll remove the test skip on them, and add a couple more test cases 👍

lazartravica (approved):
LGTM 💯

Vuksan (approved):
💯

PR #103, opened 2021-07-28
Fix transaction receipt failed

lazartravica:
# Description

Things that were changed in order of impact:

- Fixed a bug where invalid transactions still end up in the block and have transaction receipts.
- Refactored the way VM (Runtime) executions are reported throughout the code to use the `ExecutionResult` struct.
- Added comments and made it clearer which errors are due to VM execution and which due to breaking the consensus rules while processing transactions.
- `Txn.SubBalance` now returns an error itself, avoiding checks whenever it is called
- Added overflow checks when calculating zeroes & ones cost in data in the intrinsic gas cost 
- General code cleanup, refactor and variable renaming.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -205,8 +205,14 @@ func TestEthTransfer(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			defer cancel()
 			receipt, err := srv.WaitForReceipt(ctx, txnHash)
-			assert.NoError(t, err)
-			assert.NotNil(t, receipt)
+
+			if testCase.shouldSuccess {
+				assert.NoError(t, err)
+				assert.NotNil(t, receipt)
+			} else { // When an invalid transaction is supplied, there should be no receipt.
+				assert.Error(t, err)
+				assert.Nil(t, receipt)
+			}
 
 			if testCase.shouldSuccess {
 				fee = new(big.Int).Mul(
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -214,13 +214,6 @@ func TestEthTransfer(t *testing.T) {
 				assert.Nil(t, receipt)
 			}
 
-			if testCase.shouldSucceed {
-				fee = new(big.Int).Mul(
-					big.NewInt(int64(receipt.GasUsed)),
-					big.NewInt(int64(txnObject.GasPrice)),
-				)
-			}
-
 			// Fetch the balances after sending
 			balanceSender, err = rpcClient.Eth().GetBalance(
 				web3.Address(testCase.sender),
@@ -236,6 +229,11 @@ func TestEthTransfer(t *testing.T) {
 
 			expectedSenderBalance := previousSenderBalance
 			if testCase.shouldSucceed {
+				fee = new(big.Int).Mul(
+					big.NewInt(int64(receipt.GasUsed)),
+					big.NewInt(int64(txnObject.GasPrice)),
+				)
+
 				expectedSenderBalance = previousSenderBalance.Sub(
 					previousSenderBalance,
 					new(big.Int).Add(testCase.amount, fee),
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -128,7 +128,7 @@ func TestEthTransfer(t *testing.T) {
 		sender        types.Address
 		recipient     types.Address
 		amount        *big.Int
-		shouldSuccess bool
+		shouldSucceed bool
 	}{
 		{
 			// ACC #1 -> ACC #3
@@ -206,7 +206,7 @@ func TestEthTransfer(t *testing.T) {
 			defer cancel()
 			receipt, err := srv.WaitForReceipt(ctx, txnHash)
 
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
 				assert.NoError(t, err)
 				assert.NotNil(t, receipt)
 			} else { // When an invalid transaction is supplied, there should be no receipt.
@@ -214,7 +214,7 @@ func TestEthTransfer(t *testing.T) {
 				assert.Nil(t, receipt)
 			}
 
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
 				fee = new(big.Int).Mul(
 					big.NewInt(int64(receipt.GasUsed)),
 					big.NewInt(int64(txnObject.GasPrice)),
@@ -235,14 +235,14 @@ func TestEthTransfer(t *testing.T) {
 			assert.NoError(t, err)
 
 			expectedSenderBalance := previousSenderBalance
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
 				expectedSenderBalance = previousSenderBalance.Sub(
 					previousSenderBalance,
 					new(big.Int).Add(testCase.amount, fee),
 				)
 			}
 			expectedReceiverBalance := previousReceiverBalance
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
 				expectedReceiverBalance = previousReceiverBalance.Add(
 					previousReceiverBalance,
 					testCase.amount,
--- a/minimal/server.go
+++ b/minimal/server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/0xPolygon/minimal/state/runtime"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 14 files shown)


e2e/transaction_test.go
```diff
@@ -205,8 +205,14 @@ func TestEthTransfer(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			defer cancel()
 			receipt, err := srv.WaitForReceipt(ctx, txnHash)
-			assert.NoError(t, err)
-			assert.NotNil(t, receipt)
+
+			if testCase.shouldSuccess {
```
zivkovicmilos: I think we have this typo in other tests as well, and it just got propagated.
This should be `shouldSucceed` instead of `shouldSuccess`

minimal/server.go
```diff
@@ -277,25 +278,21 @@ func (j *jsonRPCHub) GetCode(hash types.Hash) ([]byte, error) {
...
 
-	_, failed, err := transition.Apply(txn)
-
-	if err != nil {
-		return nil, false, err
-	}
+	result, err = transition.Apply(txn)
 
-	return transition.ReturnValue(), failed, nil
+	return
```
zivkovicmilos, line 296: In case anyone was as confused as me, these are called named return statements:
https://golang.org/ref/spec#Return_statements

zivkovicmilos (approved):
We definitely needed some kind of error standardization in the execution layer.
I also learned something new about Go, which is a plus 🙂

Everything looks good to me 💯

state/runtime/evm/evm.go
```diff
@@ -17,7 +17,7 @@ func NewEVM() *EVM {
 }
 
 // CanRun implements the runtime interface
-func (e *EVM) CanRun(c *runtime.Contract, host runtime.Host, config *chain.ForksInTime) bool {
+func (e *EVM) CanRun(_ *runtime.Contract, _ runtime.Host, _ *chain.ForksInTime) bool {
```
Vuksan: You can omit the underscores as well if we don't use any parameter 🙂

state/runtime/precompiled/precompiled.go
```diff
@@ -96,20 +92,33 @@ func (p *Precompiled) Name() string {
...
+	result := &runtime.ExecutionResult{
+		ReturnValue: returnValue,
+		GasLeft:     c.Gas,
+		Err:         err,
 	}
-	return ret, c.Gas, err
+
+	if result.Failed() {
+		result.GasLeft = 0
+		result.ReturnValue = nil
```
Vuksan, line 118: What do you think of not overwriting `ReturnValue`, but leaving it populated?
It might provide better insight to the caller. He can see that it failed from the returned error.

state/executor.go
```diff
@@ -276,9 +267,9 @@ func (t *Transition) GetTxnHash() types.Hash {
 }
 
 // Apply applies a new transaction
-func (t *Transition) Apply(msg *types.Transaction) (uint64, bool, error) {
+func (t *Transition) Apply(msg *types.Transaction) (result *runtime.ExecutionResult, err error) {
```
Vuksan, line 276: Would it be wise to name this `err` a bit more descriptive, since `ExecutionResult` also contains an error?

Might be I'm overthinking it. If you find no reason for this disregard this comment, just want you to take another look.

state/executor.go
```diff
@@ -309,70 +299,115 @@ func (t *Transition) transactionGasCost(msg *types.Transaction) uint64 {
...
 	if err != nil {
-		return nil, 0, false, err
+		return
 	}
-	// TODO: Check if this is even possible
-	if leftoverGas > msg.Gas {
-		return nil, 0, false, errorVMOutOfGas
+
+	// 5. the purchased gas is enough to cover intrinsic usage
+	gasAvailable := msg.Gas - intrinsicGasCost
```
Vuksan: Did you forget to rename this one to `gasLeft`?
I see in other places we stick to `gasLeft` nomenclature.

state/executor.go
```diff
@@ -495,63 +520,79 @@ func (t *Transition) hasCodeOrNonce(addr types.Address) bool {
...
+			Err:     runtime.ErrMaxCodeSizeExceeded,
+		}
 	}
 
-	gasCost := uint64(len(code)) * 200
+	gasCost := uint64(len(result.ReturnValue)) * 200
+
+	if result.GasLeft < gasCost {
+		result.Err = runtime.ErrCodeStoreOutOfGas
+		result.ReturnValue = nil
```
Vuksan, line 546: Same thought as in `precompiled.go`:
What do you think of not overwriting `ReturnValue`, but leaving it populated?
It might provide better insight to the caller. He can see that it failed from the returned error.

e2e/transaction_test.go
```diff
@@ -205,8 +205,14 @@ func TestEthTransfer(t *testing.T) {
...
+				assert.NoError(t, err)
+				assert.NotNil(t, receipt)
```
Vuksan: You can merge this with the next if:
```suggestion
			if testCase.shouldSuccess {
				assert.NoError(t, err)
				assert.NotNil(t, receipt)
				fee = new(big.Int).Mul(
					big.NewInt(int64(receipt.GasUsed)),
					big.NewInt(int64(txnObject.GasPrice)),
				)
```

state/executor.go
```diff
@@ -198,11 +188,12 @@ func (t *Transition) Write(txn *types.Transaction) error {
 	// Make a local copy and apply the transaction
 	msg := txn.Copy()
 
-	gasUsed, failed, err := t.Apply(msg)
+	result, err := t.Apply(msg)
 	if err != nil {
 		fmt.Printf("Apply err: %v", err)
+		return err
```
Vuksan, line 200: Not sure that we should exit here.
Otherwise, we might never produce a receipt with a `Failed` status.

I might be wrong, please have another look.

Vuksan (changes requested):
Great job on this refactor!
Code is much more readable now. 💯 

Loved the way you decided to write `Succeeded` method although `Failed` was there. Lowering the cognitive load required to understand the code is never a bad thing 🍻 

Would've loved to see a test where we expect failed receipts, but I guess we will cover that as part of other functionality in the future.

Most of the comments are really nitpicks and there for discussion, aside from [this one](https://github.com/0xPolygon/polygon-sdk/pull/103#discussion_r679389001). Please double-check that. If I'm right I hope we can improve the tests to catch that in the future.

state/runtime/runtime.go
```diff
@@ -67,11 +67,37 @@ type Host interface {
...
+	GasLeft     uint64 // Total gas left as result of execution
+	GasUsed     uint64 // Total gas used as result of execution
+	Err         error  // Any error encountered during the execution, listed below
+}
+
+func (r *ExecutionResult) Succeeded() bool { return r.Err == nil }
+func (r *ExecutionResult) Failed() bool    { return r.Err != nil }
+func (r *ExecutionResult) Reverted() bool  { return r.Err == ErrExecutionReverted }
+
+func (r *ExecutionResult) CalculateGasUsed(gasLimit uint64, refund uint64) {
```
Kourin1996: It may be more proper name begins with `Update`or`Apply`?

Kourin1996 (approved):
Looks good to me 👏
It'll make the codes simpler and easier to understand.

state/runtime/precompiled/precompiled.go
lazartravica: Although that makes sense, I am not sure of the side effects that may happen due to that, as the EVM execution that stopped due to an error should have no `returnValue`, as seen in polygon-sdk and in geth (https://github.com/ethereum/go-ethereum/blob/f68a68a313a7a5806b0e414965e500fef453306a/core/vm/interpreter.go#L295)

state/executor.go
lazartravica: I have left the reply on the first comment

e2e/transaction_test.go
lazartravica: 👍 https://github.com/0xPolygon/polygon-sdk/pull/103/commits/397ef22344def383649dce484f776a463e4cc428

state/runtime/runtime.go
lazartravica: 👍 https://github.com/0xPolygon/polygon-sdk/pull/103/commits/6d20c6500cdc85274366e5029673535f37cefca3

state/runtime/evm/evm.go
lazartravica: 👍 didn't know that, great https://github.com/0xPolygon/polygon-sdk/pull/103/commits/efd2c2232d3b1c341d5227ec84ccdf0acf1b4674

state/executor.go
lazartravica: @Vuksan the `err` that we are checking against here is a violation of a consensus rule, and not a VM error which we check with `ExecutionResult.Failed()`. Because of that, the presence of the `err` value should not end up creating a receipt as it broke the consensus rules.

state/executor.go
lazartravica: 👍 https://github.com/0xPolygon/polygon-sdk/pull/103/commits/5f3e45780a1d65ccf4768e4fcfab102e3130b7c6

e2e/transaction_test.go
lazartravica: 👍 https://github.com/0xPolygon/polygon-sdk/pull/103/commits/d1cf11131d621b758cadb8ec2ee18fc13b3cb630

lazartravica:
@Vuksan I have merged your changes and this PR is ready for review. https://github.com/0xPolygon/polygon-sdk/pull/103/commits/fa39947e99facaeb4078c1d4b42a4f9a95cfe017

@zivkovicmilos The issue with failing websocket tests was a benign one, the premined accounts' balances were too low to do anything. https://github.com/0xPolygon/polygon-sdk/pull/103/commits/72676c05bc7ed4918304eafc1b3bee89e98f4bef

state/executor.go
Vuksan: You're completely right. 💯

Vuksan (approved):
Thanks for making the requested changes 1 commit each 🙏

Issue #91, opened 2021-07-20
Memory leak 


mrwillis:
### Description

It seems like there is a memory leak somewhere in the code. Happens after a while of producing blocks leaving us to believe it's something related to block production. CPU usage remains constant. 

### Your environment

* OS and version
AWS linux
* version of the Polygon SDK
`6906ed9`
* branch that causes this issue
`develop`

### Steps to reproduce

* Tell us how to reproduce this issue <br />
`polygon-sdk server run` with IBFT consensus of 4 nodes running for about 12 hours. 
* Where the issue is, if you know <br />
* Which commands triggered the issue, if any
`polygon-sdk server run`..

### Expected behaviour

No memory leak

### Actual behaviour

Memory leak causing an OOM eerror causing a crash

### Logs

```
9053 2021-07-20T16:42:52.019Z [INFO]  polygon.consensus.ibft.acceptState: Accept state: sequence=2210
 9054 2021-07-20T16:42:52.019Z [INFO]  polygon.consensus.ibft: current snapshot: validators=4 votes=0
 9055 2021-07-20T16:42:52.020Z [INFO]  polygon.consensus.ibft: proposer calculated: proposer=0x2d48d27cf3eec87c82f39527934df7bd2124b30a block=2210
 9056 2021-07-20T16:42:52.020Z [DEBUG] polygon.consensus.syncer: enqueue block: peer=16Uiu2HAm1k16mVYWaG1aoG7rZkbuaETqZ6vhV85vB9vtsjXmYbyz number=2209 hash=0x921643448243c439e2db6c1414c7254f2f4e05d78aa7ecb742bec4cd04259a36
 9057 fatal error: runtime: out of memory
 9058
 9059 runtime stack:
 9060 runtime.throw(0x1006f20, 0x16)
 9061         /usr/local/go/src/runtime/panic.go:1117 +0x72
 9062 runtime.sysMap(0xc008000000, 0x4000000, 0x1a90370)
 9063         /usr/local/go/src/runtime/mem_linux.go:169 +0xc6
 9064 runtime.(*mheap).sysAlloc(0x1a76840, 0x400000, 0x42dc97, 0x1a76848)
 9065         /usr/local/go/src/runtime/malloc.go:729 +0x1e5
 9066 runtime.(*mheap).grow(0x1a76840, 0x4, 0x0)
 9067         /usr/local/go/src/runtime/mheap.go:1346 +0x85
[...]

PR #105, opened 2021-07-29
Fix stuck goroutines issue in IBFT

Kourin1996:
# Description

Relevant issue: #91

This PR fixes the issue that goroutines spawned by `randomTimeout()` in consensus/ibft/ibft.go will be stuck.
This function returns a channel that emits signal after the time calculated by current IBFT round in order to use timeout. It make new channel and spawn a goroutine, emit signal to the channel from the gorountine. Note that this goroutine will not terminate until someone receives signal from the channel once. 

https://github.com/0xPolygon/polygon-sdk/blob/a0618c48b8896fb994f8ef42699617cb81fc4af0/consensus/ibft/ibft.go#L847-L862

Currently this returned channel will be passed to `getNextMessage` and used there. This function waits for a message with current view number by timeout. This function will never consume any signal from `stopCh` if it can receive expected message before timeout.

https://github.com/0xPolygon/polygon-sdk/blob/a0618c48b8896fb994f8ef42699617cb81fc4af0/consensus/ibft/ibft.go#L948-L974

As a result, in many cases, no one receives data from the channel, therefore goroutine in the function will never terminate.

This PR fixes the issue. `randomTimeout` will not spawn gorountine and just return the amount of timeout.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

You can check behaviour with sample code:
https://play.golang.org/p/Z0MpDCvFHPM

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -506,9 +506,9 @@ func (i *Ibft) runAcceptState() { // start new round
 	// we are NOT a proposer for the block. Then, we have to wait
 	// for a pre-prepare message from the proposer
 
-	timerCh := i.randomTimeout()
+	timeout := i.randomTimeout()
 	for i.getState() == AcceptState {
-		msg, ok := i.getNextMessage(timerCh)
+		msg, ok := i.getNextMessage(timeout)
 		if !ok {
 			return
 		}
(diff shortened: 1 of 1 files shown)


zivkovicmilos (approved):
This bug has been around for quite some time, but only under certain conditions that largely went unnoticed.

Great stuff, really like the change from returning a channel to using `time.After`. It's sleeker this way 💯

lazartravica (approved):
Simplified a lot, great 💯

Vuksan (approved):
Excellent PR, and nice catch! 🙇

PR #111, opened 2021-08-02
Transaction validation

zivkovicmilos:
# Description

This PR includes basic transaction validation rules, located within `validateTx`.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -169,7 +169,7 @@ func TestEthTransfer(t *testing.T) {
 	rpcClient := srv.JSONRPC()
 	for indx, testCase := range testTable {
 		t.Run(testCase.name, func(t *testing.T) {
-			// @Vuksan Please remove this skip statement when JSON-RPC Error wrapping is added
+			// TODO @Vuksan Please remove this skip statement when JSON-RPC Error wrapping is added
 			if indx == 1 {
 				t.Skip()
 			}
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -129,7 +129,7 @@ func TestEthTransfer(t *testing.T) {
 		sender        types.Address
 		recipient     types.Address
 		amount        *big.Int
-		shouldSuccess bool
+		shouldSucceed bool
 	}{
 		{
 			// ACC #1 -> ACC #3
@@ -211,14 +211,13 @@ func TestEthTransfer(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			defer cancel()
 			receipt, err := srv.WaitForReceipt(ctx, txnHash)
-			assert.NoError(t, err)
-			assert.NotNil(t, receipt)
 
-			if testCase.shouldSuccess {
-				fee = new(big.Int).Mul(
-					big.NewInt(int64(receipt.GasUsed)),
-					big.NewInt(int64(txnObject.GasPrice)),
-				)
+			if testCase.shouldSucceed {
+				assert.NoError(t, err)
+				assert.NotNil(t, receipt)
+			} else { // When an invalid transaction is supplied, there should be no receipt.
+				assert.Error(t, err)
+				assert.Nil(t, receipt)
 			}
 
 			// Fetch the balances after sending
@@ -235,14 +234,19 @@ func TestEthTransfer(t *testing.T) {
 			assert.NoError(t, err)
 
 			expectedSenderBalance := previousSenderBalance
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
+				fee = new(big.Int).Mul(
+					big.NewInt(int64(receipt.GasUsed)),
+					big.NewInt(int64(txnObject.GasPrice)),
+				)
+
 				expectedSenderBalance = previousSenderBalance.Sub(
 					previousSenderBalance,
 					new(big.Int).Add(testCase.amount, fee),
 				)
 			}
 			expectedReceiverBalance := previousReceiverBalance
-			if testCase.shouldSuccess {
+			if testCase.shouldSucceed {
 				expectedReceiverBalance = previousReceiverBalance.Add(
 					previousReceiverBalance,
 					testCase.amount,
--- a/e2e/transaction_test.go
+++ b/e2e/transaction_test.go
@@ -167,8 +167,13 @@ func TestEthTransfer(t *testing.T) {
 	srv := srvs[0]
 
 	rpcClient := srv.JSONRPC()
-	for _, testCase := range testTable {
+	for indx, testCase := range testTable {
 		t.Run(testCase.name, func(t *testing.T) {
+			// @Vuksan Please remove this skip statement when JSON-RPC Error wrapping is added
+			if indx == 1 {
+				t.Skip()
+			}
+
 			// Fetch the balances before sending
 			balanceSender, err := rpcClient.Eth().GetBalance(
 				web3.Address(testCase.sender),
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -1,0 +1,186 @@
+package e2e
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"io"
+	"math/big"
+	"testing"
+
+	"github.com/0xPolygon/minimal/crypto"
+	"github.com/0xPolygon/minimal/e2e/framework"
+	"github.com/0xPolygon/minimal/helper/tests"
+	"github.com/0xPolygon/minimal/txpool"
+	txpoolOp "github.com/0xPolygon/minimal/txpool/proto"
+	"github.com/0xPolygon/minimal/types"
+	"github.com/golang/protobuf/ptypes/any"
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	oneEth = framework.EthToWei(1)
+	signer = crypto.NewEIP155Signer(100)
+)
+
+func waitForBlock(t *testing.T, srv *framework.TestServer, expectedBlocks int, index int) int64 {
+	systemClient := srv.Operator()
+	ctx, cancelFn := context.WithCancel(context.Background())
+	stream, err := systemClient.Subscribe(ctx, &empty.Empty{})
+	if err != nil {
+		cancelFn()
+		t.Fatalf("Unable to subscribe to blockchain events")
+	}
+
+	evnt, err := stream.Recv()
+	if err == io.EOF {
+		t.Fatalf("Invalid stream close")
+	}
+	if err != nil {
+		t.Fatalf("Unable to read blockchain event")
+	}
+
+	if len(evnt.Added) != expectedBlocks {
+		t.Fatalf("Invalid number of blocks added")
+	}
+
+	cancelFn()
+
+	return evnt.Added[index].Number
+}
+
+type generateTxReqParams struct {
+	nonce         uint64
+	referenceAddr types.Address
+	referenceKey  *ecdsa.PrivateKey
+	toAddress     types.Address
+	gasPrice      *big.Int
+	value         *big.Int
+	t             *testing.T
+}
+
+func generateTx(params generateTxReqParams) *types.Transaction {
+	signedTx, signErr := signer.SignTx(&types.Transaction{
+		Nonce:    params.nonce,
+		From:     params.referenceAddr,
+		To:       &params.toAddress,
+		GasPrice: params.gasPrice,
+		Gas:      1000000,
+		Value:    params.value,
+		V:        1, // it is necessary to encode in rlp
+	}, params.referenceKey)
+
+	if signErr != nil {
+		params.t.Fatalf("Unable to sign transaction, %v", signErr)
+	}
+
+	return signedTx
+}
+
+func generateReq(params generateTxReqParams) *txpoolOp.AddTxnReq {
+	msg := &txpoolOp.AddTxnReq{
+		Raw: &any.Any{
+			Value: generateTx(params).MarshalRLP(),
+		},
+		From: types.ZeroAddress.String(),
+	}
+
+	return msg
+}
+
+func TestTxpool_ErrNonceTooLow(t *testing.T) {
+	// Test scenario:
+	// Add tx with nonce 0
+	// -> Check if tx has been parsed
+	// Add tx with nonce 0
+	// -> tx shouldn't be added, since the nonce is too low
+
+	// Predefined values
+	gasPrice := big.NewInt(10000)
+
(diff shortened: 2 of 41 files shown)


Kourin1996 (approved):
Looks good to me 👍 It's clear.

e2e/transaction_test.go
```diff
@@ -167,8 +167,13 @@ func TestEthTransfer(t *testing.T) {
 	srv := srvs[0]
 
 	rpcClient := srv.JSONRPC()
-	for _, testCase := range testTable {
+	for indx, testCase := range testTable {
 		t.Run(testCase.name, func(t *testing.T) {
+			// @Vuksan Please remove this skip statement when JSON-RPC Error wrapping is added
```
Vuksan: Put `TODO:` in front pls

txpool/txpool.go
```diff
@@ -25,12 +26,18 @@ const (
 )
 
 var (
-	ErrIntrinsicGas = errors.New("intrinsic gas too low")
+	ErrIntrinsicGas      = errors.New("intrinsic gas too low")
+	ErrNegativeValue     = errors.New("negative value")
+	ErrNonEncryptedTxn   = errors.New("non-encrypted transaction")
+	ErrInvalidSender     = errors.New("invalid sender")
+	ErrNonceTooLow       = errors.New("nonce too low")
+	ErrInsufficientFunds = errors.New("insufficient funds for gas * price + value")
```
Vuksan: Why didn't you include these errors as well?
If they are not an easy fix, we should open new issues for each of them.
```
        // ErrAlreadyKnown is returned if the transactions is already contained
	// within the pool.
	ErrAlreadyKnown = errors.New("already known")
	
	// ErrUnderpriced is returned if a transaction's gas price is below the minimum
	// configured for the transaction pool.
	ErrUnderpriced = errors.New("transaction underpriced")
	
	// ErrTxPoolOverflow is returned if the transaction pool is full and can't accpet
	// another remote transaction.
	ErrTxPoolOverflow = errors.New("txpool is full")
	
	// ErrReplaceUnderpriced is returned if a transaction is attempted to be replaced
	// with a different one without the required price bump.
	ErrReplaceUnderpriced = errors.New("replacement transaction underpriced")
	
	// ErrGasLimit is returned if a transaction's requested gas limit exceeds the
	// maximum allowance of the current block.
	ErrGasLimit = errors.New("exceeds block gas limit")
	
	// ErrOversizedData is returned if the input data of a transaction is greater
	// than some meaningful limit a user might use. This is not a consensus error
	// making the transaction invalid, rather a DOS protection.
	ErrOversizedData = errors.New("oversized data")
```

txpool/txpool.go
```diff
@@ -285,6 +272,44 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
...
+
+	// Check if the sender has enough funds to execute the transaction
+	if accountBalance.Cmp(tx.Cost()) < 0 {
+		return ErrInsufficientFunds
+	}
+
+	// Check nonce ordering
+	if t.store.GetNonce(stateRoot, tx.From) > tx.Nonce {
+		return ErrNonceTooLow
+	}
```
Vuksan: Maybe do nonce check first, it should be faster
```suggestion
	// Grab the state root for the latest block
	stateRoot := t.store.Header().StateRoot
	// Check nonce ordering
	if t.store.GetNonce(stateRoot, tx.From) > tx.Nonce {
		return ErrNonceTooLow
	}

	// Check if the sender has enough funds to execute the transaction
	accountBalance := t.store.GetBalance(stateRoot, tx.From)
	if accountBalance.Cmp(tx.Cost()) < 0 {
		return ErrInsufficientFunds
	}
```

txpool/txpool.go
```diff
@@ -285,6 +272,44 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
...
+
+	// Make sure the transaction doesn't exceed the block limit
+	// TODO: Awaiting separate PR
+
```
Vuksan, line 332: Please move the `intrinsic gas cost` check to be the last, since that one takes most resources.

txpool/txpool.go
Vuksan: I see the `TODO` down there for the `ErrGasLimit`, let's just add the task for that one so we won't forget it.

minimal/server.go
```diff
@@ -190,6 +191,24 @@ func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
 	return account.Nonce
 }
 
+func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) *big.Int {
+	zeroValue := big.NewInt(0)
+
+	snap, err := t.state.NewSnapshotAt(root)
+	if err != nil {
+		return zeroValue
```
Vuksan: I think we should return an error here if we cannot find snapshot at all.

minimal/server.go
```diff
@@ -190,6 +191,24 @@ func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
...
+	}
+	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
+	if !ok {
+		return zeroValue
+	}
+	var account state.Account
+	if err = account.UnmarshalRlp(result); err != nil {
+		return zeroValue
```
Vuksan: I think we should return an error here if we cannot unmarshal the account.
The only time where returning 0 makes sense IMO is when we cannot find the account in the snapshot that we know is there.

lazartravica (approved):
LGTM 👍

txpool/txpool_test.go
```diff
@@ -20,6 +20,26 @@ var forks = &chain.Forks{
...
+
+func (m *mockStore) GetNonce(root types.Hash, addr types.Address) uint64 {
+	return 0
+}
+
+func (m *mockStore) GetBlockByHash(types.Hash, bool) (*types.Block, bool) {
+	return nil, false
+}
+
+func (m *mockStore) GetBalance(root types.Hash, addr types.Address) *big.Int {
```
Vuksan: Nitpick, but you can omit both of those:
```suggestion
func (m *mockStore) GetBalance(types.Hash, types.Address) *big.Int {
```

txpool/txpool_test.go
```diff
@@ -20,6 +20,26 @@ var forks = &chain.Forks{
 	Istanbul:  chain.NewFork(0),
 }
 
+type mockStore struct {
+}
+
+func (m *mockStore) GetNonce(root types.Hash, addr types.Address) uint64 {
```
Vuksan: Nitpick, but you can omit both of those:
```suggestion
func (m *mockStore) GetNonce(types.Hash, types.Address) uint64 {
```

Vuksan (changes requested):
I don't see any new tests added for all those checks.
We should have a separate test for each check added.

e2e/transaction_test.go
zivkovicmilos: Added 👍 
519ab3b

txpool/txpool.go

74b6cc3

txpool/txpool.go
zivkovicmilos: Moved 👍 

519ab3b

minimal/server.go
zivkovicmilos: Agreed.

Fixed 👍 

2771012

minimal/server.go
zivkovicmilos: Fixed 👍 

2771012

txpool/txpool_test.go
zivkovicmilos: Fixed 👍 

6c05726

txpool/txpool_test.go
zivkovicmilos: Fixed 👍 

6c05726

txpool/txpool.go
zivkovicmilos: I'll try to see if I can add the check for `ErrAlreadyKnown` and `ErrReplaceUnderpriced` 👍 

`ErrUnderpriced` is unsupported because we don't have any configurable transaction pool minimums.
`ErrTxPoolOverflow` is unsupported because we don't enforce any txpool limits (whereas geth does)
`ErrGasLimit` is already part of the block limit task 👍 
`ErrOversizedData` is unsupported because we don't enforce transaction sizes in the txpool

I'll try to write individual tests for reach check, just so we know they're being handled properly, even though we've seen them work indirectly 👍

zivkovicmilos: I've added tests covering all txpool error codes, apart from `ErrAlreadyKnown` and `ErrReplaceUnderpriced`.
`ErrAlreadyKnown` is an error I believe we can never have due to our current txpool add flow.
`ErrReplaceUnderpriced` would require a refactor for our current txpool add flow, which should be a part of much [bigger task ](https://polygon-sdk.myjetbrains.com/youtrack/issue/PSDK-81)(created on YouTrack).

Added unit tests where applicable, and e2e tests for error codes which required finer state management 👍 
2a26ae9

txpool/txpool.go
Vuksan: I see that you've moved it but the crucial part is not moved.
You still call `t.store.GetBalance(stateRoot, tx.From)` before checking the `nonce`.
If nonce is invalid you have fetched balance for nothing.
Please move that part after the nonce ordering check.

minimal/server.go
```diff
@@ -190,6 +191,25 @@ func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
...
+func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, error) {
+	snap, err := t.state.NewSnapshotAt(root)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get snapshot for root, %v", err)
+	}
+
+	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
+	if !ok {
+		return nil, fmt.Errorf("unable to get result from snapshot, %v", err)
+	}
```
Vuksan: I think that in this middle condition it makes sense to return zero value, because for non existing accounts we should return 0 AFAIK.
```suggestion
	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
	if !ok {
		return big.NewInt(0), nil
	}
```

txpool/txpool.go
Vuksan: GJ
I've added tasks for other errors, to have them in the backlog.

txpool/txpool_test.go
```diff
@@ -302,3 +319,93 @@ func TestTxnQueue_Heap(t *testing.T) {
...
+	txn := generateTx(refAddress, big.NewInt(0))
+
+	assert.ErrorIs(t, pool.addImpl("", txn), ErrNonEncryptedTxn)
+
+	assert.Nil(t, pool.queue[refAddress])
+	assert.Equal(t, pool.Length(), uint64(0))
+}
+
+func TestTxPool_ErrInvalidSender(t *testing.T) {
+	// Unencrypted transactions should be discarded if not in dev mode
```
Vuksan: Wrong comment, c/p from previous test

txpool/txpool_test.go
```diff
@@ -302,3 +319,93 @@ func TestTxnQueue_Heap(t *testing.T) {
...
+func (fms faultyMockStore) GetBlockByHash(hash types.Hash, b bool) (*types.Block, bool) {
+	return nil, false
+}
+
+func (fms faultyMockStore) GetBalance(root types.Hash, addr types.Address) (*big.Int, error) {
+	return nil, fmt.Errorf("unable to fetch account state")
+}
+
+func TestTxPool_ErrInvalidAccountState(t *testing.T) {
+	// Transactions with a negative value should be discarded
```
Vuksan: Wrong comment, c/p from `TestTxPool_ErrNegativeValue` test

txpool/txpool_test.go
```diff
@@ -302,3 +319,93 @@ func TestTxnQueue_Heap(t *testing.T) {
...
+	pool.EnableDev()
+
+	refAddress := types.Address{0x1}
+	txn := generateTx(refAddress, big.NewInt(1))
+
+	assert.ErrorIs(t, pool.addImpl("", txn), ErrInvalidAccountState)
+
+	assert.Nil(t, pool.queue[refAddress])
+	assert.Equal(t, pool.Length(), uint64(0))
+}
```
Vuksan: I am certain you can make these tests into a single test with a `testTable`.
Btw gj on the last test with `faultyMockStore` 👏

e2e/txpool_test.go
```diff
@@ -0,0 +1,186 @@
...
+		gasPrice:      gasPrice,
+		value:         framework.EthToWei(5),
+		t:             t,
+	})
+
+	// The transaction shouldn't be added because of low funds
+	_, addErr := clt.AddTxn(context.Background(), addReq)
+	assert.NotNil(t, addErr)
+	assert.Contains(t, addErr.Error(), txpool.ErrInsufficientFunds.Error())
+}
```
Vuksan: GJ on these tests.
Could you maybe make one test from them with a test table?

Vuksan (changes requested):
GJ on the tests. Please check out the comments.
It seems there are conflicts with the `develop` branch 👀

txpool/txpool.go
zivkovicmilos: I thought I've moved it properly, but I've only moved out part of it.

Sorry about that 👍 

Fixed 👍 

4423ebb

minimal/server.go
zivkovicmilos: Confirming we have this exact same functionality you've suggested in `eth_endpoint.go` for fetching the account balances:
https://github.com/0xPolygon/polygon-sdk/blob/6d40a0e5108374ef2dfe5ea2d49aaac261779101/jsonrpc/eth_endpoint.go#L439-L443

Fixed 👍 

6d40a0e

txpool/txpool_test.go
zivkovicmilos: Fixed 👍 

a6fee97

txpool/txpool_test.go
zivkovicmilos: Fixed 👍 

a6fee97

txpool/txpool_test.go
zivkovicmilos: You're right 💯 

Fixed 👍
 
d8214e6

e2e/txpool_test.go
zivkovicmilos: Done 👍 

d14bf63

Vuksan (approved):
Thank you for resolving comments in a separate commit 🙏 
Tests now look much nicer 🙂 
Just a tip for future efforts: Feel free to name tests based on what's happening during their execution, and not just error that they're testing - whoever sees a test fail will be thankful for that 🙂 
For example `Transactions with a negative value should be discarded` is a better name for the test than `ErrNegativeValue`

PR #115, opened 2021-08-04
Correctly format filter ws requests

mrwillis:
# Description

Currently the websocket subscriptions fail because the unmarshaller unmarshals JSON numbers into float64 and not integer, which is what the formatting function expects, resulting in stuff like this

```
CONNECTED

SENT: {"id": 1, "method": "eth_subscribe", "params": ["newHeads", {}]}

RECEIVED: {"jsonrpc":"2.0","id":%!d(float64=1),"result":"4e31841a-c61e-442a-b1a8-de998db7f164"}
```

Moreover, it does not correctly consider strings, which are also valid under the spec. 

https://www.jsonrpc.org/specification

This results in websocket requests over JSON-RPC to timeout because it never sees a response with the same `id` sent in.

See this as to the unmarshalling behaviour

https://stackoverflow.com/questions/55436628/json-decoded-value-is-treated-as-float64-instead-of-int/55436758. 

I will also add error checking so that the `id` field is the correct format. 

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -3,6 +3,7 @@ package jsonrpc
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"math/big"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


jsonrpc/dispatcher_test.go
```diff
@@ -25,6 +25,26 @@ func expectJSONResult(data []byte, v interface{}) error {
 	return nil
 }
 
+func TestDispatcherWebsocketStringId(t *testing.T) {
```
Kourin1996: This and `TestDispatcherWebsocketInvalidRequestFormats` are similar to each other and the targets are same. Can we put this 2 cases into one test by using subtests?

zivkovicmilos (approved):
Looks good to me 💯

jsonrpc/dispatcher.go
```diff
@@ -113,6 +114,26 @@ type wsConn interface {
...
+	case string:
+		return fmt.Sprintf(`{"jsonrpc":"2.0","id":%s,"result":"%s"}`, t, resp), nil
+	case float64:
+		if t == math.Trunc(t) {
+			return fmt.Sprintf(`{"jsonrpc":"2.0","id":%d,"result":"%s"}`, int(t), resp), nil
+		} else {
+			return "", invalidJSONRequest
+		}
+	case nil:
+		return fmt.Sprintf(`{"jsonrpc":"2.0","result":"%s"}`, resp), nil
```
Vuksan: I think that by the [JSON-RPC spec](https://www.jsonrpc.org/specification) `id` is required in the `response object`, at least as `Nil`:
Screen Shot 2021-08-05 at 16 16 21 PM

Vuksan (commented):
Thank you for noticing this and adding the fix.
I left a single comment, just take another look at the spec

jsonrpc/dispatcher.go
mrwillis: Nice catch will fix

jsonrpc/dispatcher_test.go
mrwillis: Done

jsonrpc/dispatcher.go
mrwillis: Okay added that

Kourin1996 (approved):
LGTM 👏 Thank you for reporting the problem and fixing.

Vuksan (approved):
🙌

PR #116, opened 2021-08-04
Fix failed tests using Gossip protocol in Github Actions

Kourin1996:
# Description

This PR fixes failed tests in Github Actions. According to specification of Gossip protocol in libp2p https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#heartbeat, there is a delay to get peer's subscribed topics because nodes exchange this by heartbeat every 1 second as default.

This PR adds a function to wait until peer subscribe topic in unit test and sleep in e2e.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/e2e/broadcast_test.go
+++ b/e2e/broadcast_test.go
@@ -67,6 +67,9 @@ func TestBroadcast(t *testing.T) {
 				}
 			}
 
+			// wait until gossip protocol build mesh network
+			time.Sleep(time.Second * 2)
+
 			tx, err := signer.SignTx(&types.Transaction{
 				Nonce:    0,
 				From:     senderAddr,
--- a/e2e/broadcast_test.go
+++ b/e2e/broadcast_test.go
@@ -67,7 +67,7 @@ func TestBroadcast(t *testing.T) {
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


e2e/broadcast_test.go
```diff
@@ -67,6 +67,9 @@ func TestBroadcast(t *testing.T) {
 				}
 			}
 
+			// wait until gossip protocol build mesh network
```
lazartravica:
```suggestion
			// wait until gossip protocol build mesh network (https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#heartbeat)
```

lazartravica (approved):
Usually as we discussed we don't like adding new timeouts to make tests pass. But since you've done research as to why the timeouts are mandatory, I will only say excellent job as always :)

Vuksan (approved):
Gj on tackling this. 🥇 
Let's hope we don't see timeouts in our tests for some time at least.

zivkovicmilos (approved):
Great catch, as always 💯

PR #117, opened 2021-08-04
Fixed issues with eth_call, estimateGas and sendTransaction

### LINKED ISSUE #108: `from` is required in the SDK but optional in the spec causing `eth_call` to often fail
### Description

In the spec https://eth.wiki/json-rpc/API#eth_call and in the geth docs https://geth.ethereum.org/docs/rpc/ns-eth, the `from` parameter is marked optional, presumably for methods where the caller's data is not required (for example, retrieving a token name, retrieving a token's decimal count), but it is required here https://github.com/0xPolygon/polygon-sdk/blob/develop/jsonrpc/dispatcher.go#L405

This causes calls like this (from metamask)

```
2021-07-29T14:40:40.962-0400 [DEBUG] polygon-dev.jsonrpc: handle request: request={"id":"109b6e95-32ed-4c01-a842-3ea8ee1e5897","jsonrpc":"2.0","method":"eth_call","params":[{"to":"0xDA4e8F6A6af5909dDC6E38eA6e490099d88c0139","data":"0x95d89b41"},"0x1e"]}
```

to fail, which is actually the call to retrieve the token symbol:

```
  /**
     * @dev Returns the symbol of the token, usually a shorter version of the
     * name.
     */
    function symbol() public view returns (string memory) {
        return _symbol;
    }
```
    
SDK logs:
    
```
    2021-07-29T18:15:33.538Z [DEBUG] polygon.dispatcher: request: method=eth_call id=0b198c7b-9cf9-4373-8ae8-552805ec3d91
2021-07-29T18:15:33.538Z [ERROR] polygon.dispatcher: failed to dispatch: method=eth_call err="from is empty"
```
    

### Your environment

* OS and version Ubuntu 20 
* version of the Polygon SDK a0618c48b8896fb994f8ef42699617cb81fc4af0

### Steps to reproduce

https://github.com/0xPolygon/polygon-sdk/blob/develop/jsonrpc/dispatcher.go#L405 should be marked optional. Maybe a new `decodeTxn` is needed for `eth_call` specifically? Not sure how this affects the tx execution down the line. 
Going into metamask and adding a token deployed on the polygon SDK in dev mode. It'll error when trying to auto populate the token name and decimals

### Expected behaviour

Tell us what should happen

### Actual behaviour

Tell us what happens instead

### Logs

Please paste any logs here that demonstrate the issue, if they exist

### Proposed solution

lazartravica:
# Description

Please provide a detailed description of what was done in this PR:

## Removing missing from check

`decodeTxn` function was relying on having the from address set which is a weird behaviour when you're trying to do a simple eth_call, for example to get the `symbol()` of an ERC20 as mentioned in https://github.com/0xPolygon/polygon-sdk/issues/108

As the decodeTxn is only used in cases of:
 - EstimateGas => which now has handling for when from address is not set, it just skips the balance checks
 - SendTransaction => which will either way override the From by recovering the address from the signature
 - Call (eth_call) => which caused issues in this case as the from address is never read

It was safe to remove the missing `from` field check from the `decodeTxn`.

## Setting sane defaults for transaction `gas` (gasLimit) field

Currently, if we don't specify the gas when sending the transactions to the 3 functions mentioned above (estimateGas, sendTransaction, eth_call) it was set by default to an extremely large value.

This had 2 implications:
- A valid sendTransaction call would be able to spend a large amount of your funds on gas, which should never be the default.
- The default value specified could be larger than your blockLimit which would prevent you from executing eth_calls or estimateGas calls on blockchains with lower blockLimits.

The issue was fixed by setting the default gasLimit when not specified in `eth_call` & `estimateGas` to blockLimit, and leaving it as 0 when `sendTransaction` is in question, letting your transaction fail with the intrinsic gas check.
 
# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -403,7 +403,7 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 	// set default values
 	if arg.From == nil {
-		return nil, fmt.Errorf("from is empty")
+		arg.From = &types.Address{}
 	}
 	if arg.Data != nil && arg.Input != nil {
 		return nil, fmt.Errorf("both input and data cannot be set")
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -477,7 +477,7 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 	// set default values
 	if arg.From == nil {
-		arg.From = &types.Address{}
+		arg.From = &types.ZeroAddress
 	}
 	if arg.Data != nil && arg.Input != nil {
 		return nil, fmt.Errorf("both input and data cannot be set")
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -440,8 +440,7 @@ func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 	}
 
 	if arg.Gas == nil {
-		// TODO
-		arg.Gas = argUintPtr(1000000)
+		arg.Gas = argUintPtr(0)
 	}
 
 	txn := &types.Transaction{
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -275,7 +275,7 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 	valueInt := new(big.Int).Set(transaction.Value)
 
 	// If the sender address is present, recalculate the ceiling to his balance
-	if transaction.GasPrice != nil && gasPriceInt.BitLen() != 0 {
+	if transaction.From != types.ZeroAddress && transaction.GasPrice != nil && gasPriceInt.BitLen() != 0 {
 
 		// Get the account balance
 		acc, err := e.d.store.GetAccount(header.StateRoot, transaction.From)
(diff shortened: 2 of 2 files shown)


jsonrpc/dispatcher.go
```diff
@@ -403,7 +403,7 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 	// set default values
 	if arg.From == nil {
-		return nil, fmt.Errorf("from is empty")
+		arg.From = &types.Address{}
```
Kourin1996: `&types.ZeroAddress` is more clear?

zivkovicmilos (approved):
Looks great 💯

jsonrpc/eth_endpoint.go
```diff
@@ -220,6 +220,11 @@ func (e *Eth) Call(arg *txnArgs, number BlockNumber) (interface{}, error) {
 		return nil, err
 	}
 
+	// If the caller didn't supply the gas limit in the message, then we set it to maximum possible => block gas limit
+	if transaction.Gas == 0 {
+		transaction.Gas = header.GasLimit
+	}
```
Vuksan, line 226: Shouldn't we set the default gasLimit to 0 for `eth_call` per [specification](https://eth.wiki/json-rpc/API#eth_call)?

Vuksan: Now that I read it, it only states that it consumes 0 gas, so you can leave it like this if you want.

Vuksan (approved):
> SendTransaction => which will either way override the From by recovering the address from the signature

This is the case currently, but under different consensus implementations we might have the issue in the future. However, let's listen to Knuth and move on. 👍

jsonrpc/eth_endpoint.go
lazartravica: Yep, the reason why we set it to the gasLimit is because eth_call would fail the intrinsic gas checks before executing (since it is basically simulating transactions).

jsonrpc/dispatcher.go
lazartravica: 👍 https://github.com/0xPolygon/polygon-sdk/pull/117/commits/11b0825ee878b1ca0ce3337ff79356e54d87d450

PR #119, opened 2021-08-04
Improve various JSON-RPC/Geth responses

dankostiuk:
# Description

While working with some of the json-rpc response bodies, I noticed that several fields caused problems when testing with an explorer as they do not adhere to the official eth and geth json-rpc specs. This PR contains fixes for those fields and also addresses the scenario where users would always see an error response when attempting to call `eth_getBlockByNumber` on block `0x0` (see number 1 below). 

Changes include:

1. Currently, requests made to the method `eth_getBlockByNumber` with the block number '0' will always result in an internal error. Now, the following request will only attempt to return the header for the genesis block as expected, avoiding the error and returning a valid response:
```
{"id":1,"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x0", true]}
```

2. Currently, calling `eth_getTransactionReceipt` will return a response body including a `status` field as an integer decimal string. The eth spec (see https://eth.wiki/json-rpc/API#eth_gettransactionreceipt) expects that this field be a hex string which is now the case.

3. Currently, calling `net_version` will return a response body including a `result` value as a hex string. The eth spec (see https://eth.wiki/json-rpc/API#net_version) expects that this field be an integer decimal string which is now the case.

4. Currently, calling `eth_getBlockByNumber` or `eth_getBlockByHash` returns a block object that doesn't contain `totalDifficulty`, `size`, and an `uncles` array. The eth spec (see https://eth.wiki/json-rpc/API#eth_getblockbyhash) expects that these fields be present which is now the case. 

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/jsonrpc/types.go
+++ b/jsonrpc/types.go
@@ -46,42 +46,46 @@ func toTransaction(t *types.Transaction, b *types.Block, txIndex int) *transacti
 }
 
 type uncle struct {
-	ParentHash    types.Hash    `json:"parentHash"`
-	Sha3Uncles    types.Hash    `json:"sha3Uncles"`
-	Miner         types.Address `json:"miner"`
-	StateRoot     types.Hash    `json:"stateRoot"`
-	TxRoot        types.Hash    `json:"transactionsRoot"`
-	ReceiptsRoot  types.Hash    `json:"receiptsRoot"`
-	LogsBloom     types.Bloom   `json:"logsBloom"`
-	Difficulty    argUint64     `json:"difficulty"`
-	Number        argUint64     `json:"number"`
-	GasLimit      argUint64     `json:"gasLimit"`
-	GasUsed       argUint64     `json:"gasUsed"`
-	Timestamp     argUint64     `json:"timestamp"`
-	ExtraData     argBytes      `json:"extraData"`
-	MixHash       types.Hash    `json:"mixHash"`
-	Nonce         types.Nonce   `json:"nonce"`
-	Hash          types.Hash    `json:"hash"`
+	ParentHash        types.Hash    `json:"parentHash"`
+	Sha3Uncles        types.Hash    `json:"sha3Uncles"`
+	Miner             types.Address `json:"miner"`
+	StateRoot         types.Hash    `json:"stateRoot"`
+	TxRoot            types.Hash    `json:"transactionsRoot"`
+	ReceiptsRoot      types.Hash    `json:"receiptsRoot"`
+	LogsBloom         types.Bloom   `json:"logsBloom"`
+	Difficulty        argUint64     `json:"difficulty"`
+  TotalDifficulty   argUint64     `json:"totalDifficulty"`
+  Size              argUint64     `json:"size"`
+	Number            argUint64     `json:"number"`
+	GasLimit          argUint64     `json:"gasLimit"`
+	GasUsed           argUint64     `json:"gasUsed"`
+	Timestamp         argUint64     `json:"timestamp"`
+	ExtraData         argBytes      `json:"extraData"`
+	MixHash           types.Hash    `json:"mixHash"`
+	Nonce             types.Nonce   `json:"nonce"`
+	Hash              types.Hash    `json:"hash"`
 }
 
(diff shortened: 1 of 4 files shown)


Kourin1996 (approved):
Looks good to me 👍 Thank you for your improvements. It'd be great if you could format codes. Thanks

jsonrpc/types.go
```diff
@@ -45,50 +45,103 @@ func toTransaction(t *types.Transaction, b *types.Block, txIndex int) *transacti
...
+	ParentHash        types.Hash    `json:"parentHash"`
+	Sha3Uncles        types.Hash    `json:"sha3Uncles"`
+	Miner             types.Address `json:"miner"`
+	StateRoot         types.Hash    `json:"stateRoot"`
+	TxRoot            types.Hash    `json:"transactionsRoot"`
+	ReceiptsRoot      types.Hash    `json:"receiptsRoot"`
+	LogsBloom         types.Bloom   `json:"logsBloom"`
+	Difficulty        argUint64     `json:"difficulty"`
+  TotalDifficulty   argUint64     `json:"totalDifficulty"`
+  Size              argUint64     `json:"size"`
```
zivkovicmilos: Minor nitpick: Please format this file

zivkovicmilos (approved):
Looks good 💯

jsonrpc/types.go
dankostiuk: Thanks @zivkovicmilos , now that I know how to apply proper formatting this won't happen again 😄

jsonrpc/types.go
```diff
@@ -45,67 +45,120 @@ func toTransaction(t *types.Transaction, b *types.Block, txIndex int) *transacti
...
+		Number:          argUint64(h.Number),
+		GasLimit:        argUint64(h.GasLimit),
+		GasUsed:         argUint64(h.GasUsed),
+		Timestamp:       argUint64(h.Timestamp),
+		ExtraData:       argBytes(h.ExtraData),
+		MixHash:         h.MixHash,
+		Nonce:           h.Nonce,
+		Hash:            h.Hash,
+		Transactions:    []*transaction{},
+		Uncles:          []*uncle{},
```
Vuksan: If you accept [this comment](https://github.com/0xPolygon/polygon-sdk/pull/119#discussion_r684119840) here we can do something like this (not sure if I got the alignment correct, github can't format 🙂):
```suggestion
		uncle:           *toUncle(h),
		Transactions:    []*transaction{},
		Uncles:          []*uncle{},
```

jsonrpc/types.go
```diff
@@ -45,67 +45,120 @@ func toTransaction(t *types.Transaction, b *types.Block, txIndex int) *transacti
...
+	Number          argUint64      `json:"number"`
+	GasLimit        argUint64      `json:"gasLimit"`
+	GasUsed         argUint64      `json:"gasUsed"`
+	Timestamp       argUint64      `json:"timestamp"`
+	ExtraData       argBytes       `json:"extraData"`
+	MixHash         types.Hash     `json:"mixHash"`
+	Nonce           types.Nonce    `json:"nonce"`
+	Hash            types.Hash     `json:"hash"`
+	Transactions    []*transaction `json:"transactions"`
+	Uncles          []*uncle       `json:"uncles"`
```
Vuksan: What do you think of using a nested anonymous struct to reduce duplication?
```suggestion
	uncle
	Transactions    []*transaction `json:"transactions"`
	Uncles          []*uncle       `json:"uncles"`
```

Vuksan (commented):
Thank you very much for a nice PR description and for your contribution! 🥇 
Take a look at the comments I left, possible minor improvement.

jsonrpc/types.go
dankostiuk: Done. Thanks for the suggestion @Vuksan 👍

lazartravica (approved):
Excellent work! I like how you composed uncle into block, although it's a pattern we don't often find in the codebase, I find it easier to follow as it implies clear relationships between objects.

Vuksan (approved):

PR #120, opened 2021-08-05
Add support for higher nonce transactions

zivkovicmilos:
# Description

This PR adds support for keeping added transactions that have a higher nonce value that what is expected.
For example, transactions with these nonces are waiting to be added to the pool:
`TX 0 [Nonce 0]`
`TX 1 [Nonce 2]`
`TX 2 [Nonce 1]`

The flow of them being added for execution should be:
* TX 0 is valid because the nextNonce is 0. Increment the nextNonce to 1
* TX 1 has a higher nonce value than expected, keep it for now, since it is not valid, nextNonce is still 1
* TX 2 has a nonce that is equal to the expected next nonce, it is deemed as a valid transaction. Furthermore, after TX 2 is deemed valid, TX 1 which was previously in the pool because of the higher nonce, is now deemed valid as well. nextNonce is now 3.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -2,7 +2,6 @@ package framework
 
 import (
 	"context"
-	"crypto/ecdsa"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -29,14 +28,6 @@ func EthToWei(ethValue int64) *big.Int {
 	return new(big.Int).Mul(
 		big.NewInt(ethValue),
 		new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil))
-}
-
-func GenerateKeyAndAddr(t *testing.T) (*ecdsa.PrivateKey, types.Address) {
-	t.Helper()
-	key, err := crypto.GenerateKey()
-	assert.NoError(t, err)
-	addr := crypto.PubKeyToAddress(&key.PublicKey)
-	return key, addr
 }
 
 // GetAccountBalance is a helper method for fetching the Balance field of an account
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -2,6 +2,7 @@ package framework
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -17,6 +18,9 @@ import (
 	txpoolProto "github.com/0xPolygon/minimal/txpool/proto"
 	"github.com/0xPolygon/minimal/types"
 	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/stretchr/testify/assert"
+	"github.com/umbracle/go-web3"
+	"github.com/umbracle/go-web3/jsonrpc"
 	"golang.org/x/crypto/sha3"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
@@ -25,6 +29,30 @@ func EthToWei(ethValue int64) *big.Int {
 	return new(big.Int).Mul(
 		big.NewInt(ethValue),
 		new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil))
+}
+
+func GenerateKeyAndAddr(t *testing.T) (*ecdsa.PrivateKey, types.Address) {
+	t.Helper()
+	key, err := crypto.GenerateKey()
+	assert.NoError(t, err)
+	addr := crypto.PubKeyToAddress(&key.PublicKey)
+	return key, addr
+}
+
+// GetAccountBalance is a helper method for fetching the Balance field of an account
+func GetAccountBalance(
+	address types.Address,
+	rpcClient *jsonrpc.Client,
+	t *testing.T,
+) *big.Int {
+	accountBalance, err := rpcClient.Eth().GetBalance(
+		web3.Address(address),
+		web3.Latest,
+	)
+
+	assert.NoError(t, err)
+
+	return accountBalance
 }
 
 func EcrecoverFromBlockhash(hash types.Hash, signature []byte) (types.Address, error) {
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -2,9 +2,9 @@ package e2e
 
 import (
 	"context"
+	"io"
 	"math/big"
 	"testing"
-	"time"
 
 	"github.com/0xPolygon/minimal/crypto"
 	"github.com/0xPolygon/minimal/e2e/framework"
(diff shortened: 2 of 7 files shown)


txpool/txpool.go
```diff
@@ -288,89 +302,128 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
...
 
 	// Promote elements
 	tx := t.Peek()
 	if tx == nil || tx.Nonce != t.nextNonce {
+		// Nothing to promote
 		return nil
 	}
 
 	promote := []*types.Transaction{}
+	higherNonceTx := []*types.Transaction{}
```
lazartravica: var name should be plural

txpool/txpool.go
```diff
@@ -288,89 +302,128 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
...
 	for {
 		promote = append(promote, tx)
 		t.Pop()
 
-		tx2 := t.Peek()
-		if tx2 == nil || tx.Nonce+1 != tx2.Nonce {
+		nextTx := t.Peek()
+		if nextTx == nil {
+			break
+		}
```
lazartravica: perhaps it's cleaner to include this in the `for` statement itself such as:

```
for nextTx := t.Peek(); nextTx != nil {
```

txpool/txpool.go
```diff
@@ -160,6 +175,8 @@ func (t *TxPool) AddTx(tx *types.Transaction) error {
 	return nil
 }
 
+// addImpl validates the tx and adds it to the appropriate account pool.
+// Additionally, it updates the global valid transactions list
```
lazartravica: usage of `pool` and `list` in this sense introduces more confusion, imho

txpool/txpool.go
```diff
@@ -72,13 +88,13 @@ func NewTxPool(
...
-		idlePeriod: defaultIdlePeriod,
-		queue:      make(map[types.Address]*txQueue),
-		sorted:     newTxPriceHeap(),
-		sealing:    sealing,
-		forks:      forks,
+		logger:           logger.Named("txpool"),
+		store:            store,
+		idlePeriod:       defaultIdlePeriod,
+		accountTxHeapMap: make(map[types.Address]*txHeapWrapper),
+		validTxHeap:      newTxPriceHeap(),
```
lazartravica: Heap is a data structure commonly used to implement queues (abstract data structure).
It's perhaps easier to reason about what our intention is if we talk about queues rather than heaps.

Also, I don't think that it is helpful if we include the explicit type in the name of the variable (e.g. map, etc.)

My suggestion is to rename `accountTxHeapMap` to `accountQueues` and `validTxHeap` to `queue`/`pendingQueue`.

What do you think?

lazartravica (changes requested):
Excellent job on the changes.

I have minor style comments and some comments around renaming of the heap fields in the TxPool struct.

txpool/txpool.go
zivkovicmilos: I'm not sure what you mean by this `for` syntax. Did you mean `if`?

txpool/txpool.go
508b0c6

txpool/txpool.go
zivkovicmilos: Simplified 👍 
b812710

txpool/txpool.go
zivkovicmilos: Agreed. Renamed 👍 
1f614a8

txpool/txpool.go
zivkovicmilos: I completely agree. I wasn't exactly sure what to name them as well, but I guess it can be figured out from the comments as well as the renamed fields 👍 
44c44bb

e2e/framework/helper.go
```diff
@@ -27,6 +31,30 @@ func EthToWei(ethValue int64) *big.Int {
 		new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil))
 }
 
+func GenerateKeyAndAddr(t *testing.T) (*ecdsa.PrivateKey, types.Address) {
+	t.Helper()
+	key, err := crypto.GenerateKey()
+	assert.NoError(t, err)
+	addr := crypto.PubKeyToAddress(&key.PublicKey)
+	return key, addr
+}
```
Kourin1996: If I remember correctly, this function was moved to `helper/tests` package.

Kourin1996 (changes requested):
I have left one comment. There seems to be no problem apart from that 👍

e2e/framework/helper.go
zivkovicmilos: You're right 👍 
I was confused, because we standardized all of these helper methods on the IBFT branches, and here they're either nonexistent or in some other place.

Fixed 👍 
da0a77d

txpool/txpool.go
lazartravica: @zivkovicmilos My bad, what I meant to suggest is:

```
for nextTx := t.Peek(); nextTx; nextTx != nil {
```

but it doesn't look as good 👍

lazartravica (approved):
Thank you for resolving every comment with a commit!

Kourin1996 (approved):
LGTM 👍

txpool/txpool.go
```diff
@@ -288,89 +302,128 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
...
+
+		tx = nextTx
 	}
+
 	if len(promote) == 0 {
+		// Nothing to promote
+		reinsertFunc()
+
 		return nil
 	}
```
Vuksan: Can this `if` even happen?
inside of the `for` loop on the first line we append the `promote` slice with the tx, which can't be `nil` because we checked that on L349

e2e/txpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+		if signErr != nil {
+			t.Fatalf("Unable to sign transaction, %v", signErr)
+		}
+
+		return signedTx
+	}
+
+	generateReq := func(nonce uint64) *txpoolOp.AddTxnReq {
+		var msg *txpoolOp.AddTxnReq
+		unstakeTxn := generateTx(nonce)
```
Vuksan: `unstakeTxn`? 😕

e2e/txpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+		addReq := generateReq(nonces[i])
+
+		_, addErr := clt.AddTxn(context.Background(), addReq)
+		if addErr != nil {
+			t.Fatalf("Unable to add txn, %v", addErr)
+		}
+	}
+
+	// Mandatory sleep for the dev consensus and executor to go through the txns
+	time.Sleep(time.Duration(devInterval+1) * time.Second)
```
Vuksan: I think you already implemented a better way than this on another branch, so if that's the case, just wait for that branch to be merged and use it here

e2e/txpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// Add the transaction with the gap nonce value
+	addReq := generateReq(1)
+
+	_, addErr := clt.AddTxn(context.Background(), addReq)
+	if addErr != nil {
+		t.Fatalf("Unable to add txn, %v", addErr)
+	}
+
+	// Mandatory sleep for the dev consensus and executor to go through the txns
+	time.Sleep(time.Duration(devInterval+1) * time.Second)
```
Vuksan: I think you already implemented a better way than this on another branch, so if that's the case, just wait for that branch to be merged and use it here

Vuksan (changes requested):
Kudos for renaming the queues in the txpool , they are much clearer now! 🚀 
GJ on the test as well, it's on point.
I left a couple of comments, but the only one important is the one in test where you wait for seconds to pass. If that can be improved, it would be excellent, if not, I guess we'll live for now.

e2e/txpool_test.go
zivkovicmilos: Whoops, it was leftover :(

Fixed 👍 

a6edca6

e2e/txpool_test.go
zivkovicmilos: Yep. I figured out on other branches that are still awaiting PR reviews that it's better to subscribe to new blocks.

Fixed 👍 

a6edca6

e2e/txpool_test.go
zivkovicmilos: Fixed 👍 

https://github.com/0xPolygon/polygon-sdk/pull/120#discussion_r686199705

txpool/txpool.go
zivkovicmilos: I think you're actually right.

Promoted will never be empty after the for loop, since we'll always be adding at least 1 element here:
https://github.com/0xPolygon/polygon-sdk/blob/a6edca6741245aa54f1210e56556dcc136d0bd1c/txpool/txpool.go#L365

Good catch 💯 

Removing now 👍 

30caf6e

Vuksan (approved):
👏

PR #124, opened 2021-08-13
Fix eth call default values

### LINKED ISSUE #123: `eth_call` without `from` field fails during nextNonce calculation in decodeTxn
### Description

When making an http request to the `eth_call` endpoint and the `from` field isn't provided, we now override the `from` address with a ZeroAddress after recent changes last week (see https://github.com/0xPolygon/polygon-sdk/commit/386571800e75bbe65c96ca368dad2edba3c99852).

Further downstream in `dispatcher.decodeTxn` however, we attempt to call `getNextNonce` with the new `from` address (which is now a 0x0 address) and as a result encounter the `"given root and slot not found in storage"` error after failing to get the snapshot of the 0x0 address.

### Your environment

* OS and version `OSX 11.4`
* version of the Polygon SDK `3a72f15 `

### Steps to reproduce

1. Make request to `eth_call` with `from` missing from the params, for example:
```
{"id":"28eee10a-18db-40ce-a924-37d2f9f69631","jsonrpc":"2.0","method":"eth_call","params":[{"to":"0x011e4e51998e97bd74d0a8ae08b428f51b58eaba","data":"0x70a0823100000000000000000000000061a197c895ff5ceaacf7f800746541f99da4fe79"},"0xec"]}
```
(this was found using a local instance of https://metamask.github.io/test-dapp/ and attempting to add token to wallet)

### Expected behaviour

We should not see a ``"given root and slot not found in storage"`` error message and a `{"code":-32603,"message":"internal error"}` response when trying to add a token to MetaMask (calling `eth_call`)

### Actual behaviour

We see a ``"given root and slot not found in storage"`` error message and a `{"code":-32603,"message":"internal error"}` response.

### Proposed solution

We probably need to default to a `nonce` value of 0 within the `decodeTxn` function so that we can reach the `ApplyTxn` logic.

lazartravica:
# Description

Fixing the issue reported in: https://github.com/0xPolygon/polygon-sdk/issues/123

After more research, it doesn't make sense that we are fetching the average gas price when it wasn't explicitly provided as a part of the message for `eth_call`, `eth_sendTransaction` or `eth_estimateGas`, this will only cause the call to fail when not specifying the from address.

Also, when the from is not specified, we will no longer try to fetch the nonce.

# Changes include

# Checklist

- [x] I have tested this code
- [ ] I have added sufficient documentation both in code, as well as in the README

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -475,6 +475,10 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 }
 
 func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
+	if arg.Data != nil && arg.Input != nil {
+		return nil, fmt.Errorf("both input and data cannot be set")
+	}
+
 	// set default values
 	if arg.From == nil {
 		arg.From = &types.ZeroAddress
@@ -486,10 +490,6 @@ func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 			return nil, err
 		}
 		arg.Nonce = argUintPtr(nonce)
-	}
-
-	if arg.Data != nil && arg.Input != nil {
-		return nil, fmt.Errorf("both input and data cannot be set")
 	}
 	if arg.Value == nil {
 		arg.Value = argBytesPtr([]byte{})
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -495,8 +495,7 @@ func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
 		arg.Value = argBytesPtr([]byte{})
 	}
 	if arg.GasPrice == nil {
-		// use the suggested gas price
-		arg.GasPrice = argBytesPtr(d.store.GetAvgGasPrice().Bytes())
+		arg.GasPrice = argBytesPtr([]byte{})
 	}
 
 	var input []byte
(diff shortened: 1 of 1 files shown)


jsonrpc/dispatcher.go
```diff
@@ -478,24 +478,24 @@ func (d *Dispatcher) decodeTxn(arg *txnArgs) (*types.Transaction, error) {
...
 		nonce, err := d.getNextNonce(*arg.From, LatestBlockNumber)
 		if err != nil {
 			return nil, err
 		}
 		arg.Nonce = argUintPtr(nonce)
 	}
+
+	if arg.Data != nil && arg.Input != nil {
+		return nil, fmt.Errorf("both input and data cannot be set")
+	}
```
lazartravica: This was not added, git diff can be weird sometimes.

Kourin1996 (approved):
LGTM 👍

zivkovicmilos (approved):
Looks good™ 💯 

Good catch 👍

jsonrpc/dispatcher.go
Vuksan: Looking at the `develop`, it seems it was moved after the nonce check.
On `develop` branch this check is occurring before.

Vuksan: What do you think of moving this to be the first condition here?
That way we avoid fetching the nonce if these 2 are nil.

Vuksan (approved):
Small nitpick, nothing major. GJ 👍

jsonrpc/dispatcher.go
lazartravica: 👍  https://github.com/0xPolygon/polygon-sdk/pull/124/commits/0d86f9045d7b012a9d0aeed2d98b81dd1eadcc30

PR #125, opened 2021-08-13
Add support of block gas limit

Kourin1996:
# Description

This PR adds Block Gas Limit Calculation and change this value by demand. 

This codes are based on go-ethereum implementation. (https://github.com/ethereum/go-ethereum/blob/master/core/block_validator.go#L108-L129) 

Block gas limit is calculated by last block gas limit and gas used. Formula is as follows:
0775319bcb79fdbce7bb5515b5de14f9

28f9080a3004207ae9e1b1d5d0f77948

(The value 1024 is the constant value called `GasLimitBoundDivisor`. It's defined as hardcoded)

Basic block gas limit is calculated by first formula. In most cases, this value will be next block gas limit. Next block gas limit will be increased if more than 2/3 gas limit is consumed at last block. Otherwise, next block gas limit will be decreased. In addition, each miner can choose target range of block gas limit (`GasFloor` and `GasCeil`). If new block gas limit is out of this range, new block gas limit is determined by only parent block gas limit regardless the value calculated by formula 1. Finally block gas limit is moving into the range. But the delta value one block can increase will be limited to GasLimit/1024.

## New Command Options

This PR also adds command options so that user can specify some parameters as follows:

### `genesis` command
--genesis-gas-limit [GAS_LIMIT]: specifies gas limit of genesis block

### `server` & `dev` command
--gas-floor [GAS_FLOOR]: specifies lower limit of target block gas limit
--gas-ceil [GAS_CEIL]: specifies upper limit of target block gas limit

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


Kourin1996:
I finished to implement code but we're still discussing whether we add this block gas limit calculation which go-ethereum uses.

Kourin1996:
Just quick research
- go-ethereum
    - Follows a formula to change block gas limit on demand
    - Miner can specify `--gas-floor-target` and `--gaslimit` options to converge towards a specific block gas limit.
    - constant params (hardcoded parameters)
        - GasLimitBoundDivisor: 1024
            - The bound divisor of the gas limit, used in update calculations
            - Diff of block gas limit is up to (Last Block Gas Limit)/GasLimitBoundDivisor
        - MinGasLimit: 5000
            - Lowest block gas limit that can be accepted
            - If next block gas limit is less than this, then next block gas limit will be set to MinGasLimit
- Bor (Side chain in Polygon)
    - Same as go-ethereum
- Binance Smart Chain
    - Same as go-ethereum, but constant params are different
    - constant params
        - GasLimitBoundDivisor: 256
        - MinGasLimit: 5000
- GoQuorum
    - Same as go-ethereum, but constant params are different
    - constant params
        - GasLimitBoundDivisor: 4096
        - MinGasLimit: 700000000
- Celo
    - Block Gas Limit is defined in Smart Contract

zivkovicmilos (approved):
Looks good to me 💯 

Good work on adding test coverage as well.


One question I have:
Does it make sense to check the block limits here in the `validateTx`, since I see we're checking it outside?
If not, please remove this comment

https://github.com/0xPolygon/polygon-sdk/blob/000a4ebbed35df29a7867ff6167d2e695a50ee15/txpool/txpool.go#L330-L331

lazartravica:
> Looks good to me 💯
> 
> Good work on adding test coverage as well.
> 
> One question I have:
> Does it make sense to check the block limits here in the `validateTx`, since I see we're checking it outside?
> If not, please remove this comment
> 
> https://github.com/0xPolygon/polygon-sdk/blob/000a4ebbed35df29a7867ff6167d2e695a50ee15/txpool/txpool.go#L330-L331

It doesn't

> Looks good to me 💯
> 
> Good work on adding test coverage as well.
> 
> One question I have:
> Does it make sense to check the block limits here in the `validateTx`, since I see we're checking it outside?
> If not, please remove this comment
> 
> https://github.com/0xPolygon/polygon-sdk/blob/000a4ebbed35df29a7867ff6167d2e695a50ee15/txpool/txpool.go#L330-L331
@zivkovicmilos 

It doesn't make much sense to me to check for the block limit when validating transactions to be added to the txpool.
Since you left the TODO originally, do you remember what were the thoughts behind it?

Also, not sure what do you mean by `outside` can you please elaborate?

zivkovicmilos:
@lazartravica @Kourin1996 

It makes much more sense for us to be checking the block limit `outside` of the txpool only when building the block, like Yoshiki does it now, for example in `ibft buildBlock`:

https://github.com/0xPolygon/polygon-sdk/blob/000a4ebbed35df29a7867ff6167d2e695a50ee15/consensus/ibft/ibft.go#L398-L402

The original comment was left as a sort of note for us to keep in mind these limits need to be checked, as for example geth does this check when validating the transaction addition:

https://github.com/ethereum/go-ethereum/blob/0a68558e7e025afebf67b81bf48ecb8b0fa7c06d/core/tx_pool.go#L589-L592


@Kourin1996 Please remove the //TODO for the blocklimit inside `validateTx`

blockchain/blockchain.go
```diff
@@ -226,6 +232,45 @@ func (b *Blockchain) Genesis() types.Hash {
 	return b.genesis
 }
 
+// CalculateGasLimit returns the gas limit of the next block after parent
+func (b *Blockchain) CalculateGasLimit(number uint64) (uint64, error) {
+	parent, ok := b.GetHeaderByNumber(number - 1)
+	if !ok {
+		return 0, fmt.Errorf("parent of (%d) not found", number)
+	}
+	// todo: EIP1559 support for London hardfolk
```
Vuksan: Funny typo 😄 
```suggestion
	// todo: EIP1559 support for London hardfork
```

consensus/ibft/ibft.go
```diff
@@ -918,6 +936,35 @@ func (i *Ibft) VerifyHeader(parent, header *types.Header) error {
 	return nil
 }
 
+func (i *Ibft) verifyGasLimit(header *types.Header) error {
+	if header.GasLimit < blockchain.MinGasLimit {
+		return fmt.Errorf("block gas limit falls %d", blockchain.MinGasLimit)
```
Vuksan: Maybe:
```suggestion
		return fmt.Errorf("block gas limit below %d", blockchain.MinGasLimit)
```

consensus/ibft/ibft.go
```diff
@@ -918,6 +936,35 @@ func (i *Ibft) VerifyHeader(parent, header *types.Header) error {
...
+	}
+	if header.GasLimit > blockchain.MaxGasLimit {
+		return fmt.Errorf("block gas limit exceeds %d", blockchain.MaxGasLimit)
+	}
+	if header.GasUsed > header.GasLimit {
+		return fmt.Errorf("block gas limit exceeds gas used, limit = %d, used=%d", header.GasLimit, header.GasUsed)
```
Vuksan:
```suggestion
		return fmt.Errorf("block gas used exceeds gas limit, limit = %d, used=%d", header.GasLimit, header.GasUsed)
```

chain/chain.go
```diff
@@ -18,6 +18,12 @@ var (
 
 	// GenesisDifficulty is the default difficulty of the Genesis block.
 	GenesisDifficulty = big.NewInt(131072)
+
+	// DefaultGasFloor is default value of target gas floor for mined blocks
+	DefaultGasFloor uint64 = 8000000
+
+	// DefaultGasCeil is default value of target gas ceiling for mined blocks
+	DefaultGasCeil uint64 = 8000000
```
Vuksan, line 26: Is it normal that they are both the same value?

blockchain/blockchain_test.go
```diff
@@ -561,3 +562,82 @@ func TestBlockchainWriteBody(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := NewTestBlockchain(t, nil)
+			b.writeGenesis(&chain.Genesis{
+				GasLimit: tt.parentGasLimit,
+				GasUsed:  tt.parentGasUsed,
+			})
```
Vuksan: Maybe catch the error that `writeGenesis` can throw:
```suggestion
			err := b.writeGenesis(&chain.Genesis{
				GasLimit: tt.parentGasLimit,
				GasUsed:  tt.parentGasUsed,
			})
			if err != nil {
				t.Fatal(err)
			}
```

consensus/ibft/ibft.go
```diff
@@ -388,6 +395,12 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 		if txn == nil {
 			break
 		}
+		if gas := transition.AvailableGas(); gas < state.TxGas {
+			i.logger.Info("Block doesn't have enough gas to process new transaction", "remaining", gas, "minimum required gas", state.TxGas)
+			retFn()
+			break
+		}
```
Vuksan: What do you think of moving this check and the one in `dev.go` into `Write()` method of `executor.go`'s Transition object?
That way we will end up with only one check instead of two.

Vuksan (changes requested):
Good job on the PR! Commit messages are 👍 
I left some minor comments, so please check them out

Vuksan:
> @lazartravica @Kourin1996
> 
> It makes much more sense for us to be checking the block limit `outside` of the txpool only when building the block, like Yoshiki does it now, for example in `ibft buildBlock`:
> 
> https://github.com/0xPolygon/polygon-sdk/blob/000a4ebbed35df29a7867ff6167d2e695a50ee15/consensus/ibft/ibft.go#L398-L402
> 
> The original comment was left as a sort of note for us to keep in mind these limits need to be checked, as for example geth does this check when validating the transaction addition:
> 
> https://github.com/ethereum/go-ethereum/blob/0a68558e7e025afebf67b81bf48ecb8b0fa7c06d/core/tx_pool.go#L589-L592
> 
> @Kourin1996 Please remove the //TODO for the blocklimit inside `validateTx`

@zivkovicmilos @lazartravica 
That check in the tx_pool is there to prevent the transactions with gas higher than **block gas limit** to enter the tx_pool. You can see that if you take a look at the part of code that initializes this `currentMaxGas` variable -> [link](https://github.com/ethereum/go-ethereum/blob/0a68558e7e025afebf67b81bf48ecb8b0fa7c06d/core/tx_pool.go#L1262)

As for the `outside` part, Yoshiki's code tackles checking block gas limit when block is being constructed from transactions from tx_pool. This is a separate issue which should prevent txs with invalid gas from entering tx_pool in the first place.

@Kourin1996 Please add that simple check to the `validateTx` method and remove TODO. Just check if the transaction consumes more gas than what block gas limit allows.

chain/chain.go
Kourin1996: If we use same value for them, Block Gas Limit is trying to converge on specific value. If they are different each other, there will be tolerance. But in usually, miner aims to set block gas limit to specific value.

consensus/ibft/ibft.go
Kourin1996: I think this is great idea since can do this check when other node appends new block.

Kourin1996:
@Vuksan @lazartravica @zivkovicmilos Thank you for discussion and some feedback.
I fixed codes based on your feedback. I didn't change behavior except for moving pre check to make sure block gas limit has enough amount to add new tx. 

> @Kourin1996 Please add that simple check to the validateTx method and remove TODO. Just check if the transaction consumes more gas than what block gas limit allows.

I think can't detect how much gas tx consumes before evaluating. And gas pool check this and terminates process if all block gas limit is consumed.

blockchain/blockchain.go
Kourin1996: Fixed

consensus/ibft/ibft.go
Kourin1996: Fixed

consensus/ibft/ibft.go
Kourin1996: Fixed

blockchain/blockchain_test.go
Kourin1996: Added new check

Vuksan: It's not the same, since `t.Fatal()` will stop execution, while `assert.NoError()` will display the error and continue program execution, but that's minor thing.

txpool/txpool.go
```diff
@@ -352,9 +352,6 @@ func (t *TxPool) validateTx(tx *types.Transaction) error {
 		return ErrInsufficientFunds
 	}
 
-	// Make sure the transaction doesn't exceed the block limit
-	// TODO: Awaiting separate PR
-
```
Vuksan, line 357: I would leave this `TODO` as it is.
AFAIK this check is there to make sure that the gas sent with the transaction doesn't exceed current block limit.
Not just to check if we can fit one regular 21k gas transaction into the block, as this new check does.

Vuksan (changes requested):

lazartravica:
Closing this PR in favor of #184

PR #127, opened 2021-08-17
Fix structure of result for eth_getBlockByNumber and eth_getBlockByHash

Kourin1996:
# Description

This PR fixes result of `eth_getBlockByNumber` and `eth_getBlockByHash`
+ Sets proper value to `size` field in block
+ Lets it return array of transactions when second arg is `true`, but return array of tx hash if second args is `false`
+ Lets it return array of uncle hash in `uncles` field
  - Note it'll be empty array if you use `dev` or `ibft` consensus because fork doesn't appear
+ Lets it return `null` if block doesn't exist, not error

Please check the spec for details: https://eth.wiki/json-rpc/API#eth_getblockbyhash

In addition,
+ Adds Content-Type to JSONRPC response header

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -19,7 +19,7 @@ func (e *Eth) ChainId() (interface{}, error) {
 }
 
 // GetBlockByNumber returns information about a block by block number
-func (e *Eth) GetBlockByNumber(number BlockNumber, full bool) (interface{}, error) {
+func (e *Eth) GetBlockByNumber(number BlockNumber, fullTx bool) (interface{}, error) {
 	var num uint64
 	switch number {
 	case LatestBlockNumber:
@@ -35,20 +35,20 @@ func (e *Eth) GetBlockByNumber(number BlockNumber, full bool) (interface{}, erro
 		num = uint64(number)
 	}
 
-	block, ok := e.d.store.GetBlockByNumber(num, full)
+	block, ok := e.d.store.GetBlockByNumber(num, true)
 	if !ok {
 		return nil, fmt.Errorf("unable to get block by num %v", num)
 	}
-	return toBlock(block), nil
+	return toBlock(block, fullTx), nil
 }
 
 // GetBlockByHash returns information about a block by hash
-func (e *Eth) GetBlockByHash(hash types.Hash, full bool) (interface{}, error) {
-	block, ok := e.d.store.GetBlockByHash(hash, full)
+func (e *Eth) GetBlockByHash(hash types.Hash, fullTx bool) (interface{}, error) {
+	block, ok := e.d.store.GetBlockByHash(hash, true)
 	if !ok {
 		return nil, fmt.Errorf("unable to get block by hash %v", hash)
 	}
-	return toBlock(block), nil
+	return toBlock(block, fullTx), nil
 }
 
 // BlockNumber returns current block number
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -37,7 +37,7 @@ func (e *Eth) GetBlockByNumber(number BlockNumber, fullTx bool) (interface{}, er
 
 	block, ok := e.d.store.GetBlockByNumber(num, true)
 	if !ok {
-		return nil, fmt.Errorf("unable to get block by num %v", num)
+		return nil, nil
 	}
 	return toBlock(block, fullTx), nil
 }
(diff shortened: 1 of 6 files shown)


jsonrpc/types.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/0xPolygon/minimal/types"
 )
 
+// For union type of transaction and types.Hash
+type transactionOrHash interface {
+	getHash() types.Hash
```
Kourin1996, line 14: Actually this method is not used, but I want to limit types of value to be set.

jsonrpc/eth_endpoint.go
```diff
@@ -35,20 +35,20 @@ func (e *Eth) GetBlockByNumber(number BlockNumber, full bool) (interface{}, erro
 		num = uint64(number)
 	}
 
-	block, ok := e.d.store.GetBlockByNumber(num, full)
+	block, ok := e.d.store.GetBlockByNumber(num, true)
 	if !ok {
-		return nil, fmt.Errorf("unable to get block by num %v", num)
+		return nil, nil
```
zivkovicmilos, line 43: Is there any specific reason why we're not returning an error here?

jsonrpc/eth_endpoint.go
```diff
@@ -35,20 +35,20 @@ func (e *Eth) GetBlockByNumber(number BlockNumber, full bool) (interface{}, erro
...
 }
 
 // GetBlockByHash returns information about a block by hash
-func (e *Eth) GetBlockByHash(hash types.Hash, full bool) (interface{}, error) {
-	block, ok := e.d.store.GetBlockByHash(hash, full)
+func (e *Eth) GetBlockByHash(hash types.Hash, fullTx bool) (interface{}, error) {
+	block, ok := e.d.store.GetBlockByHash(hash, true)
 	if !ok {
-		return nil, fmt.Errorf("unable to get block by hash %v", hash)
+		return nil, nil
```
zivkovicmilos, line 52: Is it standard to not return an error when we can't find the block?

zivkovicmilos (approved):
Looks great 💯 
I've left a few comments, but I'm probably missing something.

Also, good job on expanding some of the tests 👍

jsonrpc/eth_endpoint.go
Kourin1996: I checked jsonrpc spec and tried to get from ethereum mainnet via infura. I just got null when block not found. So I think it's ok we don't need to return error.

lazartravica (approved):
LGTM 💯 that uncle composition didn't live in the codebase for too long, but that's fine as the concept of `uncle` blocks is closely tied to non-IBFT consensuses.

jsonrpc/eth_endpoint_test.go
```diff
@@ -152,22 +152,28 @@ func TestEth_Block_GetBlockByNumber(t *testing.T) {
...
-		{EarliestBlockNumber, true},
-		{BlockNumber(-50), true},
-		{BlockNumber(0), false},
-		{BlockNumber(2), false},
-		{BlockNumber(50), true},
+		{LatestBlockNumber, true, false},
+		{EarliestBlockNumber, false, true},
+		{BlockNumber(0), true, false},
+		{BlockNumber(2), true, false},
+		{BlockNumber(50), false, false},
```
Vuksan: I noticed that you removed `-50` test. Do you think it's bad if we have it or was that an oversight?

Vuksan (approved):
Fantastic work! 👏 
I left a single question, both the implementation and PR description are on point 🎯

jsonrpc/eth_endpoint_test.go
Kourin1996: I'm thinking whether we should return error, or not in this case. But actually negative number won't be passed to this function. (because parsing arguments is failed) If negative number is passed to this function, it just returns null in current implementation.

Kourin1996: I let function not to return error when transaction doesn't exist. For that reason, this function doesn't return error if -50 is passed. So I'm thinking if I should add new check to detect arg is negative.

Kourin1996: I tested with mainnet and they return error if first bit is set in hex arg. So I let this function return error in this case.

Kourin1996: @Vuksan If you agree with my opinion and the change, please just click `Resolve conversation`. Thanks

jsonrpc/types.go
Sunset17: Thank you

PR #128, opened 2021-08-17
Add a check for overwriting IBFT init data

zivkovicmilos:
# Description

This PR introduces a safety check when initializing ibft directories:
- If there is already some kind of initialized ibft data, then the `ibft init` command should fail

# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (approved):
It works completely 👌

command/ibft/ibft_init.go
```diff
@@ -72,20 +80,30 @@ func (p *IbftInit) Run(args []string) int {
 		return 1
 	}
 
-	if err := minimal.SetupDataDir(dataDir, []string{"consensus", "libp2p"}); err != nil {
+	subDirectories := []string{"consensus", "libp2p"}
```
lazartravica: Please refactor these to constants, no need for a slice. It would be much more readable below to reference LIBP2P_DIR than subDirectories[1].

lazartravica (changes requested):
Just a minor comment. Everything else looks good.

command/helper/helper.go
```diff
@@ -483,3 +483,19 @@ func FormatKV(in []string) string {
...
+	if err != nil {
+		return false
+	}
+
+	// Check if the directory exists, and that it's actually a directory if there is a hit
+	if fileInfo, statErr := os.Stat(pathAbs); os.IsNotExist(statErr) || !fileInfo.IsDir() {
+		return false
+	}
+
+	return true
```
Vuksan: What happens if we get another error that isn't `not exist` one?
If we get it, and `fileInfo` is nil I think that we're gonna get null pointer exception from `fileInfo.IsDir()`

Vuksan (changes requested):

command/ibft/ibft_init.go
zivkovicmilos: Agreed 👍 

Fixed 👍 
1f7127d

command/helper/helper.go
zivkovicmilos: You're right, we're not handling that kind of case.
I checked and fileInfo will indeed be nil if any kind of error occurs that's not `not exists`.

Fixed 👍 
ba7e15e

command/ibft/ibft_init.go
```diff
@@ -56,6 +56,19 @@ func (p *IbftInit) GetBaseCommand() string {
...
+func generateAlreadyInitializedError(directory string) string {
+	output := "\n[IBFT INIT ERROR]\n"
+	output += fmt.Sprintf("Directory %s has previously initialized IBFT data\n", directory)
+	return output
+}
+
+var (
+	CONSENSUS_DIR = "consensus"
+	LIBP2P_DIR    = "libp2p"
+)
```
Kourin1996: Just small suggestions
(1) In Golang, name of constant variable is usually named by CamelCase style. (If you don't want to export constant, then can use lowerCamelCase)
https://golang.org/doc/effective_go#constants
https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

(2) `"consensus"` and `"libp2p"` are literals, so you can use `const` instead of `var`.

Vuksan (approved):
LGTM

lazartravica (changes requested):
Please take a look at @Kourin1996 comment regarding const naming and that's it!

command/ibft/ibft_init.go
zivkovicmilos: You're right, completely forgot about this 👍 

Fixed 👍 

f684a99

lazartravica (approved):

PR #132, opened 2021-08-18
required block parameter for snapshot eth endpoints

mrwillis:
# Description

According to https://eth.wiki/json-rpc/API#the-default-block-parameter, passing a block parameter for snapshot calls is required. This PR makes them required by erroring if the unmarshalling of the request returns a 0 block number. Ideally blockNumber is passed in as `nil` to these eth endpoints and we can check for that, but it seems like the reflection makes it a 0.

If you guys are OK with this, I can add some tests 

# Changes include

# Breaking changes

Calling any of the 

`eth_getBalance`
`eth_getCode`
`eth_getTransactionCount`
`eth_getStorageAt`
`eth_call`

will return an error now if the block param is not passed as an extra arg. In geth:

https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/rpc/json.go#L303

# Checklist

- [x] I have tested this code

# Additional comments


Vuksan (commented):
The only thing I'm worried about is that if we do this, we lose the ability to do these requests for the genesis block since its number is 0.

@Kourin1996 @zivkovicmilos @lazartravica Do you guys think we can refactor [`stringToBlockNumber`](https://github.com/0xPolygon/polygon-sdk/blob/b8659d19bb432c74be9310055084ebd7c26fcd2b/jsonrpc/codec.go#L50-L70) to somehow return `nil` as a block number if it wasn't passed through or if it was explicitly `nil`?

We need to be careful though since on some calls like `eth_newFilter` and `eth_getLogs` this parameter should still be optional with the default value.

Kourin1996:
> The only thing I'm worried about is that if we do this, we lose the ability to do these requests for the genesis block since its number is 0.
> 
> @Kourin1996 @zivkovicmilos @lazartravica Do you guys think we can refactor [`stringToBlockNumber`](https://github.com/0xPolygon/polygon-sdk/blob/b8659d19bb432c74be9310055084ebd7c26fcd2b/jsonrpc/codec.go#L50-L70) to somehow return `nil` as a block number if it wasn't passed through or if it was explicitly `nil`?
> 
> We need to be careful though since on some calls like `eth_newFilter` and `eth_getLogs` this parameter should still be optional with the default value.

I guess value is set to 0 (it is default value for BlockNumber=int64) if block number is not in json arguments. And unmarshalling is done by reflect automatically. So my suggestion is to create new type for Optional Block Number (it's same to *int64)

Kourin1996:
Difficult to know json params has required values because json.Unmarshal set default value if the value doesn't exist in json. Let's imagine json.Unmarshal parses `["0x...."]` and map to [types.Address, BlockNumber] type value. (parsing second value would be skipped)

https://github.com/0xPolygon/polygon-sdk/blob/b8659d19bb432c74be9310055084ebd7c26fcd2b/jsonrpc/dispatcher.go#L300-L309

I guess `stringToBlockNumber` won't be called in this case. A simple solution is that we can create new type for Optional BlockNumber (alias of *int64) because default value of pointer type is nil.

mrwillis:
Happy to do whatever. I like @Kourin1996's approach

lazartravica:
> Happy to do whatever. I like @Kourin1996's approach

Agree, I think we can go forward with @Kourin1996 's approach

mrwillis:
@Kourin1996 Bit stuck here. Probably my go-noobness! Wondering if you can help.

I created the `OptionalBlockNumber` type alias, but it still seems like it is struggling with passing "latest", "earliest", etc. Those are now failing. My guess is that the `UnmarshalJSON` middleware is no longer being called and hence it can't convert the string "latest" -> OptionalBlockNumber.

Any ideas how I could do this? I tried editing `UnmarshalJSON` to something like this, but it doesn't seem to like it

```go
// UnmarshalJSON automatically decodes the user input for the block number, when a JSON RPC method is called
func (b OptionalBlockNumber) UnmarshalJSON(buffer []byte) error {
	num, err := stringToBlockNumber(string(buffer))
	if err != nil {
		return err
	}
	*b = num
	return nil
}
```

Kourin1996:
@mrwillis Thank you for your investigating! I checked this problem little bit. Error happens if we define OptionalBlockNumber and use this. But when just change type of argument in endpoint handler to pointer, it works. As of now, we can use this way.
Sorry for confusing and Thanks for taking your effort!

```go
func (e *Eth) GetTransactionCount(address types.Address, number *BlockNumber) (interface{}, error) {
	fmt.Printf("\n\nGetTransactionCount %+v, %v\n\n", address, number)
        // => number should be nil when second arguments are not passed

	if number == nil {
		return nil, fmt.Errorf("block parameter is required")
	}
	nonce, err := e.d.getNextNonce(address, BlockNumber(*number))
	if err != nil {
		if err == ErrStateNotFound {
			return argUintPtr(0), nil
		}
		return nil, err
	}
	return argUintPtr(nonce), nil
}
```

mrwillis:
Thanks @Kourin1996 figured it out!

jsonrpc/eth_endpoint.go
```diff
@@ -216,13 +219,16 @@ func (e *Eth) GasPrice() (interface{}, error) {
...
+	if number == nil {
+		return nil, fmt.Errorf("block parameter is required")
+	}
 	transaction, err := e.d.decodeTxn(arg)
 	if err != nil {
 		return nil, err
 	}
 	// Fetch the requested header
-	header, err := e.d.getBlockHeaderImpl(number)
+	header, err := e.d.getBlockHeaderImpl(BlockNumber(*number))
```
Kourin1996: I think you don't have to wrap with `BlockNumber` here.

Kourin1996 (commented):
Thank you for fixing the codes! I have left a small comment.

Kourin1996 (approved):
Thank you for fixing!

lazartravica (approved):
LGTM, thanks @Kourin1996 for the assistance

PR #135, opened 2021-08-20
Fix invalid V value parsing from transaction

### LINKED ISSUE #130: Cannot send txs using chainId other than 100 when running `server`
### Description

When trying to send txs using MetaMask (eth_sendRawTransaction), tx signature encoding fails when using a genesis file created with `--chainid 647`. I observed the error message below after trying to do a simple transaction (sending the premined account 100eth from itself):

```
2021-08-17T20:07:47.398Z [ERROR] polygon.dispatcher: failed to dispatch: method=eth_sendRawTransaction err="invalid sender"
```

Upon testing again using chainId 100, I no longer saw this error.

It looks like this is related to an issue reported earlier (see https://github.com/0xPolygon/polygon-sdk/issues/99).

### Your environment

* OS and version `OSX 11.4`
* version of the Polygon SDK `d8b80b9 `

### Steps to reproduce

1. Set chainId in genesis file to be `647`. 
2. Spin up 4 validator nodes using `server`, with validator-1 being the premined account.
3. Using MetaMask, connect to the RPC URL, inputting `647` for the chainId.
4. Import premined account privateKey and see that I have the starting balance in my account.
5. Try sending 100ETH to the same account, observe a brief MetaMask error and errors in the logs indicating an "invalid sender".

### Expected behaviour

The transaction should have went through without any issues when using a chainId other than 100.

### Actual behaviour

The transaction failed to go through.

### Proposed solution

Looking at https://github.com/0xPolygon/polygon-sdk/issues/99, I see that using chainId 100 will avoid this issue however we'd like to use a number other than 100 when making our chain public.

zivkovicmilos:
# Description

This PR fixes an issue with how V values larger than 8 bits are being processed when unmarshalling the transaction RLP object.

## Mini crash course on ECDSA

The ECDSA (Elliptic Curve Digital Signature Algorithm) is a cryptographically secure digital signature scheme, based on the elliptic-curve cryptography (ECC). ECDSA relies on the math of the cyclic groups of elliptic curves over finite fields and on the difficulty of the ECDLP problem (elliptic-curve discrete logarithm problem).

### ECDSA Signing

The ECDSA signing algorithm (RFC 6979) takes as input a message msg + a private key privKey and produces as output a signature, which consists of pair of integers {r, s}. The ECDSA signing algorithm is based on the ElGamal signature scheme and works as follows (with minor simplifications):

1. Calculate the message hash, using a cryptographic hash function like SHA-256: `h = hash(msg)`
2. Generate securely a random number `k` in the range [1..n-1]
2.1. In case of deterministic-ECDSA, the value `k` is HMAC-derived from `h + privKey` (see RFC 6979)
3. Calculate the random point `R = k * G` and take its x-coordinate: `r = R.x`
4. Calculate the signature proof: `s`
5. Return the signature `{r, s}`.

TL;DR: 
* `r` - x coordinate of the random point `R`
* `s` - signature proof
* `v` - last byte of the signature (explained in the next section)

### ECDSA Public Key from signature

It is important to know that the ECDSA signature scheme allows the public key to be recovered from the signed message together with the signature. The recovery process is based on some mathematical computations (described in the SECG: SEC 1 standard) and returns 0, 1 or 2 possible EC points that are valid public keys, corresponding to the signature. To avoid this ambiguity, some ECDSA implementations add one additional bit `v` to the signature during the signing process and it takes the form `{r, s, v}`. From this extended ECDSA signature {r, s, v} + the signed message, the signer's public key can be restored with confidence.

[📚Reference](https://cryptobook.nakov.com/digital-signatures/ecdsa-sign-verify-messages)

## EIP-155

In short, EIP-155 provides basic replay protection by having the transaction signature take into account the chain ID it was meant for in the first place, meaning if someone signs a transaction on chain A, that transaction cannot be added / verified on chain B.

[📚Reference](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md)

## What was actually the issue?

When the hash of the raw transaction gets decoded, we need to unmarshal it's RLP value.
During RLP unmarshalling, we'd unmarshal the `v` value as a **byte**, but as it can be seen from the EIP-155 spec, this V value can be larger than 8 bits, since it's calculated as `{0,1} + CHAIN_ID * 2 + 35`.

Tests were added that cover this specific fix.

### Additional clarification regarding Ethereum magic numbers
In theory, the `v` value should be either `0` or `1`, indicating the parity of the `y` coordinate of the curve point.
However, in Ethereum, this scales to be either `27` / `28`; `35` / `36`, as explained here:
https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739

# Changes include

# Checklist

- [x] I have tested this code


crypto/txsigner.go
```diff
@@ -21,6 +22,9 @@ type TxSigner interface {
 
 	// SignTx signs a transaction
 	SignTx(tx *types.Transaction, priv *ecdsa.PrivateKey) (*types.Transaction, error)
+
+	// calculateV calculates the V value based on the type of signer used
+	calculateV(parity byte) []byte
```
Kourin1996: It's just a curious question. `calculateV` needs to be exported? Currently this method is called by own object.

Kourin1996 (approved):
Looks good to me 👍 I like the introduction about signature and EIP-155 in description.
I just left a small comment. It'd be great if you would check.

crypto/txsigner.go
lazartravica: If all the implementations of the TxSigner interface are contained inside this package, then it's all good. @zivkovicmilos though I don't see why exactly you would intend that.

lazartravica (approved):
LGTM, thank you for the excellent Crash course :thumbsup:

crypto/txsigner.go
zivkovicmilos: This is definitely a method all signers should have, but since it's only used as part of local processing I didn't export it.

You're right to point out that if we have different implementations across packages, this wouldn't work as intended. 

Fixed, so it's exported :+1:

554e26a

PR #136, opened 2021-08-24
Fix wrong response when account doesn't exist in jsonrpc endpoint

Kourin1996:
# Description

Related issue: https://github.com/0xPolygon/polygon-sdk/issues/131

This PR fixes wrong response in some endpoints of jsonrpc

1. `eth_getStorageAt(address, index, block number)`, `eth_getTransactionCount(address, block number)`
     These endpoint returned error in both cases of (1) the header with given number doesn't exist and (2) data for the account doesn't exist in storage. But they should return zero value in case of (2). So this PR adds new special error and return zero value when account doesn't exist in storage.

2. `GetLogs(options)`
     This returned null when there are no logs to return. But expect to return empty array. So this PR lets this function return empty array in this case.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/jsonrpc/errors.go
+++ b/jsonrpc/errors.go
@@ -1,0 +1,7 @@
+package jsonrpc
+
+import "errors"
+
+var (
+	ErrStateNotFound = errors.New("given root and slot not found in storage")
+)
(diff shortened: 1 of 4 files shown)


zivkovicmilos (approved):
Looks good 💯 

Would love to have mini tests that cover these errors as well, if it's not too much of a hassle 👍

Kourin1996:
@zivkovicmilos I added some small tests.

zivkovicmilos:
@Kourin1996 
Great work on these tests, now we have 💯 coverage 🙂

lazartravica (approved):
Great job on the tests, and the PR description 👍

PR #137, opened 2021-08-24
Allow omitted `params` in json-rpc request

dankostiuk:
# Description

According to the official json-rpc spec [here](https://www.jsonrpc.org/specification#request_object), the `params` member "MAY be omitted". 

This bug was found after testing the integration with a blockchain bridge component that expected a non-error response when calling:
```
{"id":"1","jsonrpc":"2.0","method":"eth_chainId"}
```

The above request returns a successful response when testing on Polygon Mainnet.

# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (changes requested):
Thank you for your improvement! Looks good to me and it works locally.

Probably we can also skip preparing argument array for unmarshalling in this case. I would be very grateful if you could put following codes into if block:
https://github.com/0xPolygon/polygon-sdk/blob/d7aa445e2da21c146790ece84050af7fa19f05af/jsonrpc/dispatcher.go#L300-L305

Kourin1996:
@dankostiuk Thank you for modifying. After some tests, I noticed I could call an endpoint which needs arguments without params. (for example, you can call `eth_getBlockByNumber` without params) I think it should reject in this case.
@lazartravica @zivkovicmilos What do you think about this?

lazartravica:
> @dankostiuk Thank you for modifying. After some tests, I noticed I could call an endpoint which needs arguments without params. (for example, you can call `eth_getBlockByNumber` without params) I think it should reject in this case.
> @lazartravica @zivkovicmilos What do you think about this?

@Kourin1996 I'd say to leave it as it is allowing for optional params. If omitting the params object doesn't cause errors in certain endpoints, but rather has undocumented functionality we can fix it later and approve this PR.

Kourin1996 (approved):
Looks good to me. Thank you for creating PR!

lazartravica (approved):

PR #138, opened 2021-08-25
updated json error responses

munna0908:
# Description
Current JSON-RPC implementation returns the error responses without nesting it with proper response object, as part of this PR, I have defined various error objects and modified the handler methods to return a well structured error response. 

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


jsonrpc/dispatcher.go
```diff
@@ -242,20 +243,24 @@ func (d *Dispatcher) Handle(reqBody []byte) ([]byte, error) {
 		if err := json.Unmarshal(reqBody, &req); err != nil {
 			return nil, invalidJSONRequest
 		}
+		if req.Method == "" {
+			//TODO:Add id if required
```
zivkovicmilos: Is this TODO still relevant?

jsonrpc/dispatcher.go
```diff
@@ -242,20 +243,24 @@ func (d *Dispatcher) Handle(reqBody []byte) ([]byte, error) {
...
 	}
 
 	// handle batch requests
 	var requests []Request
 	if err := json.Unmarshal(reqBody, &requests); err != nil {
-		return nil, d.internalError("batch method", err)
+		return nil, d.internalError(nil, "batch method", err)
 	}
-	var responses []Response
+	var responses []interface{}
```
zivkovicmilos: I would really try to step away from using this interface umbrella for multiple different response types. It would be great if you could create a common interface that `ErrorResponse` and `Response` could implement.

jsonrpc/dispatcher.go
```diff
@@ -305,21 +310,23 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, error) {
...
-		return nil, invalidJSONRequest
+		invalidParams.Data = req.ID
+		return nil, invalidParams
 	}
 
 	output := fd.fv.Call(inArgs)
 	err = getError(output[1])
 	if err != nil {
-		return nil, d.internalError(req.Method, err)
+		fmt.Println("he", err)
```
zivkovicmilos: Please remove this leftover print 🙂

jsonrpc/dispatcher.go
```diff
@@ -330,13 +337,15 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, error) {
...
-		return nil, d.internalError(req.Method, err)
+		return nil, d.internalError(req.ID, req.Method, err)
 	}
 	return respBytes, nil
 }
 
-func (d *Dispatcher) internalError(method string, err error) error {
+func (d *Dispatcher) internalError(id interface{}, method string, err error) error {
 	d.logger.Error("failed to dispatch", "method", method, "err", err)
+	internalError.Data = id
```
zivkovicmilos: I think we do this method a couple of times here in the JSON-RPC Layer: 
We have a global error object that we modify before sending. Why wouldn't we create a new one each time, instead of modifying the same one, and by that risking some data to be leftover?

We can use the same principle from txpool: Have constant error messages, but create new errors each time

jsonrpc/jsonrpc.go
```diff
@@ -220,7 +225,14 @@ func (j *JSONRPC) handle(w http.ResponseWriter, req *http.Request) {
...
 	if err != nil {
-		handleErr(err)
+		id := err.(*ErrorObject).Data
+		err.(*ErrorObject).Data = nil
+		errorResponse := ErrorResponse{
+			JSONRPC: "2.0",
+			ID:      id,
+			Error:   err.(*ErrorObject),
+		}
+		handleErr(errorResponse)
```
zivkovicmilos: Why is the actual error being unwrapped here, modified, then being put back? What is the purpose of setting the `Data` field to `nil` in the `Error` field of `ErrorResponse`?

zivkovicmilos (changes requested):
I've left some comments that need to be cleared up, otherwise it looks good 👍

jsonrpc/codec.go
```diff
@@ -12,13 +12,20 @@ import (
...
+	JSONRPC string       `json:"jsonrpc"`
+	ID      interface{}  `json:"id,omitempty"`
+	Error   *ErrorObject `json:"error"`
 }
 
 // Response is a jsonrpc response
 type Response struct {
-	ID      interface{}     `json:"id"`
 	JSONRPC string          `json:"jsonrpc"`
+	ID      interface{}     `json:"id"`
```
zivkovicmilos, line 56: I know this was not modified as part of this task, but we should really consider if it is worthwhile to have the `ID` field as type interface.

As far as I can remember, this field can be either a number or a string (uuid).

jsonrpc/jsonrpc.go
```diff
@@ -220,7 +225,14 @@ func (j *JSONRPC) handle(w http.ResponseWriter, req *http.Request) {
...
 	resp, err := j.dispatcher.Handle(data)
 	if err != nil {
-		handleErr(err)
+		id := err.(*ErrorObject).Data
+		err.(*ErrorObject).Data = nil
+		errorResponse := ErrorResponse{
+			JSONRPC: "2.0",
+			ID:      id,
+			Error:   err.(*ErrorObject),
+		}
```
Kourin1996: This `err` is always the value of *ErrorObject? If not, you might want to add some check here. or we should consider to let `Handle` return *ErrorObject.

jsonrpc/dispatcher.go
lazartravica: @munna0908 Please let the reviewer resolve the comments 🙏

jsonrpc/dispatcher.go
lazartravica: This is a very dangerous thing to do, please instantiate new error every time if they need to differ  👍

lazartravica (changes requested):
@munna0908 Please provide a textual description in the body of the pull request to what was changed and why. 

This is not mandatory if the commit messages are clear and separated, but since you have only one commit, you should provide a couple of sentences to explain your reasoning.

Thanks! 🙏

jsonrpc/dispatcher.go
munna0908: it not relevant, removed it

jsonrpc/dispatcher.go
munna0908: Fixed it. Here is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/138/commits/192a250666919d1bf1d69174b60d340941de149c)

jsonrpc/dispatcher.go
munna0908: cleaned it up

jsonrpc/dispatcher.go
munna0908: Declared a new interface for response, `ErrorResponse` and `SuccessResponse` implements `Response` interface. here is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/138/commits/6bf58ba45a7548edb4e111aff19f33c4baf9f2b7)

jsonrpc/jsonrpc.go
munna0908: Modified this logic. Here is the [link](https://github.com/0xPolygon/polygon-sdk/pull/138/commits/db5007ae038262ad8ae1340ef59a713fd1144354)

jsonrpc/codec.go
```diff
@@ -78,3 +133,30 @@ func (b *BlockNumber) UnmarshalJSON(buffer []byte) error {
...
+// NewRpcResponse returns Success/Error response object
+func NewRpcResponse(id interface{}, jsonrpcver string, reply []byte, err Error) Response {
+
+	var response Response
+	switch err.(type) {
+	case nil:
+		response = &SuccessResponse{JSONRPC: jsonrpcver, ID: id, Result: reply}
+	default:
+		jsonerr := &ErrorObject{Code: err.ErrorCode(), Message: err.Error()}
+		response = &ErrorResponse{JSONRPC: jsonrpcver, ID: id, Error: jsonerr}
```
Kourin1996: Why don't you use `NewRpcErrorResponse`?

Kourin1996 (commented):
Hi, Thank you for your fixes! I like the way to define each error as struct because can hold error code into struct and it's possible to understand what error function return clearly.
As of now, I found you defined a function which is not used. It'd be great if you could check.

jsonrpc/dispatcher.go
```diff
@@ -305,39 +288,32 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, error) {
...
-	}
-	respBytes, err := json.Marshal(resp)
-	if err != nil {
-		return nil, d.internalError(req.Method, err)
-	}
-	return respBytes, nil
 }
 
-func (d *Dispatcher) internalError(method string, err error) error {
+func (d *Dispatcher) internalError(method string, err error) {
```
Kourin1996: It's better to name this function as `log...` because it just output a log here?

jsonrpc/jsonrpc.go
munna0908: Fixed this

jsonrpc/codec.go
munna0908: currently, i am sticking to interface, i will open a new task for this specially

jsonrpc/dispatcher.go
munna0908: Fixed it, [Commit-link](https://github.com/0xPolygon/polygon-sdk/pull/138/commits/8dcef833eb2f8067388f23ba87a42c40f9022a40)

jsonrpc/codec.go
munna0908: Updated it, c[ommit-link](https://github.com/0xPolygon/polygon-sdk/pull/138/commits/8dcef833eb2f8067388f23ba87a42c40f9022a40)

zivkovicmilos (approved):
Thanks for fixing up the minor things we've mentioned in the commends 🙏 

Looks good to me 💯

jsonrpc/codec.go
Kourin1996: Thank you!

jsonrpc/dispatcher.go
Kourin1996: Thank you!

Kourin1996 (approved):
LGTM! Thank you for taking your time to fix :)

munna0908:
> @munna0908 Please provide a textual description in the body of the pull request to what was changed and why.
> 
> This is not mandatory if the commit messages are clear and separated, but since you have only one commit, you should provide a couple of sentences to explain your reasoning.
> 
> Thanks! 🙏

@lazartravica i have added the description, Thanks ..!!

lazartravica (approved):
Good job @munna0908 ! 💯 Feel free to merge.

PR #139, opened 2021-08-26
Ensure eth_logs handles case where fromBlock == toBlock

dankostiuk:
# Description

While attempting to integrate a bridge with nodes running the SDK, json-rpc calls going out to `eth_logs` which had the same value for the `fromBlock` and `toBlock` fields within the request were returning empty.

For example:
```
{"jsonrpc":"2.0","id":3061,"method":"eth_getLogs","params":[{"address":["0x2da3dcaa07453dfdb69ce7a4ce3674e4084a2234"],"fromBlock":"0x21b7","toBlock":"0x21b7","topics":[["0xdbb69440df8433824a026ef190652f29929eb64b4d1d5d2a69be8afe3e6eaed8"]]}]}
```

would previously return a response body such as:
```
{
    "id": 3061,
    "jsonrpc": "2.0",
    "result": null
}
```

I adjusted the for-loop comparator so that queries with identical `fromBlock` and `toBlock` values return logs for that single block.

I also brought back some commented out tests I saw and added the case where `fromBlock == toBlock`.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/jsonrpc/eth_endpoint_test.go
+++ b/jsonrpc/eth_endpoint_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"math/big"
+	"strconv"
 	"testing"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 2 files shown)


jsonrpc/eth_endpoint_test.go
```diff
@@ -105,6 +106,77 @@ func (m *mockBlockStore2) add(blocks ...*types.Block) {
 }
 
 func (m *mockBlockStore2) GetReceiptsByHash(hash types.Hash) ([]*types.Receipt, error) {
+	switch hash {
```
Kourin1996, line 110: Is it possible to set mock value when you use mockBlockStore2 in test?

jsonrpc/eth_endpoint_test.go
```diff
@@ -214,104 +286,77 @@ func TestEth_Block_BlockNumber(t *testing.T) {
...
+				// If there is an error and test isn't expected to fail
+				t.Fatalf("Error: %v", logError)
+			} else if !testCase.shouldFail {
+				switch testCase.name {
+				case "Found matching logs, fromBlock < toBlock":
+					assert.Lenf(t, foundLogs, 3, "Invalid number of logs found")
+				case "Found matching logs, fromBlock == toBlock":
+					assert.Lenf(t, foundLogs, 1, "Invalid number of logs found")
+				case "No logs found":
+					assert.Lenf(t, foundLogs, 0, "Invalid number of logs found")
```
Kourin1996: I think it's better to give expected length in testCase.

Kourin1996 (changes requested):
Hello, Thank you for PR! It works locally and it is a good catch.
I have left some small comments about test. It'd be great if you could check.

jsonrpc/eth_endpoint_test.go
lazartravica: @Kourin1996 not exactly sure what do you mean by this?

lazartravica (approved):
Excellent catch and good job on writing the tests!

jsonrpc/eth_endpoint_test.go
Kourin1996: @lazartravica This function returns constant values actually. We need to write test case which matches these return value. So mock should be able to set value the mock returns in test code like as other mock if possible.

jsonrpc/eth_endpoint_test.go
```diff
@@ -214,104 +281,78 @@ func TestEth_Block_BlockNumber(t *testing.T) {
...
+				// If there is an error and test isn't expected to fail
+				t.Fatalf("Error: %v", logError)
+			} else if !testCase.shouldFail {
+				switch testCase.name {
+				case "Found matching logs, fromBlock < toBlock":
+					assert.Lenf(t, foundLogs, testCase.expectedLength, "Invalid number of logs found")
+				case "Found matching logs, fromBlock == toBlock":
+					assert.Lenf(t, foundLogs, testCase.expectedLength, "Invalid number of logs found")
+				case "No logs found":
+					assert.Lenf(t, foundLogs, testCase.expectedLength, "Invalid number of logs found")
```
Kourin1996: Same args are passed in every cases, can we remove switch? That was intended to do that.

Kourin1996 (approved):
Thank you for fixing! LGTM

Vuksan (approved):
LGTM

zivkovicmilos (approved):
Looking good 💯

PR #145, opened 2021-09-06
Revert "Allow omitted `params` in json-rpc request"

lazartravica:
Reverts 0xPolygon/polygon-sdk#137

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -280,16 +280,16 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, Error) {
 	inArgs := make([]reflect.Value, fd.inNum)
 	inArgs[0] = service.sv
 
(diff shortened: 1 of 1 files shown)


zivkovicmilos (approved):
Looks good considering what it fixes 💯

Vuksan (approved):

Kourin1996 (approved):

dankostiuk:
Out of curiosity, why is https://github.com/0xPolygon/polygon-sdk/pull/137 being reverted? Thanks in advance

munna0908:
> Out of curiosity, why is #137 being reverted? Thanks in advance

hi @dankostiuk , After the recent changes, if we omit `params` in the json RPC request  `call` function in handleReq will panic. 
Hence to avoid this we are reverting back and will fix this soon.

PR #146, opened 2021-09-06
Fix/get storage at response

munna0908:
# Description

 Modified `getStorageAt` RPC method to return the RLP decode value.  

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments
 `TestEth_State_GetStorageAt` is also modified to test the new response

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/0xPolygon/polygon-sdk/helper/hex"
 	"github.com/0xPolygon/polygon-sdk/types"
+	"github.com/umbracle/fastrlp"
 )
 
 // Eth is the eth jsonrpc endpoint
@@ -202,8 +203,17 @@ func (e *Eth) GetStorageAt(address types.Address, index types.Hash, number Block
 		}
 		return nil, err
 	}
-
-	return argBytesPtr(result), nil
+	//Parse the RLP value
+	p := &fastrlp.Parser{}
+	v, err := p.Parse(result)
+	if err != nil {
+		return argBytesPtr(types.ZeroHash[:]), nil
+	}
+	data, err := v.Bytes()
+	if err != nil {
+		return argBytesPtr(types.ZeroHash[:]), nil
+	}
+	return argBytesPtr(data), nil
 }
 
 // GasPrice returns the average gas price based on the last x blocks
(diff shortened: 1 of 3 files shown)


Kourin1996 (commented):
Thank you for the PR! I test locally and node returns correct value.
I checked Ethereum json-rpc wiki and result will be padded with zeros so that result is 32 bytes data.
Let's say storage data is 0x4, then result would be 0x00000000000000000000000000000004.

https://eth.wiki/json-rpc/API#eth_getstorageat

zivkovicmilos (approved):
Looks good to me 💯

munna0908:
> Thank you for the PR! I test locally and node returns correct value.
> I checked Ethereum json-rpc wiki and result will be padded with zeros so that result is 32 bytes data.
> Let's say storage data is 0x4, then result would be 0x00000000000000000000000000000004.
> 
> https://eth.wiki/json-rpc/API#eth_getstorageat

@Kourin1996 I have tested with ganache and local geth, both are returning result without padding bits

Kourin1996:
> @Kourin1996 I have tested with ganache and local geth, both are returning result without padding bits

@munna0908 Understood. I tested with Ganache and your saying is correct. Thank you for checking!

Kourin1996 (approved):

lazartravica (approved):

munna0908:
@Vuksan  can i merge this ?

jsonrpc/eth_endpoint.go
```diff
@@ -202,8 +203,17 @@ func (e *Eth) GetStorageAt(address types.Address, index types.Hash, number Block
...
+	//Parse the RLP value
+	p := &fastrlp.Parser{}
+	v, err := p.Parse(result)
+	if err != nil {
+		return argBytesPtr(types.ZeroHash[:]), nil
+	}
+	data, err := v.Bytes()
+	if err != nil {
+		return argBytesPtr(types.ZeroHash[:]), nil
+	}
```
Vuksan, line 215: Any reason why we don't return `nil, err` instead of `argBytesPtr(types.ZeroHash[:]), nil` here?
I see that in the previous `if` we return this result, but that is only for the `ErrStateNotFound` error.

Vuksan (commented):
LGTM. 👍 
I left a single question for discussion.

jsonrpc/eth_endpoint.go
munna0908: In case of any internal error related to getStorageAt call, geth return the zero/empty hash, I followed the same. 
Looking forward to your inputs. 😅
@Vuksan

Vuksan (approved):
👍

PR #148, opened 2021-09-07
Fix/json rpc error responses

munna0908:
# Description

Added additional checks and updated the error response messages for eth,net,web API's 

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Yet to perform few checks on `call` method.

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -435,16 +435,6 @@ func (d *Dispatcher) getBlockHeaderImpl(number BlockNumber) (*types.Header, erro
 	}
 }
 
-func (d *Dispatcher) isAccountAvailable(addr types.Address) (isAvailable bool) {
-	number := LatestBlockNumber
-	header, err := d.getBlockHeaderImpl(number)
-	if err != nil {
-		return false
-	}
-	_, err = d.store.GetAccount(header.StateRoot, addr)
-	return err == nil
-
-}
 func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (uint64, error) {
 	if number == PendingBlockNumber {
 		res, ok := d.store.GetNonce(address)
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -442,10 +442,7 @@ func (d *Dispatcher) isAccountAvailable(addr types.Address) (isAvailable bool) {
 		return false
 	}
 	_, err = d.store.GetAccount(header.StateRoot, addr)
-	if err != nil {
-		return false
-	}
-	return true
+	return err == nil
 
 }
 func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (uint64, error) {
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -239,6 +239,7 @@ func (d *Dispatcher) Handle(reqBody []byte) ([]byte, error) {
 		if req.Method == "" {
 			return NewRpcResponse(req.ID, "2.0", nil, NewInvalidRequestError("Invalid json request")).Bytes()
 		}
+
 		resp, err := d.handleReq(req)
 
 		return NewRpcResponse(req.ID, "2.0", resp, err).Bytes()
@@ -286,16 +287,16 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, Error) {
 		inputs[i] = val.Interface()
 		inArgs[i+1] = val.Elem()
 	}
-
-	if err := json.Unmarshal(req.Params, &inputs); err != nil {
-
-		return nil, NewInvalidParamsError("Invalid Params")
+	if fd.numParams() > 0 {
+		if err := json.Unmarshal(req.Params, &inputs); err != nil {
+			return nil, NewInvalidParamsError("Invalid Params")
+		}
 	}
 
 	output := fd.fv.Call(inArgs)
 	if err := getError(output[1]); err != nil {
 		d.logInternalError(req.Method, err)
-		return nil, NewInternalError("Internal error")
+		return nil, NewInvalidRequestError(err.Error())
 	}
 
 	var data []byte
@@ -434,6 +435,19 @@ func (d *Dispatcher) getBlockHeaderImpl(number BlockNumber) (*types.Header, erro
 	}
 }
 
+func (d *Dispatcher) isAccountAvailable(addr types.Address) (isAvailable bool) {
+	number := LatestBlockNumber
+	header, err := d.getBlockHeaderImpl(number)
+	if err != nil {
+		return false
+	}
+	_, err = d.store.GetAccount(header.StateRoot, addr)
+	if err != nil {
+		return false
+	}
+	return true
+
+}
 func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (uint64, error) {
 	if number == PendingBlockNumber {
 		res, ok := d.store.GetNonce(address)
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -191,7 +191,7 @@ func (e *Eth) GetTransactionReceipt(hash types.Hash) (interface{}, error) {
 func (e *Eth) GetStorageAt(address types.Address, index types.Hash, number *BlockNumber) (interface{}, error) {
 	//Set the block number to latest
 	if number == nil {
-		*number = LatestBlockNumber
+		number = createBlockNumberPointer(LatestBlockNumber)
 	}
 	// Fetch the requested header
 	header, err := e.d.getBlockHeaderImpl(*number)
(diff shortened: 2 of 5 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -188,6 +193,10 @@ func (e *Eth) GetTransactionReceipt(hash types.Hash) (interface{}, error) {
 
 // GetStorageAt returns the contract storage at the index position
 func (e *Eth) GetStorageAt(address types.Address, index types.Hash, number BlockNumber) (interface{}, error) {
+	//Set the block number to latest
+	if BlockNumber(0) == number {
+		number = LatestBlockNumber
+	}
```
Kourin1996, line 199: `number` will be overwritten to LatestBlockNumber when I give `0x0` to 3rd params. I think better to distinguish between (1) number is zero and (2) number is not given.

Kourin1996: `number` will be overwritten to `LatestBlockNumber` when I give `0x0` to 3rd parameter. I think better to distinguish between (1) `0x0` is given (2) `number` is not given.

FYI, I tried Ganache and it can return genesis state.

lazartravica: @munna0908 please take a look at @Kourin1996 's comment 🙏

jsonrpc/eth_endpoint.go
```diff
@@ -471,6 +489,12 @@ func (e *Eth) GetTransactionCount(address types.Address, number BlockNumber) (in
 
 // GetCode returns account code at given block number
 func (e *Eth) GetCode(address types.Address, number BlockNumber) (interface{}, error) {
+
+	//Set the block number to latest
+	if BlockNumber(0) == number {
+		number = LatestBlockNumber
+	}
```
lazartravica: @munna0908 same as the above. we need to have a way to query the genesis block storage/balances/code since our entire PoS implementation depends on predeployed contracts.

lazartravica (changes requested):
Good job on the params. Please take a look at my comments for block number 0

jsonrpc/eth_endpoint.go
munna0908: sure @lazartravica , will modify this accordingly

jsonrpc/dispatcher.go
```diff
@@ -434,6 +435,16 @@ func (d *Dispatcher) getBlockHeaderImpl(number BlockNumber) (*types.Header, erro
 	}
 }
 
+func (d *Dispatcher) isAccountAvailable(addr types.Address) (isAvailable bool) {
```
zivkovicmilos: Minor nitpick, there is no need to use a named return if we're not utilizing the actual name of the return variable (should return `bool`, not `isAvailable bool`, since we don't leverage `isAvailable` within the method)

jsonrpc/eth_endpoint.go
```diff
@@ -81,6 +83,9 @@ func (e *Eth) SendRawTransaction(input string) (interface{}, error) {
 
 // SendTransaction creates new message call transaction or a contract creation, if the data field contains code.
 func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
+	if !e.d.isAccountAvailable(*arg.From) {
+		return nil, errors.New("unknown account")
```
zivkovicmilos: Confirming GETH and possibly other clients have this logic as well, as it seemed fishy to me when I first looked at it 👍

jsonrpc/eth_endpoint.go
zivkovicmilos: Good catch @lazartravica 💯

jsonrpc/dispatcher.go
```diff
@@ -434,6 +435,16 @@ func (d *Dispatcher) getBlockHeaderImpl(number BlockNumber) (*types.Header, erro
 	}
 }
```
zivkovicmilos: I would go even a step further and suggest we add a comment for this method that explains what it does. What does `available` mean in this context? I think it would be clearer if there was an explanation

zivkovicmilos (changes requested):
Looks good, just a few minor nitpicks to be solved 💯

jsonrpc/eth_endpoint.go
munna0908: Fixed it, Here is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/148/commits/fea3a3c8428e7c23ce1cb5d46a0465a4e781ee8c)

jsonrpc/eth_endpoint.go
munna0908: Fixed it, Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/148/commits/fea3a3c8428e7c23ce1cb5d46a0465a4e781ee8c)

jsonrpc/eth_endpoint.go
munna0908: This is not required, since we don't support account manager to sign transactions. Fixed it [Commit-link](https://github.com/0xPolygon/polygon-sdk/commit/7ed9279cd30c72299f4073dc9bf906cde80e4724)

jsonrpc/dispatcher.go
munna0908: This method is no more required, Removed It. Commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/148/commits/41c54a5ed786a9241552993650ae88bdfcfdd47b)

jsonrpc/dispatcher.go
munna0908: This method is removed

jsonrpc/eth_endpoint.go
```diff
@@ -547,3 +547,6 @@ func (e *Eth) Unsubscribe(id string) (bool, error) {
 	ok := e.d.filterManager.Uninstall(id)
 	return ok, nil
 }
+func createBlockNumberPointer(x BlockNumber) *BlockNumber {
+	return &x
+}
```
zivkovicmilos: Please remove this method from the `eth_endpoint.go` file as this file is exclusively meant to house the dispatch endpoints.

You can move it to a more appropriate place (for example `codec.go`, where the constant block number values reside)

zivkovicmilos: I would go even as far as to refactor this method to only support the constant block numbers that we have defined in `codec.go` as we don't use it for anything other than getting the reference to these constants.

munna0908: Sure Milos, will do the required refactoring.

munna0908: @zivkovicmilos  this method can also be used to  return a reference of  non-constant blockNumber value, so i am just moving this into  `codec.go` for now.

Kourin1996 (approved):
LGTM 👍

jsonrpc/eth_endpoint.go
zivkovicmilos: Why would we need a separate function to get a reference to a non-constant `BlockNumber` type?

lazartravica: @munna0908 tagging for visibility

lazartravica (approved):

jsonrpc/eth_endpoint.go
munna0908: @lazartravica Thanks for tagging, will close it ASAP

jsonrpc/dispatcher.go
```diff
@@ -295,7 +296,7 @@ func (d *Dispatcher) handleReq(req Request) ([]byte, Error) {
 	output := fd.fv.Call(inArgs)
 	if err := getError(output[1]); err != nil {
 		d.logInternalError(req.Method, err)
-		return nil, NewInternalError("Internal error")
+		return nil, NewInvalidRequestError(err.Error())
 	}
```
Vuksan: We log internal error and then raise invalid request error?
Why is it `invalid request` error if it is the result of the `Call` operation?

Vuksan (commented):
Wasn't this task supposed to:
```
Review all JSON RPC API methods and update the error response messages to meet infura standard's.
```

I don't see many response messages updated, but I may be wrong.

_Note:_ Please, in the future, let's not reuse existing branches for new PRs (this one was used for [PR#138](https://github.com/0xPolygon/polygon-sdk/pull/138)). Let's always open a new branch for a new PR, otherwise, GitHub tends to get funny with displayed changes.

jsonrpc/dispatcher.go
munna0908: Ganache 

> Wasn't this task supposed to:
> 
> ```
> Review all JSON RPC API methods and update the error response messages to meet infura standard's.
> ```
> 
> I don't see many response messages updated, but I may be wrong.
> 
> _Note:_ Please, in the future, let's not reuse existing branches for new PRs (this one was used for [PR#138](https://github.com/0xPolygon/polygon-sdk/pull/138)). Let's always open a new branch for a new PR, otherwise, GitHub tends to get funny with displayed changes.

@Vuksan most of the error messages are aligned, only few need to be update, I have done that.

munna0908: @Vuksan  It should be internal error optimally 😅, but ganache is returning an invalid request error.

Vuksan (approved):

jsonrpc/eth_endpoint.go
munna0908: @zivkovicmilos Fixed it Commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/148/commits/37f93254e322b6c583fb816435ad0b0f5563baae)

zivkovicmilos: Looks good 👍

zivkovicmilos (approved):
Looks much cleaner now.

Just make sure to resolve any standing conflicts with the develop branch, and we're good to go 🚀

PR #151, opened 2021-09-10
Add PriceLimit setting in TxPool

Kourin1996:
# Description

This PR adds PriceLimit check in TxPool so that TxPool reject incoming transactions with low GasPrice.
This change contains following new command line options for the setting:
* --price-limit [PRICE_LIMIT]
  + Sets PriceLimit which is the threshold of Tx GasPrice for acceptable transaction in TxPool
  + This limitation is applied to only remote transactions (txs via Gossip or Reorg) when only this option is given.
* --nolocals
  + Enable check of GasPrice for local transaction just like remote transaction.
* --locals [COMMA_SEPARATED_ADDRESSES]
  + Skip GasPrice check for the transactions sent from given addresses

Doc PR: 0xPolygon/polygon-sdk-docs#17

### Case 1, Test gas price checking for remote transaction with low gas price
(1) Run first IBFT node without `--seal` and remaining IBFT nodes with `--seal` and `--price-limit [PRICE_LIMIT]`
(2) Send transaction with lower gas price than `PRICE_LIMIT` to node1 by JSON-RPC
(3) Transaction shouldn't be processed because first node won't build block and remaining node won't take the transaction because gas price is too low.

### Case 2, Test gas price checking for remote transaction with high gas price
(1) Start nodes as in the Case 1
(2) Send transaction with larger gas price than `PRICE_LIMIT` to node1 by JSON-RPC
(3) Transaction should be processed because the Node 2-4 will take the tx into TxPool and one of them will process it.

### Case3, Test skipping GasPrice checking for local transaction
(1) Run all IBFT nodes with just `--seal`
(2) Send transaction with lower gas price than `PRICE_LIMIT` to node1 by JSON-RPC
(3) Transaction should be processed because first node can take the tx into txpool and process it.

### Case4, Test gas price checking for local transaction when `NoLocals` is enabled
(1) Run all IBFT nodes with `--seal` and `--nolocals`
(2) Send transaction with lower gas price than `PRICE_LIMIT` to node1 by JSON-RPC
(3) Server should return error because failed to validate tx due to low gas price in node1

### Case5, Test gas price checking for local transaction when `NoLocals` is enabled
(1) Run first IBFT nodes with `--seal`, `--nolocals`, `--locals [ADDRESS]` and run remaining nodes with `--seal`, `--nolocals`
(2) Send transaction with lower gas price than `PRICE_LIMIT` to node1 by JSON-RPC
(3) Transaction should be processed because first node can take the tx into txpool and process it. (From in Tx is registered in local address list)

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -279,7 +279,7 @@ func TestTxPool_TransactionCoalescing(t *testing.T) {
 	)
 }
 
-func TestGasLimit(t *testing.T) {
+func TestPriceLimit(t *testing.T) {
 	signer := &crypto.FrontierSigner{}
 	senderKey, senderAddr := tests.GenerateKeyAndAddr(t)
 	_, receiverAddr := tests.GenerateKeyAndAddr(t)
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -298,17 +298,17 @@ func TestPriceLimit(t *testing.T) {
 	}{
 		{
 			name:             "tx should propagate",
-			numNodes:         5,
+			numNodes:         3,
 			locals:           nil,
 			noLocals:         false,
 			priceLimit:       10,
 			gasPrice:         100,
 			err:              nil,
-			numReceivedNodes: 5,
+			numReceivedNodes: 3,
 		},
 		{
 			name:             "tx should be rejected due to low gas price",
-			numNodes:         5,
+			numNodes:         3,
 			locals:           nil,
 			noLocals:         true,
 			priceLimit:       10,
@@ -318,7 +318,7 @@ func TestPriceLimit(t *testing.T) {
 		},
 		{
 			name:             "tx should not propagate",
-			numNodes:         5,
+			numNodes:         3,
 			locals:           nil,
 			noLocals:         false,
 			priceLimit:       10,
@@ -328,13 +328,13 @@ func TestPriceLimit(t *testing.T) {
 		},
 		{
 			name:             "tx should propagate because sender address is treated as local transaction",
-			numNodes:         5,
+			numNodes:         3,
 			locals:           []string{senderAddr.String()},
 			noLocals:         true,
 			priceLimit:       10,
 			gasPrice:         0,
 			err:              nil,
-			numReceivedNodes: 5, // only first node receive
+			numReceivedNodes: 3,
 		},
 	}
 
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"fmt"
 	"io"
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/0xPolygon/polygon-sdk/crypto"
 	"github.com/0xPolygon/polygon-sdk/e2e/framework"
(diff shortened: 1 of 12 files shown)


txpool/txpool.go
```diff
@@ -29,12 +29,21 @@ var (
...
 	ErrAlreadyKnown        = errors.New("already known")
 )
 
+type TxOrigin = string
+
+const (
+	OriginAddTxn TxOrigin = "addTxn"
+	OriginReorg  TxOrigin = "reorg"
+	OriginGossip TxOrigin = "gossip"
+)
```
zivkovicmilos, line 48: Love that it's structured like this, instead of just having it baked into the code

zivkovicmilos (approved):
Looks great 💯

Kourin1996:
~~Need to check whether tx comes from node operator or other user. If tx is emitted by node operator, then can ignore checking gasPrice.~~

e2e/txpool_test.go
```diff
@@ -276,3 +278,124 @@ func TestTxPool_TransactionCoalescing(t *testing.T) {
 		"To address balance mismatch after gap transaction",
 	)
 }
+
+func TestGasLimit(t *testing.T) {
```
lazartravica: This should be TestPriceLimit (as below), not gaslimit?

txpool/txpool.go
```diff
@@ -184,12 +208,16 @@ func (t *TxPool) AddTx(tx *types.Transaction) error {
 
 // addImpl validates the tx and adds it to the appropriate account transaction queue.
 // Additionally, it updates the global valid transactions queue
-func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
+func (t *TxPool) addImpl(ctx TxOrigin, tx *types.Transaction) error {
```
lazartravica: I'd suggest to rename the `ctx` to something like `origin` to avoid confusion with the context package

txpool/txpool.go
```diff
@@ -641,3 +679,38 @@ func (t *txPriceHeapImpl) Pop() interface{} {
 	*t = old[0 : n-1]
 	return job
 }
+
+type localAccounts struct {
+	accounts map[types.Address]bool
+	mutex    sync.RWMutex
```
lazartravica, line 706: Do we really need a mutex if we are only setting the localAccounts when starting up the node, and after that we are only reading from different threads?

lazartravica (changes requested):
The tests are beautiful @Kourin1996 nice job! Please see my comments

txpool/txpool.go
Kourin1996: There is a possibility that accounts are added to this struct after starting. If noLocals is disabled, address is added to local accounts when new local transaction is inserted to txPool. This is used when reorg happens (In IBFT, it won't happen)

txpool/txpool.go
Kourin1996: Fixed in https://github.com/0xPolygon/polygon-sdk/pull/151/commits/d5e92b6f828f644bfbf9e025fc62d6bea106cf42

e2e/txpool_test.go
Kourin1996: Fixed in https://github.com/0xPolygon/polygon-sdk/pull/151/commits/d5e92b6f828f644bfbf9e025fc62d6bea106cf42

zivkovicmilos (approved):

txpool/txpool.go
lazartravica: Got it, thanks for the answers, we shouldn't count on the consensus having instant finality 👍

lazartravica (approved):

lazartravica:
@Kourin1996 we need a corresponding PR  on the https://github.com/0xPolygon/polygon-sdk-docs repo before merging this PR

You can adopt the structure around CLI commands which Milos has created yesterday.

Also, we will go with the locals and nolocals options you've described above.

Thanks!

Kourin1996:
Doc PR: https://github.com/0xPolygon/polygon-sdk-docs/pull/17

PR #153, opened 2021-09-10
Fix/bootnodes self connection

munna0908:
# Description

This PR avoids node to self-connect, if specified as a boot node.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/network/server.go
+++ b/network/server.go
@@ -149,6 +149,10 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 			if err != nil {
 				return nil, fmt.Errorf("failed to parse bootnode %s: %v", raw, err)
 			}
+			if node.ID == srv.host.ID() {
+				srv.logger.Info("Omitting bootnode with same ID as host", node.ID)
+				continue
+			}
 			// add the bootnode to the peerstore
 			srv.host.Peerstore().AddAddr(node.ID, node.Addrs[0], peerstore.AddressTTL)
 			bootnodes = append(bootnodes, node)
--- a/network/server_test.go
+++ b/network/server_test.go
@@ -223,6 +223,10 @@ func TestSelfConnection_WithBootNodes(t *testing.T) {
 	assert.NoError(t, err)
 	peerAddressInfo, err := StringToAddrInfo("/ip4/127.0.0.1/tcp/10001/p2p/16Uiu2HAmJxxH1tScDX2rLGSU9exnuvZKNM9SoK3v315azp68DLPW")
 	assert.NoError(t, err)
+	//remove the temporary directory
+	t.Cleanup(func() {
+		assert.NoError(t, os.RemoveAll(directoryName))
+	})
 	tests := []struct {
 		name         string
 		bootNodes    []string
@@ -255,6 +259,4 @@ func TestSelfConnection_WithBootNodes(t *testing.T) {
 		})
 	}
 
-	//remove the temporary directory
-	assert.NoError(t, os.RemoveAll(directoryName))
 }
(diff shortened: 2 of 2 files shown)


network/server.go
```diff
@@ -149,6 +149,10 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 			if err != nil {
 				return nil, fmt.Errorf("failed to parse bootnode %s: %v", raw, err)
 			}
+			if node.ID == srv.host.ID() {
+				srv.logger.Info("Omitting bootnode with same ID as host", node.ID)
```
zivkovicmilos, line 153: This can probably be a `Warning` instead of an `Info` log

network/server_test.go
```diff
@@ -209,3 +211,40 @@ func TestNat(t *testing.T) {
...
+		hostSelfAddr := "/ip4/127.0.0.1/tcp/10001/p2p/" + peerId.Pretty()
+		peerAddr := "/ip4/127.0.0.1/tcp/10001/p2p/16Uiu2HAmJxxH1tScDX2rLGSU9exnuvZKNM9SoK3v315azp68DLPW"
+		conf := func(c *Config) {
+			c.NoDiscover = false
+			c.DataDir = directoryName
+			c.Chain.Bootnodes = []string{hostSelfAddr, peerAddr}
+		}
+		srv0 := CreateServer(t, conf)
+		assert.NotContains(t, srv0.discovery.bootnodes, hostSelfAddr)
+	})
```
zivkovicmilos: Good job on the tests 💯 

One nitpick:
We can probably merge out these two test runs into a test table and run them as test cases, like we do in other parts of the SDK.

zivkovicmilos (approved):
Lookin' good 💯

network/server.go
munna0908: I felt `Info` makes more sense, since that particular bootnode entry would be a valid one for other nodes in the cluster.

network/server_test.go
munna0908: Cool will merge them.

network/server_test.go
```diff
@@ -209,3 +211,50 @@ func TestNat(t *testing.T) {
...
+			}
+			srv0 := CreateServer(t, conf)
+
+			assert.Equal(t, srv0.discovery.bootnodes, tt.expectedList)
+
+		})
+	}
+
+	//remove the temporary directory
+	assert.NoError(t, os.RemoveAll(directoryName))
```
Kourin1996: You might want to use `t.Cleanup` or `defer` instead of deleting at the end of test to make sure to delete temporary directory even if `panic()` or `t.Fatal()` is called while testing.

```go
directoryName, err := ioutil.TempDir(os.TempDir(), "")
assert.NoError(t, err)
t.Cleanup(func() {
        assert.NoError(t, os.RemoveAll(directoryName))
})
```

Kourin1996 (changes requested):
It works locally. Thank you. I have left a small comment for improving test. It'd be great if you could check and fix it. Thanks!

network/server_test.go
munna0908: Sure, Thanks for the input.

munna0908: @Kourin1996 Commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/153/commits/61f2ab712310387b46b0480c7a3e44c4f09920da)

network/server_test.go
munna0908: Here's the commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/153/commits/299ab40a6bbfd561b2430672db0b58a6f2d90ab0)

Kourin1996 (approved):
LGTM 👏

network/server.go
lazartravica: Agree with @munna0908 this is a low-severity message.

lazartravica (approved):

Issue #150, opened 2021-09-09
Intrinsic gas too low Error while deploying the contract


poprygun:
### Description

Getting an error using hardhat contract deployment procedure.  I followed the steps in [Getting Started Guide](https://sdk-docs.polygon.technology/docs/how-tos/howto-setup-ibft/howto-set-ibft-locally), and was able to start all nodes without any errors, however, I am not able to deploy a simple contract.

Contract code:

```javascript
//SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

contract Greeter {
  string greeting;

  constructor(string memory _greeting) {
    greeting = _greeting;
  }

  function greet() public view returns (string memory) {
    return greeting;
  }

  function setGreeting(string memory _greeting) public {
    greeting = _greeting;
  }
}

```

hardhat.config.js section with network:

```javascript
module.exports = {
  solidity: "0.8.0",
  defaultNetwork: "polygon",
  networks: {
    polygon: {
      chainId: 100,
      url: 'http://localhost:30002',
      accounts: ['0x${private-key-hardcoded}'],
      gas: "auto",
      gasPrice: "auto",
      timeout: 20000
    }
  }
};
```

deploy.js script:

```javascript
async function main() {
  // We get the contract to deploy
  const [deployer] = await ethers.getSigners();

  console.log("Deploying contracts with the account:", deployer.address);

  console.log("Account balance:", (await deployer.getBalance()).toString());

  const Greeter = await ethers.getContractFactory("Greeter");
  const greeter = await Greeter.deploy("Hello, Hardhat!");

  console.log("Greeter deployed to:", greeter.address);
}

main()
  .then(() => process.exit(0))
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });
```

### Your environment

* macOS Catalina
* version of the Polygon SDK
* current [develop branch](https://github.com/0xPolygon/polygon-sdk.git)

### Steps to reproduce

* Tell us how to reproduce this issue <br />

- Follow the steps  in [Getting Started Guide](https://sdk-docs.polygon.technology/docs/how-tos/howto-setup-ibft/howto-set-ibft-locally) to configure and run polygon-sdk locally
- Configure  sample hardhat project (see hardhat.config.js code above)
- Create sample contract (see provided source above)
- Run `npx hardhat run scripts/deploy.js --network polygon`

* Where the issue is, if you know <br />

* Which commands triggered the issue, if any

- attempt to deploy the contract

[...]

PR #154, opened 2021-09-14
Fix invalid reference values

zivkovicmilos:
# Description

This PR fixes issue #150.

## Problem with `eth_gasPrice`
Within the Polygon SDK, we keep track of an average gas price for transactions, using a rolling average method.
This rolling average method previously didn't take into account the genesis block, so the initial value was set to `0`, causing any `eth_gasPrice` call to return a `0` value if no transactions have been previously executed (the chain has just been started, for example)

## Problem with `eth_estimateGas`
The `eth_estimateGas` issue is indirectly linked to the `eth_gasPrice` issue mentioned before.
The Polygon SDK estimates the gas for a certain transaction by running the transaction multiple times, while keeping the gas within certain bounds. Then, it tried to determine what is the minimum gas amount required for the transaction to go through.

If the transaction had a defined `Gas` field, `eth_estimateGas` would take this value as the high end for its bounding.

The `Gas` field in the transaction is usually a result of a previous `eth_gasPrice` call, and if the value returned from that call is faulty, than the `eth_estimateGas` call would be faulty, as well.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -248,6 +248,9 @@ func (b *Blockchain) writeGenesisImpl(header *types.Header) error {
 		return err
 	}
 
+	// Update the average gas price to take into account the genesis block
+	b.UpdateGasPriceAvg(new(big.Int).SetUint64(header.GasUsed))
+
 	// Advance the head
 	if _, err := b.advanceHead(header); err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


lazartravica (approved):
Good job on the homestead fix as well for estimateGas 👏

Kourin1996 (approved):
LGTM 👏

munna0908 (approved):
LGTM 🚀

PR #155, opened 2021-09-14
Fix/max size of transaction

munna0908:
# Description

This PR adds a check on the transaction size to avoid DOS attacks.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/txpool/txpool_test.go
+++ b/txpool/txpool_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/big"
+	"math/rand"
 	"strconv"
 	"testing"
 
@@ -374,13 +375,16 @@ func TestTxnQueue_Heap(t *testing.T) {
 	})
 }
 
-func generateTx(from types.Address, value *big.Int) *types.Transaction {
+func generateTx(from types.Address, value *big.Int, size uint64) *types.Transaction {
+	data := make([]byte, size)
+	rand.Read(data)
 	return &types.Transaction{
 		From:     from,
 		Nonce:    0,
 		Gas:      validGasLimit,
 		GasPrice: big.NewInt(1),
 		Value:    value,
+		Input:    data,
 	}
 }
 
@@ -461,7 +465,7 @@ func TestTxPool_ErrorCodes(t *testing.T) {
 			pool.AddSigner(poolSigner)
 
 			refAddress := testCase.refAddress
-			txn := generateTx(refAddress, testCase.txValue)
+			txn := generateTx(refAddress, testCase.txValue, 0)
 
 			assert.ErrorIs(t, pool.addImpl("", txn), testCase.expectedError)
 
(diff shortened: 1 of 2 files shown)


lazartravica (approved):

txpool/txpool_test.go
```diff
@@ -374,13 +375,16 @@ func TestTxnQueue_Heap(t *testing.T) {
 	})
 }
 
-func generateTx(from types.Address, value *big.Int) *types.Transaction {
+func generateTx(from types.Address, value *big.Int, size uint64) *types.Transaction {
+	data := make([]byte, size)
+	rand.Read(data)
```
zivkovicmilos: `generateTx` is supposed to be a general purpose method for transaction generation. If tomorrow we have a use-case where we want to add actual input to this helper method, it would render it useless since you've overwritten the `Input` storage with your predefined input within the method.

I suggest 2 options for fixing this:
1. Add an `Input` field as a parameter which will be passed to the transaction `Input` field.
2. Leave the function the way it was, but set the `Input` field from within your test case. This way, you save the additional parameter pass for tests which don't even utilize the `Input` field, but use this method

zivkovicmilos (changes requested):
Looks good 💯 

Just a minor change requested

txpool/txpool_test.go
munna0908: That makes sense, will do the required modifications. Thanks @zivkovicmilos

txpool/txpool_test.go
```diff
@@ -470,7 +474,46 @@ func TestTxPool_ErrorCodes(t *testing.T) {
...
+	}{
 
+		{
+			name:    "Tx_Data is greater than  MAX_SIZE",
+			address: types.Address{0x1},
+			succeed: false,
+			size:    132096,
+		},
+		{
+			name:    "Tx_Data is less than  MAX_SIZE",
```
Kourin1996: Just curious questions, double spaces between `than` and `MAX_SIZE` is intentional?

Kourin1996 (approved):
Looks good to me. I have left a small comment.

txpool/txpool_test.go
munna0908: No its not 😅

txpool/txpool_test.go
munna0908: Fixed it. Here's the commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/155/commits/ea46815256cb0fa65352cf667ec28b87c1e87c40)

txpool/txpool_test.go
munna0908: Fixed it @Kourin1996

zivkovicmilos (approved):
Thank you for making the requested change 🙏 

Looks great now 💯

munna0908:
@Vuksan can u please approve this PR

lazartravica:
@munna0908 Feel free to merge this PR 👍

lazartravica (approved):

PR #159, opened 2021-09-20
Add eth_getBlockTransactionCountByNumber

mrwillis:
# Description

Added https://eth.wiki/json-rpc/API#eth_getblocktransactioncountbynumber

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -19,26 +19,31 @@ func (e *Eth) ChainId() (interface{}, error) {
 	return argUintPtr(e.d.chainID), nil
 }
 
-// GetBlockByNumber returns information about a block by block number
-func (e *Eth) GetBlockByNumber(number BlockNumber, fullTx bool) (interface{}, error) {
-	var num uint64
+func GetNumericBlockNumber(number BlockNumber, e *Eth) (uint64, error) {
 	switch number {
 	case LatestBlockNumber:
-		num = e.d.store.Header().Number
+		return e.d.store.Header().Number, nil
 
 	case EarliestBlockNumber:
-		return nil, fmt.Errorf("fetching the earliest header is not supported")
+		return 0, fmt.Errorf("fetching the earliest header is not supported")
 
 	case PendingBlockNumber:
-		return nil, fmt.Errorf("fetching the pending header is not supported")
+		return 0, fmt.Errorf("fetching the pending header is not supported")
 
 	default:
 		if number < 0 {
-			return nil, fmt.Errorf("invalid argument 0: block number larger than int64")
+			return 0, fmt.Errorf("invalid argument 0: block number larger than int64")
 		}
-		num = uint64(number)
+		return uint64(number), nil
 	}
+}
 
+// GetBlockByNumber returns information about a block by block number
+func (e *Eth) GetBlockByNumber(number BlockNumber, fullTx bool) (interface{}, error) {
+	num, err := GetNumericBlockNumber(number, e)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


lazartravica (approved):

jsonrpc/eth_endpoint.go
```diff
@@ -55,6 +55,31 @@ func (e *Eth) GetBlockByHash(hash types.Hash, fullTx bool) (interface{}, error)
...
+
+	case PendingBlockNumber:
+		return nil, fmt.Errorf("fetching the pending header is not supported")
+
+	default:
+		if number < 0 {
+			return nil, fmt.Errorf("invalid argument 0: block number larger than int64")
+		}
+		num = uint64(number)
+	}
```
Kourin1996: I guess this part is the same to the one in `GetBlockByNumber`. Could I trouble you to define new function and move this?

Kourin1996 (commented):
Thank you for the PR! It works locally. I have left a small comment just for code improvement. I'd appreciated if you could check. Thank you!

jsonrpc/eth_endpoint.go
mrwillis: Done

mrwillis:
@Kourin1996 Is it possible to support the `pending` block parameter? I notice this is not supported but part of the spec 

https://eth.wiki/json-rpc/API#the-default-block-parameter

I assume `pending` in this context actually means blocks that aren't yet mined?

Kourin1996:
> @Kourin1996 Is it possible to support the pending block parameter? I notice this is not supported but part of the spec
> https://eth.wiki/json-rpc/API#the-default-block-parameter
> I assume pending in this context actually means blocks that aren't yet mined?

Hello, @mrwillis. Thank you for your suggestions! I think so, `pending` means a block which is created but not completed mining yet. In IBFT, there is no such block. (But there is a block which has been produced but not confirmed)
We're starting discussion about this. And we can do this task in different PR.

mrwillis:
> > @Kourin1996 Is it possible to support the pending block parameter? I notice this is not supported but part of the spec
> > https://eth.wiki/json-rpc/API#the-default-block-parameter
> > I assume pending in this context actually means blocks that aren't yet mined?
> 
> Hello, @mrwillis. Thank you for your suggestions! I think so, `pending` means a block which is created but not completed mining yet. In IBFT, there is no such block. (But there is a block which has been produced but not confirmed)
> We're starting discussion about this. And we can do this task in different PR.

Thanks @Kourin1996! Sounds good.

zivkovicmilos:
> 
> 
> @Kourin1996 Is it possible to support the `pending` block parameter? I notice this is not supported but part of the spec
> 
> https://eth.wiki/json-rpc/API#the-default-block-parameter
> 
> I assume `pending` in this context actually means blocks that aren't yet mined?

Hey @mrwillis,

Currently we don't have any support for `pending` or I think `earliest`. 
Would this be something that you'd need soon?

zivkovicmilos (approved):
Looks good to me, thank you for the contribution 🙏 💯

mrwillis:
> > @Kourin1996 Is it possible to support the `pending` block parameter? I notice this is not supported but part of the spec
> > https://eth.wiki/json-rpc/API#the-default-block-parameter
> > I assume `pending` in this context actually means blocks that aren't yet mined?
> 
> Hey @mrwillis,
> 
> Currently we don't have any support for `pending` or I think `earliest`. Would this be something that you'd need soon?

Hey @zivkovicmilos, so I only ask for this as it seems to be used by block explorers to see the transactions in the "pending" block

Kourin1996 (approved):
Approved. Sorry actually I checked latest PR before but I didn't change status.

PR #161, opened 2021-09-22
Fix/ibft syncer issue

### LINKED ISSUE #152: Error `polygon.consensus.syncer: failed to notify` after adding + removing peers/validators
### Description

I've seen many occurrences of this new error after adding and removing peers and validators to the original validator set:
```
2021-09-10T16:32:00.056Z [ERROR] polygon.consensus.syncer: failed to notify: err="rpc error: code = Unavailable desc = connection error: desc = "transport: failed to write client preface: stream reset""
```

It seems as if the active nodes may somehow still maintain a reference to old (removed) peers and/or validators even so they are no longer visible when running the SDK commands:

```
./main peers list
```
and
```
./main ibft snapshot
```

I've made sure to run `./main ibft propose --vote drop` on greater than 50% of active validators (removing it from the list that `ibft snapshot` returns, as expected) but it seems I only see an even greater frequency of the same error -- for reference below, validators were removed at 12:08 and 12:14:
Screen Shot 2021-09-10 at 12 40 42 PM

### Your environment

* OS and version `OSX Big Sur 11.5.2 (20G95)`
* version of the Polygon SDK `0b2cc26`

### Steps to reproduce

1. Run `ibft server` for 4 bootnodes
2. Add peer `new_node1` using `peers add`
3. Vote in new validator `new_node1` using `ibft propose --vote auth`
4. Vote in new validator `new_node2` using `ibft propose --vote auth`
5. Stop validators `new_node1`, `new_node2`
6. Drop both new validators `new_node1`, `new_node2` using `ibft propose --vote drop`
7. Verify that `peers list` shows 3 peers for each node (as expected for 4 bootnodes)
8. Verify that `ibft snapshot` shows 4 validator addresses total (as expected for 4 bootnodes)
9. See new errors in logs

### Expected behaviour

I shouldn't any of these errors.

### Actual behaviour

I see many of these errors in the logs on all 4 bootnodes.

I initially suspected that the validators were stopped prematurely without being dropped by the active validators but when attempting to remove them from the validator list, it seems that I only see more of the same error in the logs.

munna0908:
# Description

This PR resolves issue #152 

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/protocol/syncer.go
+++ b/protocol/syncer.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"math/big"
 	"sync"
+	"sync/atomic"
 
 	"github.com/0xPolygon/polygon-sdk/blockchain"
 	"github.com/0xPolygon/polygon-sdk/network"
@@ -144,8 +145,8 @@ type Syncer struct {
 	logger     hclog.Logger
 	blockchain blockchainShim
 
-	peers     map[peer.ID]*syncPeer
-	peersLock sync.RWMutex
+	peers    sync.Map // Maps peer.ID -> syncPeer
+	peersNum int64    // Number of entries in the peers map
 
 	serviceV1 *serviceV1
 	stopCh    chan struct{}
@@ -160,7 +161,7 @@ type Syncer struct {
 func NewSyncer(logger hclog.Logger, server *network.Server, blockchain blockchainShim) *Syncer {
 	s := &Syncer{
 		logger:     logger.Named("syncer"),
-		peers:      map[peer.ID]*syncPeer{},
+		peersNum:   0,
 		stopCh:     make(chan struct{}),
 		blockchain: blockchain,
 		server:     server,
@@ -220,18 +221,15 @@ const syncerV1 = "/syncer/0.1"
 // enqueueBlock adds the specific block to the peerID queue
 func (s *Syncer) enqueueBlock(peerID peer.ID, b *types.Block) {
 	s.logger.Debug("enqueue block", "peer", peerID, "number", b.Number(), "hash", b.Hash())
-	s.peersLock.RLock()
-	p, ok := s.peers[peerID]
-	s.peersLock.RUnlock()
+
+	foundPeer, ok := s.peers.Load(peerID)
 	if ok {
-		p.appendBlock(b)
+		foundPeer.(*syncPeer).appendBlock(b)
 	}
 }
 
 // Broadcast broadcasts a block to all peers
 func (s *Syncer) Broadcast(b *types.Block) {
-	s.peersLock.RLock()
-	defer s.peersLock.RUnlock()
 	// diff is number in ibft
 	diff := new(big.Int).SetUint64(b.Number())
 
@@ -246,11 +244,19 @@ func (s *Syncer) Broadcast(b *types.Block) {
 			Value: b.MarshalRLP(),
 		},
 	}
-	for _, p := range s.peers {
-		if _, err := p.client.Notify(context.Background(), req); err != nil {
+
+	s.peers.Range(func(peerID, foundPeer interface{}) bool {
+		if _, err := foundPeer.(*syncPeer).client.Notify(context.Background(), req); err != nil {
 			s.logger.Error("failed to notify", "err", err)
 		}
-	}
+
+		return true
+	})
+}
+
+// NumSyncPeers returns the number of peers in the sync map. [Thread safe]
+func (s *Syncer) NumSyncPeers() int64 {
+	return atomic.LoadInt64(&s.peersNum)
 }
 
 // Start starts the syncer protocol
@@ -281,9 +287,8 @@ func (s *Syncer) Start() {
 			}
 
 			if evnt.Type == network.PeerEventDisconnected {
-				s.peersLock.Lock()
-				delete(s.peers, evnt.PeerID)
-				s.peersLock.Unlock()
+				s.peers.Delete(evnt.PeerID)
+				atomic.AddInt64(&s.peersNum, -1)
 				continue
 			}
 			if evnt.Type != network.PeerEventConnected {
@@ -306,28 +311,31 @@ func (s *Syncer) Start() {
 func (s *Syncer) BestPeer() *syncPeer {
 	var bestPeer *syncPeer
 	var bestTd *big.Int
-	s.peersLock.RLock()
-	defer s.peersLock.RUnlock()
-	for _, p := range s.peers {
-		status := p.status
+
+	s.peers.Range(func(peerID, foundPeer interface{}) bool {
+		status := foundPeer.(*syncPeer).status
 		if bestPeer == nil || status.Difficulty.Cmp(bestTd) > 0 {
-			bestPeer, bestTd = p, status.Difficulty
+			bestPeer, bestTd = foundPeer.(*syncPeer), status.Difficulty
 		}
-	}
+
+		return true
+	})
+
 	if bestPeer == nil {
 		return nil
 	}
+
 	curDiff := s.blockchain.CurrentTD()
+
 	if bestTd.Cmp(curDiff) <= 0 {
 		return nil
 	}
+
 	return bestPeer
 }
 
 // HandleUser is a helper method that is used to handle new user connections within the Syncer
 func (s *Syncer) HandleUser(peerID peer.ID, conn *grpc.ClientConn) error {
-	s.peersLock.Lock()
-	defer s.peersLock.Unlock()
 	// watch for changes of the other node first
 	clt := proto.NewV1Client(conn)
 
@@ -339,12 +347,15 @@ func (s *Syncer) HandleUser(peerID peer.ID, conn *grpc.ClientConn) error {
 	if err != nil {
 		return err
 	}
-	s.peers[peerID] = &syncPeer{
+
+	s.peers.Store(peerID, &syncPeer{
 		peer:      peerID,
 		client:    clt,
 		status:    status,
 		enqueueCh: make(chan struct{}),
-	}
+	})
+	atomic.AddInt64(&s.peersNum, 1)
+
 	return nil
 }
 
--- a/protocol/syncer.go
+++ b/protocol/syncer.go
@@ -144,7 +144,8 @@ type Syncer struct {
 	logger     hclog.Logger
 	blockchain blockchainShim
 
-	peers map[peer.ID]*syncPeer // TODO: Remove
+	peers     map[peer.ID]*syncPeer
+	peersLock sync.RWMutex
 
 	serviceV1 *serviceV1
 	stopCh    chan struct{}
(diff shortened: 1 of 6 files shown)


protocol/syncer.go
```diff
@@ -315,6 +326,8 @@ func (s *Syncer) BestPeer() *syncPeer {
 
 // HandleUser is a helper method that is used to handle new user connections within the Syncer
 func (s *Syncer) HandleUser(peerID peer.ID, conn *grpc.ClientConn) error {
+	s.peersLock.Lock()
+	defer s.peersLock.Unlock()
```
Kourin1996: There is a lag until you get the status from peer. So it might be a good idea to lock after you get response from peer.

Kourin1996 (changes requested):

protocol/syncer.go
```diff
@@ -144,7 +144,8 @@ type Syncer struct {
 	logger     hclog.Logger
 	blockchain blockchainShim
 
-	peers map[peer.ID]*syncPeer // TODO: Remove
+	peers     map[peer.ID]*syncPeer
+	peersLock sync.RWMutex
```
zivkovicmilos: I think we can improve performance here even more than with using a RW mutex.

Since our peers are inside of a map, we currently have only 1 mutex that, when you grab it, you hold the access to the entire map.
This means that if I'm accessing the `syncPeer` at key value `peerID1` (not actual peer.ID), I'll be holding up any process that wants to access the `syncPeer` at key value `peerID2`.

I have two suggestions:
1. Create a separate mux map, where you would map the peer.ID -> mux. This way, you hold the mux to a single entry, not the entire map
2. Use `sync.Map` to accomplish the behavior mentioned in the first suggestion. [Godoc](https://pkg.go.dev/sync#Map). This is definitely the better route, as this solution takes into account any optimizations that we might miss.

zivkovicmilos (changes requested):
Please check out the proposed changes, otherwise it looks good 💯

protocol/syncer.go
```diff
@@ -219,15 +220,18 @@ const syncerV1 = "/syncer/0.1"
...
+	s.peersLock.RUnlock()
 	if ok {
 		p.appendBlock(b)
 	}
 }
 
 // Broadcast broadcasts a block to all peers
 func (s *Syncer) Broadcast(b *types.Block) {
+	s.peersLock.RLock()
+	defer s.peersLock.RUnlock()
```
Vuksan: Please, only lock+unlock around the statements that actually use the resource.
In this case that would be around the for loop on L249.

Although in human time initialization is fast, there is no need to have a resource lock and a thread waiting when the resource is not yet demanded.

protocol/syncer.go
```diff
@@ -296,7 +306,8 @@ func (s *Syncer) Start() {
 func (s *Syncer) BestPeer() *syncPeer {
 	var bestPeer *syncPeer
 	var bestTd *big.Int
-
+	s.peersLock.RLock()
+	defer s.peersLock.RUnlock()
```
Vuksan: Please move `RUnlock()` call to ~L316, just after the for loop. (no need for `defer` then)

protocol/syncer.go
Vuksan: I agree with Yoshiki.

Please, only lock+unlock around the statements that actually use the resource.
In this case that would be around the statement on L342.

protocol/syncer_test.go
```diff
@@ -0,0 +1,142 @@
...
+	go sync1.Start()
+	go sync2.Start()
+
+	network.MultiJoin(t, srv0, srv1, srv0, srv2, srv1, srv2)
+	time.Sleep(1 * time.Second)
+	//Disconnect peer2
+	srv1.Disconnect(srv2.AddrInfo().ID, "testing")
+	time.Sleep(1 * time.Second)
+	//server1 syncer should not have disconnected peer
+	assert.NotContains(t, sync1.peers, srv2.AddrInfo().ID)
```
Vuksan: You meant `should have disconnected` right? Without `not`?
```suggestion
	// server1 syncer should have disconnected from server2 peer
	assert.NotContains(t, sync1.peers, srv2.AddrInfo().ID)
```

Vuksan (changes requested):
GJ on the test. 👏 
As for comments, all of them are concentrate on lowering the area in which certain resource is under lock.

protocol/syncer.go
zivkovicmilos: Irrelevant after commit 9593ada

protocol/syncer.go
zivkovicmilos: Irrelevant after commit 9593ada

protocol/syncer.go
zivkovicmilos: Irrelevant after commit 9593ada

protocol/syncer_test.go
zivkovicmilos: I've refactored the entire test 👍 

9593ada

zivkovicmilos (approved):

Kourin1996 (approved):
LGTM!

protocol/syncer.go
```diff
@@ -245,17 +247,17 @@ const syncerV1 = "/syncer/0.1"
 func (s *Syncer) enqueueBlock(peerID peer.ID, b *types.Block) {
 	s.logger.Debug("enqueue block", "peer", peerID, "number", b.Number(), "hash", b.Hash())
 
-	p, ok := s.peers[peerID]
+	foundPeer, ok := s.peers.Load(peerID)
```
lazartravica: Be consistent in naming. I suggest calling it simply `peer` instead of `foundPeer`

protocol/syncer.go
```diff
@@ -245,17 +247,17 @@ const syncerV1 = "/syncer/0.1"
...
+		foundPeer.(*syncPeer).appendBlock(b)
 	}
 }
 
 func (s *Syncer) updatePeerStatus(peerID peer.ID, status *Status) {
 	s.logger.Debug("update peer status", "peer", peerID, "latest block number", status.Number, "latest block hash", status.Hash, "difficulty", status.Difficulty)
 
-	if p, ok := s.peers[peerID]; ok {
-		p.updateStatus(status)
+	if p, ok := s.peers.Load(peerID); ok {
```
lazartravica: Be consistent in naming. I suggest calling it simply `peer` instead of `p`

protocol/syncer.go
```diff
@@ -275,11 +277,19 @@ func (s *Syncer) Broadcast(b *types.Block) {
 			Value: b.MarshalRLP(),
 		},
 	}
-	for _, p := range s.peers {
-		if _, err := p.client.Notify(context.Background(), req); err != nil {
+
+	s.peers.Range(func(peerID, foundPeer interface{}) bool {
```
lazartravica: Be consistent in naming. I suggest calling it simply `peer` instead of `foundPeer`

protocol/syncer.go
```diff
@@ -334,19 +344,25 @@ func (s *Syncer) BestPeer() *syncPeer {
 	var bestPeer *syncPeer
 	var bestTd *big.Int
 
-	for _, p := range s.peers {
-		status := p.status
+	s.peers.Range(func(peerID, foundPeer interface{}) bool {
```
lazartravica: Be consistent in naming. I suggest calling it simply `peer` instead of `foundPeer`

protocol/syncer_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// wait until gossip protocol builds the mesh network (https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md)
+	waitCtx, cancelWait := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancelWait()
+
+	numPeers, err := WaitUntilSyncPeersNumber(waitCtx, syncers[1], 2)
+	if err != nil {
+		t.Fatalf("Unable to add sync peers, %v", err)
+	}
+	assert.Equal(t, int64(len(servers)-1), numPeers)
```
lazartravica, line 206: Please provide message for assert

protocol/syncer_test.go
```diff
@@ -0,0 +1,209 @@
...
+	peerToDisconnect := servers[2].AddrInfo().ID
+	servers[1].Disconnect(peerToDisconnect, "testing")
+
+	waitCtx, cancelWait = context.WithTimeout(context.Background(), time.Second*10)
+	defer cancelWait()
+	numPeers, err = WaitUntilSyncPeersNumber(waitCtx, syncers[1], 1)
+	if err != nil {
+		t.Fatalf("Unable to disconnect sync peers, %v", err)
+	}
+	assert.Equal(t, int64(len(servers)-2), numPeers)
```
lazartravica, line 220: Please provide message for assert

protocol/syncer_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// server1 syncer should have disconnected from server2 peer
+	_, found := syncers[1].peers.Load(peerToDisconnect)
+	assert.False(t, found)
```
lazartravica, line 227: Please provide message for assert

lazartravica (changes requested):
Good job on how the test is setup.

I have left minor naming and message related comments.

I don't like the implementation with the `sync.Map` and `sync/atomic.LoadInt64`. It's harder to understand what's going on than a `RWMutex` and we are using 2 synchronization primitives instead of one.

The most critical thing about using `sync.Map` and `sync/atomic.LoadInt64` is that they are 2 atomic operations, not one as it could be if we were using the RWMutex. In this case I don't see how it could affect us in a bad way, but in some other scenario if somebody was to implement the same method for updating a map and the length something like this could happen:

Thread A: Add entry to map
Thread B: Get map
Thread B: Get the length variable
Thread A: Increase the length variable by 1

This would result in Thread B working with a map of N elements, but with a length variable with a value of N-1.

protocol/syncer.go
zivkovicmilos: Fixed 👍 

1943ea5

protocol/syncer.go
zivkovicmilos: Fixed 👍 

1943ea5

protocol/syncer.go
zivkovicmilos: Fixed 👍 

1943ea5

protocol/syncer.go
zivkovicmilos: Fixed 👍 

1943ea5

protocol/syncer_test.go
zivkovicmilos: Fixed 👍 

d89d88e

protocol/syncer_test.go
zivkovicmilos: Fixed 👍 

d89d88e

protocol/syncer_test.go
zivkovicmilos: Fixed 👍 

d89d88e

lazartravica (approved):

Vuksan (approved):
👏

PR #163, opened 2021-09-22
Split up data directories into keystore and data. Add READ_ONLY flag.

mrwillis:
# Description

I split up the data directories so that not all the data is under ./chain for example. Now it is split between `--data-dir` and `--key-dir` which seemed like a more reasonable way of going about it, especially given the fact that you might transfer "the whole chain" to another system or back it up.

The second rationale for this and for the READ_ONLY flag is to allow clients to share state and not participate in consensus. The reasoning for this is to create a sort of "light client" that doesn't validate blocks and only serves to gossip new txns. I dug a bit and realized that I think the only thing that writes to disk is in the IBFT consensus on block writes.

This allows for easy horizontal scaling from a JSON-RPC point of view as there is no longer the requirement for each polygon-sdk node to store the whole chain, but can read from a shared volume instead. 

The idea is kind of like a master write, slave read idea. The primary node would have READ_ONLY=false while the slaves would have READ_ONLY=true.

# Changes include

# Breaking changes

There is a new required flag called `--key-dir` which is mandatory and stores the `consensus` and `libp2p` sub directories previously in the data dir. 

# Checklist

- [ ] I have tested this code

# Additional comments

### DIFF
--- a/command/helper/config.go
+++ b/command/helper/config.go
@@ -27,6 +27,8 @@ type Config struct {
 	Dev         bool
 	DevInterval uint64
 	Join        string
+	KeyDir      string
+	ReadOnly    bool
 }
 
 // Network defines the network configuration params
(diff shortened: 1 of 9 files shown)


mrwillis:
So after some more investigation, this strategy doesn't work too well. Leveldb locks the whole database and it's not meant to be shared between processes. Rocksdb seems to somewhat have support for this but doesn't seem super well tested. 

I think in the future when the SDK is more pluggable (https://sdk-docs.polygon.technology/docs/reference/modules/storage), one could plug in something like dynamoDB if shared state/concurrent reads was important

PR #173, opened 2021-10-05
Extend the genesis command to allow a configurable block gas limit

0xpanoramix:
# Description

This PR extends the genesis command flags by adding a `--block-gas-limit` parameter which will be included in the `genesis.json` file.

It includes tests for a user-defined limit and the default one.

# Changes include

# Checklist

- [X] I have tested this code


command/genesis/genesis.go
```diff
@@ -106,6 +106,15 @@ func (c *GenesisCommand) DefineFlags() {
 		ArgumentsOptional: false,
 		FlagOptional:      true,
 	}
+
+	c.FlagMap["blockLimit"] = helper.FlagDescriptor{
```
lazartravica: All CLI arguments we have are `snake-case`, change this to `block-limit`

command/genesis/genesis_test.go
```diff
@@ -0,0 +1,133 @@
+package genesis_test
```
lazartravica: Although your test is completely correct, it's too much code just to check that our code is writting JSON files correctly, in general we want to test the entire implications of running a command and not just the immediate result (notice how none of the subdirectories under `command/` have any `*_test.go` files). 

A test that I think would make much more sense is a full on e2e test. You can take a look at [TestPreminedBalance](https://github.com/0xPolygon/polygon-sdk/blob/96be90487c144f7b1589ac65aad3f741c81644e3/e2e/transaction_test.go#L56-L103).

What I encourage you to try to do is to create a test which will setup the TestServer (just as in the above linked test), call the JSON-RPC getBlockByNumber method for the genesis block and assert that the gasLimit is what you expect it to be.

lazartravica (changes requested):
Excellent job @PtitLuca! 

I think this is the perfect time to get acquainted with our e2e testing framework for spinning up Test Servers with custom configurations.

Please take a look at the comments. 🤘

command/genesis/genesis_test.go
zivkovicmilos: I think it's clever that he wrote out the command output to the file and later checked it. 
We should probably keep it now that it's written out, but add the e2e test that you've mentioned 👍 

One thing I'd change is to just place these generated files in a sub-directory that makes it clear it was not a part of the original codebase, but just a result of the tests, for example `genesis-tests/genesis.json`

zivkovicmilos (approved):
Lookin' good 💯 

Please check out @lazartravica's comments regarding the tests.

Kourin1996 (commented):
Hi, Thank you for the PR! I have noticed 2 things and It'd be great if you could check them.

(1) dev command has similar command option with different name.
When you run node with dev mode, you don't have to run `genesis` command before node starts. In this case, genesis.json will be created automatically if it doesn't exist. `dev` command supports `gas-limit` option already. I think we can unify the option names.

(2) Block Gas Limit is not set is subsequent blocks.
When you give `--block-gas-limit` option on genesis command, block gas limit in genesis.json and genesis block change. But blocks from 1 are still using `100000000` due to hardcoded value in `buildBlock` in consensus/ibft.go or `writeNewBlock` in consensus/dev.go. A block gas limit adjustment would be another task but I think you can inherit parent's block gas limit for now.

Kourin1996 (approved):
Thank you for fixing. Looks good to me 👌

command/genesis/genesis.go
lazartravica: @PtitLuca thanks for the fix. Please include the commit hash in the comment to the requested change, it's way easier to keep track of what and why was changed that way.

Thanks

lazartravica (approved):
Excellent job Luca!

Congratulations on your first ready-to-merge PR :)

Feel free to merge.

Keep it coming 🤘

PR #175, opened 2021-10-06
Add unit tests for syncer

Kourin1996:
# Description

This PR adds unit tests for syncer in protocol package

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/blockchain/storage/leveldb/leveldb_test.go
+++ b/blockchain/storage/leveldb/leveldb_test.go
@@ -19,6 +19,9 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
 		t.Fatal(err)
 	}
 	close := func() {
+		if err := s.Close(); err != nil {
+			t.Fatal(err)
+		}
 		if err := os.RemoveAll(path); err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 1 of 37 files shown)


zivkovicmilos (approved):
Lookin' good 💯

lazartravica (changes requested):
👌 💪 

But before approving, just 2 questions out of curiousity:

1. `purgeBlocks` fix in protocol/syncer.go => how did the problem appear, what can the given error cause?
2. Why did you in some tests use the term `User` and in others `Peer`? Shouldn't they all be Peers?

Kourin1996:
> purgeBlocks fix in protocol/syncer.go => how did the problem appear, what can the given error cause?

purgeBlocks removes a block with given hash and before that, because syncer expects next block from latest block. For that reason, I fixed. I think original code should work because syncer just try to write latest block to own chain again in the first time.

> Why did you in some tests use the term User and in others Peer? Shouldn't they all be Peers?

I just match test name with the target function.
Renamed at https://github.com/0xPolygon/polygon-sdk/pull/175/commits/0780729f98c3c1f0c843588b0973c1223aac3e8f

lazartravica (approved):
@Kourin1996 LGTM! Feel free to merge

PR #181, opened 2021-10-14
Enable non-validator to broadcast block + Fix blocking issue in syncer

Kourin1996:
# Description

This PR fixes 2 issues:
- Fix issue that non-validator wasn't working as syncer
- Fix blocking issue when non-validator chooses another non-validator which isn't advancing block as sync peer

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -139,6 +139,24 @@ func WaitUntilTxPoolFilled(ctx context.Context, srv *TestServer, requiredNum uin
 	return res.(*txpoolProto.TxnPoolStatusResp), nil
 }
 
+// WaitUntilBlockMined waits until server mined block with bigger height than given height
+// otherwise returns timeout
+func WaitUntilBlockMined(ctx context.Context, srv *TestServer, desiredHeight uint64) (uint64, error) {
+	clt := srv.JSONRPC().Eth()
+	res, err := RetryUntilTimeout(ctx, func() (interface{}, bool) {
+		height, err := clt.BlockNumber()
+		if err == nil && height >= desiredHeight {
+			return height, false
+		}
+		return nil, true
+	})
+
+	if err != nil {
+		return 0, err
+	}
+	return res.(uint64), nil
+}
+
 func RetryUntilTimeout(ctx context.Context, f func() (interface{}, bool)) (interface{}, error) {
 	type result struct {
 		data interface{}
--- a/e2e/syncer_test.go
+++ b/e2e/syncer_test.go
@@ -1,0 +1,46 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/0xPolygon/polygon-sdk/e2e/framework"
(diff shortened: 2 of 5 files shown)


lazartravica (approved):

e2e/framework/helper.go
```diff
@@ -140,6 +140,24 @@ func WaitUntilTxPoolFilled(ctx context.Context, srv *TestServer, requiredNum uin
 	return res.(*txpoolProto.TxnPoolStatusResp), nil
 }
 
+// WaitUntilBlockMined waits until server mined block with bigger height than given height
+// otherwise returns timeout
+func WaitUntilBlockMined(ctx context.Context, srv *TestServer, desiredHeight uint64) (uint64, error) {
```
zivkovicmilos, line 145: We had a similar func here, but this is also good 👍 :
https://github.com/0xPolygon/polygon-sdk/blob/2813f9b85a8408285e07e197a04fe5e5c93f2e88/e2e/txpool_test.go#L27-L51

e2e/syncer_test.go
```diff
@@ -0,0 +1,46 @@
...
+	// Wait until some blocks are mined
+	ctx2, cancel2 := context.WithTimeout(context.Background(), time.Minute)
+	t.Cleanup(func() {
+		cancel2()
+	})
+	_, err := framework.WaitUntilBlockMined(ctx2, ibftManager.GetServer(0), desiredHeight)
+	assert.NoError(t, err)
+
+	// Get latest block height
+	latestBlockHeight, err := ibftManager.GetServer(0).JSONRPC().Eth().BlockNumber()
```
zivkovicmilos: We should probably export this out into a separate func (fetching the latest block height)

protocol/syncer.go
```diff
@@ -67,18 +79,25 @@ func (s *syncPeer) purgeBlocks(lastSeen types.Hash) {
 	}
 }
 
+// ErrPopTimeout
```
zivkovicmilos: Leftover?

zivkovicmilos (approved):
Looks great 💯

e2e/framework/helper.go
Kourin1996: Actually it's different from it because waitForBlock waits for next event while WaitUntilBlockMined waits the block with specific height is added to chain. I tried to integrate but didn't

protocol/syncer.go
Kourin1996: Removed at https://github.com/0xPolygon/polygon-sdk/pull/181/commits/d2609cbb85c058030a1acbe938e58bb5c99d7d3d

e2e/syncer_test.go
Kourin1996: Added at https://github.com/0xPolygon/polygon-sdk/pull/181/commits/d2609cbb85c058030a1acbe938e58bb5c99d7d3d

Issue #179, opened 2021-10-07
Block writing still stops in non-sealing nodes due to nil BestPeer


dankostiuk:
### Description

It looks like we are still observing the same issue as originally described in https://github.com/0xPolygon/polygon-sdk/issues/167

After investigating, this seems to be related to some inconsistencies in how we determine the best peer to connect to within `BestPeer()`. We are seeing that at times, `curDiff` is always greater than `bestTd`, resulting in this function always returning nil.

We've adding some extra logging in `BestPeer` to help investigate:
```go 
func (s *Syncer) BestPeer() *syncPeer {
	var bestPeer *syncPeer
	var bestTd *big.Int

	for _, p := range s.peers {
		status := p.status
		if bestPeer == nil || status.Difficulty.Cmp(bestTd) > 0 {
			bestPeer, bestTd = p, status.Difficulty
		}
	}
	if bestPeer == nil {
		return nil
	}
	curDiff := s.blockchain.CurrentTD()
	fmt.Println("BestPeer() curDiff = ", curDiff)
	fmt.Println("BestPeer() bestTd = ", bestTd)
	if bestTd.Cmp(curDiff) <= 0 {
		return nil
	}
	return bestPeer
}
```

Although the change https://github.com/0xPolygon/polygon-sdk/commit/5170883f9de116c18ad32640c46af8e9de5f1a84 was put in (to broadcast the header's difficulty instead of the block number), it appears that we are actually initializing the header's `Difficulty` to be the same value as the header's `Number`, resulting in a broadcast of the same value:

https://github.com/0xPolygon/polygon-sdk/blob/bba205e6ed4aad9522450558ed3dbad67664e723/consensus/ibft/ibft.go#L349-L353

Upon attempting to correct L353 above, we are still seeing the same issue as can be seen in our logs. Below the bestTd is always 464670654277. Furthermore, the difficulty outputted in the `enqueue block` debug statement has the same value as the block number:
```
2021-10-07T14:11:43.046Z [DEBUG] polygon.consensus.syncer: enqueue block: peer=16Uiu2HAmBv5UkctkZK1ui2tMip6XqpykvMnrzDuLGNjN38YDztmW number=982519 hash=0xfbda9f200b380eeab2eeb9fb2e80bb25a925929b55c1f2a35d38dbc085f1c831
[...]

PR #182, opened 2021-10-14
Total Difficulty in IBFT

dbrajovic:
# Description

Defining total difficulty in our IBFT implementation.
Addresses [#179](https://github.com/0xPolygon/polygon-sdk/issues/179)

Update:

Hardcoding `GenesisDifficulty` to 1 breaks backwards compatibility for existing chains due to new nodes having to generate the genesis block with difficulty 1.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

This PR is about coming to terms on what Total Difficulty means in IBFT.

In other consensuses, TD is what helps identify the correct chain (fork) based on the sum of all blocks' `difficulty` field (which varies from block to block). 
In IBFT, TD has nothing to do with the `difficulty` field, but serves the same purpose, in a sense that it decides what is the latest block (number) - hence this value should represent the block's `number` field.

In short (our IBFT impl):
```
Block's difficulty -> Block's number
Block's TD -> Sum of all previous block numbers (up to current)
Chain TD -> TD of the latest block in the chain
```

Related issues:
[https://github.com/0xPolygon/polygon-sdk/issues/179](#179)

This fix should provide the `Syncer` with the correct value (fetched with `CurrentTD()`) in order to resolve its best peer.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -290,8 +290,9 @@ func (b *Blockchain) writeCanonicalHeader(event *Event, h *types.Header) error {
 		return fmt.Errorf("parent difficulty not found")
 	}
 
+	_ = td
 	// calculate new TD (IBFT)
-	newTD := big.NewInt(0).Add(td, big.NewInt(1))
+	newTD := big.NewInt(0).SetUint64(h.Difficulty)
 	if err := b.db.WriteCanonicalHeader(h, newTD); err != nil {
 		return err
 	}
@@ -332,8 +333,9 @@ func (b *Blockchain) advanceHead(newHeader *types.Header) (*big.Int, error) {
 		currentDiff = td
 	}
 
+	_ = currentDiff
 	// Calculate the new difficulty
-	diff := big.NewInt(1).Add(currentDiff, new(big.Int).SetUint64(newHeader.Difficulty))
+	diff := big.NewInt(0).SetUint64(newHeader.Difficulty)
 	if err := b.db.WriteDiff(newHeader.Hash, diff); err != nil {
 		return nil, err
 	}
@@ -736,10 +738,10 @@ func (b *Blockchain) dispatchEvent(evnt *Event) {
 
 // writeHeaderImpl writes a block and the data, assumes the genesis is already set
 func (b *Blockchain) writeHeaderImpl(evnt *Event, header *types.Header) error {
-	head := b.Header()
+	currentHead := b.Header()
 
 	// Write the data
-	if header.ParentHash == head.Hash {
+	if header.ParentHash == currentHead.Hash {
 		// Fast path to save the new canonical header
 		return b.writeCanonicalHeader(evnt, header)
 	}
@@ -748,7 +750,7 @@ func (b *Blockchain) writeHeaderImpl(evnt *Event, header *types.Header) error {
 		return err
 	}
 
-	headerDiff, ok := b.readDiff(head.Hash)
+	currentDiff, ok := b.readDiff(currentHead.Hash)
 	if !ok {
 		panic("failed to get header difficulty")
 	}
@@ -762,13 +764,13 @@ func (b *Blockchain) writeHeaderImpl(evnt *Event, header *types.Header) error {
 		)
 	}
 
+	_ = parentDiff
+	incomingDiff := big.NewInt(0).SetUint64(header.Difficulty)
+
 	// Write the difficulty
 	if err := b.db.WriteDiff(
 		header.Hash,
-		big.NewInt(1).Add(
-			parentDiff,
-			new(big.Int).SetUint64(header.Difficulty),
-		),
+		incomingDiff,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


dbrajovic:
After tuning the unit tests, I discovered that TD is calculated in other places as well (`advanceHead()`).  At all points, there is a fetching of the parent difficulty used only to calculate new TD (which is redundant in this new approach). 

I've left `_ = variableName` to suppress compiler warnings - I intend to remove these once proven that they were not meant for anything else.

Regarding the TD field in the units test (blockchain_test.go), this is the value we would like to see hardcoded to 1, in which case, all that is needed is to hardcode `GetChainTD()` to always return 1.

dbrajovic:
After revising the aforementioned calculation of TD, I decided to go with the less invasive approach of hardcoding each block's difficulty to 1, keeping the existing logic (summing of difficulties) intact, such that the following are true:

> Block difficulty = 1
> Block TD = Block number (1 + 1 + ... + 1)
> ChainTD = TD of the latest block (its number)

consensus/ibft/ibft.go
```diff
@@ -880,8 +880,8 @@ func (i *Ibft) verifyHeaderImpl(snap *Snapshot, parent, header *types.Header) er
 		return fmt.Errorf("invalid sha3 uncles")
 	}
 
-	// difficulty has to match number
-	if header.Difficulty != header.Number {
+	// difficulty has to be 1 (except for genesis block)
+	if header.Difficulty != 1 && header.Number != 0 {
```
zivkovicmilos: Why does the genesis block have a difficulty != 1?

zivkovicmilos (changes requested):
Looks great to me 💯 

Left a minor question that needs clarification

consensus/ibft/ibft.go
dbrajovic: Misplaced difficulty check. The failing `TestTransition_AcceptState_Validator_VerifyCorrect` unit test lead me to that conclusion.

[Fix](https://github.com/0xPolygon/polygon-sdk/blob/a39a21bb547f99417d645b59ecf087fec2835371/chain/chain.go#L20) (breaking change)

zivkovicmilos (approved):

Looks great now 💯

blockchain/blockchain.go
```diff
@@ -285,21 +285,21 @@ func (b *Blockchain) GetTD(hash types.Hash) (*big.Int, bool) {
...
 func (b *Blockchain) writeCanonicalHeader(event *Event, h *types.Header) error {
-	td, ok := b.readDiff(h.ParentHash)
+	parentTD, ok := b.readDiff(h.ParentHash)
 	if !ok {
 		return fmt.Errorf("parent difficulty not found")
 	}
 
-	diff := big.NewInt(1).Add(td, new(big.Int).SetUint64(h.Difficulty))
-	if err := b.db.WriteCanonicalHeader(h, diff); err != nil {
+	currentTD := big.NewInt(1).Add(parentTD, new(big.Int).SetUint64(h.Difficulty))
```
Kourin1996: I think this won't cause a misleading. (Just I thought why `1` here?)

```go
currentTD := new(big.Int).Add(parentTD, new(big.Int).SetUint64(h.Difficulty))
```

or

```go
currentTD := big.NewInt(0).Add(parentTD, new(big.Int).SetUint64(h.Difficulty))
```

blockchain/blockchain.go
```diff
@@ -322,25 +322,25 @@ func (b *Blockchain) advanceHead(newHeader *types.Header) (*big.Int, error) {
...
 			return nil, fmt.Errorf("parent difficulty not found")
 		}
-		currentDiff = td
+		parentTD = td
 	}
 
 	// Calculate the new difficulty
-	diff := big.NewInt(1).Add(currentDiff, new(big.Int).SetUint64(newHeader.Difficulty))
-	if err := b.db.WriteDiff(newHeader.Hash, diff); err != nil {
+	newTD := big.NewInt(1).Add(parentTD, new(big.Int).SetUint64(newHeader.Difficulty))
```
Kourin1996: This is also

```go
newTD := new(big.Int).Add(parentTD, new(big.Int).SetUint64(newHeader.Difficulty))
```

Kourin1996 (commented):
Thank you for the PR. I like renaming variable to proper one. I've left small comments. I'd appreciated if you could check and fix if need.

blockchain/blockchain.go
dbrajovic: Previous implementation logic. 

Addressed in this [commit](https://github.com/0xPolygon/polygon-sdk/pull/182/commits/8f320b13ff4c911d5c7165d8c2259910ba05c495)

blockchain/blockchain.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/182/commits/8f320b13ff4c911d5c7165d8c2259910ba05c495)

Kourin1996 (approved):
Thank you for fixing. LGTM👍

zivkovicmilos (approved):
Approving for a second time because Github reset my approval for some reason 🙂

lazartravica:
@dbrajovic feel free to close this PR and delete the branch if you plan on opening a new one 👍

lazartravica:
Closing this PR in favour of https://github.com/0xPolygon/polygon-sdk/pull/208

cc @dbrajovic

PR #178, opened 2021-10-07
Add concurrency support for account specific txns in txpool

zivkovicmilos:
# Description

This PR adds basic concurrency support for account specific queues in the txpool.

# Changes include

- [x] Bugfix (non-breaking change that solves an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [ ] I have updated the README and other relevant documents (guides...)
- [x] I have added sufficient documentation both in code, as well as in the READMEs

PR #183, opened 2021-10-14
TxPool gas limit verification + txn error recover

### LINKED ISSUE #174: Tx gasLimit > 100000000 causes gas limit reached error and stuck tx
### Description

It seems that setting a tx gasLimit higher than 100000000 results in a pending tx that will never confirm. Any txs after this from the same account will also get stuck with no way of overwriting them.

https://github.com/0xPolygon/polygon-sdk/blob/335e78199b52c768e9b00c3f955e9a42a3ce5ed4/consensus/ibft/ibft.go#L356

Looking at the code above, a block header is initialized with its gasLimit set to 100000000. All tx gasLimits within that same block must not surpass this limit or else this issue occurs.

Aside from these transactions getting 'stuck', I observed many occurrences of the following log messages:
```
[ERROR] polygon: failed to apply tx: err=\"gas limit reached in the pool\""}

[31mERROR\u001b[0m\tbasichost\tfailed to resolve local interface addresses\t{\"error\": \"route ip+net: netlinkrib: too many open files\"}"}

[ERROR] polygon.network: failed to dial bootnode: err=\"peer rejected; insufficient capacity\""}
```

### Your environment

* OS and version `Ubuntu 20`
* branch that causes this issue `96be904`

### Steps to reproduce

1. Send a tx with gasLimit 5,000,000,000
2. Observe the `gas limit reached in the pool` log messages
3. Observe that subsequent txs from the same address cannot be processed
4. Attempting to override the nonce of the stuck tx (with a lower gasLimit but higher gasPrice) had no effect

### Expected behaviour

I expect there to be some sort of threshold that a tx's gasLimit must be under before it can be accepted. This threshold should be well under the hardcoded header gasLimit (aka Transition.gasPool) of 100000000.

### Actual behaviour

A tx with an extremely high gasLimit still gets picked up resulting in indefinite pending txs and error messages.

0xpanoramix:
# Description

This PR resolves #174 .

It adds two features :

1. A new verification whenever a transaction is submitted to the transaction pool operator.
2. Any detected error while writing transaction to a block is now marked as recoverable or not recoverable, indicating if the transaction has to be discarded or not. 

# Changes include

# Checklist

- [X] I have tested this code

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -389,7 +389,11 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 			break
 		}
 		if err := transition.Write(txn); err != nil {
-			retFn()
+			if err.IsRecoverable {
+				retFn()
+			} else {
+				i.txpool.DecreaseAccountNonce(txn)
+			}
 			break
 		}
 		txns = append(txns, txn)
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -388,15 +388,20 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 		if txn == nil {
 			break
 		}
-		if err := transition.Write(txn); err != nil {
-			if err.IsRecoverable {
-				retFn()
-			} else {
-				i.txpool.DecreaseAccountNonce(txn)
+
+		if txn.ExceedsBlockGasLimit(header.GasLimit) {
+			i.txpool.DecreaseAccountNonce(txn)
+		} else {
+			if err := transition.Write(txn); err != nil {
+				if err.IsRecoverable {
+					retFn()
+				} else {
+					i.txpool.DecreaseAccountNonce(txn)
+				}
+				break
 			}
-			break
+			txns = append(txns, txn)
 		}
-		txns = append(txns, txn)
 	}
 	i.logger.Info("picked out txns from pool", "num", len(txns), "remaining", i.txpool.Length())
 
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -279,6 +279,13 @@ func TestTxPool_TransactionCoalescing(t *testing.T) {
 }
 
 func TestInvalidTransactionRecover(t *testing.T) {
+	// Test scenario :
+	//
+	// 1. Send a transaction with gasLimit > block gas limit.
+	//		-> The transaction should not be applied, and the nonce should not be incremented.
+	//
+	// 2. Send a second transaction with a gasLimit < block gas limit.
+	//		-> The transaction should be applied, and the receiver balance increased.
 	senderKey, senderAddress := tests.GenerateKeyAndAddr(t)
 	_, receiverAddress := tests.GenerateKeyAndAddr(t)
 	testTable := []struct {
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"github.com/umbracle/go-web3"
 	"io"
 	"math/big"
 	"testing"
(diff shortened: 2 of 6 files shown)


state/executor.go
```diff
@@ -349,36 +362,62 @@ func (t *Transition) apply(msg *types.Transaction) (result *runtime.ExecutionRes
...
+			IsRecoverable: true,
+		}
+	}
+
+	// the amount of gas required is lower than the specified block gas limit
+	if msg.Gas > t.gasPool {
+		return nil, &TransactionApplicationError{
+			Err:           ErrBlockLimitExceeded,
+			IsRecoverable: false,
+		}
```
Kourin1996: Just confirm. Why this is not recoverable? This error happens if you can't add this tx to block because block doesn't have enough remaining gas to take this tx. So I thought you could add this tx in next block.

Did you want to check if tx gas doesn't exceed BlockGasLimit?

txpool/txpool.go
```diff
@@ -219,6 +219,16 @@ func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
...
 }
 
+// DecreaseAccountNonce resets the nonce attached to an account whenever a transaction produce an error which is not
+// recoverable, meaning the transaction will be discarded.
+//
+// Since any discarded transaction should not affect the world state, the nextNonce should be reset to the value
+// it was set to before the transaction appeared.
+func (t *TxPool) DecreaseAccountNonce(tx *types.Transaction) {
+	txnsQueue := t.accountQueues[tx.From]
+	txnsQueue.nextNonce -= 1
```
Kourin1996: you might want to check if accountQueues has record for tx.From just in case

Kourin1996 (changes requested):
Hi, Thank you for the PR. I've left small comments. I'd appreciate if you could check and fix if need. Thanks.

state/executor.go
0xpanoramix: > Did you want to check if tx gas doesn't exceed BlockGasLimit ?

Exactly, if the block has a gas limit of 5, and the gas provided in the message is 6, it will never be able to fit in a block.
This check is a special case of the third check (`the amount of gas required is available in the block`).

Kourin1996: Please note that t.gasPool doesn't equal to BlockGasLimit. If some transactions are processed already, t.gasPool has consumed the total of used gas already.

e2e/txpool_test.go
```diff
@@ -276,3 +277,185 @@ func TestTxPool_TransactionCoalescing(t *testing.T) {
 		"To address balance mismatch after gap transaction",
 	)
 }
+
+func TestInvalidTransactionRecover(t *testing.T) {
```
zivkovicmilos, line 281: Would love to see a description / test scenario above this test 🙏

state/executor.go
```diff
@@ -330,12 +333,22 @@ var (
...
 	ErrNotEnoughFundsForGas  = fmt.Errorf("not enough funds to cover gas costs")
 	ErrBlockLimitReached     = fmt.Errorf("gas limit reached in the pool")
+	ErrBlockLimitExceeded     = fmt.Errorf("specified gas exceed block gas limit")
 	ErrIntrinsicGasOverflow  = fmt.Errorf("overflow in intrinsic gas calculation")
 	ErrNotEnoughIntrinsicGas = fmt.Errorf("not enough gas supplied for intrinsic gas costs")
 	ErrNotEnoughFunds        = fmt.Errorf("not enough funds for transfer with given value")
 )
 
-func (t *Transition) apply(msg *types.Transaction) (result *runtime.ExecutionResult, err error) {
+type TransactionApplicationError struct {
```
zivkovicmilos: What do you think about renaming this to `TransitionApplicationError`, since it applies to state transitions instead of just regular transactions?

@Kourin1996 @lazartravica @Vuksan

state/executor.go
```diff
@@ -330,12 +333,22 @@ var (
 	ErrNonceIncorrect        = fmt.Errorf("incorrect nonce")
 	ErrNotEnoughFundsForGas  = fmt.Errorf("not enough funds to cover gas costs")
 	ErrBlockLimitReached     = fmt.Errorf("gas limit reached in the pool")
+	ErrBlockLimitExceeded     = fmt.Errorf("specified gas exceed block gas limit")
```
zivkovicmilos: Typo here, should be `exceeds`

txpool/txpool.go
```diff
@@ -219,6 +219,18 @@ func (t *TxPool) addImpl(ctx string, tx *types.Transaction) error {
 	return nil
 }
 
+// DecreaseAccountNonce resets the nonce attached to an account whenever a transaction produce an error which is not
+// recoverable, meaning the transaction will be discarded.
+//
+// Since any discarded transaction should not affect the world state, the nextNonce should be reset to the value
+// it was set to before the transaction appeared.
+func (t *TxPool) DecreaseAccountNonce(tx *types.Transaction) {
+	if t.accountQueues[tx.From] != nil {
```
zivkovicmilos, line 228: Please note that this should be thread safe after we merge PR #178

state/executor.go
```diff
@@ -184,25 +184,28 @@ func (t *Transition) Receipts() []*types.Receipt {
...
 	var err error
 	if txn.From == emptyFrom {
 		// Decrypt the from address
 		txn.From, err = signer.Sender(txn)
 		if err != nil {
-			return err
+			return &TransactionApplicationError{
+				Err:           err,
+				IsRecoverable: false,
+			}
```
zivkovicmilos: I see that we repeat this construction logic a lot of times when returning.

What do you think about separating it out into a small helper function like `generateTransitionErr`, or something along those lines?

zivkovicmilos (changes requested):
Looks alright so far 👍 
Excellent work on the tests 💯 

Left a few comments we should discuss

state/executor.go
Kourin1996: Agree with @zivkovicmilos's idea

state/executor.go
0xpanoramix: Thank you for reporting the error, I've moved this verification outside the apply function since I don't have access to the block gas limit inside of it.
It is now in the `writeNewBlock` function in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/ecd78d43b1ef64d0357f50ed924f89b926f5f6bc.

txpool/txpool.go
0xpanoramix: Good to know, thank you !

state/executor.go
```diff
@@ -330,12 +330,29 @@ var (
...
+	ErrBlockLimitExceeded     = fmt.Errorf("specified gas exceeds block gas limit")
 	ErrIntrinsicGasOverflow  = fmt.Errorf("overflow in intrinsic gas calculation")
 	ErrNotEnoughIntrinsicGas = fmt.Errorf("not enough gas supplied for intrinsic gas costs")
 	ErrNotEnoughFunds        = fmt.Errorf("not enough funds for transfer with given value")
 )
 
-func (t *Transition) apply(msg *types.Transaction) (result *runtime.ExecutionResult, err error) {
+type TransitionApplicationError struct {
+	Err				error
+	IsRecoverable	bool // Should the transaction be discarded, or put back in the queue.
```
Kourin1996: I think you can format this parr

consensus/ibft/ibft.go
```diff
@@ -388,11 +388,20 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 		if txn == nil {
 			break
 		}
-		if err := transition.Write(txn); err != nil {
-			retFn()
-			break
+
+		if txn.ExceedsBlockGasLimit(header.GasLimit) {
+			i.txpool.DecreaseAccountNonce(txn)
```
Kourin1996, line 393: Just confirm. You don't need to put tx back in this case?

e2e/txpool_test.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/041dac89638041f72990dd1db784ba4a74b498cf

state/executor.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/1c2736e2686c3503d22ac12cc23a9206a4c28066

state/executor.go
0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/081f3c0feb45ed866bc62681f88479ced7fdc1f0

state/executor.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/1c2736e2686c3503d22ac12cc23a9206a4c28066

zivkovicmilos (approved):
Thank you for resolving these mini changes 🙏 

Good to go 🚀

state/executor.go
0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-sdk/pull/183/commits/56679fcc33ff00437b754ddeeeabab08d1a94787

Kourin1996 (approved):
LGTM 👏

lazartravica (approved):
Well done @PtitLuca 

I love this pattern with recoverable/unrecoverable errors an I'm glad the tests are easy to read as well.

🤘

PR #184, opened 2021-10-14
Block gas target

zivkovicmilos:
# Description

This PR adds support for a target block gas limit.
The user sets the target block gas limit when starting the chain, and strives towards it during chain execution.

Most of the code has been adapted from @Kourin1996's [PR](https://github.com/0xPolygon/polygon-sdk/pull/125).

## Relevant docs PR
Docs have also been updated, [PR #16](https://github.com/0xPolygon/polygon-sdk-docs/pull/16).

# Changes include

# Checklist

- [x] I have tested this code


command/helper/config.go
```diff
@@ -99,6 +102,19 @@ func (c *Config) BuildConfig() (*server.Config, error) {
 		conf.Chain = cc
 	}
 
+	// Target gas limit
+	if c.BlockGasTarget != "" {
```
Kourin1996, line 133: `c.BlockGasTarget` is set when you run `server` command but I don't see command option for BlockGasTarget. For that reason, I can't the change of BlockGasTarget even if I set BlockGasTarget into genesis.

Kourin1996 (changes requested):
Hi, code looks ok. But I can't run node with custom BlockGasTarget because of loading config issue. Can you check it again? Thanks

command/helper/config.go
zivkovicmilos: You're right.

I added it in as a genesis command, instead of adding support for the `server` command.

Fixed now 👍 

e6a3db4

Kourin1996 (approved):
LGTM 👍

chain/chain.go
```diff
@@ -18,6 +18,9 @@ var (
 
 	// GenesisDifficulty is the default difficulty of the Genesis block.
 	GenesisDifficulty = big.NewInt(131072)
+
+	// DefaultBlockGasTarget is the default value for the block gas target for new blocks
+	DefaultBlockGasTarget uint64 = 8000000
```
lazartravica: Remove the DefaultBlockGasTarget.

If the server doesn't have the block gas target set, then just copy over the limit from the parent block when it is proposing a block.

blockchain/blockchain_test.go
```diff
@@ -561,3 +562,48 @@ func TestBlockchainWriteBody(t *testing.T) {
...
+			expectedGasLimit: 20000000/1024 + 20000000,
+		},
+		{
+			name:             "should decrease next gas limit towards target",
+			blockGasTarget:   25000000,
+			parentGasLimit:   26000000,
+			expectedGasLimit: 26000000 - 26000000/1024,
+		},
+		{
+			name:             "should not alter gas limit",
```
lazartravica: rename to "should not alter gas limit when exactly the same"

blockchain/blockchain_test.go
```diff
@@ -561,3 +562,48 @@ func TestBlockchainWriteBody(t *testing.T) {
...
+			blockGasTarget:   25000000,
+			parentGasLimit:   25000000,
+			expectedGasLimit: 25000000,
+		},
```
lazartravica, line 590: you are missing a test case where the diff between block gas limits is less than the delta

state/executor.go
```diff
@@ -196,6 +203,16 @@ func (t *Transition) Write(txn *types.Transaction) error {
...
+	if gas := t.AvailableGas(); gas < TxGas {
+		t.logger.Info(
+			"Block doesn't have enough gas to process new transaction",
+			"remaining",
+			gas, "minimum required gas",
+			TxGas,
+		)
+		return ErrBlockLimitReached
+	}
+
```
lazartravica: Why do we need this? Can't we rely on https://github.com/0xPolygon/polygon-sdk/blob/bfc7d94d077accabbbbb2d878574398453874e0a/state/executor.go#L379 ?

lazartravica (changes requested):
Please take a look at the comments.

P.S. And also resolve mere conflicts

chain/chain.go
lazartravica: I think that having default block gas target would cause 2 usability issues:

1. It would confuse users because they didn't know about it and suddenly they see the limit increasing to the default value from the initial gas limit in the genesis.
2. If I want to have a permissioned network with an unchanging block limit, then I need to set the initial block gas limit in the genesis, and set the additional flag in each server command that I run.

blockchain/blockchain_test.go
zivkovicmilos: Renamed 👍 

8b04228

blockchain/blockchain_test.go
zivkovicmilos: Added the 2 missing test cases, good catch 👍 

fe278e5

chain/chain.go
zivkovicmilos: This is completely correct, and both are valid points 👍 

I overlooked the usability aspect of this, as it would be frustrating to have the chain move in a random direction even though the genesis value has been set.

Fixed 👍 

a8b0343

state/executor.go
zivkovicmilos: You're right, the check you've highlighted is redundant because we have `subGasPool` when applying the txn, which literally does the same thing.

Fixed 👍 

48264b5

lazartravica (approved):
LGTM :metal: Take a look at the comments on the https://github.com/0xPolygon/polygon-sdk-docs/pull/16#pullrequestreview-799566003 and feel free to merge after resolving them.

PR #186, opened 2021-10-18
Fix/peer reconnection

munna0908:
# Description

This PR fixes the issue that the non bootnodes keep disconnected after loosing all connections. 
As part of this change, every node initiates a new routine which periodically(Every 30 sec)attempts to make new connection, if peercount is less than the minimumPeercount (currently set to 1 )

# Changes include

# Checklist

- [x] I have tested this code


network/server_test.go
```diff
@@ -212,6 +214,58 @@ func TestNat(t *testing.T) {
 	})
 }
 
+func WaitUntilPeerReconnects(ctx context.Context, srv *Server, id peer.ID) (bool, error) {
```
Kourin1996: I think `WaitUntilPeerConnectsTo` is better? because you don't check node disconnects from some peer in this function.

protocol/syncer.go
```diff
@@ -484,9 +490,11 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
...
 
 		b := p.popBlock()
-		if err := s.blockchain.WriteBlocks([]*types.Block{b}); err != nil {
-			s.logger.Error("failed to write block", "err", err)
-			break
+		if b != nil {
+			if err := s.blockchain.WriteBlocks([]*types.Block{b}); err != nil {
+				s.logger.Error("failed to write block", "err", err)
+				break
+			}
```
Kourin1996: `handler` may expect a block, so I think you shouldn't pass nil to handler. So you might want to break if popBlock returns nil

network/server.go
```diff
@@ -182,6 +185,27 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 	return srv, nil
 }
 
+// checkPeerCount will attempt to make new connections if the active peer count is lesser than the specified limit.
+func (s *Server) checkPeerConnections() {
+	for {
+
+		select {
+		case <-time.After(30 * time.Second):
```
Kourin1996, line 193: I'm sure polling approach will work. So it's just curious question => is it possible for us to use any Event Subscription? But PeerEventDisconnected is emitted before peers is deleted from store. So we may add new event?

I want some comments about this, @lazartravica

Kourin1996 (changes requested):
Hello, Thank you for the PR. I've left some comments. It'd be great if you could check and fix if need. Thanks!

protocol/syncer.go
Kourin1996: @munna0908 and me talked outside. I think we can pass nil to handler. But we do double check if peer is closed or not in WatchSyncWithPeer and popBlock. Can we skip p.IsClosed() in `WatchSyncWithPeer`? In this case, you modify handler little bit: if block is nill then handler return false

lazartravica: I agree with @Kourin1996's comments, we can skip the first `if p.IsClosed() {` in the `WatchSyncWithPeer`, and rely on the existing check in `popBlock` and add a new one to the `handler`.

@munna0908 could you also please refactor so that the callback is not called `handler` but rather `isValidator`, and avoid the double negation that we have (`return !isValidator` and `if !handler(b)`)?

network/server.go
lazartravica: @munna0908 the `checkPeerConnections` can block on receiving the `PeerEventDisconnected` event and only after that check if the `s.numPeers() < MinimumPeerConnections` and try to reconnect to the bootnode.

Feel free to create a new subscription. 

Also, if you have any concerns about this (you mentioned offline somethng among the lines of cascading events) => could you care to explain how that might happen and what would be the result?

TLDR: Polling every 30 seconds is something that we should try to avoid since we have a straightforward way of subscribng to events.

lazartravica (changes requested):
Please see my comments.

protocol/syncer.go
munna0908: @lazartravica current implementation of pop block, blocks until we receive a block, but it's not safe to indefinitely wait for a block from remote node, because the remote node may be malicious. 
So in this [PR](https://github.com/0xPolygon/polygon-sdk/pull/181/commits/fbc80d4703080fd00a189c2d30c9ccaf0a5f8701) @Kourin1996  added a timeout, and he is returning a nil value, so i felt having an explict check i.e p.isClose() is required. 
Looking forward to your inputs.

munna0908: @lazartravica sure will refactor handler callback func

network/server.go
munna0908: Fixed it Commit-[Link](https://github.com/0xPolygon/polygon-sdk/pull/186/commits/efd124d6f9bff3282e598c556dbcb01e47429b62)

network/server_test.go
munna0908: @Kourin1996 Fixed it [link](https://github.com/0xPolygon/polygon-sdk/pull/186/commits/7d56b77aff6c6622aa530a9a7c47cbf686582e65)

network/server.go
```diff
@@ -182,6 +185,32 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+	}
+	for {
+
+		event, ok := <-updateCh
+		if !ok {
+			return
+		} else if event.Type == PeerEventDisconnected {
+			if s.numPeers() < MinimumPeerConnections {
+				if s.config.NoDiscover || len(s.discovery.bootnodes) == 0 {
+					//TODO: dial peers from the peerstore
```
zivkovicmilos: What's up with this TODO?

network/server.go
```diff
@@ -182,6 +185,32 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+				} else {
+					s.dialQueue.add(s.discovery.bootnodes[rand.Intn(len(s.discovery.bootnodes))], 10)
```
zivkovicmilos: Can we please separate this out into multiple lines, it's messy to read? 🙈

network/server_test.go
```diff
@@ -212,6 +214,58 @@ func TestNat(t *testing.T) {
...
+		}
+		return true, false
+
+	})
+	if err != nil {
+		return false, err
+	}
+	return res.(bool), nil
+}
+func TestPeerReconnection(t *testing.T) {
```
zivkovicmilos, line 336: Please provide some kind of description / test scenario for this test 🙏

zivkovicmilos (changes requested):
Looks good 💯 

Left minor comments that should be resolved

network/server.go
munna0908: @lazartravica  suggested to create a new issue in youtrack which can be handled later.

zivkovicmilos (approved):
Thank you for the fixes 🙏 

Looks good 🚀

lazartravica:
We have discussed offline to bring back the polling approach without the events, to avoid code complexity.

protocol/syncer.go
lazartravica: Thanks Rahul, I agree with you. If @Kourin1996 is on the same line, he can resolve the conversation.

munna0908:
> We have discussed offline to bring back the polling approach without the events, to avoid code complexity.

@lazartravica reverted back to polling approach here is the [link](https://github.com/0xPolygon/polygon-sdk/pull/186/commits/50554df7736c9ff8356abd3e3c3f897bd890dce2)

e2e/syncer_test.go
```diff
@@ -19,7 +19,7 @@ func TestSyncer(t *testing.T) {
 	// Start IBFT cluster (4 Validator + 2 Non-Validator)
 	ibftManager := framework.NewIBFTServersManager(t, IBFTMinNodes+numNonValidators, IBFTDirPrefix, func(i int, config *framework.TestServerConfig) {
 		config.SetSeal(i < IBFTMinNodes)
-		config.SetShowsLog(i == IBFTMinNodes+numNonValidators-1)
+		//config.SetShowsLog(i == IBFTMinNodes+numNonValidators-1)
```
lazartravica: Please remove this comment, we already have a PR which removed it.

lazartravica (approved):
Just take a look at my minor nitpick and that's it, feel free to merge

Kourin1996 (approved):
LGTM 👍

munna0908:
@lazartravica 

> Just take a look at my minor nitpick and that's it, feel free to merge

Done

PR #189, opened 2021-11-01
Remove type casting for the transition error

zivkovicmilos:
# Description

This PR introduces a hot-fix for error casting when applying transactions.
Even though no error would occur, because of type casting where a non-empty object is created, a condition check if an error has occurred would fail, causing unexpected behavior.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -393,7 +393,7 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 			i.txpool.DecreaseAccountNonce(txn)
 		} else {
 			if err := transition.Write(txn); err != nil {
-				if err.IsRecoverable {
+				if err.(*state.TransitionApplicationError).IsRecoverable {
 					retFn()
 				} else {
 					i.txpool.DecreaseAccountNonce(txn)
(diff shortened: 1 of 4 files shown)


zivkovicmilos:
@lazartravica @Kourin1996 
The tests keep failing on Github actions for some reason, but I've ran them at least 5+ times locally and they passed every time:
[Test Results — Run all tests.pdf](https://github.com/0xPolygon/polygon-sdk/files/7454796/Test.Results.Run.all.tests.pdf)

lazartravica (approved):
LGTM

consensus/ibft/ibft.go
```diff
@@ -393,7 +393,7 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 			i.txpool.DecreaseAccountNonce(txn)
 		} else {
 			if err := transition.Write(txn); err != nil {
-				if err.IsRecoverable {
+				if err.(*state.TransitionApplicationError).IsRecoverable {
```
Kourin1996: Just a comment. It looks to be ok now. But you might want to add safe-check in order to avoid panic

```go
if appErr, ok := err.(*state.TransitionApplicationError); ok && appErr.IsRecoverable {
```

Kourin1996 (approved):
LGTM

consensus/ibft/ibft.go
zivkovicmilos: Good call 👍 

Even though we know what kind of error the call will return, it's always better to add that additional type assertion check layer 💯 

Added:
f9bfe91

PR #191, opened 2021-11-03
Feature/dns support

munna0908:
# Description

This PR adds the DNS support in following cases

1. Bootnode multiaddr supports dns/dsn4/dns6
2. Nodes can listen and advertise an external DNS address, if specified using a `--dns` flag

# Changes include

# Checklist

- [x] I have tested this code


lazartravica (approved):
LGTM

command/helper/config.go
```diff
@@ -221,3 +238,22 @@ func readConfigFile(path string) (*Config, error) {
 
 	return &config, nil
 }
+
+func parseDNSString(s string) (version string, name string) {
+	s = strings.Trim(s, "/")
+	fmt.Println(s)
```
zivkovicmilos: Please remove the leftover print

command/helper/config.go
```diff
@@ -221,3 +238,22 @@ func readConfigFile(path string) (*Config, error) {
...
+	case "dns":
+		version = "dns"
+	case "dns4":
+		version = "dns4"
+	case "dns6":
+		version = "dns6"
+	}
+	name = split[1]
+	return
+}
```
zivkovicmilos: I would appreciate it if this func had error handling and tests written for it.
In my experience, we should never expect a string to be the correct format, as it's not always the case.

zivkovicmilos (changes requested):
Left a few minor comments.

Otherwise it looks good 💯

Kourin1996 (approved):
LGTM 👍

command/helper/config.go
munna0908: @zivkovicmilos Here is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/191/commits/bb373293bbfc4c97524dbc250daa0dae653e3d10)

command/helper/config.go
munna0908: cleaned up

lazartravica: Resolving this conversation. @munna0908 don't forget to include commit msg

cc @zivkovicmilos

command/helper/config.go
lazartravica: @munna0908 we're missing the tests that @zivkovicmilos  mentioned. We don't have any test cases for bogus DNS names.

zivkovicmilos: @munna0908 
I think it would be wise if we added a regex check to the func as well, this way we can keep the format air-tight

munna0908: @zivkovicmilos @lazartravica  Added regex to match the dns string commit [link](https://github.com/0xPolygon/polygon-sdk/pull/191/commits/fd76aaf24fdfdb8156ee2f600f92f6826f339e07)

munna0908: Added test cases as well commit [link](https://github.com/0xPolygon/polygon-sdk/pull/191/commits/f4307b4977f08088f65e675dc29a5c27938b0931)

zivkovicmilos (approved):
Thank you for making the DNS string parse air-tight 🙏 

Looks good 🚀

PR #197, opened 2021-11-05
Add loadbot to run stress tests

0xpanoramix:
# Description

This PR includes a loadbot able to run a configurable number of transactions, with custom parameters such as :

- Txn gas
- Txn gas price
- Txn value
- Accounts used by the loadbot
- Number of txn to send between provided accounts
- Number of txn to send per second

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments

Feel free to check the documentation on the loadbot usage here.

### DIFF
--- a/command/loadbot/execution.go
+++ b/command/loadbot/execution.go
@@ -15,16 +15,18 @@ import (
 	"time"
 )
 
+// Configuration represents the loadbot run configuration.
+// It contains the required parameters to run the stress test.
 type Configuration struct {
-	TPS       uint64
-	Value     *big.Int
-	Gas       uint64
-	GasPrice  *big.Int
-	Accounts  []types.Address
-	RPCURLs   []string
+	TPS       uint64          // Number of transactions per second
+	Value     *big.Int        // The value sent in each transaction
+	Gas       uint64          // The transaction's gas
+	GasPrice  *big.Int        // The transaction's gas price
+	Accounts  []types.Address // All the accounts used by the bot to send transaction
+	RPCURLs   []string        // The JSON RPC endpoints, used to send transactions
 	ChainID   uint64
-	TxnToSend uint64 // The number of transactions to send
-	GRPCUrl   string
+	TxnToSend uint64 // The number of transaction to send
+	GRPCUrl   string // The gRPC url, used while verifying transactions to check if the TxPool is empty
 }
 
 type Metrics struct {
@@ -32,9 +34,11 @@ type Metrics struct {
 	Total     uint64        // The total number of transactions processed
 	Failed    uint64        // The number of failed transactions
 	Duration  time.Duration // The execution time of the loadbot
-	TxnHashes []web3.Hash
+	TxnHashes []web3.Hash   // The hashes of the transactions sent
 }
 
+// createClients will create JSON RPC clients using the provided addresses in the CLI.
+// Each one of these clients will send transaction(s), each one after another.
 func (c *Configuration) createClients() ([]*jsonrpc.Client, error) {
 	var clients []*jsonrpc.Client
 
@@ -48,19 +52,27 @@ func (c *Configuration) createClients() ([]*jsonrpc.Client, error) {
 	return clients, nil
 }
 
+// We create the transactions using createTransactionObjects before the loadbot send them.
 func (c *Configuration) createTransactionObjects() ([]*web3.Transaction, error) {
 	var transactions []*web3.Transaction
 	var nonces = make(map[types.Address]uint64)
 	var numberOfAccounts = uint64(len(c.Accounts))
 
+	// Initialize accounts nonces at 0.
 	for _, account := range c.Accounts {
 		nonces[account] = 0
 	}
 
+	// Each loop create one transaction.
 	for i := uint64(0); i < c.TxnToSend; i++ {
+		// Get sender and receiver.
 		from := c.Accounts[i%numberOfAccounts]
 		to := c.Accounts[(i+1)%numberOfAccounts]
+
+		// Get sender nonce.
 		nonce := nonces[from]
+
+		// Create the transaction object.
 		txn := &web3.Transaction{
 			From:     web3.Address(from),
 			To:       (*web3.Address)(&to),
@@ -78,6 +90,8 @@ func (c *Configuration) createTransactionObjects() ([]*web3.Transaction, error) 
 	return transactions, nil
 }
 
+// run is the main method of the loadbot.
+// The TPS is used to determine the rate at which every transaction is sent.
 func (c *Configuration) run(clients []*jsonrpc.Client, txns []*web3.Transaction) *Metrics {
 	ticker := time.NewTicker(1 * time.Second / time.Duration(c.TPS))
 	defer ticker.Stop()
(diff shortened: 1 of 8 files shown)


lazartravica (changes requested):
@PtitLuca please solve the linter errors, mark the PR as ready for review if it shouldn't be a draft and merge the develop again onto this branch to solve the failing tests due to timeout.

Otherwise, it looks good, I'd only ask you to add some usage examples to the documentation.

command/loadbot/execution.go
```diff
@@ -0,0 +1,224 @@
...
+	TxnHashes []web3.Hash   // The hashes of the transactions sent
+}
+
+// createClients will create JSON RPC clients using the provided addresses in the CLI.
+// Each one of these clients will send transaction(s), each one after another.
+func (c *Configuration) createClients() ([]*jsonrpc.Client, error) {
+	var clients []*jsonrpc.Client
+
+	for _, url := range c.RPCURLs {
+		conn, err := jsonrpc.NewClient(fmt.Sprintf("http://%s", url))
```
Kourin1996: There is a possibility that URL with protocol would be given. And impossible to specify https. You might want to add `http://` only if url doesn't have protocol.

command/loadbot/execution.go
```diff
@@ -0,0 +1,224 @@
...
+// We create the transactions using createTransactionObjects before the loadbot send them.
+func (c *Configuration) createTransactionObjects() ([]*web3.Transaction, error) {
+	var transactions []*web3.Transaction
+	var nonces = make(map[types.Address]uint64)
+	var numberOfAccounts = uint64(len(c.Accounts))
+
+	// Initialize accounts nonces at 0.
+	for _, account := range c.Accounts {
+		nonces[account] = 0
+	}
```
Kourin1996: FYR, you don't have to initialize map every time. map returns zero-value for non-existing key.

```go
m := make(map[string]int)

// map access for existing key
m["foo"] = 10
fmt.Println(m["foo"]) // => 10
x, ok := m["foo"]
fmt.Println(x, ok) // => 10, true

// map access for non-existing key
fmt.Println(m["bar"]) // => 0
x, ok = m["bar"]
fmt.Println(x, ok) // => 0, false
```

command/loadbot/loadbot_command.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if err := flags.Parse(args); err != nil {
+		l.UI.Error(fmt.Sprintf("failed to parse args: %v", err))
+		return 1
+	}
+
+	// Parse accountsRaw
+	var addresses = []types.Address{}
+	if accountsRaw == nil {
+		l.UI.Error("failed to parse accountsRaw used by the loadbot")
```
Kourin1996: account is more proper name to show to user (accountsRaw is just variable name)

command/loadbot/loadbot_command.go
```diff
@@ -0,0 +1,213 @@
...
+		if err := placeholder.UnmarshalText([]byte(account)); err != nil {
+			l.UI.Error(fmt.Sprintf("Failed to decode account address: %v", err))
+			return 1
+		}
+		addresses = append(addresses, placeholder)
+	}
+
+	// Parse urls
+	if len(urls) == 0 {
+		l.UI.Error("please provide at least one node url to run the loabot")
```
Kourin1996: Probably typo (loabot -> loadbot)

Kourin1996 (changes requested):
Good stuff 👍 I've left some small comments. It'd be great if you could check and fix if need. Thank you!

command/loadbot/loadbot_command.go
0xpanoramix: I think its a typo due to refactor, thanks for mentioning it !

command/loadbot/loadbot_command.go
lazartravica: Sharp eye! :)

command/loadbot/loadbot_command.go
```diff
@@ -0,0 +1,213 @@
...
+	l.FlagMap["url"] = helper.FlagDescriptor{
+		Description: "The URL used to make the transaction submission",
+		Arguments: []string{
+			"URL",
+		},
+		ArgumentsOptional: false,
+		FlagOptional:      false,
+	}
+
+	l.FlagMap["chain-id"] = helper.FlagDescriptor{
```
zivkovicmilos: Consider renaming this flag to chainid, since we have it without the dash in the rest of the codebase

command/loadbot/loadbot_command.go
```diff
@@ -0,0 +1,213 @@
...
+	})
+	if err != nil {
+		l.UI.Error(fmt.Sprintf("failed to execute loadbot: %v", err))
+		return 1
+	}
+
+	l.UI.Info("Loadbot execution finished. Got following metrics :")
+	l.UI.Info(fmt.Sprintf("Transactions submitted: %v", metrics.Total))
+	l.UI.Info(fmt.Sprintf("Transactions failed: %v", metrics.Failed))
+	l.UI.Info(fmt.Sprintf("Duration: %v", metrics.Duration))
```
zivkovicmilos: Consider packing this into a KV output, like we do for example in the ibft-init command

command/loadbot/execution.go
```diff
@@ -0,0 +1,219 @@
...
+		conn, err := jsonrpc.NewClient(url)
+		if err != nil {
+			return nil, fmt.Errorf("failed to connect to server: %v", err)
+		}
+		clients = append(clients, conn)
+	}
+	return clients, nil
+}
+
+// We create the transactions using createTransactionObjects before the loadbot send them.
```
zivkovicmilos: Just a nitpick, please avoid adding dots to end of code comments, it's not consistent with the rest of the code style we already have

command/loadbot/execution.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		fmt.Printf("TxPool not empty, %d transactions remaining..\n", res.Length)
+		return nil, true
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return res.(*txpoolOp.TxnPoolStatusResp), nil
+}
```
zivkovicmilos: Don't we already have this helper method somewhere in the testing framework? If not on the develop branch, then on some other branch definitely. Please consider moving it to a more appropriate place -> helper/tests/testing.go

zivkovicmilos (approved):
Great PR, everything is pretty detailed 💯 

Please check my comments, as far as I'm concerned it's solid 👍

Kourin1996 (approved):

brkomir (approved):

command/loadbot/loadbot_command.go
0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-sdk/pull/197/commits/45b67ce0874d8dbd2120ee3c409098b5a1f99c15

command/loadbot/loadbot_command.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-sdk/pull/197/commits/45b67ce0874d8dbd2120ee3c409098b5a1f99c15

command/loadbot/execution.go
0xpanoramix: Removed in https://github.com/0xPolygon/polygon-sdk/pull/197/commits/45b67ce0874d8dbd2120ee3c409098b5a1f99c15

command/loadbot/execution.go
0xpanoramix: Thanks, I have used the E2E method here. Since we use it outside of the E2E scope, I agree with your idea of moving it to a more appropriate place :+1: 

Added in https://github.com/0xPolygon/polygon-sdk/pull/197/commits/c4b96d1419a2d4800e1f3d952d141b9e0138d1e6

lazartravica (changes requested):
Sorry but I don't see any docs on the [polygon-sdk-docs ](https://github.com/0xPolygon/polygon-sdk-docs) repo as we discussed. Feel free to link the PR here once you have it ready.

lazartravica (approved):

lazartravica (approved):

PR #205, opened 2021-11-08
Fix/peer reconnection

munna0908:
# Description

This PR fixes issue in PR #186 .

# Changes include

# Checklist

- [x] I have tested this code


protocol/syncer_test.go
```diff
@@ -284,7 +283,7 @@ func TestWatchSyncWithPeer(t *testing.T) {
 			go func() {
 				syncer.WatchSyncWithPeer(peer, func(b *types.Block) bool {
 					// sync until latest block
-					return b.Header.Number < latestBlock.Header.Number
+					return !(b.Header.Number < latestBlock.Header.Number)
```
lazartravica: @munna0908 please flip the `<`, no reason to `!` the entire expression, very confusing that way

lazartravica (changes requested):

protocol/syncer_test.go
munna0908: @lazartravica fixed it [link](https://github.com/0xPolygon/polygon-sdk/pull/205/commits/07d18eb1a029eb68952a217913493ce537ee954c)

lazartravica (approved):

PR #207, opened 2021-11-09
Test/build with race flag

munna0908:
# Description

Outcome of this PR is to create two different binaries. 

- **polygon-sdk-race**: This binary is generated by setting the race flag, it is used to run all e2e tests.
- **polygon-sdk**: This binary is generated without setting the race flag, it is used for further deployments.  

Changes made to the Yaml file 

1. Added new step to build a binary `polygon-sdk-race` with race flag set. I have enabled CGO as specified by the github documentation to generate this binary.

Changes made to source code 

1. Modified e2e test server code to utilise `polygon-sdk-race` binary instead of `polygon-sdk` to run all e2e tests. 

 
# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (approved):
LGTM 👌

lazartravica (changes requested):
Hey @munna0908, I have a couple of questions and recommendations:

Why did we need to enable CGO in order to enable race monitoring in the artifact?

The `-race` flag will make the polygon-sdk exit with a non-zero exit code when a race is detected, are our integration tests properly detecting if the process exits in that way (`TestServer`)?

Why didn't you include the `-race` flag in running the test themselves, but only in building the binary?

I think it would be good to build 2 versions of the polygon-sdk binary, one with and one without the race detector. The reason is that we don't want to release the binary tomorrow consuming 10x resources due to the race detector. The best approach I can think of is to build an artifact called `polygon-sdk-race` (alongside the regularly built `polygon-sdk`) which will be called from the tests. We also don't want to have the binary with the race detector on our testnet at the moment.

munna0908:
> Hey @munna0908, I have a couple of questions and recommendations:
> 
> Why did we need to enable CGO in order to enable race monitoring in the artifact?
> 
> The `-race` flag will make the polygon-sdk exit with a non-zero exit code when a race is detected, are our integration tests properly detecting if the process exits in that way (`TestServer`)?
> 
> Why didn't you include the `-race` flag in running the test themselves, but only in building the binary?
> 
> I think it would be good to build 2 versions of the polygon-sdk binary, one with and one without the race detector. The reason is that we don't want to release the binary tomorrow consuming 10x resources due to the race detector. The best approach I can think of is to build an artifact called `polygon-sdk-race` (alongside the regularly built `polygon-sdk`) which will be called from the tests. We also don't want to have the binary with the race detector on our testnet at the moment.

@lazartravica Initially, I tried without enabling CGO, but github actions prompted an error, that CGO should be enabled.

If we include `race` flag in build, we can find more races by running the binary under a realistic workload, Since our tests have incomplete coverage.

Yeah,That makes sense to have a different binary, which can be used for testing.

lazartravica:
> If we include `race` flag in build, we can find more races by running the binary under a realistic workload, Since our tests have incomplete coverage.

Got it, but please include it in the `-race` in the tests so that we may also detect any race conditions in our unit tests as well.

brkomir (approved):

munna0908:
> > If we include `race` flag in build, we can find more races by running the binary under a realistic workload, Since our tests have incomplete coverage.
> 
> Got it, but please include it in the `-race` in the tests so that we may also detect any race conditions in our unit tests as well.

@lazartravica added `race` flag in the tests and discovered that tests seems to fail due to race condition

.github/workflows/main.yml
```diff
@@ -39,13 +39,17 @@ jobs:
           submodules: recursive
 
       - name: Go build
-        run: CGO_ENABLED=0 GOOS=linux go build -a -o artifacts/polygon-sdk .
+        run: CGO_ENABLED=1 GOOS=linux go build -race -a -o artifacts/polygon-sdk .
```
zivkovicmilos: I thought the point of having a separate build routine `Go build with race` is that the regular `Go build` would be _without_ the _race_ flag?

.github/workflows/main.yml
```diff
@@ -39,13 +39,17 @@ jobs:
...
+      - name: Go build with race 
+        run: CGO_ENABLED=1 GOOS=linux go build -race -a -o artifacts/polygon-sdk-race .
+      
       - name: Add artifacts directory to the path
         run: echo "$(pwd)/artifacts" >> $GITHUB_PATH
 
       - name: Go test
-        run: go test -timeout 20m ./...
+        run: go test -timeout 28m ./...
+        run: ls
```
zivkovicmilos: What happens when you put two runs like this? Why `1s`?

zivkovicmilos (changes requested):
I still have some questions that need to be resolved and I think there are minor mistakes.

We could've avoided all the confusion if we had at least some kind of documentation for how the Github builds work and what are the params, so I'm requesting that you update the PR description with the changes made to the YAML file, what they mean and why we made them.

.github/workflows/main.yml
munna0908: Fixed it, Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/207/commits/4ae18f2be64714f933d5e5b421737c831b2e9e6a)

.github/workflows/main.yml
munna0908: It was mistake, Removed it. Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/207/commits/4ae18f2be64714f933d5e5b421737c831b2e9e6a)

e2e/framework/testserver.go
```diff
@@ -38,7 +39,7 @@ type TestServerConfigCallback func(*TestServerConfig)
 
 const (
 	initialPort   = 12000
-	polygonSDKCmd = "polygon-sdk"
+	polygonSDKCmd = "polygon-sdk-race"
```
zivkovicmilos: This is a problem.

We can't just overwrite the command we use for the testing server and consider that every local build is going to have a race safe binary.

This is in relation to the step where we build out the binary the test server will use:
https://github.com/0xPolygon/polygon-sdk/blob/test/build_with_race_flag/e2e/README.md#step-1-build-the-polygon-sdk

I'd like to discuss this @munna0908 @lazartravica @Kourin1996

zivkovicmilos:
I don't think this achieves what we wanted the PR to achieve.

If I understood correctly, the point of the PR was to add a separate GH action that would run the tests _again_ to check for any races.

The way we set up the yaml configuration now, we're just building two binaries, but we never added the race action (`Go test race`, like we have `Go test`).

Additionally, we just assume that every local build has a binary `polygon-sdk-race`, which is definitely not the case, nor is it in any of the documentation that we have (concretely the README file for e2e tests).

Why do we need the two binaries? Why couldn't we just slap on a new GH action `Go test race` with the code `go test -race -timeout 28m ./...`?

I'd like us to discuss this, I'm probably missing something.

EDIT:

After communicating with the team, I've concluded that this PR does in fact achieve the thing we set out - run the race-secure binary for our e2e tests.

@lazartravica @munna0908 @Kourin1996

zivkovicmilos (approved):
Looks good to me now 💯 

Thank you for resolving all of the comments and concerns 🙏

lazartravica (changes requested):
@munna0908 Since we don't need the race artifact for anything else apart from the tests, why don't we simply do the following (without changing the `polygonSDKCmd`):

1. Build the `polygon-sdk` binary with race enabled
2. Run the tests on this binary
3. If all goes well, build the binary without race enabled, effectively overwriting the `polygon-sdk`.

munna0908:
> @munna0908 Since we don't need the race artifact for anything else apart from the tests, why don't we simply do the following (without changing the `polygonSDKCmd`):
> 
> 1. Build the `polygon-sdk` binary with race enabled
> 2. Run the tests on this binary
> 3. If all goes well, build the binary without race enabled, effectively overwriting the `polygon-sdk`.

@lazartravica Agree, Here's is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/207/commits/5ac3c95d28020de54a460934ffcc477c875ecb69). Thanks for your input.

lazartravica (approved):
LGTM

PR #213, opened 2021-11-15
Implement slotGauge for managing TxPool capacity

dbrajovic:
# Description

Previously, despite having atomic reads/writes, the `slots` field of `TxPool` was concurrently unsafe in the case of pool resizing, should the incoming transaction overflow it. `slotGauge` abstracts away this capacity concept and handles overflow in a concurrently safe manner.

# Changes include

# Checklist

- [x] I have tested this code


txpool/txpool.go
```diff
@@ -64,6 +64,44 @@ type signer interface {
...
+}
+
+func (g *slotGauge) height() uint64 {
+	g.lock()
+	defer g.unlock()
+
+	return g.current
+}
+
+// must come before unlock()
```
brkomir: These methods are unnecessary since you embed sync.Mutex.

txpool/txpool.go
```diff
@@ -555,38 +576,76 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
 // increaseSlots decreases number of taken slots
 func (t *TxPool) decreaseSlots(slots uint64) {
-	atomic.AddUint64(&t.slots, ^(slots - 1))
+	t.gauge.decrease(slots)
+	// atomic.AddUint64(&t.slots, ^(slots - 1))
+}
+
+// Checks if the incoming tx would cause an overflow
+// and attempts to allocate space for it
+func (t *TxPool) gaugeCheck(tx *types.Transaction, isLocal bool) error {
```
brkomir: This function does more than its name implies, consider more appropriate name.

txpool/txpool.go
```diff
@@ -555,38 +576,76 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
+	t.gauge.lock()
+	defer t.gauge.unlock()
+
+	if t.gauge.current+slots(tx) <= t.gauge.limit {
+		// no overflow
+		return nil
+	}
+
+	if !isLocal && t.Underpriced(tx) {
```
brkomir, line 622: This doesn't seem related to the rest of the method code, this should be done in other place.

txpool/txpool.go
```diff
@@ -956,7 +1015,7 @@ func (a *localAccounts) addAddr(addr types.Address) {
 	a.accounts[addr] = true
 }
 
-// numSlots calculates the number of slots for given transaction
-func numSlots(tx *types.Transaction) uint64 {
+// slots() calculates the number of slots for given transaction
+func slots(tx *types.Transaction) uint64 {
```
brkomir: Maybe rename to slotsRequired to improve readability.

txpool/txpool.go
```diff
@@ -473,12 +489,17 @@ func (t *TxPool) ProcessEvent(evnt *blockchain.Event) {
...
 	// remove the mined transactions from the pendingQueue list
 	for _, txn := range delTxns {
-		t.decreaseSlots(numSlots(txn))
+		t.decreaseSlots(slots(txn))
 		t.pendingQueue.Delete(txn)
 		t.remoteTxns.Delete(txn)
 	}
 }
 
+// Returns the number of slots currently occupying the pool
```
brkomir: You don't need this method. Use gauge.height directly or rename it.

txpool/txpool_test.go
```diff
@@ -855,7 +856,139 @@ func TestSizeLimit(t *testing.T) {
...
 			assert.Equal(t, tt.err, err)
 			assert.Equal(t, tt.len, pool.Length())
-			assert.Equal(t, tt.slots, pool.slots)
+			assert.Equal(t, tt.slots, pool.slotsOccupied())
+		})
+	}
+}
+
+func TestGaugeCheck(t *testing.T) {
+	type AddTx struct {
```
brkomir: There is a similar structure in the test above. We should strive to code reusability. Consider taking common things out of the test methods

txpool/txpool_test.go
```diff
@@ -855,7 +856,139 @@ func TestSizeLimit(t *testing.T) {
...
+		nonce    uint64
+		gasPrice *big.Int
+		slotSize uint64
+		origin   TxOrigin
+	}
+
+	genTx := func(t *testing.T, arg AddTx) *types.Transaction {
```
brkomir: If this is a helper function mark it as such. I don't feel that we should define helper functions inside the tests themselves though.

txpool/txpool_test.go
```diff
@@ -855,7 +856,139 @@ func TestSizeLimit(t *testing.T) {
...
+					defer wg.Done()
+					tx := genTx(t, incoming)
+					assert.NoError(t, pool.addImpl(incoming.origin, tx))
+				}(incomingTx)
+			}
+			wg.Wait()
+
+			// In whichever order the incoming txs came in
+			// they should not break the gauge limit invariant
+			assert.Equal(t, tt.maxSlots, pool.slotsOccupied())
```
brkomir: I am not sure what we are trying to accomplish here, but if you run the tests several times you will get a failing test...

brkomir (changes requested):

txpool/txpool.go
```diff
@@ -64,6 +64,44 @@ type signer interface {
...
+
+	g.current -= slots
+}
+
+func (g *slotGauge) height() uint64 {
+	g.lock()
+	defer g.unlock()
+
+	return g.current
+}
```
Kourin1996: Just curious, do you have any special reasons to rename `current` to `height`? It seems to be getter method.

txpool/txpool.go
```diff
@@ -555,38 +576,76 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
-	atomic.AddUint64(&t.slots, slots)
+	t.gauge.increase(slots)
+	// atomic.AddUint64(&t.slots, slots)
 }
 
 // increaseSlots decreases number of taken slots
 func (t *TxPool) decreaseSlots(slots uint64) {
-	atomic.AddUint64(&t.slots, ^(slots - 1))
+	t.gauge.decrease(slots)
+	// atomic.AddUint64(&t.slots, ^(slots - 1))
```
Kourin1996: Why are you keepping old code?

txpool/txpool.go
```diff
@@ -555,38 +576,76 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
 	}
+
 	return dropped, true
 }
 
 // increaseSlots increases number of taken slots
 func (t *TxPool) increaseSlots(slots uint64) {
-	atomic.AddUint64(&t.slots, slots)
+	t.gauge.increase(slots)
+	// atomic.AddUint64(&t.slots, slots)
```
Kourin1996: Why are you keepping old code?

Kourin1996 (changes requested):
Hi, I've left small comments. It'd be great if you could check and fix if need. Thanks

txpool/txpool.go
dbrajovic: Agreed.

[Fixed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/268b2c90d24759c95da79201eda539751a253842)

txpool/txpool.go
dbrajovic: Perhaps @Kourin1996 can provide more detail on the original idea behind that specific logic.

As it is [right now](https://github.com/0xPolygon/polygon-sdk/blob/6efebc6a92f885082620ed2a8d59e37d4d27074c/txpool/txpool.go#L310), that part can be translated to: "In case of an overflow, no room will be made for remote transactions with lower gas fee than the remote transaction with the lowest fee currently present in the pool".  Resulting in a dropped transaction.

It's a decision-making move tied to the event of an overflow. I'm not entirely sure if it should be put elsewhere, but it definitely calls for a discussion.

txpool/txpool.go
dbrajovic: An alternative I had in mind is `slotWeight()`.

txpool/txpool_test.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/ce80cf9d857d332b6c1258016cbccee67526cfbd)

txpool/txpool_test.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/ce80cf9d857d332b6c1258016cbccee67526cfbd)

txpool/txpool.go
dbrajovic: Nothing in particular, maybe this approach is more readable:

````
type slots uint64

type slotGauge struct {
	sync.Mutex
	height slots
	max    slots
}
````

and then rename `height()` to `currentHeight()` ?

File: txpool/txpool.go (hunk 10 above)
Comment (dbrajovic) [reply txpool/txpool.go]: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/12090081c1dc6220d415c77562363fcb3c804715)

File: txpool/txpool.go (hunk 11 above)
Comment (dbrajovic) [reply txpool/txpool.go]: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/12090081c1dc6220d415c77562363fcb3c804715)

File: txpool/txpool.go (hunk 2 above)
Comment (dbrajovic) [reply txpool/txpool.go]: What about `handlePossibleOverflow()` ?

File: txpool/txpool.go (hunk 9 above)
Comment (Kourin1996) [reply txpool/txpool.go]: I don't think it's good to rename or add new meaning unless you modify functionality
But actually I'd like to have someone's opinion @zivkovicmilos @lazartravica

File: txpool/txpool.go (hunk 5 above)
Comment (dbrajovic) [reply txpool/txpool.go]: I guess it's more a matter of what we'd like to export from a package (since the same could be said for [(*Txpool) Length()](https://github.com/0xPolygon/polygon-sdk/blob/6efebc6a92f885082620ed2a8d59e37d4d27074c/txpool/txpool.go#L405).

Maybe I'm going about this wrong, but when writing unit tests I feel like we should be testing API calls that the package is providing (much like the rest of the code is importing `TxPool` and using it). This could the part that makes  tests somewhat unstandardised and harder to "read".

File: txpool/txpool_test.go (hunk 8 above)
Comment (dbrajovic) [reply txpool/txpool_test.go]: I can't say I was able to reproduce that scenario. What steps did you take ?

File: txpool/txpool.go [hunk 12]
```diff
@@ -64,6 +64,34 @@ type signer interface {
...
+	g.current -= slots
+}
+
+func (g *slotGauge) height() uint64 {
+	g.Lock()
+	defer g.Unlock()
+
+	return g.current
+}
+
```
zivkovicmilos: Please document these methods

txpool/txpool.go
```diff
@@ -555,38 +566,74 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
 			t.remoteTxns.Push(tx)
 		}
 		return nil, false
 	}
+
 	return dropped, true
 }
 
 // increaseSlots increases number of taken slots
 func (t *TxPool) increaseSlots(slots uint64) {
```
zivkovicmilos: I think this method is completely useless since you can directly call `t.gauge.increase(slots)`

txpool/txpool.go
```diff
@@ -555,38 +566,74 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
-	atomic.AddUint64(&t.slots, slots)
+	t.gauge.increase(slots)
 }
 
 // increaseSlots decreases number of taken slots
 func (t *TxPool) decreaseSlots(slots uint64) {
```
zivkovicmilos: I think this method is completely useless since you can directly call `t.gauge.decrease(slots)`

txpool/txpool.go
zivkovicmilos: Agree with @brkomir that this func should be named more appropriately 👍 

Maybe something like `processSlots`.

txpool/txpool.go
zivkovicmilos: I think we should leave it in this method as it is now, since it was all in `addImpl` anyways, we just moved it to a helper.
Moving it out as the txpool currently stands would just complicate spaghetti the logic even more.

When we finally do a refactor of the txpool, this will definitely get outsourced to a more appropriate place

txpool/txpool.go
zivkovicmilos: +1 on `slotsRequired`

txpool/txpool.go
zivkovicmilos: This method is not exported anywhere, that's why @brkomir is suggesting we just get rid of it

txpool/txpool.go
zivkovicmilos: Agree with @Kourin1996, I don't think there's a reason to make a func named `height` return `current`.

I suggest you rename both the field and getter function name. `current` just seems ambiguous.

zivkovicmilos (changes requested):
Left a few minor comments, which I'd like to discuss.

The implementation is solid, it's just the naming that's a bit off

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/1cf77747ebc905fe88b13247fcd0357562973f9c)

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/1cf77747ebc905fe88b13247fcd0357562973f9c)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/97b98b295063247033647f8cfca2d7dd5b82378a)

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/1cf77747ebc905fe88b13247fcd0357562973f9c)

txpool/txpool.go
dbrajovic: Agreed, `increaseSlots/decreaseSlots` is redundant now.

[Removed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/b0b4046798b2672bb5746693333c71f0f2f7ee49)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/b0b4046798b2672bb5746693333c71f0f2f7ee49)

txpool/txpool.go
dbrajovic: [Documented](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/3cc877447981839e5fb76f69e2aacb357459a249)

zivkovicmilos (approved):
Thank you for making the requested cosmetic changes 🙏 

The implementation is solid, looks good to me 💯

Kourin1996 (approved):
LGTM 👍

txpool/txpool_test.go
brkomir: just run the test several times, this is the error i got right now
Screenshot 2021-11-23 at 16 44 36

brkomir (approved):

txpool/txpool.go
```diff
@@ -555,38 +567,60 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
-	atomic.AddUint64(&t.slots, slots)
-}
+// Checks if the incoming tx would cause an overflow
+// and attempts to allocate space for it
+func (t *TxPool) processSlots(tx *types.Transaction, isLocal bool) error {
+	if t.gauge.height+slotsRequired(tx) <= t.gauge.limit {
+		return nil // no overflow
+	}
+
+	if !isLocal && t.Underpriced(tx) {
```
lazartravica, line 622: I don't think this check here is related to slots themselves, and as such shouldn't be in the processSlots method.

txpool/txpool.go
```diff
@@ -555,38 +567,60 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
 	return underpriced
 }
 
-func (t *TxPool) Discard(slots uint64, force bool) ([]*types.Transaction, bool) {
+func (t *TxPool) Discard(remaining uint64, force bool) ([]*types.Transaction, bool) {
```
lazartravica: The name `remaining` is correct from the context of the below for loop, but not exactly when calling the method. If I were to see this method signature, I wouldn't know what the author meant by `remaining`. I propose `numberOfSlotsToRemove` or something less explicit but still straight to the point.

txpool/txpool.go
```diff
@@ -555,38 +567,60 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
 
-		txSlots := numSlots(tx.tx)
-		if slots >= txSlots {
-			slots -= txSlots
+		txSlots := slotsRequired(tx.tx)
+		if remaining >= txSlots {
+			remaining -= txSlots
 		} else {
-			slots = 0
+			remaining = 0
```
lazartravica: We can do an early return here for more readability, and with that we don't need the `else` branch.

txpool/txpool.go
```diff
@@ -345,16 +360,13 @@ func (t *TxPool) addImpl(origin TxOrigin, tx *types.Transaction) error {
 		t.locals.addAddr(tx.From)
 	}
 
-	// Skip check of GasPrice in the future transactions created by same address when TxPool receives transaction by Gossip or Reorg
-	if isLocal && !t.locals.containsAddr(tx.From) {
-		t.locals.addAddr(tx.From)
-	}
-
+	// Move promotable txs to the pending queue
```
lazartravica, line 362: Not saying it is wrong, but why did we get rid of this locals functionality?

txpool/txpool.go
```diff
@@ -306,27 +333,16 @@ func (t *TxPool) addImpl(origin TxOrigin, tx *types.Transaction) error {
...
-				wrapper.accountQueue.Remove(tx.Hash)
-			}
-			mux.unlock()
-
-			t.pendingQueue.Delete(tx)
-			t.decreaseSlots(numSlots(tx))
-		}
+	// check for slot overflow and handle accordingly
+	t.gauge.Lock()
+	if err := t.processSlots(tx, isLocal); err != nil {
```
lazartravica, line 341: Why don't you put the Lock & Unlock as well as increasing it by slots required inside the `processSlots` method itself? That way we have isolated all slots logic inside this method (which I think was the intention) and also we have eliminated the possibility of somebody else invoking the `processSlots` method without properly locking and unlocking around it.

lazartravica (changes requested):
Good job on refactoring for more clear code.

Please take a look at my comments

txpool/txpool.go
dbrajovic: It was discussed in @brkomir 's third review conversation (now resolved). 

If you still feel it should be moved elsewhere, I can do so.

txpool/txpool.go
dbrajovic: We didn't. It was duplicate code.

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/0da6d39f81009802cb0ee3253d634dfe9c0204a6)

txpool/txpool.go
dbrajovic: ~[Fixed](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/0da6d39f81009802cb0ee3253d634dfe9c0204a6)~

[New fix](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/7b2c3e3fe2f9abd373538df74c4a1eedcfd3bf75)

txpool/txpool.go
dbrajovic: That's the way it was originally written a few commits back (https://github.com/0xPolygon/polygon-sdk/pull/213/commits/bab1157a912d15930dc152b2a0b5cf520b8ee9c7). But the `TestGaugeCheck` test showed on multiple runs that a race condition is still possible. 

This code fixes that at the expense of readability. And adds to the point of refactoring TxPool in general.

txpool/txpool.go
lazartravica: @zivkovicmilos @dbrajovic I understand that you want to avoid adding extra Lines of Code to the addImpl() itself, but we don't extract code into methods/functions because we want to hide things under the carpet, but because we want to contextually organize in this case.

My opinion is that in the long run, this will make it more easier to refactor since we will be aware of this piece of logic, and not put it away in the processSlots.

txpool/txpool.go
```diff
@@ -570,38 +580,61 @@ func (t *TxPool) Underpriced(tx *types.Transaction) bool {
...
 	dropped := make([]*types.Transaction, 0)
-	for t.remoteTxns.Length() > 0 && slots > 0 {
+	for t.remoteTxns.Length() > 0 && slotsToRemove > 0 {
 		tx := t.remoteTxns.Pop()
 		dropped = append(dropped, tx.tx)
 
-		txSlots := numSlots(tx.tx)
-		if slots >= txSlots {
-			slots -= txSlots
+		txSlots := slotsRequired(tx.tx)
```
lazartravica: You forgot to eliminate the `else`.

Lines 589-593 can be replaced with
```
txSlots := slotsRequired(tx.tx)
if slotsToRemove < txSlots {
  return dropped, false
}

slotsToRemove -= txSlots
```

Every additional negation we need to do in our heads to figure out what's happening is a wasted thought process :)

lazartravica (changes requested):

txpool/txpool.go
lazartravica: We have discussed this on a huddle, this check for Underpriced needs to occur only if we see there's not enough space in the current slots, that's why it should be in the processSlots. Thank you @dbrajovic for explaining.

@dbrajovic You can perhaps add a comment of similar contents: `If we need to find space for this tx, first let's check if its' gasPrice is less than the tx with the smallest gasPrice in the current txPool`

txpool/txpool.go
dbrajovic: Refactored that entire piece in https://github.com/0xPolygon/polygon-sdk/pull/213/commits/7b2c3e3fe2f9abd373538df74c4a1eedcfd3bf75

txpool/txpool.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/7b2c3e3fe2f9abd373538df74c4a1eedcfd3bf75)

txpool/txpool.go
dbrajovic: [Added comments](https://github.com/0xPolygon/polygon-sdk/pull/213/commits/7b2c3e3fe2f9abd373538df74c4a1eedcfd3bf75)

lazartravica (approved):
LGTM 👍

PR #216, opened 2021-11-17
Feature/prometheus metrics

munna0908:
# Description

This PR adds prometheus instrumentation feature, it currently captures following metrics

-  No.of Validators 
-  No.of Round
-  No.of Pending Transactions 
-  No.of Transactions for block 
-  Block Interval 

The prometheus listener address should be specified using `-metrics` flag 

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments


munna0908:
@lazartravica @zivkovicmilos @Kourin1996  Please take a look and let me know if we need to add unit test cases for this.

command/helper/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	BlockGasTarget string                 `json:"block_gas_target"`
 	GRPCAddr       string                 `json:"rpc_addr"`
 	JSONRPCAddr    string                 `json:"jsonrpc_addr"`
+	METRICSAddr    string                 `json:"metrics_addr"`
```
ferranbt: Could we lowercase this?

consensus/metrics.go
```diff
@@ -0,0 +1,64 @@
+package consensus
+
+import (
+	"github.com/go-kit/kit/metrics"
```
ferranbt, line 4: Not sure If i would use one of this big aggregator libraries. I prefer to use something more lightweight and specific for the problem. This is what I normally use https://github.com/armon/go-metrics.

command/helper/config.go
ferranbt: I would also create a type Telemetry and include Prometheus there because you will likely have more things on the future related to telemetry (like OpenTracing).
```
type Telemetry struct {
   Prometheus string
}
```

command/helper/helper.go
```diff
@@ -414,6 +414,7 @@ func ReadConfig(baseCommand string, args []string) (*Config, error) {
 	flags.StringVar(&cliConfig.JSONRPCAddr, "jsonrpc", "", "")
 	flags.StringVar(&cliConfig.Join, "join", "", "")
 	flags.StringVar(&cliConfig.Network.Addr, "libp2p", "", "")
+	flags.StringVar(&cliConfig.METRICSAddr, "metrics", "", "")
```
ferranbt: This should be called `prometheus`.

txpool/txpool.go
```diff
@@ -117,6 +117,8 @@ type TxPool struct {
 
 	// Indicates which txpool operator commands should be implemented
 	proto.UnimplementedTxnPoolOperatorServer
+
+	metrics *Metrics
```
ferranbt, line 121: Metrics is a specific type of object because it is needed in many of the modules of the code. Then, it is usually set as a global variable like [this](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/eth/downloader/metrics.go#L26). If you use metrics this way it will simplify a lot the PR since you will not need to move around a Metrics type.

command/helper/config.go
munna0908: @ferranbt Thanks for your input, will create a new type called telemetry.

txpool/txpool.go
munna0908: @ferranbt I felt like having a separate metric type for each subsystem (**_consensusMetrics_**,**_txpoolMetrics_**,**_p2pMetrics_**) would be more cleaner and give us more control down the line. But I do agree with you that it makes the PR more simpler.
@lazartravica @Vuksan Whats your take ?

lazartravica: @munna0908 I'd be happy to discuss more, I am leaning more towards what Ferran said.

If you could argument what kind of control would be given by current implementation, that would be excellent.

My main point is that here we should try to make non-functional features such as Metrics make as less changes to the signatures of our methods as possible. In the end this would simplify testing as well, not only usage.

lazartravica (changes requested):
I don't feel that tests are needed on this PR, the biggest test will be seeing that the metrics are showing correctly on the testnet.

As for the comments Ferran and I made, please take a look at them 🙏

txpool/txpool.go
munna0908: @lazartravica As discussed, i feel this approach will make future enhancements very easily(when we would want to support different instrumentations)and it will not be lot of overhead for writing other unit test cases, since the constructor to individual metric type can return a nilmetric if the instrumentation type is set to nil.
Looking forward for your inputs @ferranbt

consensus/metrics.go
munna0908: @ferranbt Thanks for suggesting new package,I think this package doesn't support histogram metric, which is required to capture block interval metric.

txpool/txpool.go
lazartravica: @munna0908 Can you please provide some examples about this? Got it for the tests.

munna0908: @lazartravica  For example, for consensus metric type, currently we have a [function](https://github.com/0xPolygon/polygon-sdk/blob/34c58d0340c500938a6487846884eb2e8e9d7136/consensus/metrics.go#L21) to return prometheus metrics, likewise we can define functions to return other metrics like openMetrics etc.

command/helper/helper.go
munna0908: Done, Here's the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/ce11ec1c9f644ed16644db8da3718955cbf82e86).

command/helper/config.go
munna0908: Commit [Link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/23a9953c3222ebfd6de9d46123e3e478e89bb7c4)

txpool/txpool.go
lazartravica: @munna0908 why wouldn't you be able to do it in the example that ferran linked in the first reply, I still don't understand?

munna0908: > @munna0908 why wouldn't you be able to do it in the example that ferran linked in the first reply, I still don't understand?

@lazartravica we can absolutely do it,but i feel it will not be very clean and it is cumbersome to maintain.

lazartravica (approved):
LGTM

command/helper/config.go
```diff
@@ -33,6 +34,11 @@ type Config struct {
 	Join           string
 }
 
+// Tele
```
zivkovicmilos: Please add a better description of this struct

command/helper/config.go
```diff
@@ -219,6 +232,11 @@ func (c *Config) mergeConfigWith(otherConfig *Config) error {
 		c.GRPCAddr = otherConfig.GRPCAddr
 	}
 
+	if otherConfig.Metrics.PrometheusAddr != "" {
+
```
zivkovicmilos: Nitpick warning: please remove the empty space

consensus/ibft/ibft.go
```diff
@@ -637,12 +646,28 @@ func (i *Ibft) runValidateState() {
...
 		} else {
+			// update metrics
+			i.updateMetrics(block)
+
 			// move ahead to the next block
 			i.setState(AcceptState)
 		}
 	}
 }
+func (i *Ibft) updateMetrics(block *types.Block) {
```
zivkovicmilos, line 671: Please document this method better

consensus/consensus.go
```diff
@@ -55,4 +55,5 @@ type Factory func(
 	*state.Executor,
 	*grpc.Server,
 	hclog.Logger,
+	*Metrics,
```
zivkovicmilos: Why can't we have the metrics field be part of the consensus Config?

server/config.go
```diff
@@ -34,5 +34,11 @@ func DefaultConfig() *Config {
 		JSONRPCAddr: &net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: DefaultJSONRPCPort},
 		GRPCAddr:    &net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: DefaultGRPCPort},
 		Network:     network.DefaultConfig(),
+		Metrics:     &Telemetry{PrometheusAddr: nil},
 	}
 }
+
+// Telemetry holds the config details for metric services.
```
zivkovicmilos: Nitpick warning: Please don't add punctuation to comments

server/config.go
```diff
@@ -18,14 +18,14 @@ type Config struct {
...
 	LibP2PAddr  *net.TCPAddr
-
-	Network    *network.Config
-	DataDir    string
-	Seal       bool
-	Locals     []types.Address
-	NoLocals   bool
-	PriceLimit uint64
-	MaxSlots   uint64
+	Metrics     *Telemetry
```
zivkovicmilos: Why don't we call this field `Telemetry` instead of `Metrics`?

consensus/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func NilMetrics() *Metrics {
+	return &Metrics{
+		Validators:    discard.NewGauge(),
+		Rounds:        discard.NewGauge(),
+		NumTxs:        discard.NewGauge(),
+		BlockInterval: discard.NewHistogram(),
+	}
+}
```
zivkovicmilos: Please document this entire file better (methods, implementation details...)

server/config.go
zivkovicmilos: Not just here, but also in the `helper/config.go`

consensus/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+type Metrics struct {
+	Validators metrics.Gauge
+
+	Rounds metrics.Gauge
+
+	NumTxs metrics.Gauge
+
+	//Time between current block and the previous block in seconds
+	BlockInterval metrics.Histogram
+}
```
zivkovicmilos: For some reason, you define a `Metrics` struct here with the methods `GetPrometheusMetrics` and `NilMetrics`, but you do the exact same thing in the `txpool` package.

Please define an interface and force these **specific implementations** to implement it.

zivkovicmilos (changes requested):
Overall it looks good, but there are some design choices I think we need to reconsider.

Please see the comments I've left

Kourin1996 (approved):
LGTM 👏

server/config.go
munna0908: Renamed it Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/dec1ff280243cbd04ebf13eb154c4955685b314f)

command/helper/config.go
munna0908: Removed it

consensus/metrics.go
munna0908: Done, Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/40e745fb9520b6e8aaae856a1f0a978803abe2ec)

server/config.go
munna0908: Done, Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/40e745fb9520b6e8aaae856a1f0a978803abe2ec)

consensus/ibft/ibft.go
munna0908: Done, commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/40e745fb9520b6e8aaae856a1f0a978803abe2ec)

command/helper/config.go
munna0908: Done, commit [link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/40e745fb9520b6e8aaae856a1f0a978803abe2ec)

consensus/metrics.go
munna0908: @zivkovicmilos, `GetPrometheusMetrics` & `NilMetrics` are more like constructor for individual metric subsystems, currently metrics type doesn't have any concrete methods, thats the reason why i didn't go with an interface.

consensus/consensus.go
munna0908: I felt like having metrics in Config will make code more clumsy, since fields of metrics type will be used frequently to update the counters across the flow.

zivkovicmilos (approved):
Thank you for resolving the requested changes, and for documenting the code better 🙏 

Looks good to me 💯

command/helper/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	BlockGasTarget string                 `json:"block_gas_target"`
 	GRPCAddr       string                 `json:"rpc_addr"`
 	JSONRPCAddr    string                 `json:"jsonrpc_addr"`
+	Telemetry      *Telemetry             `json:"telemetrics"`
```
lazartravica: I think this should be `telemetry`, not `telemetrics`, as far as I understand from the docs: https://github.com/0xPolygon/polygon-sdk-docs/pull/25/files

lazartravica (changes requested):
@munna0908 please resolve merge conflicts and take a look at my comment around th econfig.

command/helper/config.go
munna0908: Fixed it commit-[link](https://github.com/0xPolygon/polygon-sdk/pull/216/commits/af59b874784e4644f8292c4bf3b089d89dad8909)

lazartravica (approved):

lazartravica (approved):

PR #217, opened 2021-11-18
[IBFT] Block difficulty is one 

dbrajovic:
# Description

Originally starting as the idea in (now closed) [PR#182](https://github.com/0xPolygon/polygon-sdk/pull/182), this PR is the third (and final) part of the series, addressing the block difficulty in IBFT. The first two being [PR#201](https://github.com/0xPolygon/polygon-sdk/pull/201) and [PR#208](https://github.com/0xPolygon/polygon-sdk/pull/208).

Essentially, it conforms to the standard described in the original [IBFT consensus](https://github.com/ethereum/EIPs/issues/650#:~:text=DEFAULT_DIFFICULTY%3A%20Default%20block%20difficulty%2C%20which%20is%20set%20to%200x0000000000000001%20.)

# Changes include

# Breaking changes

The breaking change is introduced [here](https://github.com/0xPolygon/polygon-sdk/pull/217/commits/f6dff5c290ae8a5cc39af4907f52af6cc53ac00c).

A previously running node restarted with this build might still be able to produce blocks, but the danger of it failing to verify headers (during bulk sync) at some point is never absent. 

Newly added nodes to the cluster will not be able to move past genesis and can never sync with the chain.

# Checklist

- [ ] I have tested this code


consensus/ibft/ibft.go
```diff
@@ -351,7 +351,7 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 		Miner:      types.Address{},
 		Nonce:      types.Nonce{},
 		MixHash:    IstanbulDigest,
-		Difficulty: parent.Number + 1,   // we need to do this because blockchain needs difficulty to organize blocks and forks
+		Difficulty: 1,                   // we need to do this because blockchain needs difficulty to organize blocks and forks
```
zivkovicmilos: We should probably make this a global constant of the `ibft` package

chain/chain.go
```diff
@@ -17,7 +17,7 @@ var (
 	GenesisGasLimit uint64 = 4712388
 
 	// GenesisDifficulty is the default difficulty of the Genesis block.
-	GenesisDifficulty = big.NewInt(131072)
+	GenesisDifficulty = big.NewInt(1)
```
zivkovicmilos, line 20: Have we confirmed that modifying this value will cause the breaking change, and leaving it alone will not?

zivkovicmilos (approved):
Looks good to me 💯 

I've left some minor comments.

Thank you for leaving the chronological description of the changes in the PR, now people can follow along 🙏

Kourin1996 (approved):
Looks good to me, apart from the reviews by Milos 👍

consensus/ibft/ibft.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/217/commits/cd26d0c65120515afd6c6e956c0e74a4542d01d9)

chain/chain.go
dbrajovic: Yes, I have tested the change with a local 4 validator node setup under this build. I updated the PR description on that matter.

brkomir (approved):

mrwillis:
hi @dbrajovic @lazartravica, will this be added as a flag or something like --ibft-difficulty in `develop` given that it's breaking? Is there any actual benefits to using this other than conforming to the spec?

dbrajovic:
> hi @dbrajovic @lazartravica, will this be added as a flag or something like --ibft-difficulty in `develop` given that it's breaking? Is there any actual benefits to using this other than conforming to the spec?

The original idea was just conforming to the spec - no additional benefits other than that. If this gets released after the point the chain cannot be reset for you, we'll probably include a forks-in-time support

zivkovicmilos:
Closing and archiving under `archive/bug/block-difficulty-is-one`:

https://github.com/0xPolygon/polygon-sdk/releases/tag/archive%2Fbug%2Fblock-difficulty-is-one

PR #218, opened 2021-11-18
Increased timeout of popBlock method to 20s and decreased log level t…

brkomir:
…o INFO if error occurs

# Description

Purpose of this PR is to decrease the log level for the unnecessary ERROR logs for popBlock method timeout on the non-validator nodes. This occurs when not all of the validator nodes are up and running and the block production time can vary depending of the number of nodes down, causing the non-validator node to encounter a timeout while waiting for new block to arrive.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/protocol/syncer.go
+++ b/protocol/syncer.go
@@ -513,9 +513,14 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
 			s.logger.Info("Connection to a peer has closed already", "id", p.peer)
 			break
 		}
+
 		b, err := p.popBlock(popTimeout)
 		if err != nil {
-			s.logger.Info("failed to pop block", "err", err)
+			if errors.Is(err, ErrPopTimeout) {
+				s.logger.Debug(fmt.Sprintf("failed to pop block within %ds", popTimeout))
+			} else {
+				s.logger.Info("failed to pop block", "err", err)
+			}
 			break
 		}
 		if err := s.blockchain.WriteBlocks([]*types.Block{b}); err != nil {
--- a/protocol/syncer.go
+++ b/protocol/syncer.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	maxEnqueueSize = 50
-	popTimeout     = time.Second * 10
+	popTimeout     = 20 * time.Second
 )
 
 var (
@@ -515,7 +515,7 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
 		}
 		b, err := p.popBlock(popTimeout)
 		if err != nil {
-			s.logger.Error("failed to pop block", "err", err)
+			s.logger.Info("failed to pop block", "err", err)
 			break
 		}
 		if err := s.blockchain.WriteBlocks([]*types.Block{b}); err != nil {
--- a/protocol/syncer.go
+++ b/protocol/syncer.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	maxEnqueueSize = 50
-	popTimeout     = 20 * time.Second
+	popTimeout     = 10 * time.Second
 )
 
 var (
@@ -516,11 +516,7 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
 
 		b, err := p.popBlock(popTimeout)
 		if err != nil {
-			if errors.Is(err, ErrPopTimeout) {
-				s.logger.Debug(fmt.Sprintf("failed to pop block within %ds", popTimeout))
-			} else {
-				s.logger.Info("failed to pop block", "err", err)
-			}
+			s.logSyncPeerPopBlockError(err, p)
 			break
 		}
 		if err := s.blockchain.WriteBlocks([]*types.Block{b}); err != nil {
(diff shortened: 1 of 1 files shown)


lazartravica (approved):

lazartravica:
@brkomir Please try to supply better information in the body of the pull request as to why the PR was made. (under description).

protocol/syncer.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	maxEnqueueSize = 50
-	popTimeout     = time.Second * 10
+	popTimeout     = 20 * time.Second
```
zivkovicmilos: How did you come up with this number? `20s`?

protocol/syncer.go
```diff
@@ -513,9 +513,14 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
...
 		}
+
 		b, err := p.popBlock(popTimeout)
 		if err != nil {
-			s.logger.Error("failed to pop block", "err", err)
+			if errors.Is(err, ErrPopTimeout) {
+				s.logger.Debug(fmt.Sprintf("failed to pop block within %ds", popTimeout))
+			} else {
+				s.logger.Info("failed to pop block", "err", err)
+			}
```
zivkovicmilos: We can probably separate this out into a helper method for syncer, this func is already as complicated as it is 👍

zivkovicmilos (approved):
Looks good to me 💯 

I've left a few minor comments, nothing major 🙂

mrwillis:
@brkomir 

Not sure if this was relevant for an issue, but is there any way to make this non-blocking or connect to a new peer? We've seen cases of this causing "slow transaction confirmations". This is the scenario we've seen

1. Node is connected to a bad peer which has pop block timeouts every 10-15 blocks (still trying to reproduce why this is happening)
2. Client submits txn to the node
3. Client is continually polling `eth_getTransactionReceipt` on node
4. Node is going to timeout on pop block, which takes a while, while blocking new blocks from being processed, and hence `eth_getTransactionReceipt` keeps returning a null receipt to the client
5. Bunch of blocks come in after the pop block, `eth_getTransactionReceipt` finally returns non-null receipt

Client "thinks" txn took 20-30 seconds.

protocol/syncer.go
brkomir: When you want to grow something you usually double it :) I remember we mentioned this on the meeting, but it really doesn't matter how long this is.

protocol/syncer.go
```diff
@@ -513,9 +513,14 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
 			s.logger.Info("Connection to a peer has closed already", "id", p.peer)
 			break
 		}
+
 		b, err := p.popBlock(popTimeout)
 		if err != nil {
-			s.logger.Error("failed to pop block", "err", err)
+			if errors.Is(err, ErrPopTimeout) {
+				s.logger.Debug(fmt.Sprintf("failed to pop block within %ds", popTimeout))
```
Kourin1996: I think it's ok this log has info level, too

Kourin1996 (approved):
LGTM! I've only left small comments. We can continue to discuss about this. Thank you.

protocol/syncer.go
lazartravica: @brkomir We reached a consensus of having the Warn level with a message saying: "failed to pop block within %ds, please check if all the validators are running"

protocol/syncer.go
lazartravica: @brkomir We talked about decreasing this to the previous value.

protocol/syncer.go
lazartravica: Also include the ID of the Peer.

protocol/syncer.go
```diff
@@ -513,9 +513,14 @@ func (s *Syncer) WatchSyncWithPeer(p *syncPeer, handler func(b *types.Block) boo
...
+			} else {
+				s.logger.Info("failed to pop block", "err", err)
```
lazartravica: Also include the ID of the peer in this msg.

protocol/syncer.go
brkomir: https://github.com/0xPolygon/polygon-sdk/pull/218/commits/a63e3b1996d3790840e45d3c8f4f203d8eecf428

protocol/syncer.go
brkomir: Done. https://github.com/0xPolygon/polygon-sdk/pull/218/commits/a63e3b1996d3790840e45d3c8f4f203d8eecf428

protocol/syncer.go
brkomir: Done https://github.com/0xPolygon/polygon-sdk/pull/218/commits/a63e3b1996d3790840e45d3c8f4f203d8eecf428

protocol/syncer.go
brkomir: https://github.com/0xPolygon/polygon-sdk/pull/218/commits/a63e3b1996d3790840e45d3c8f4f203d8eecf428

lazartravica (approved):

dbrajovic (approved):

PR #219, opened 2021-11-18
Combined IBFT PoA + PoS

zivkovicmilos:
# Description

This PR adds support for combining both the current PoA implementation and PoS implementation into the main feature branch of the Polygon SDK.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments
The documentation PR for these changes is up as well:
https://github.com/0xPolygon/polygon-sdk-docs/pull/37


command/helper/helper.go
```diff
@@ -531,3 +543,67 @@ func DirectoryExists(directoryPath string) bool {
...
+	stakingAccount.Storage = storageMap
+
+	// Set the Staking SC balance to numValidators * defaultStakedBalance
+	stakingAccount.Balance = stakedAmount
+
+	// Add the account to the premine map so the executor can apply it to state
+	premineMap[types.StringToAddress("1001")] = stakingAccount
+
+	return nil
+}
```
Kourin1996: Shall we move this function into a specific package for PoS?

zivkovicmilos: Great idea 💯 

There's no need for this PoS logic to be bundled with other helper methods. We can move it to a separate package

zivkovicmilos: Fixed 👍 

7575d59

command/genesis/genesis.go
```diff
@@ -205,15 +241,45 @@ func (c *GenesisCommand) Run(args []string) int {
...
 		extraData = ibftExtra.MarshalRLPTo(extraData)
 	}
 
+	// Check if any addresses should be prestaked in the dev consensus
+	if consensus == devConsensus {
+		if len(ibftValidators) != 0 {
+			for _, val := range ibftValidators {
+				validators = append(validators, types.StringToAddress(val))
+			}
+		}
```
Kourin1996: You don't need to check if ibftValidators is not zero because for loop is skipped when it's zero

command/genesis/genesis.go
```diff
@@ -205,15 +241,45 @@ func (c *GenesisCommand) Run(args []string) int {
...
+
+		if isPos {
+			return map[string]interface{}{
+				"type": ibft.PoS,
+			}
+		}
+
+		return map[string]interface{}{
+			"type": ibft.PoA,
+		}
```
Kourin1996: Can we define each map as constant value?

command/genesis/genesis.go
zivkovicmilos: Fixed 👍 

c0fa6eb

command/helper/helper.go
```diff
@@ -276,6 +277,21 @@ func FillPremineMap(
...
+	// For each map, iterate over the keys and save the values
+	// into the resulting map
+	for _, m := range maps {
+		for key, value := range m {
+			mergedMap[key] = value
+		}
+	}
+	return mergedMap
+}
+
```
brkomir, line 291: Just a side note, I think we don't need this much granularity in our comments, since I would argue that there is no benefit in any comment on this method, the method is short and self-explanatory and that's great, but comments are just reiterating what the code does and bring unneccessary clutter.

command/genesis/genesis.go
zivkovicmilos: Fixed 👍 

c0fa6eb

command/helper/helper.go
zivkovicmilos: Agreed, I figure the method description is good enough 👍 

Fixed 👍
 
9723dd6

command/ibft/ibft_candidates.go
```diff
@@ -9,50 +9,51 @@ import (
...
-// IbftCandidates is the command to query the snapshot
+// IbftCandidates is the command to get a current list of IBFT candidates
+// and their corresponding votes
 type IbftCandidates struct {
 	helper.Meta
 }
 
 // GetHelperText returns a simple description of the command
-func (p *IbftCandidates) GetHelperText() string {
+func (c *IbftCandidates) GetHelperText() string {
```
brkomir, line 27: Another example, this function is very clear in its intentions, no need for the comment above since it brings no value to understanding what the method does.

consensus/ibft/ibft.go
```diff
@@ -82,12 +83,136 @@ type Ibft struct {
...
 
 	secretsManager secrets.SecretsManager
+
+	// nolint:unused
+	mechanism ConsensusMechanism // IBFT ConsensusMechanism used (PoA / PoS)
+}
+
+// Define the type of the iBFT consensus
+
+type Type string
```
brkomir: This name "Type" seems too general, consider renaming it to Consensus for example, or something more specific.

consensus/ibft/ibft.go
```diff
@@ -82,12 +83,136 @@ type Ibft struct {
...
+// ConsensusMechanismFactory is the factory function to create a consensus mechanism
+type ConsensusMechanismFactory func(ibft *Ibft) (ConsensusMechanism, error)
+
+var mechanismBackends = map[Type]ConsensusMechanismFactory{
+	PoA: PoAFactory,
+	PoS: PoSFactory,
+}
+
+// runHook runs a specified hook if it is present in the hook map
+// nolint:unused
```
brkomir: remove this "nolint" instructions

consensus/ibft/poa.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Register the ProcessHeadersHook
+	poa.hookMap[ProcessHeadersHook] = poa.processHeadersHook
+
+	// Register the CandidateVoteHook
+	poa.hookMap[CandidateVoteHook] = poa.candidateVoteHook
+}
+
+// ShouldWriteTransactions indicates if transactions should be written to a block
+func (poa *PoAMechanism) ShouldWriteTransactions(blockNumber uint64) bool {
```
brkomir, line 243: Why this method isn't included in the hook mechanism? Is it because the hook funcs don't return a value other than error? This is what I was talking about in our previous discussion of hooks vs interface methods.

consensus/ibft/pos.go
```diff
@@ -0,0 +1,196 @@
...
+import (
+	"errors"
+	"fmt"
+	"github.com/0xPolygon/polygon-sdk/contracts/staking"
+	"github.com/0xPolygon/polygon-sdk/types"
+)
+
+// PoSMechanism defines specific hooks for the Proof of Stake IBFT mechanism
+type PoSMechanism struct {
+	// Reference to the main IBFT implementation
```
brkomir, line 12: unnecessary comments

consensus/ibft/pos.go
```diff
@@ -0,0 +1,196 @@
...
+
+// PoSFactory initializes the required data
+// for the Proof of Stake mechanism
+func PoSFactory(ibft *Ibft) (ConsensusMechanism, error) {
+	pos := &PoSMechanism{
+		mechanismType: PoS,
+		ibft:          ibft,
+	}
+
+	// Initialize the hook map
```
brkomir: unnecessary comment

brkomir (approved):
Overall looks great! I have left a few comments just as a discussion, nothing major.

consensus/ibft/ibft.go
lazartravica: @brkomir I'd leave this up to Milos to decide, since the package name is IBFT, it isn't that unclear what `Type` means.

lazartravica (approved):

command/ibft/ibft_candidates.go
zivkovicmilos: Completely agree 👍 

All of our command interface methods are documented in this way, just to keep them standardized.

We can discuss removing these useless comments for all commands on some future PR

consensus/ibft/ibft.go
zivkovicmilos: I think it's fine we leave it as `Type` for now, as @lazartravica noted it's in the IBFT package.
We have countless examples of naming these helper types something generic in the PSDK.

Noticed a typo from your suggestion that I fixed 👍 

87df4ce

consensus/ibft/ibft.go
zivkovicmilos: Good catch 💯 

This is leftover from the stub merges.

Fixed 👍 

150a31d

consensus/ibft/poa.go
zivkovicmilos: Exactly.

Hooks are supposed to only execute logic and report an error in execution, not return values.

I felt like this method was a good compromise to be included as an interface method, because we might want to have additional logic in the future for deciding on block writes and this allows us to be flexible 👍

consensus/ibft/pos.go
zivkovicmilos: I like to enforce the practice of documenting every section of a structure we're using. It would seem messy to just document only certain parts of it, and leave others undocumented, even if they're clear in their use

consensus/ibft/pos.go
zivkovicmilos: Fixed it for the PoA mechanism as well 👍 

4a75f49

consensus/ibft/pos.go
brkomir: again, not sure how much value these comments bring, but in my experience, people just ignore these comments. Furthermore, if any time in the future the type of the field struct changes, people tend to forget to change the comments... Anyway, this Is ok for now.

command/ibft/ibft_candidates.go
brkomir: agreed.

consensus/ibft/ibft.go
```diff
@@ -364,6 +513,8 @@ func (i *Ibft) runSyncState() {
 			return isValidator
 		})
 
+		updateSnapshotCallback(oldLatestNumber)
+
```
Kourin1996: `oldLatestNumber` doesn't change after bulk sync is done. Is this ok?
(Let's think we start sync from 10 and could do bulk sync to 40 and synced one by one to 48. In this case, first `updateSnapshotCallback` is called with 10 and second should be called with 40? You don't need to update oldLatestNumber?)

zivkovicmilos: You're right 👍 

It would be updated only in the next loop.

I've also noticed that we can jump out of the loop quicker if we updated the snapshot store on a per block basis with PoS.

Fixed 👍 

d24f46e

Kourin1996 (approved):
Thank you for fixing the things. LGTM!

consensus/ibft/ibft.go
Vuksan: I am with @brkomir on this one. IMO we should strive away from generic types which increase the cognitive load to understand (take a look at the package to understand the full picture).
`ConsensusType` would seem more appropriate.

consensus/ibft/poa.go
```diff
@@ -0,0 +1,231 @@
...
+		len(snap.Votes),
+	)
+
+	return nil
+}
+
+// verifyHeadersHook verifies that the header nonce conforms to the IBFT PoA proposal format
+func (poa *PoAMechanism) verifyHeadersHook(nonceParam interface{}) error {
+	// Cast the param to the nonce
+	nonce := nonceParam.(types.Nonce)
```
Vuksan: Shouldn't we check if the cast has passed, and return some error otherwise?
`interface{}` can be anything.

consensus/ibft/poa.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Check the nonce format
+	// Because you must specify either AUTH or DROP vote, it is confusing how to have a block without any votes.
+	// 		This is achieved by specifying the miner field to zeroes,
+	// 		because then the value in the Nonce will not be taken into consideration.
+	if nonce != nonceDropVote && nonce != nonceAuthVote {
```
Vuksan, line 79: Let's move `nonceDropVote` and `nonceAuthVote` definitions to this file since they're only used here. 🙂

consensus/ibft/poa.go
```diff
@@ -0,0 +1,231 @@
...
+
+// verifyHeadersHook verifies that the header nonce conforms to the IBFT PoA proposal format
+func (poa *PoAMechanism) verifyHeadersHook(nonceParam interface{}) error {
+	// Cast the param to the nonce
+	nonce := nonceParam.(types.Nonce)
+
+	// Check the nonce format
+	// Because you must specify either AUTH or DROP vote, it is confusing how to have a block without any votes.
+	// 		This is achieved by specifying the miner field to zeroes,
+	// 		because then the value in the Nonce will not be taken into consideration.
```
Vuksan: I don't understand this statement, especially `it is confusing how to have a block without any votes.`
If everyone else understands this, don't feel like you need to change anything.

consensus/ibft/poa.go
```diff
@@ -0,0 +1,231 @@
...
+	snap       *Snapshot
+	parentSnap *Snapshot
+	proposer   types.Address
+	saveSnap   func(h *types.Header)
+}
+
+// processHeadersHook does the required logic for PoA header processing
+func (poa *PoAMechanism) processHeadersHook(hookParam interface{}) error {
+	// Cast the params to processHeadersHookParams
+	params := hookParam.(*processHeadersHookParams)
```
Vuksan: What if the cast fails?

consensus/ibft/poa.go
```diff
@@ -0,0 +1,231 @@
...
+type candidateVoteHookParams struct {
+	header *types.Header
+	snap   *Snapshot
+}
+
+// candidateVoteHook checks if any candidate is up for voting by the operator
+// and casts a vote in the Nonce field of the block being built
+func (poa *PoAMechanism) candidateVoteHook(hookParams interface{}) error {
+	// Cast the params to candidateVoteHookParams
+	params := hookParams.(*candidateVoteHookParams)
```
Vuksan: What if the cast fails?

consensus/ibft/pos.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Log the info message
+	pos.ibft.logger.Info(
+		"current snapshot",
+		"validators",
+		len(snap.Set),
+	)
+
+	return nil
+}
```
Vuksan: Does it make sense to also log their stake?
If that proves to be too complex compared to the value it brings, drop this request.

consensus/ibft/ibft.go
```diff
@@ -82,12 +83,135 @@ type Ibft struct {
...
+func (i *Ibft) runHook(hookName string, hookParams interface{}) error {
+	// Grab the hook map
+	hookMap := i.mechanism.GetHookMap()
+
+	// Grab the actual hook if it's present
+	hook, ok := hookMap[hookName]
+	if !ok {
+		// hook not found, continue
+		return nil
+	}
```
Vuksan: IMO it isn't ok to return nil as an error if the hook isn't found and trick the caller.
It would be better to return `HookNotFoundError` and then let the caller decide what he wants to do with that information (continue execution or halt).

consensus/ibft/pos.go
```diff
@@ -0,0 +1,198 @@
...
+
+// insertBlockHook checks if the block is the last block of the epoch,
+// in order to update the validator set
+func (pos *PoSMechanism) insertBlockHook(numberParam interface{}) error {
+	headerNumber := numberParam.(uint64)
```
Vuksan: What if the cast fails?

contracts/staking/query.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, errors.New("validators method doesn't exist in Staking contract ABI")
+	}
+
+	selector := method.ID()
+	res, err := t.Apply(&types.Transaction{
+		From:     from,
+		To:       &AddrStakingContract,
+		Value:    big.NewInt(0),
+		Input:    selector,
+		GasPrice: big.NewInt(0),
```
Vuksan, line 61: Is it ok to hardcode the price to 0 here?
What happens if the network contains transactions which have bigger gas price? Will it stop this one from being included?

contracts/staking/query.go
```diff
@@ -0,0 +1,73 @@
...
+func DecodeValidators(method *abi.Method, returnValue []byte) ([]types.Address, error) {
+	decodedResults, err := method.Outputs.Decode(returnValue)
+	if err != nil {
+		return nil, err
+	}
+	results, ok := decodedResults.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("failed type assertion from decodedResults to map")
+	}
+	web3Addresses, ok := results["0"].([]web3.Address)
```
Vuksan, line 31: How can you be sure there will be `results["0"]`?
Possible nil pointer exception?

consensus/ibft/pos.go
```diff
@@ -0,0 +1,198 @@
...
+		return number / i.epochSize
+	}
+
+	return number/i.epochSize + 1
+}
+
+// IsFirstOfEpoch checks if the block number is the first of the epoch
+func (i *Ibft) IsFirstOfEpoch(number uint64) bool {
+	return number%i.epochSize == 1
+}
```
Vuksan: Can we remove this method?
I see it's not used anywhere, except in `pos_test.go` where we test its behavior.

consensus/ibft/pos.go
```diff
@@ -0,0 +1,198 @@
...
+	); err != nil {
+		pos.ibft.logger.Error("failed to bulk update validators", "err", err)
+	}
+
+	return nil
+}
+
+// verifyBlockHook checks if the block is an epoch block and if it has any transactions
+func (pos *PoSMechanism) verifyBlockHook(blockParam interface{}) error {
+	block := blockParam.(*types.Block)
```
Vuksan: What if the cast fails?

consensus/ibft/pos_test.go
```diff
@@ -0,0 +1,146 @@
+package ibft
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
Vuksan, line 7: Why are we using this package for assertions?
Aren't the ones in pkg.go.dev good enough?

cc @lazartravica

helper/staking/staking.go
```diff
@@ -0,0 +1,176 @@
...
...0602082019050919050565b600081519050919050565b6000602082019050919050565b600082825260208201905092915050565b600082825260208201905092915050565b6000610ba682610c73565b9150610bb183610c73565b9250827fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff03821115610be657610be5610c7d565b5b828201905092915050565b6000610bfc82610c73565b9150610c0783610c73565b925082821015610c1a57610c19610c7d565b5b828203905092915050565b6000610c3082610c53565b9050919050565b60006fffffffffffffffffffffffffffffffff82169050919050565b600073ffffffffffffffffffffffffffffffffffffffff82169050919050565b6000819050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b600080fd5b7f4f6e6c79207374616b65722063616e2063616c6c2066756e6374696f6e000000600082015250565b7f696e646578206f7574206f662072616e67650000000000000000000000000000600082015250565b610d6a81610c73565b8114610d7557600080fd5b5056fea26469706673582212200732d127c069747c9fba40cde21178279bf118074d859232ff634cd4c0c2678964736f6c63430008070033"
```
Vuksan: Is there no better way to fetch this?

helper/staking/staking.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return bigKeccak.Bytes()
+}
+
+// GetStorageIndexes is a helper function for getting the correct indexes
+// of the storage slots which need to be modified during bootstrap.
+//
+// It is SC dependant, and based on the SC located at:
+// https://github.com/0xPolygon/staking-contracts/
+func GetStorageIndexes(address types.Address, index int64) *StorageIndexes {
```
Vuksan: This can be a private function.

helper/staking/staking.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	// Save the storage map
+	stakingAccount.Storage = storageMap
+
+	// Set the Staking SC balance to numValidators * defaultStakedBalance
+	stakingAccount.Balance = stakedAmount
+
+	// Add the account to the premine map so the executor can apply it to state
+	premineMap[types.StringToAddress("1001")] = stakingAccount
```
Vuksan: I think it would be better if we return this account from the function so that it can be more explicit for the caller.
I am not a big fan of modifying the object passed in the function, especially when it's implicit like this.

command/genesis/genesis.go
```diff
@@ -174,21 +215,29 @@ func (c *GenesisCommand) Run(args []string) int {
...
+	var validators []types.Address
+
+	if consensus == ibftConsensus {
+		// Epoch size must be greater than 1, so new transactions have a chance to be added to a block.
+		// Otherwise, every block would be an endblock (meaning it will not have any transactions).
+		// Check is placed here to avoid additional parsing if epochSize < 2
+		if epochSize < 2 {
+			c.UI.Error("Epoch size must be greater than 1")
+			return 1
+		}
```
Vuksan: Shouldn't we check whether consensus is `PoS` since this only applies in that case?

command/genesis/genesis.go
```diff
@@ -199,15 +248,39 @@ func (c *GenesisCommand) Run(args []string) int {
...
 		extraData = make([]byte, ibft.IstanbulExtraVanity)
 		extraData = ibftExtra.MarshalRLPTo(extraData)
 	}
 
+	// Check if any addresses should be prestaked in the dev consensus
+	if consensus == devConsensus {
+		for _, val := range ibftValidators {
+			validators = append(validators, types.StringToAddress(val))
+		}
+	}
```
Vuksan: Comment does not match the code.

command/helper/helper.go
```diff
@@ -274,6 +275,18 @@ func FillPremineMap(
...
+func MergeMaps(maps ...map[string]interface{}) map[string]interface{} {
+	mergedMap := make(map[string]interface{})
+	for _, m := range maps {
+		for key, value := range m {
+			mergedMap[key] = value
+		}
+	}
+
+	return mergedMap
+}
```
Vuksan: Wouldn't this overwrite all the values that first map has set?
Given that two (or more maps) have same keys.

I know it won't happen now, but one day it might, since the method signature is not really clear on this. 🙂

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+	expectedBalance.Add(expectedBalance, stakeAmount)
+
+	assert.Equal(t, expectedBalance.String(), scBalance.String())
+
+	stakedAmount, stakedAmountErr := framework.GetStakedAmount(stakerAddr, client)
+	if stakedAmountErr != nil {
+		t.Fatalf("Unable to get staked amount, %v", stakedAmountErr)
+	}
+
+	assert.Equal(t, expectedBalance.String(), stakedAmount.String())
```
Vuksan, line 136: Was your intention to check the total amount that is staked in the smart contract or the amount that this stakerAddr has staked?

Because it does the first, which was already checked in the previous assert if I'm not mistaken.

e2e/framework/helper.go
```diff
@@ -50,6 +54,123 @@ func GetAccountBalance(
...
+		Value:    amount,
+		Input:    MethodSig("stake"),
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	_, err := srv.SendRawTx(ctx, txn, senderKey)
+
+	if err != nil {
+		return fmt.Errorf("unable to call Staking contract method, %v", err)
```
Vuksan: Just in case:
```suggestion
		return fmt.Errorf("unable to call Staking contract method stake, %v", err)
```

e2e/framework/helper.go
```diff
@@ -50,6 +54,123 @@ func GetAccountBalance(
...
+		Value:    big.NewInt(0),
+		Input:    MethodSig("unstake"),
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	receipt, err := srv.SendRawTx(ctx, txn, senderKey)
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to call Staking contract method, %v", err)
```
Vuksan:
```suggestion
		return nil, fmt.Errorf("unable to call Staking contract method unstake, %v", err)
```

e2e/framework/helper.go
```diff
@@ -50,6 +54,123 @@ func GetAccountBalance(
...
+			From:     web3.Address(from),
+			To:       &toAddress,
+			Data:     selector,
+			GasPrice: 100000000,
+			Value:    big.NewInt(0),
+		},
+		web3.Latest,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to call Staking contract method, %v", err)
```
Vuksan:
```suggestion
		return nil, fmt.Errorf("Unable to call Staking contract method stakedAmount, %v", err)
```

e2e/framework/helper.go
```diff
@@ -50,6 +54,123 @@ func GetAccountBalance(
...
```
Vuksan:
```suggestion
		return nil, fmt.Errorf("Unable to call Staking contract method validators, %v", err)
```

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+		expBalance.String(),
+		actualAccountBalance.String(),
+		"Account balance mismatch after unstake exploit",
+	)
+
+	assert.Equalf(t,
+		zeroEth.String(),
+		actualStakingSCBalance.String(),
+		"Staked address balance mismatch after unstake exploit",
+	)
```
Vuksan: `framework.GetStakedAmount(senderAddr, client)` on L318 returns total staked amount of the whole network.
It shouldn't be zero right? Only one validator performed unstaking here, so other 3 should still have their balances staked

Vuksan: If we are running this with only 1 validator, let's change the test to have at least 2 of them please, to make the test realistic.

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+	defaultBalance := framework.EthToWei(100)
+	bigGasPrice := big.NewInt(framework.DefaultGasPrice)
+
+	senderKey, senderAddr := tests.GenerateKeyAndAddr(t)
+	numDummyStakers := 100
+
+	devInterval := 5 // s
+
+	// Set up the test server
+	srvs := framework.NewTestServers(t, 1, func(config *framework.TestServerConfig) {
```
Vuksan, line 373: More validators please. At least 2.

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+		expBalance.String(),
+		actualAccountBalance.String(),
+		"Account balance mismatch after stake / unstake exploit",
+	)
+}
+
+// Test scenario:
+// User has 0 ETH staked and a balance of 100 ETH
+// Stake 2 ETH -> Unstake
+// Expected result for tests: Staked: 0 ETH; Balance: ~100 ETH
```
Vuksan: Also add this check, please:
```suggestion
// Expected result for tests: Staked: 0 ETH; Balance: ~100 ETH; not a validator
```

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+	defaultBalance := framework.EthToWei(100)
+	bigGasPrice := big.NewInt(framework.DefaultGasPrice)
+
+	senderKey, senderAddr := tests.GenerateKeyAndAddr(t)
+	numDummyStakers := 10
+
+	devInterval := 5 // s
+
+	// Set up the test server
+	srvs := framework.NewTestServers(t, 1, func(config *framework.TestServerConfig) {
```
Vuksan, line 512: More validators please so that the initial staking amount doesn't only consist of this user's tokens.

e2e/pos_test.go
```diff
@@ -0,0 +1,617 @@
...
+				},
+				From: types.ZeroAddress.String(),
+			}
+		}
+
+		_, addErr := txpoolClient.AddTxn(context.Background(), msg)
+		if addErr != nil {
+			t.Fatalf("Unable to add txn, %v", addErr)
+		}
+	}
```
Vuksan: why the for loop for these 2 transactions? Is it because of the c/p from the previous one? 🙂

Vuksan (changes requested):
Good job on this effort. 👏 
Please take a look at the comments I've left.

consensus/ibft/ibft.go
zivkovicmilos: Got it 👍 

Renamed it to something more appropriate:

e965909

consensus/ibft/poa.go
zivkovicmilos: You're right 👍 

Added cast checks for all hooks:

1218296

consensus/ibft/poa.go
zivkovicmilos: You're right 👍 

Added cast checks for all hooks:

1218296

consensus/ibft/poa.go
zivkovicmilos: You're right 👍 

Added cast checks for all hooks:

1218296

consensus/ibft/pos.go
zivkovicmilos: You're right 👍 

Added cast checks for all hooks:

1218296

consensus/ibft/pos.go
zivkovicmilos: You're right 👍 

Added cast checks for all hooks:

1218296

consensus/ibft/poa.go
zivkovicmilos: Good catch 💯 

Moved: 

e506815

consensus/ibft/poa.go
zivkovicmilos: You're right, it's confusing 👍 

Fixed it:

8913803

consensus/ibft/pos.go
zivkovicmilos: Currently, we don't have support for grabbing the staked amount for an individual address. Adding that is trivial once we have it as a method on the SC.

I've added support for printing out the total staked amount (which we can do):

d67f984

consensus/ibft/ibft.go
zivkovicmilos: I agree 👍 

I've added explicit error handling with hooks:

b8c46b9

contracts/staking/query.go
zivkovicmilos: I think it's completely fine to leave it to 0, since there isn't any validator for it in `transition.Apply`, apart from the regular check for `gasPrice * gasLimit`. We've tested this out in non-zero gas networks, and it works fine 👍 

cc @Kourin1996

contracts/staking/query.go
zivkovicmilos: The `results` map will always be of type `map[string]interface{}`, with the first field being the actual result.

If the result wasn't populated, the call would error out.

@Kourin1996 please confirm

consensus/ibft/pos.go
zivkovicmilos: Removed 👍 


5bb4bc1

consensus/ibft/pos_test.go
zivkovicmilos: We use the `testify` package because it gives us a lot of flexibility that the regular `pkg.go.dev` couldn't provide.

If we were stuck using the regular golang package, we'd need to write all of our custom asserts - which `testify` provides for us since it is a wrapper over the standard go package.

helper/staking/staking.go
zivkovicmilos: Sadly no, the only way to inject this sort of value into the SDK is to have it hardcoded like this, since the Polygon SDK is shipped as-is.

One other approach would be to write it to disk, and read from there - but that just introduces an additional overhead every time we want to grab the bytecode.

helper/staking/staking.go
zivkovicmilos: Good catch 💯 

Fixed 👍 

93ecbc0

helper/staking/staking.go
zivkovicmilos: Fixed 👍 

4cfd106

command/genesis/genesis.go
zivkovicmilos: Good catch 💯 

Fixed 👍 

50113f7

command/genesis/genesis.go
zivkovicmilos: Fixed 👍 

090281b

command/helper/helper.go
zivkovicmilos: You're right, the wording isn't really clear.

Fixed 👍 

8d1b964

e2e/pos_test.go
zivkovicmilos: You're right - we can only check the total staked amount on the Staking SC.

The first check asserts that the account (address) balance is correct for the staking address.
The second check asserts that the `stakedAmount` value is set correctly on the contract itself.

These two values need to be the same

e2e/framework/helper.go
zivkovicmilos: Fixed 👍 

742a179

e2e/framework/helper.go
zivkovicmilos: Fixed 👍 

742a179

e2e/framework/helper.go
zivkovicmilos: Fixed 👍 

742a179

e2e/framework/helper.go
zivkovicmilos: Fixed 👍 

742a179

e2e/pos_test.go
zivkovicmilos: You're right 💯 

I've upped the validator count and made the test better for checking:

eb35d49

e2e/pos_test.go
zivkovicmilos: This test actually has 101 stakers:
https://github.com/0xPolygon/polygon-sdk/blob/eb35d49a64ab3c8d044e3333c09e68cab7a6443e/e2e/pos_test.go#L380

e2e/pos_test.go

28f5205

e2e/pos_test.go
zivkovicmilos: This test has `11` stakers:
https://github.com/0xPolygon/polygon-sdk/blob/eb35d49a64ab3c8d044e3333c09e68cab7a6443e/e2e/pos_test.go#L516

e2e/pos_test.go

8831ce9

contracts/staking/query.go
Kourin1996: @Vuksan @zivkovicmilos Yes, this transaction is for querying the contract about validators and it will be called directly after the node added the block where the height is the end of an epoch and not be applied to the state. For that reason, it doesn't affect to the order and account balance even though we set higher gas price.

contracts/staking/query.go
Kourin1996: @Vuksan @zivkovicmilos `results` has the result of the contract call and its type is decided by ABI. But now we're expecting it has an array of addresses at the first return value based on the ABI.

btw, `ok` will be `false` when `results["0"]` is nil. panic doesn't occur here.

helper/staking/staking.go
```diff
@@ -0,0 +1,176 @@
...
...0602082019050919050565b600081519050919050565b6000602082019050919050565b600082825260208201905092915050565b600082825260208201905092915050565b6000610ba682610c73565b9150610bb183610c73565b9250827fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff03821115610be657610be5610c7d565b5b828201905092915050565b6000610bfc82610c73565b9150610c0783610c73565b925082821015610c1a57610c19610c7d565b5b828203905092915050565b6000610c3082610c53565b9050919050565b60006fffffffffffffffffffffffffffffffff82169050919050565b600073ffffffffffffffffffffffffffffffffffffffff82169050919050565b6000819050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b600080fd5b7f4f6e6c79207374616b65722063616e2063616c6c2066756e6374696f6e000000600082015250565b7f696e646578206f7574206f662072616e67650000000000000000000000000000600082015250565b610d6a81610c73565b8114610d7557600080fd5b5056fea26469706673582212200732d127c069747c9fba40cde21178279bf118074d859232ff634cd4c0c2678964736f6c63430008070033"
```
0xAleksaOpacic: Change byteCode to match new contract which supports sending money directly to contract address for staking.

consensus/ibft/ibft.go
```diff
@@ -15,14 +17,18 @@ import (
...
 
 const (
 	DefaultEpochSize = 100000
 )
 
+var (
+	ErrInvalidHookParam = errors.New("invalid hook param")
+	ErrMissingHook      = errors.New("missing hook")
+)
+
```
Vuksan: Maybe we can be a bit more descriptive. Reading these logs I would think we're building some software for pirates. 🙂

contracts/staking/query.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if res.Failed() {
+		return nil, res.Err
+	}
+
+	extractedMap, extractErr := extractResponse(method, res.ReturnValue)
+	if extractErr != nil {
+		return nil, fmt.Errorf("unable to extract results, %v", extractErr)
```
Vuksan: Use `errors` instead of `fmt`?

contracts/staking/query.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, errors.New("failed type assertion from decodedResults to map")
+	}
+
+	return results, nil
+}
+
+func DecodeValidators(method *abi.Method, returnValue []byte) ([]types.Address, error) {
+	extractedMap, extractErr := extractResponse(method, returnValue)
+	if extractErr != nil {
+		return nil, fmt.Errorf("unable to extract results, %v", extractErr)
```
Vuksan: Use `errors` instead of `fmt`?

consensus/ibft/pos.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	// Log the info message
+	pos.ibft.logger.Info(
+		"current snapshot",
+		"validators",
+		len(snap.Set),
+		"total staked amount",
+		stakedAmount.String(),
+	)
```
Vuksan: It was my mistake to ask for this extra log during the [last review](https://github.com/0xPolygon/polygon-sdk/pull/219#discussion_r773954001).
I feel like it right now increases the computational cost by too much since we have to query the SC for this. Maybe revert it back to how it was?

@lazartravica I would like your opinion on this as well.

e2e/pos_test.go
```diff
@@ -0,0 +1,613 @@
...
+		expBalance.String(),
+		actualAccountBalance.String(),
+		"Account balance mismatch after stake / unstake exploit",
+	)
+}
+
+// Test scenario:
+// User has 0 ETH staked and a balance of 100 ETH
+// Stake 2 ETH -> Unstake
+// Expected result for tests: Staked: 0 ETH; Balance: ~100 ETH; not a staker
```
Vuksan: [Previous comment link.](https://github.com/0xPolygon/polygon-sdk/pull/219#discussion_r774586278)
I meant to add the actual check that the node is no longer a validator since he has unstaked his whole amount, not only a comment. 🙂 
I really meant validator, not staker.

Vuksan (changes requested):
Great job on resolving previous issues in separate commits. 🙏

contracts/staking/query.go
lazartravica: I'd say if you want to do any kind of string interpolation it's much more readable to use `Errorf` than `Sprintf` inside of an `Errors.New`

contracts/staking/query.go
lazartravica: I'd say if you want to do any kind of string interpolation it's much more readable to use Errorf than Sprintf inside of an Errors.New

consensus/ibft/pos.go
lazartravica: I completely agree, looking at this commit: https://github.com/0xPolygon/polygon-sdk/pull/219/commits/d67f9847d240fac77e83ac99c91fd026415e0632

Let's get rid of the `total staked amount`, this is the reason why we introduced epoch blocks.

Looking at it again, I completely understand the need for printing the number of validators, but the amount staked is nothing something that I see giving any value in this log message for the acceptState eitherway.

consensus/ibft/ibft.go
zivkovicmilos: Fixed 🏴‍☠️ 

74f0c78

contracts/staking/query.go
zivkovicmilos: I agree with @lazartravica, in the codebase we usually do 2 types of error manipulation:
* `errors.New` for errors that are regular string messages and don't need additional runtime data
* `fmt.Errorf` for errors that required additional runtime data relevant for the error

That's why I opted to use `fmt.Errorf`, since it has a cleaner interface than using `errors.New(fmt.Sprintf(...))`

contracts/staking/query.go
zivkovicmilos: I agree with @lazartravica, in the codebase we usually do 2 types of error manipulation:
* `errors.New` for errors that are regular string messages and don't need additional runtime data
* `fmt.Errorf` for errors that required additional runtime data relevant for the error

That's why I opted to use `fmt.Errorf`, since it has a cleaner interface than using `errors.New(fmt.Sprintf(...))`

consensus/ibft/pos.go
zivkovicmilos: I've revered the commit 👍 
It's comforting to know that we can add this kind of query functionality really quickly in future feature additions, as we have support for it through `go-web3`


731b9c8

e2e/pos_test.go
zivkovicmilos: 🤦 I didn't get it on the first read

Fixed, and added the requested check 👍 :

7f1e06d

Vuksan (approved):
🚀 🎉

helper/staking/staking.go
zivkovicmilos: Fixed 👍 

49a2f05

0xAleksaOpacic (approved):

PR #224, opened 2021-11-18
Fix txpool's GetNonce() to properly return nextnonce for addr within global pending txs

dankostiuk:
# Description

Ethers.js and Metamask use `this.getTransactionCount("pending")` to obtain the nonce to use for subsequent txs. When testing with the SDK, it appears that `eth_getTransactionCount` was giving us inconsistent values when querying across different nodes. 

It turns out the GetNonce() function appeared to be returning the nextNonce for the given address for the node's accountQueue instead of the global pendingQueue.

This PR correctly obtains the nextNonce based on the highest nonce of the address within the global pendingTxs queue, similar to how matic does it (see https://github.com/maticnetwork/bor/blob/0a9bd7befd47f78f6ddd9104c5508ecdda726452/internal/ethapi/api.go).

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -218,14 +218,20 @@ func (a *accountQueueWrapper) unlock() {
 
 // GetNonce returns the next nonce for the account, based on the txpool
 func (t *TxPool) GetNonce(addr types.Address) (uint64, bool) {
-	mux := t.lockAccountQueue(addr, false)
-	defer mux.unlock()
-
-	wrapper, ok := t.accountQueues[addr]
-	if !ok {
+	pendingTxs, _ := t.GetTxs(false)
+	accountTxs := pendingTxs[addr]
+	if len(accountTxs) == 0 {
 		return 0, false
 	}
-	return wrapper.accountQueue.nextNonce, true
+
+	highestNonce := uint64(0)
+	for k := range accountTxs {
+		if k > highestNonce {
+			highestNonce = k
+		}
+	}
+
+	return accountTxs[highestNonce].Nonce + 1, true
 }
 
 // NumAccountTxs Returns the number of transactions in the account specific queue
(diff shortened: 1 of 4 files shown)


txpool/txpool.go
```diff
@@ -373,8 +379,8 @@ func (t *TxPool) DecreaseAccountNonce(tx *types.Transaction) {
 	}
 }
 
-// GetTxs gets both pending and queued transactions
-func (t *TxPool) GetTxs() (map[types.Address]map[uint64]*types.Transaction, map[types.Address]map[uint64]*types.Transaction) {
+// GetTxs gets pending and queued transactions
+func (t *TxPool) GetTxs(inclQueued bool) (map[types.Address]map[uint64]*types.Transaction, map[types.Address]map[uint64]*types.Transaction) {
 
 	pendingTxs := make(map[types.Address]map[uint64]*types.Transaction)
 	sortedPricedTxs := t.pendingQueue.index
```
zivkovicmilos, line 386: I don't believe this action is thread safe, since you're directly fetching a reference to a protected map and later iterating over it

zivkovicmilos (changes requested):
Hey @dankostiuk,

Thank you for raising a PR on this issue 🙏 

I've left a comment concerning the thread-safe semantics of grabbing the txPriceHeap index field.

Otherwise, the PR looks good 💯

txpool/txpool.go
dankostiuk: Thanks @zivkovicmilos , fixed - let me know if that's what you had in mind, thanks

zivkovicmilos: I would just add the `t.pendingQueue.lock.Unlock()` call to after you finish ranging over `sortedPricedTxs` (after L#394), since that is the point where you don't need to hold the lock on the pending queue index anymore

zivkovicmilos (approved):
Looks good to me 💯 

Thank you for making the requested lock changes 🙏

Kourin1996 (approved):
LGTM 👏 Thank you for the PR

lazartravica (approved):
LGTM, thank you for the contribution

munna0908:
@dankostiuk I feel the reason for mismatch in pending nonce is because the gossiped zero gas transactions are rejected by the node due to wrong configuration of txpool price limit.(Which is fixed in this [commit](https://github.com/0xPolygon/polygon-sdk/commit/5fe3259c0d5c4b6ca505287f050d25cb38a92219))And your current modifications will still not fix the issue since the count is still based on the transactions present in the pendingQueue

PR #229, opened 2021-11-19
Apply correct error on insufficient funds + gas

mrwillis:
# Description

In `apply` in `executor.go`, the function `txn.GetBalance` and comparison `gasLeft > msg.Gas` doesn't return an error and just inherits the `err` from the higher scope.  If a txn failed to be applied because there was insufficient balance, it just returns a not very useful error. This PR throws the correct errors manually 

```
2021-11-19T14:07:56.076-0500 [ERROR] polygon: failed to apply tx: err="<nil>, recoverable [true]"
```

New errors look like

```
2021-11-19T14:31:00.026-0500 [ERROR] polygon: failed to apply tx: err="not enough funds for transfer with given value, recoverable [true]"
```

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments


zivkovicmilos (approved):
Hey @mrwillis,

Thank you for opening up this PR and for fixing the issue 🙏 

Looks great to me 💯

Kourin1996 (approved):

mrwillis:
@zivkovicmilos would you like me to add tests for this?

zivkovicmilos:
That would be awesome @mrwillis 💯 

Anything apart from this base PR is considered a bonus 🙂

PR #222, opened 2021-11-18
Add flag for picking the ibft consensus mechanism [PoS / PoA]

zivkovicmilos:
# Description

This PR adds support for picking out the type of IBFT mechanism when generating the genesis file.

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [ ] I have updated the README and other relevant documents (guides...)
- [x] I have added sufficient documentation both in code, as well as in the READMEs

# Additional comments

The documentation should be updated to include this flag as soon as the `ibft-combined-modes` repo PR goes through.

PR #234, opened 2021-11-22
Implement PoA / PoS stubs

zivkovicmilos:
# Description

This PR implements the hooks introduced in PR #221.

Awaiting PR #222 before being ready for review.
Branched off from the PR #221 branch working branch, so it will show changes from that branch until it is merged into `ibft-combined-modes`.

# Changes include

# Checklist

- [x] I have tested this code


command/ibft/ibft_propose.go
```diff
@@ -0,0 +1,139 @@
...
+	"context"
+	"fmt"
+
+	"github.com/0xPolygon/polygon-sdk/command/helper"
+	"github.com/0xPolygon/polygon-sdk/consensus/ibft/proto"
+	ibftOp "github.com/0xPolygon/polygon-sdk/consensus/ibft/proto"
+	"github.com/0xPolygon/polygon-sdk/types"
+)
+
+// IbftPropose is the command to query the snapshot
```
brkomir: same as in the ibft_candidates.go

command/ibft/ibft_candidates.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/0xPolygon/polygon-sdk/command/helper"
+	ibftOp "github.com/0xPolygon/polygon-sdk/consensus/ibft/proto"
+	empty "google.golang.org/protobuf/types/known/emptypb"
+)
+
+// IbftCandidates is the command to query the snapshot
```
brkomir: I suppose this description isn't correct

command/ibft/ibft_candidates.go
```diff
@@ -0,0 +1,94 @@
...
+// GetHelperText returns a simple description of the command
+func (p *IbftCandidates) GetHelperText() string {
+	return "Queries the current set of proposed candidates, as well as candidates that have not been included yet"
+}
+
+func (p *IbftCandidates) GetBaseCommand() string {
+	return "ibft candidates"
+}
+
+// Help implements the cli.IbftCandidates interface
```
brkomir, line 27: This comment seems out of context. Same for the other comment references to cli.IbftCandidates interface in the file

command/ibft/ibft_candidates.go
```diff
@@ -0,0 +1,94 @@
...
+	return helper.GenerateHelp(p.Synopsis(), helper.GenerateUsage(p.GetBaseCommand(), p.FlagMap), p.FlagMap)
+}
+
+// Synopsis implements the cli.IbftCandidates interface
+func (p *IbftCandidates) Synopsis() string {
+	return p.GetHelperText()
+}
+
+// Run implements the cli.IbftCandidates interface
+func (p *IbftCandidates) Run(args []string) int {
```
brkomir: I'm not sure if receiver name "p" is good in this context.

command/ibft/ibft_propose.go
```diff
@@ -0,0 +1,139 @@
...
+// GetHelperText returns a simple description of the command
+func (p *IbftPropose) GetHelperText() string {
+	return "Proposes a new candidate to be added or removed from the validator set"
+}
+
+func (p *IbftPropose) GetBaseCommand() string {
+	return "ibft propose"
+}
+
+// Help implements the cli.IbftPropose interface
```
brkomir: cli.IbftPropose interface, same as in the ibft_candidates.go

consensus/ibft/ibft.go
```diff
@@ -467,7 +571,10 @@ func (i *Ibft) runAcceptState() { // start new round
 		return
 	}
 
-	i.logger.Info("current snapshot", "validators", len(snap.Set))
+	if hookErr := i.runHook(AcceptStateLogHook, snap); hookErr != nil {
+		i.logger.Error(fmt.Sprintf("Unable to run hook %s, %v", AcceptStateLogHook, hookErr))
+	}
+	//i.logger.Info("current snapshot", "validators", len(snap.Set), "votes", len(snap.Votes))
```
brkomir: should this be uncommented?

brkomir (approved):
I have left some minor comments, but overall looks good, GJ!
This hook mechanism looks interesting :)
I'm just curious, why did you opt out for it instead of just declaring the methods in the mechanism interface and implementing them in poa and pos (something like a strategy pattern)? My opinion is that we could avoid those type casts in this case, and maybe have a little more simplicity in this code.

command/ibft/ibft_propose.go
zivkovicmilos: I'm not sure what you meant by this?

command/ibft/ibft_candidates.go
zivkovicmilos: These are just generic command descriptions we have for each method commands need to implement

consensus/ibft/ibft.go
zivkovicmilos: 🤦 

Fixed 👍 

fef1864

command/ibft/ibft_candidates.go
zivkovicmilos: Fixed 👍 

fef1864

command/ibft/ibft_candidates.go
zivkovicmilos: Fixed 👍 

fef1864

command/ibft/ibft_propose.go
zivkovicmilos: Fixed 👍 

fef1864

zivkovicmilos:
@brkomir 

There is no simplicity to be gained with interface methods (strategy pattern).

The reason I went with the hook approach is that we don't need to define additional methods that don't do anything or just return dummy values (like with the strategy pattern).

You call a hook, and it's either present and it executes or nothing happens. This way, you implement only the methods your specific mechanism needs, instead of defining methods from PoA in PoS code and making them do nothing

command/ibft/ibft_propose.go
brkomir: me neither now :)

brkomir:
@zivkovicmilos 
When I said simplicity I meant in terms of following:

- no need for typecasting meaning methods can have clearly defined types for parameters and return values
- there is no need for a mechanism for registering methods in a map
- it is ok for interface methods to return and do nothing, since "do nothing" is a specific implementation in that sense. This is also what you achieved with hooks, just in a different way. The problem can be if the implementation of the interface has many "do nothing" methods which suggest that your interface is too much generalised for its purpose and probably needs to be refactored/segregated.

Your implementation, as every other implementation, has pros and cons. With hooks you gain more flexibility in terms of registering methods which are not mandatory, but you loose on type safety and the caller must know what exactly the what method parameters types are in order to call it properly, so in that sense the decoupling is "broken" if you get my point. I'm not saying that your approach isn't good, just pointing out what we gain/loose with it.

brkomir (approved):

Kourin1996 (approved):

lazartravica (approved):

zivkovicmilos:
@brkomir 

You're completely right.

The only reason we went with the hook approach was to minimize the refactor efforts we'd have later down the line when we need to modularize this package, or even adapt to a separate 3rd party package outside the SDK for consensus needs.

PR #237, opened 2021-11-24
Invalid txn stops build block

brkomir:
# Description

I have addressed the issue where any TransitionApplicationError from applying the txn causes the block to be prematurely finished 

# Changes include

# Checklist

- [x] I have tested this code


consensus/ibft/ibft.go
```diff
@@ -448,10 +432,52 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 	// is sealed after all the committed seals
 	block.Header.ComputeHash()
 
-	i.logger.Info("build block", "number", header.Number, "txns", len(txns))
+	i.logger.Info("build block", "number", header.Number, "txPoolLength", len(txns))
```
Kourin1996: What does `txPoolLength` means? len(txns) semms to be number of transactions in the block

consensus/ibft/ibft.go
```diff
@@ -448,10 +432,52 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
...
+		}
+
+		if txn.ExceedsBlockGasLimit(gasLimit) {
+			i.logger.Error(fmt.Sprintf("failed to write transaction: %v", state.ErrBlockLimitExceeded))
+			i.txpool.DecreaseAccountNonce(txn)
+			continue
+		}
+
+		if err := transition.Write(txn); err != nil {
+			if appErr, ok := err.(*state.GasLimitReachedTransitionApplicationError); ok && appErr.IsRecoverable {
```
Kourin1996: Wondering whether we should check appErr.IsRecoverable is true because it's IsRecoverable 
 in case of GasLimitReachedTransitionApplicationError

Kourin1996 (approved):
Hi, looks ok for me 😊
I've left small comments. It'd be great if you could check them. Thank you!

consensus/ibft/ibft.go
brkomir: thats an unintended change, probably due to using rename refactoring in goland, will revert this

consensus/ibft/ibft.go
brkomir: I also think this is not necessary, i was intending to remove this but obviously forgot. Tnx for the notice

brkomir: [changed](https://github.com/0xPolygon/polygon-sdk/pull/237/commits/ecdc987abb40a81960d8dbbf35b60fe59ef78c53)

consensus/ibft/ibft.go
brkomir: [reverted](https://github.com/0xPolygon/polygon-sdk/pull/237/commits/ecdc987abb40a81960d8dbbf35b60fe59ef78c53)

consensus/ibft/ibft.go
```diff
@@ -452,6 +436,48 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
...
+			}
+			continue
+		}
+
+		txns = append(txns, txn)
+	}
+
+	for _, retFunc := range returnTxnFuncs {
+		retFunc()
+	}
```
zivkovicmilos, line 479: I think it's clever that we run the callbacks later. Is there any side effect to having them run here, and not right after the txn fails?

I would love to have more in code documentation for this entire method 🙂

state/executor.go
```diff
@@ -355,6 +355,16 @@ func NewTransitionApplicationError(err error, isRecoverable bool) *TransitionApp
 	}
 }
 
+type GasLimitReachedTransitionApplicationError struct {
+	TransitionApplicationError
+}
```
zivkovicmilos, line 360: Is there any specific reason why we wrap this error type?

zivkovicmilos (approved):
Everything looks good to me 💯 

I've left some minor comments to better understand the implementation.

state/executor.go
dbrajovic: I think it's because he's trying pass that specific error up [WriteTransactions()](https://github.com/0xPolygon/polygon-sdk/blob/121570edc4ebc88532e14459886fc14f96b98d4e/consensus/ibft/ibft.go#L443) (previously in `buildBlock()`). 

I guess that's a bit problematic to achieve due to the current implementation of `TransitionApplicationError`.

dbrajovic (approved):

consensus/ibft/ibft.go
brkomir: We run them here and not immediately after the tnx fails because we don't want to return the txn to the pool and pop it again in the next iteration.

state/executor.go
brkomir: I didn't want to add additional field in the TransitionApplicationError so I created this error type. The problem was to determine when it is the time to break from the loop, and the moment we chose was when the next transaction can't fit in the block due to the block gas limit being reached.

consensus/ibft/ibft.go
brkomir: [comments added](https://github.com/0xPolygon/polygon-sdk/pull/237/commits/ce819d1981811f3daae949691081ffdb827d48d8)

PR #238, opened 2021-11-24
Add further documentation to IBFT spec functions

### LINKED ISSUE #236: Discussion: `MaxFaultyNodes` and `NumValid` wrong math?
### Description

If I'm reading the spec correctly from https://github.com/ethereum/EIPs/issues/650, the system can tolerate AT MOST F faulty nodes in a N validator nodes network, where `N = 3F + 1` implies `F = (N - 1)/3`.

The function looks like this 

```go
// MaxFaultyNodes returns the maximum number of allowed faulty nodes, based on the current validator set
func (v *ValidatorSet) MaxFaultyNodes() int {
	// numberOfValidators / 3
	return int(math.Ceil(float64(len(*v))/3)) - 1
}
```

Shouldn't this be 

```go
// MaxFaultyNodes returns the maximum number of allowed faulty nodes, based on the current validator set
func (v *ValidatorSet) MaxFaultyNodes() int {
	// numberOfValidators / 3
	return int(math.Ceil(float64((len(*v)-1)/3)))
}
```

I am not sure if this should be a `math.Ceil` or `math.Floor`. I think intuitively it should be a ceiling. 

Moreover this function

https://github.com/0xPolygon/polygon-sdk/blob/7f2e61d19b63bde38f52925dee4eefc47e03ddf5/consensus/ibft/state.go#L101

is `2F`, but according to the spec it should be `2F + 1`. Eventhough this looks wrong, it ends up being okay here because of the greater than https://github.com/0xPolygon/polygon-sdk/blob/7f2e61d19b63bde38f52925dee4eefc47e03ddf5/consensus/ibft/ibft.go#L627 and here https://github.com/0xPolygon/polygon-sdk/blob/7f2e61d19b63bde38f52925dee4eefc47e03ddf5/consensus/ibft/ibft.go#L632

but it looks incorrect here. 

https://github.com/0xPolygon/polygon-sdk/blob/7f2e61d19b63bde38f52925dee4eefc47e03ddf5/consensus/ibft/ibft.go#L784

Believe it should be greater than, or just change the `NumValid` to be consistent with the spec. 

### Your environment

* OS and version Ubuntu 20
* version of the Polygon SDK 

### Steps to reproduce

zivkovicmilos:
# Description

This PR adds addition explanation to how the maximum number of faulty nodes is calculated.

Closes issue #236.

# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (approved):
Nice supplements 👏

consensus/ibft/state.go
```diff
@@ -282,8 +286,17 @@ func (v *ValidatorSet) Includes(addr types.Address) bool {
...
+	// F -> number of faulty nodes
+	//
+	// N = 3F + 1
+	// => F = (N - 1) / 3
+	//
+	// IBFT tolerates 1 failure with 4 nodes
+	// 4 = 3 * 1 + 1
+	// To tolerate 2 failures, IBFT requires 7 nodes
+	// 7 = 3 * 2 + 1
+	return int(math.Floor(float64(len(*v)-1) / 3))
```
brkomir: This can be reduced to `(len(arr)-1) / 3`.

brkomir (approved):
LGTM

lazartravica (approved):

consensus/ibft/state.go
zivkovicmilos: Yes, it's simpler, but for a length of 0 it doesn't bind it to 0 (it goes negative):
https://goplay.tools/snippet/R3en9z64S7F

zivkovicmilos: I just noticed I've swapped the outputs 🤦 

Fixed 👍 

2df2d26

brkomir: GG :)

Issue #231, opened 2021-11-19
`addImpl` in txpool.go incorrectly reports that the pool is full, even if empty, due to overflow


mrwillis:
### Description
`addImpl` in txpool.go incorrectly reports that the pool is full, even if empty, causing broadcast failures. This causes downstream txn delays.

```
2021-11-19T16:22:26.745-0500 [ERROR] polygon.txpool: failed to add broadcasted txn: err="txpool is full"
```

It looks like the `txpool.slot` variable seems to have overflowed (see logs below). So this is two potentially related issues:

1. Node is incorrectly reporting txpool is full as a result of below
2. `txpool.slot` is overflowing

### Your environment

* OS and version Ubuntu 20
* version of the Polygon SDK 6efebc6a92f885082620ed2a8d59e37d4d27074c
* branch that causes this issue `develop`

### Steps to reproduce

* Tell us how to reproduce this issue <br />
Start a cluster with 4 nodes. Send a basic transaction (send 10 ETH) to the fourth node. Observe that on the other nodes, in `addImpl`, an error will be reported that the txpool is full.
* Where the issue is, if you know <br />

https://github.com/0xPolygon/polygon-sdk/blob/develop/txpool/txpool.go#L589 is an overflow if `t.slots = 0`

* Which commands triggered the issue, if any

### Expected behaviour

Node should accept valid gossiped non-local transaction

### Actual behaviour

Node rejects valid gossiped non-local transaction

### Logs

![image](https://user-images.githubusercontent.com/2652683/142695094-9603c7c9-d6f4-4b4f-ac63-f9fbff94619f.png)

![image](https://user-images.githubusercontent.com/2652683/142695674-16bf42f2-e585-405d-a319-8637224645f4.png)

![image](https://user-images.githubusercontent.com/2652683/142700866-c1f54e11-1125-4400-9b8a-224f3024d04c.png)

### Proposed solution

It looks like the discard function does not correctly consider the case with remoteTxns length = 0

mrwillis:
Seems related to https://github.com/0xPolygon/polygon-sdk/pull/213. I can close if that will solve this issue.

dbrajovic:
Hey @mrwillis ,

thank you for raising the issue. The bug has been identified and a fix under way.

PR #240, opened 2021-11-25
[TxPool] Fix incorrect call to decreaseSlots() when no tx was deleted

dbrajovic:
# Description

This PR addresses the issue [#231](https://github.com/0xPolygon/polygon-sdk/issues/231). The bug was occurring specifically due to a second call to `decreaseSlots()` (during `ResetWithHeader()`) for a transaction that was popped previously, thus resulting in an underflow of `t.slots`.

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments

Essentially, the underflow of `t.slots` happens exclusively when a node is a proposer - here’s the flow:
Proposer running in `AcceptState` calls `buildBlock()` , so the first `Pop()` of the transaction happens - this `Pop()` is what calls the `decreaseSlots()`.
Afterwards, proposer moves into `ValidateState` where it reaches `insertBlock()`. Inside, a call to TxPool's `ResetWithHeader()`  is made which leads to clearing out the mined transactions.
This is where the second call to `decreaseSlots()` happens, but the transaction was already gone. `t.pendingQueue.Delete(tx)` doesn’t do anything at this point (since the tx is gone),  but the decrease in slots happens anyhow resulting in the underflow.


txpool/txpool.go
```diff
@@ -473,9 +485,11 @@ func (t *TxPool) ProcessEvent(evnt *blockchain.Event) {
...
 	// remove the mined transactions from the pendingQueue list
 	for _, txn := range delTxns {
-		t.decreaseSlots(numSlots(txn))
-		t.pendingQueue.Delete(txn)
-		t.remoteTxns.Delete(txn)
+		deletedTx := t.pendingQueue.Delete(txn)
+		if deletedTx != nil {
+			t.gauge.decrease(slotsRequired(txn))
+			t.remoteTxns.Delete(txn)
+		}
```
zivkovicmilos:
```suggestion
		if deletedTx := t.pendingQueue.Delete(txn); deletedTx != nil {		        
			t.gauge.decrease(slotsRequired(txn))
			t.remoteTxns.Delete(txn)
		}
```

Just a minor nitpick 🙂

zivkovicmilos (approved):
Looks great 💯 

Let's fix this pesky issue 🚀

PR #251, opened 2021-11-26
Fix incorrect call to decreaseSlots() when no tx was deleted

### LINKED ISSUE #231: `addImpl` in txpool.go incorrectly reports that the pool is full, even if empty, due to overflow

dbrajovic:
# Description

This PR addresses the issue [#231](https://github.com/0xPolygon/polygon-sdk/issues/231). The bug was occurring specifically due to a second call to decreaseSlots() (during ResetWithHeader()) for a transaction that was popped previously, thus resulting in an underflow of t.slots.

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments

Essentially, the underflow of `t.slots` happens exclusively when a node is a proposer - here’s the flow:

Proposer running in `AcceptState` calls `buildBlock()`, so the first `Pop()` of the transaction happens implicitly calling `decreaseSlots(0`.
Afterwards, proposer moves into `ValidateState` where it reaches `insertBlock()`. Inside, a call to TxPool's `ResetWithHeader()` is made which leads to clearing out the mined transactions.
This is where the second call to `Pop()` (`decreaseSlots()`) happens, but the transaction was already gone. `t.pendingQueue.Delete(tx)` doesn’t do delete at this point (since the tx was gone already), but the decrease in slots happens anyhow resulting in the underflow.


zivkovicmilos (approved):
Lookin' good 💯

txpool/txpool.go
```diff
@@ -841,14 +842,17 @@ func (t *txPriceHeap) Length() uint64 {
 	return uint64(len(t.index))
 }
 
-func (t *txPriceHeap) Delete(tx *types.Transaction) {
+func (t *txPriceHeap) Delete(tx *types.Transaction) *types.Transaction {
```
brkomir: Maybe we should return a boolean value here instead of transaction, so we can do if statement with an "ok" check

brkomir (approved):
I have left a comment, nothing major. Overall looks ok.
Have you tried to add some tests for this?

Kourin1996 (approved):
Looks good apart from @brkomir's comment

txpool/txpool.go
dbrajovic: [Changed](https://github.com/0xPolygon/polygon-sdk/pull/251/commits/5aa22e1de9f29b6cb1dded78f184a018649569c6)

txpool/txpool.go
```diff
@@ -841,14 +842,19 @@ func (t *txPriceHeap) Length() uint64 {
...
-func (t *txPriceHeap) Delete(tx *types.Transaction) {
+func (t *txPriceHeap) Delete(tx *types.Transaction) bool {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
-	if item, ok := t.index[tx.Hash]; ok {
-		heap.Remove(t.heap, item.index)
-		delete(t.index, tx.Hash)
+	item, ok := t.index[tx.Hash]
+	if !ok {
```
brkomir, line 850: you can simplify this additionally since you basically return value of the ok
if ok {
// do delete and heap.remove
}

return ok

brkomir (approved):

lazartravica (approved):

PR #252, opened 2021-11-26
Ensure eth_getLogs response includes non-0 transactionHash

dankostiuk:
# Description

It appears that the eth_getLogs endpoint has always returned a zero-address for the transactionHash within the json-rpc response body.

It seems the original bug lies in the `keyvalue.go#ReadReceipts()` which uses the RLPUnmarshaler to obtain the Receipt.TxHash but isn't able to and as a result, a default zero-address is assigned.

This PR uses a similar approach to what eth_getTransactionReceipt() does in getting each transactionHash from the block.Transactions[] -- I realize this solution makes for a heavier endpoint since we are considering the Transactions[] for each block within a start/end block range, but without being able to properly unmarshal the RLP, this endpoint will not be able to return a valid response.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code


zivkovicmilos (approved):
I don't mind the added overhead in `eth_getLogs`. As long as it's returning the information users expect I'm fine with it 👍 
This method is something we're currently considering to make quicker using better blooms.

Thank you as always for opening up a PR and contributing 🙏

brkomir (approved):

Kourin1996 (approved):

PR #254, opened 2021-11-26
Add namespace capability to HashiCorp Vault Secrets Manager

mrwillis:
# Add namespace capability to Vault Secrets Manager

In the hosted hashicorp and enterprise version, namespaces are used by default, and the default is "admin". This was not set in the current code causing a lot of head-scratching (access denied, etc). 

It looks like the namespaces are supported in the OS version as well, so this is not a breaking change. 

This PR adds support for the namespace field and sets it on the vault client. 

If this looks okay, I can update the docs as well. 

# Changes include

# Breaking changes

# Checklist

- [ ] I have tested this code

### DIFF
--- a/command/secrets/secrets_generate.go
+++ b/command/secrets/secrets_generate.go
@@ -16,6 +16,7 @@ type SecretsGenerate struct {
 const (
 	defaultNodeName       = "polygon-sdk-node"
 	defaultConfigFileName = "./secretsManagerConfig.json"
+	defaultNamespace = "admin"
 )
 
 func (s *SecretsGenerate) DefineFlags() {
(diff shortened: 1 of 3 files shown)


zivkovicmilos (approved):
Hey @mrwillis,

Thank you for pointing it out - we've missed this feature in the initial PR 🙏 

As always, thank you for the contribution 💯 

I've opened up a corresponding docs PR:
https://github.com/0xPolygon/polygon-sdk-docs/pull/28

mrwillis:
@zivkovicmilos re getting assigned, do you want some tests for this? It doesn't look like the vault is stubbed in the tests, but I could test the abstraction

zivkovicmilos:
That would be awesome @mrwillis 💯 

Anything apart from this base PR is considered a bonus 🙂

mrwillis:
@zivkovicmilos I can't actually find where the abstraction or any of the underlying calls are tested. I see there used to be a test file for the vault but now there is only one for the local. 

https://github.com/0xPolygon/polygon-sdk/pull/209/commits/4e82daec055a122633544b50adfbff911aeac314

Kourin1996 (approved):
LGTM 👍

zivkovicmilos:
@mrwillis 
There were no mock / unit tests added for Vault, as we've tested all of it manually through a running instance.
One addition I can see being added is mocking the `SecretsManager` interface for Vault

brkomir (approved):

PR #256, opened 2021-11-27
Bug/txpool pruned txs dont decrease gauge

dbrajovic:
# Description

This PR fixes a bug that was appearing when sending transactions stale transactions (nonce is lower that what the account queue is expecting - nextNonce field). As a consequence, these low nonce transaction would be cleared during Promote(), but the gauge was not decreased.

# Changes include

# Checklist

- [x] I have tested this code


zivkovicmilos (approved):
Lookin' good 🚀

Kourin1996 (approved):

lazartravica (approved):

PR #257, opened 2021-11-27
Fix unknown account fetch

zivkovicmilos:
# Description

This PR adds support for fetching uninitialized (unknown) accounts in certain helper methods.
The helper methods would error out when failing to find the account in storage, which shouldn't have been the case.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

I've also formatted / removed useless prints from the `eth_endpoint.go` file.

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -449,7 +449,7 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 		return 0, err
 	}
 	acc, err := d.store.GetAccount(header.StateRoot, address)
-	if err != nil && errors.Is(err, ErrStateNotFound) {
+	if errors.As(err, &ErrStateNotFound) {
 		// If the account doesn't exist / isn't initialized,
 		// return a nonce value of 0
 		return 0, nil
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -3,6 +3,7 @@ package jsonrpc
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -448,9 +449,14 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 		return 0, err
 	}
 	acc, err := d.store.GetAccount(header.StateRoot, address)
-	if err != nil {
+	if err != nil && errors.Is(err, ErrStateNotFound) {
+		// If the account doesn't exist / isn't initialized,
+		// return a nonce value of 0
+		return 0, nil
+	} else if err != nil {
 		return 0, err
 	}
+
 	return acc.Nonce, nil
 }
 
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -129,6 +129,9 @@ func (e *Eth) GetTransactionByHash(hash types.Hash) (interface{}, error) {
 		}
 	}
 	// txn not found (this should not happen)
+	e.d.logger.Warn(
+		fmt.Sprintf("Transaction with hash [%s] not found", blockHash),
+	)
 	return nil, nil
 }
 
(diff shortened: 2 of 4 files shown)


ferranbt:
Good catch! It is a pity but much of these endpoints are not fully documented.

lazartravica (approved):

brkomir (commented):
Looks good overall :)
But i think we should add tests for this

jsonrpc/dispatcher.go
```diff
@@ -448,9 +449,14 @@ func (d *Dispatcher) getNextNonce(address types.Address, number BlockNumber) (ui
 		return 0, err
 	}
 	acc, err := d.store.GetAccount(header.StateRoot, address)
-	if err != nil {
+	if err != nil && errors.Is(err, ErrStateNotFound) {
```
Kourin1996: You don't need to do nil-check to avoid passing nil to errors.Nil

```suggestion
	if errors.Is(err, ErrStateNotFound) {
```

jsonrpc/eth_endpoint.go
```diff
@@ -330,12 +345,21 @@ func (e *Eth) EstimateGas(
 	// If the sender address is present, recalculate the ceiling to his balance
 	if transaction.From != types.ZeroAddress && transaction.GasPrice != nil && gasPriceInt.BitLen() != 0 {
 		// Get the account balance
+
+		// If the account is not initialized yet in state,
+		// assume it's an empty account
+		accountBalance := big.NewInt(0)
 		acc, err := e.d.store.GetAccount(header.StateRoot, transaction.From)
-		if err != nil {
+		if err != nil && !errors.Is(err, ErrStateNotFound) {
```
Kourin1996:
```suggestion
		if !errors.Is(err, ErrStateNotFound) {
```

jsonrpc/eth_endpoint.go
```diff
@@ -498,17 +524,21 @@ func (e *Eth) GetBalance(address types.Address, number *BlockNumber) (interface{
 	}
 
 	acc, err := e.d.store.GetAccount(header.StateRoot, address)
-	if err != nil {
+	if err != nil && errors.Is(err, ErrStateNotFound) {
```
Kourin1996:
```suggestion
	if errors.Is(err, ErrStateNotFound) {
```

jsonrpc/eth_endpoint.go
```diff
@@ -534,12 +563,16 @@ func (e *Eth) GetCode(address types.Address, number *BlockNumber) (interface{},
 		return nil, err
 	}
 
+	emptySlice := []byte{}
 	acc, err := e.d.store.GetAccount(header.StateRoot, address)
-	if err != nil {
+	if err != nil && errors.Is(err, ErrStateNotFound) {
```
Kourin1996:
```suggestion
	if errors.Is(err, ErrStateNotFound) {
```

Kourin1996 (approved):
LGTM 👍 
I've left small comments. Please check them. Thank you!

jsonrpc/dispatcher.go
zivkovicmilos: Fixed 👍 

33e84c7

jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍 

33e84c7

jsonrpc/eth_endpoint.go
zivkovicmilos: Fixed 👍 

33e84c7

zivkovicmilos:
@brkomir 

Added coverage tests 👍 

33e84c7

jsonrpc/eth_endpoint.go
zivkovicmilos: No need to omit the `err != nil` check here, as we need to check that in fact the error is `nil` later:

https://github.com/0xPolygon/polygon-sdk/blob/33e84c77e253db62ef35f12f49e0fa9c385db252/jsonrpc/eth_endpoint.go#L356-L360

Kourin1996 (approved):

PR #260, opened 2021-11-30
Fix/bootnode reconnection

munna0908:
# Description

This PR enforces PSDK users to specify at-least 2 bootnodes during the genesis configuration.

# Changes include

# Breaking Change

As part of this PR, we have added a new constraint to specify at-least two bootnodes while setting up the genesis condition and while starting the server. Due to this change PSDK network can't be started with a single bootnode.

# Checklist

- [x] I have tested this code

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -25,7 +25,7 @@ type Chain struct {
 	Name      string   `json:"name"`
 	Genesis   *Genesis `json:"genesis"`
 	Params    *Params  `json:"params"`
-	Bootnodes []string `json:"bootnodes"`
+	Bootnodes []string `json:"bootnodes,omitempty"`
 }
 
 // Genesis specifies the header fields, state of a genesis block
--- a/helper/flags/flags.go
+++ b/helper/flags/flags.go
@@ -24,17 +24,21 @@ func (i *ArrayFlags) Set(value string) error {
 	return nil
 }
 
-type BootnodeFlags []string
+type BootnodeFlags struct {
+	IsSet bool
+	Addrs []string
+}
 
 func (i *BootnodeFlags) String() string {
-	return formatArrayForOutput(*i)
+	return formatArrayForOutput(i.Addrs)
 }
 
 func (i *BootnodeFlags) Set(value string) error {
+	i.IsSet = true
 	if _, err := multiaddr.NewMultiaddr(value); err != nil {
 		return err
 	}
-	*i = append(*i, value)
+	i.Addrs = append(i.Addrs, value)
 	return nil
 }
 
--- a/helper/flags/flags.go
+++ b/helper/flags/flags.go
@@ -25,8 +25,8 @@ func (i *ArrayFlags) Set(value string) error {
 }
 
 type BootnodeFlags struct {
-	IsSet bool
-	Addrs []string
+	AreSet bool
+	Addrs  []string
 }
 
 func (i *BootnodeFlags) String() string {
@@ -34,7 +34,7 @@ func (i *BootnodeFlags) String() string {
 }
 
 func (i *BootnodeFlags) Set(value string) error {
-	i.IsSet = true
+	i.AreSet = true
 	if _, err := multiaddr.NewMultiaddr(value); err != nil {
 		return err
 	}
--- a/network/server.go
+++ b/network/server.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -181,6 +182,11 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 	logger.Info("LibP2P server running", "addr", AddrInfoToString(srv.AddrInfo()))
 
 	if !config.NoDiscover {
+
+		if config.Chain.Bootnodes != nil && len(config.Chain.Bootnodes) < 2 {
+			return nil, errors.New("Minimum two bootnodes are required")
+		}
+
 		// start discovery
 		srv.discovery = &discovery{srv: srv}
 		srv.discovery.setup()
--- a/network/server.go
+++ b/network/server.go
@@ -30,6 +30,9 @@ const DefaultLibp2pPort int = 1478
 
 const MinimumPeerConnections int64 = 1
 
+// MinimumBootNodes Count is set to 2 so that, a bootnode can reconnect to the network using other bootnode after restarting.
+const MinimumBootNodes int = 2
+
 type Config struct {
 	NoDiscover     bool
 	Addr           *net.TCPAddr
@@ -183,7 +186,7 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 
 	if !config.NoDiscover {
 
-		if config.Chain.Bootnodes != nil && len(config.Chain.Bootnodes) < 2 {
+		if config.Chain.Bootnodes != nil && len(config.Chain.Bootnodes) < MinimumBootNodes {
 			return nil, errors.New("Minimum two bootnodes are required")
 		}
 
--- a/network/server_test.go
+++ b/network/server_test.go
@@ -346,12 +346,14 @@ func TestNat(t *testing.T) {
 	})
 }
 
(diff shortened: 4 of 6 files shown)


chain/chain.go
```diff
@@ -25,7 +25,7 @@ type Chain struct {
 	Name      string   `json:"name"`
 	Genesis   *Genesis `json:"genesis"`
 	Params    *Params  `json:"params"`
-	Bootnodes []string `json:"bootnodes"`
+	Bootnodes []string `json:"bootnodes,omitempty"`
```
zivkovicmilos, line 28: Why do you need the `omitempty` here if this field can never be unpopulated?

network/server_test.go
```diff
@@ -363,40 +365,51 @@ func WaitUntilPeerConnectsTo(ctx context.Context, srv *Server, id peer.ID) (bool
...
 func TestPeerReconnection(t *testing.T) {
 	conf := func(c *Config) {
-		c.MaxPeers = 2
+		c.MaxPeers = 3
 		c.NoDiscover = false
 	}
 	//Create bootnode
 	bootNode := CreateServer(t, conf)
-
+	bootNode2 := CreateServer(t, conf)
```
zivkovicmilos: Please consider not naming variables by prepending a number to them.

This practice is present for some reason in other tests as well, but we should look to move away from it

helper/flags/flags.go
```diff
@@ -24,17 +24,21 @@ func (i *ArrayFlags) Set(value string) error {
 	return nil
 }
 
-type BootnodeFlags []string
+type BootnodeFlags struct {
+	IsSet bool
```
zivkovicmilos:
```suggestion
	AreSet bool
```

Since this bool is in regards to the bootnode flags (multiple) it should reference them in plural form

network/server.go
```diff
@@ -181,6 +182,11 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 	logger.Info("LibP2P server running", "addr", AddrInfoToString(srv.AddrInfo()))
 
 	if !config.NoDiscover {
+
+		if config.Chain.Bootnodes != nil && len(config.Chain.Bootnodes) < 2 {
```
zivkovicmilos: This is a breaking change - please check the PR box that this is a breaking change, and in the breaking changes section explain why this will cause clusters with 1 bootnode to halt

network/server_test.go
```diff
@@ -491,3 +499,43 @@ func TestSelfConnection_WithBootNodes(t *testing.T) {
 		})
 	}
 }
+
+func TestMinimumBootNodeCount(t *testing.T) {
+	tests := []struct {
+		name      string
+		bootNodes []string
+		isError   bool
```
zivkovicmilos:
```suggestion
		shouldFail bool
```

We use the flags `shouldFail` and `shouldSucceed` in all of the other tests, so there is no reason to change it up here

network/server_test.go
```diff
@@ -491,3 +499,43 @@ func TestSelfConnection_WithBootNodes(t *testing.T) {
...
+	}{
+		{
+			name:      "Server config with empty bootnodes",
+			bootNodes: []string{},
+			isError:   true,
+		},
+		{
+			name:      "Server config with less than two bootnodes",
+			bootNodes: []string{"/ip4/127.0.0.1/tcp/10001/p2p/16Uiu2HAmJxxH1tScDX2rLGSU9exnuvZKNM9SoK3v315azp68DLPW"},
```
zivkovicmilos: Please consider not using hard-coded values for these peerIDs, instead generate them during the test.

There's no reason to hard-code them

zivkovicmilos (changes requested):
There are a couple of things that need to be changed, but otherwise it looks good 👍 

One of the biggest changes that I've requested is changing the status of this PR to be a breaking change, since it will render networks with a single bootnode useless

network/server_test.go
```diff
@@ -363,40 +365,51 @@ func WaitUntilPeerConnectsTo(ctx context.Context, srv *Server, id peer.ID) (bool
...
-	srv1.Disconnect(bootNode.AddrInfo().ID, "bye")
 
-	assert.True(t, <-disconnectedCh1, "Failed to recieved peer disconnected event")
+	//disconnect from the first boot node
+	disconnectedCh1 := asyncWaitForEvent(srv1, 15*time.Second, disconnectedPeerHandler(bootNode.AddrInfo().ID))
+	srv1.Disconnect(bootNode.AddrInfo().ID, "Bye")
 
-	//disconnect from the second node
-	disconnectedCh2 := asyncWaitForEvent(srv1, 10*time.Second, disconnectedPeerHandler(srv2.AddrInfo().ID))
+	assert.True(t, <-disconnectedCh1, "Failed to recieve peer disconnected event")
```
zivkovicmilos: Typo here: `recieve` -> `receive`

The same typo is on line 403

chain/chain.go
munna0908: This field is unpopulated in some e2e test cases

network/server_test.go
munna0908: I felt this generation is unnecessary overhead, since this test scope is to verify whether the bootnode count constraints are being satisfied or not.

network/server_test.go
munna0908: Corrected it. Thanks for the catch. Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/260/commits/abff6809bcbc43529eff0a1b051caa06584c7c7e)

helper/flags/flags.go
munna0908: Renamed it , Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/260/commits/abff6809bcbc43529eff0a1b051caa06584c7c7e)

network/server_test.go
munna0908: Renamed them, Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/260/commits/25709ccd3d899ace25567a64f7edf7488a934531)

network/server.go
munna0908: Done

Kourin1996: Nitpick comment. Shall we define a constant for this magic number? And it's good you can write why you needed to set this value to 2.

```go
const (
    MinimumBootnodes = 2 // Why you set MinimumBootnodes to 2
)

...

		if config.Chain.Bootnodes != nil && len(config.Chain.Bootnodes) < MinimumBootnodes {
```

munna0908: Thanks for your suggestion @Kourin1996 , Here's is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/260/commits/afc4a6e9c20d683dd70fe56885ce552d37ca539b)

Kourin1996 (approved):

network/server_test.go
zivkovicmilos: I get that we're just trying to check if the size of the bootnodes array is correct.
It's just misleading to have hardcoded values - it gives the impression that these IDs are somehow relevant to the test.

The overhead of generating new keys is minimal to non-existent

chain/chain.go
zivkovicmilos: Why would we put `omitempty` because of that?

We should never modify the actual implementation to conform to some tests in this sense, but the other way around

munna0908:
@zivkovicmilos @lazartravica @Kourin1996 I would like to bring to your notice that in most of the cases, when we are trying to unmarshall the json data into go structs, we have not specified `omitempty` tags, because of this we unnecessarily create empty fields.
For Example: While Unmarshalling the chain information to Chain struct, we create empty bootnode field even when the bootnode list is empty in json file.

network/server_test.go
munna0908: That makes sense, Here's the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/260/commits/2db9caaecbc7e5e9186e16750333423a91670c54)

zivkovicmilos (approved):

Lookin' good 🚀

brkomir (approved):

lazartravica (approved):

PR #265, opened 2021-12-01
Resolve invalid nonce queries + race conditions in TxPool

zivkovicmilos:
# Description

This PR resolves some pending issues with the TxPool module:
* Race conditions when locking the account-specific queue
* Invalid next nonce read from world state when the TxPool is advanced (introduced in PR #224)
* Invalid TxPool account nonce when the world state has advanced
* Unpruned leftover transactions on block insertion

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

The PR is not meant to be a fix-all solution but rather a patch until we get up and running with the TxPool refactor.


txpool/txpool.go
```diff
@@ -471,60 +468,164 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
 		NewChain: []*types.Header{h},
 	}
+
+	// Process the txns in the event to make sure the TxPool is up-to-date
 	t.ProcessEvent(evnt)
 }
 
-// ProcessEvent processes the blockchain event and resets the txpool accordingly
+// ProcessEvent processes the blockchain event and updates the txpool accordingly
 func (t *TxPool) ProcessEvent(evnt *blockchain.Event) {
```
Kourin1996: Please note that you need to recover transactions from old block in case of other consensuses. You need to revert some transactions if reorg happens. but we don't have such consensus right now

txpool/txpool.go
```diff
@@ -471,60 +468,164 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
+		// but are submitted to the chain state
+		dropTxnCallback(txn)
+
+		// Attempt to update the next account nonce in the TxPool
+		// with the one in state if it's greater
+		wrapper := t.lockAccountQueue(txn.From, true)
+
+		// Grab the latest nonce from state
+		stateRoot := t.store.Header().StateRoot
+		stateNonce := t.store.GetNonce(stateRoot, txn.From)
```
Kourin1996: How long does it take time to call them? If the block has multiple transactions for one account, you would query for state nonce of the account multiple times. You might want to create nonce map first.

txpool/txpool.go
```diff
@@ -471,60 +468,164 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
+	// Remove the txns from the block that were just committed to state
+	// from any queues in the TxPool
+	for _, txn := range txnsInBlocks {
+		// Initially drop any txns that are leftover in the TxPool queues,
+		// but are submitted to the chain state
+		dropTxnCallback(txn)
+
+		// Attempt to update the next account nonce in the TxPool
+		// with the one in state if it's greater
+		wrapper := t.lockAccountQueue(txn.From, true)
```
Kourin1996: It's a similar one. We should avoid calling lock more than necessary.

txpool/txpool.go
```diff
@@ -471,60 +468,164 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
+					"Dropping txn [%s] due to a low nonce [%d < %d]",
+					lowestNonceTx.Hash.String(),
+					lowestNonceTx.Nonce,
+					accountNextNonce,
+				),
+			)
+			wrapper.accountQueue.txs.Pop()
+
+			txnDropCleanup(lowestNonceTx)
+		}
```
Kourin1996: Shall we create a method for this?

txpool/txpool.go
```diff
@@ -471,60 +468,164 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
+							pendingQueueTxn.tx.Hash.String(),
+						),
+					)
+
+					txnDropCleanup(pendingQueueTxn.tx)
+				}
+			}
+		}
+
+		t.pendingQueue.lock.Unlock()
```
Kourin1996: Shall we create a method for this? Are you deleting transactions with lower nonce from pending queue, right?

txpool/txpool.go
```diff
@@ -394,13 +394,10 @@ func (t *TxPool) addImpl(origin TxOrigin, tx *types.Transaction) error {
...
-	mux := t.lockAccountQueue(tx.From, true)
-	defer mux.unlock()
+	wrapper := t.lockAccountQueue(tx.From, true)
+	defer wrapper.unlock()
 
-	wrapper, ok := t.accountQueues[tx.From]
-	if ok {
-		wrapper.accountQueue.nextNonce -= 1
-	}
+	wrapper.accountQueue.nextNonce -= 1
```
munna0908: This need to improved, nextnonce should be decreased to the nonce of the transaction , Check this [link](https://github.com/ethereum/go-ethereum/blob/721c5723c0e31c4d764c61ee99c505d18466b363/core/tx_pool.go#L1021) for reference.

txpool/txpool.go
munna0908: I do agree with @Kourin1996 , it would be cleaner to have a method to remove tx from pending queue.

munna0908 (changes requested):

txpool/txpool.go
zivkovicmilos: Exactly 👍 

Do you feel like the removed section for the `OldChain` is something that we should leave intact?

@lazartravica @munna0908 curious to hear your thoughts on this as well

txpool/txpool.go
zivkovicmilos: It's not a super-costly operation, but it does add additional overhead.

I've refactored the way we do this state nonce fetching to be more optimal 👍 

01f412d & d1de5f0

txpool/txpool.go
zivkovicmilos: Extracted the logic 👍 

ecc07d8

txpool/txpool.go
zivkovicmilos: Extracted it 👍 

ecc07d8

txpool/txpool.go
zivkovicmilos: Completely agree with you, there is no need to be locking the account queue for each transaction.
I've added support for transaction bundling by account, and simplified the logic in `ProcessEvent`:

ecc07d8

txpool/txpool.go
```diff
@@ -361,14 +394,21 @@ func (t *TxPool) addImpl(origin TxOrigin, tx *types.Transaction) error {
...
-		t.remoteTxns.Push(tx)
+		pushErr := t.remoteTxns.Push(tx)
+		if pushErr != nil {
+			t.logger.Error(
+				fmt.Sprintf(
+					"Unable to push txn [%s] to the remote txns queue",
+					tx.Hash.String(),
+				),
+			)
+		}
```
dbrajovic: Use a short if statement.

txpool/txpool.go
```diff
@@ -471,60 +510,195 @@ func (t *TxPool) ResetWithHeader(h *types.Header) {
...
+						stateNonce: t.store.GetNonce(stateRoot, txn.From),
+						// Set up the transaction array
+						transactions: make([]*types.Transaction, 0),
+					}
+
+					eventWrapperMap[txn.From] = eventWrapper
+				}
+
+				// Add the transaction to the wrapper
+				eventWrapper.addTxn(txn)
```
dbrajovic: Maybe add a `continue`  in the case of `if !ok { ... }` and lose the else branch, so to make it less verbose.

txpool/txpool.go
lazartravica: @zivkovicmilos let's not alter any more than we need to right now, we will tackle this when refactoring next week.

txpool/txpool.go
```diff
@@ -200,6 +200,7 @@ type accountQueueWrapper struct {
 	lock         sync.RWMutex // lock for accessing the accountQueue
 	writeLock    int32        // flag indicating whether a write lock is held
 	accountQueue *txHeapWrapper
+	address      types.Address
```
lazartravica: No need to extend the fields on this struct just for debug log purposes. The absence of the log statement that says there was nothing to prune from the accountQueue is an information as well, we don't need an additional log.Debug statement for that. If you would like to have a log msg indicating that pruneAccountTx was called and at which time you can do a log.Debug before calling the pruneAccountTx from within the ProcessEvent method.

brkomir (approved):
Apart from comments already there, I couldn't find anything off spot

txpool/txpool.go
munna0908: Got clarity from milos

txpool/txpool.go
munna0908: @zivkovicmilos I do agree with youshiki, if we're planning to support various consensus down the line it make sense to have the oldchain logic intact

munna0908 (approved):
LGTM 🚀

txpool/txpool.go

4d1fb51

txpool/txpool.go

4d1fb51

dbrajovic (approved):
LGTM, this should resolve quite a few issues as well as any potential ones.

txpool/txpool.go
zivkovicmilos: Got it 👍 

Reverted that change here:

bd95c19

txpool/txpool.go
lazartravica: @munna0908 @zivkovicmilos 's PR does the change where the nextNonce is always in sync with the state. I don't think your comment is relevant in this case.

lazartravica (changes requested):
eth_getTransactionCount with the "pending" block param will always return the nextNonce from the state, are you aware of this, and did we plan out an issue to fix this?

txpool/txpool.go
zivkovicmilos: You're right 👍 

We already have a log that it's dropped if that situation occurs.

Fixed 👍 

6b30066

zivkovicmilos:
> eth_getTransactionCount with the "pending" block param will always return the nextNonce from the state, are you aware of this, and did we plan out an issue to fix this?

@lazartravica 
What do you mean?

The  fetch that the `dispatcher` does for pending nonces shouldn't rely on the `TxPool` to return a `not found` message, like it did earlier.

If the account is not initialized in the TxPool, it will grab the value from state, but if it is (and it's an account that has seen some uncommitted activity since the last state change), it will return the TxPool `nextNonce`.
We already had this functionality earlier, which was overwritten by PR #224.

Kourin1996 (approved):

lazartravica:
> > eth_getTransactionCount with the "pending" block param will always return the nextNonce from the state, are you aware of this, and did we plan out an issue to fix this?
> 
> @lazartravica What do you mean?
> 
> The fetch that the `dispatcher` does for pending nonces shouldn't rely on the `TxPool` to return a `not found` message, like it did earlier.
> 
> If the account is not initialized in the TxPool, it will grab the value from state, but if it is (and it's an account that has seen some uncommitted activity since the last state change), it will return the TxPool `nextNonce`. We already had this functionality earlier, which was overwritten by PR #224.

Thanks for the clarification. Got it. Approving the PR

lazartravica (approved):

PR #266, opened 2021-12-03
Refactor loop stress test to wait more accurately

zivkovicmilos:
# Description

This PR fixes a minor issue with the Test_TransactionIBFTLoop that causes it to fail on Github actions.
The test falsely assumes that the 50 or so transactions will get committed within a timeout period, when that can never be guaranteed.

The fix just changes the test to rely on transaction hashes instead of a timeout.

# Changes include

# Checklist

- [x] I have tested this code


lazartravica (approved):
💯 , let's see if the tests pass on GH and let's merge.

lazartravica:
@zivkovicmilos What are we going to do with this PR?

zivkovicmilos:
@lazartravica 
I think we should push it out to develop when people approve, since it does actually fix the test to be more accurate, even if the refactor wasn't the source of the final fix

Kourin1996 (approved):

PR #272, opened 2021-12-06
Fix reconnection issue for bootnodes

Kourin1996:
# Description

While investigating issue https://github.com/0xPolygon/polygon-sdk/issues/246 and fixing it on https://github.com/0xPolygon/polygon-sdk/pull/259, I found node can't reconnect to bootnodes again after node rejoin network.

This PR fixes this issue. 

In discovery, node checks the routing table every 5 secs and put bootnodes into table if table is empty.
https://github.com/0xPolygon/polygon-sdk/blob/70e340068731049d3a37a4d2e250b1905e01e5bc/network/discovery.go#L203-L213
https://github.com/0xPolygon/polygon-sdk/blob/70e340068731049d3a37a4d2e250b1905e01e5bc/network/discovery.go#L216-L222

Basically, a node tries to connect to a node when it's added to routing table automatically. But this behaviour prevents node from connecting bootnodes.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


munna0908:
@Kourin1996  I guess, this issue is fixed as part of this [PR](https://github.com/0xPolygon/polygon-sdk/pull/205)

network/discovery.go
```diff
@@ -131,21 +128,43 @@ func (d *discovery) setup() error {
...
+	d.srv.host.Peerstore().AddAddr(node.ID, node.Addrs[0], peerstore.AddressTTL)
+	if _, err := d.routingTable.TryAddPeer(node.ID, false, false); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *discovery) setupTable() error {
+	// Sleep until network setup is done
+	time.Sleep(5 * time.Second)
```
zivkovicmilos: Why `5s`?

zivkovicmilos (approved):
Looks great 💯

network/discovery.go
Kourin1996: Originally filling table was done 5 sec later after the network layer started. When I change this part of code, I noticed syncer couldn't connect because libp2p of peer was not started.

brkomir (approved):

network/discovery.go
lazartravica: @Kourin1996 is there a way to signal from the networking layer when it has started?

This looks very error prone, or in the best case scenario, we would delay the boot of the node by 5s always.

Kourin1996: Actually, this error doesn't occur inside the node. If You start validator and non-validators at the same time, non-validators try to connect to validator nodes for syncer. But validator's syncer service didn't begin at that time. That's why non-validator couldn't sync blocks

Kourin1996: Let me think another way to fix this issue

Kourin1996: @lazartravica @brkomir @zivkovicmilos I divided libp2p initialization into two function and make it start connecting after all gRPC server are ready

zivkovicmilos (approved):
Thank you for clearing up the discussions 🙏 

Looks great 💯

network/discovery.go
```diff
@@ -131,21 +128,41 @@ func (d *discovery) setup() error {
...
+	for _, node := range d.bootnodes {
+		if err := d.addToTable(node); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
 func (d *discovery) call(peerID peer.ID) error {
+	d.srv.logger.Debug("Quering a peer for near peers", "peer", peerID)
```
lazartravica:
```suggestion
	d.srv.logger.Debug("Querying a peer for near peers", "peer", peerID)
```

protocol/syncer.go
```diff
@@ -378,6 +384,7 @@ func (s *Syncer) BestPeer() *SyncPeer {
 
 	s.peers.Range(func(peerID, peer interface{}) bool {
 		status := peer.(*SyncPeer).status
+		fmt.Printf("status %+v\n", status)
```
lazartravica: Remove printf

lazartravica (approved):
See my minor comments, correct them and you're good to go 👍

network/discovery.go
Kourin1996: Fixed at 439ce005c87799a29848eb479702534ea338c24f

protocol/syncer.go
Kourin1996: Fixed at 439ce005c87799a29848eb479702534ea338c24f

PR #273, opened 2021-12-06
Add support for properly returning pending transactions

zivkovicmilos:
# Description

This PR adds a fix to the `getTransactionByHash` endpoint, where pending transactions were treated as `null`.

Pending transactions should have the following fields nulled out, with the rest filled in:
* `block number`
* `block hash`
* `transaction index`

`null` responses for `getTransactionByHash` should be reserved for transactions that have either been dropped, or never added to the client.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

The PR requests modifications on the `umbracle/go-web3` package.
PR containing requested changes:
https://github.com/umbracle/go-web3/pull/119


lazartravica (approved):
LGTM 👍

jsonrpc/eth_endpoint.go
```diff
@@ -111,28 +111,44 @@ func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
...
-	block, ok := e.d.store.GetBlockByHash(blockHash, true)
-	if !ok {
-		// block receipts not found
-		return nil, nil
-	}
-	for idx, txn := range block.Transactions {
-		if txn.Hash == hash {
-			return toTransaction(txn, block, idx), nil
+	// Check the TxPool for the transaction if it's pending
+	pendingTx, pendingFound := e.d.store.GetPendingTx(hash)
```
Kourin1996: I think you can return immediately after you found tx

```suggestion
	if pendingTx, pendingFound := e.d.store.GetPendingTx(hash); pendingFound {
		return toPendingTransaction(pendingTx), nil
	}
	
```

jsonrpc/eth_endpoint.go
```diff
@@ -111,28 +111,44 @@ func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
...
+		)
+		return nil, nil
 	}
-	// txn not found (this should not happen)
-	e.d.logger.Warn(
-		fmt.Sprintf("Transaction with hash [%s] not found", blockHash),
-	)
-	return nil, nil
+
+	return toPendingTransaction(pendingTx), nil
```
Kourin1996: `toPendingTransaction` is nil safe?

Kourin1996 (changes requested):

jsonrpc/eth_endpoint.go
```diff
@@ -111,28 +111,44 @@ func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
 	return transaction.Hash.String(), nil
 }
 
-// GetTransactionByHash returns a transaction by his hash
+// GetTransactionByHash returns a transaction by its hash.
+// If the transaction is still pending -> return the txn with some fields omitted
+// If the transaction is sealed into a block -> return the whole txn with all fields
 func (e *Eth) GetTransactionByHash(hash types.Hash) (interface{}, error) {
```
brkomir, line 117: This method never returns an error. If this is correct behaviour, remove the error from returned values

brkomir (approved):
Apart from the comment I've left, looks good!

jsonrpc/eth_endpoint.go
zivkovicmilos: I've refactored the entire endpoint to be more readable 👍 

ed68a97 & e5c4e1d

jsonrpc/eth_endpoint.go
zivkovicmilos: I've refactored the entire endpoint to be more readable 👍 

ed68a97 & e5c4e1d

jsonrpc/eth_endpoint.go
zivkovicmilos: I believe all of our endpoint methods need to return a `interface{}, error` response:
https://github.com/umbracle/go-web3/blob/e590eac2e9fd8e635782af2b3b9dbb49117dd003/jsonrpc/transport/http.go#L67-L69

dbrajovic (approved):

jsonrpc/eth_endpoint.go
```diff
@@ -111,27 +111,67 @@ func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
...
+					txn,
+					argUintPtr(block.Number()),
+					argHashPtr(block.Hash()),
+					&idx,
+				), true
+			}
+		}
+
+		return nil, false
 	}
```
Kourin1996: no need to return bool in 2nd return because always false when this function returns nil

jsonrpc/eth_endpoint.go
```diff
@@ -111,27 +111,67 @@ func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
...
+	// for the pending transaction with the provided hash
+	findPendingTx := func() (*transaction, bool) {
+		// Check the TxPool for the transaction if it's pending
+		if pendingTx, pendingFound := e.d.store.GetPendingTx(hash); pendingFound {
+			return toPendingTransaction(pendingTx), true
 		}
+
+		// Transaction not found in the TxPool
+		return nil, false
 	}
```
Kourin1996: it's same. You don't need to return bool

Kourin1996 (approved):
Looks good :)

jsonrpc/eth_endpoint.go
lazartravica: Excellent catch! @Kourin1996

jsonrpc/eth_endpoint.go
brkomir: I see now, all good

jsonrpc/eth_endpoint.go
zivkovicmilos: 🤦 

Good catch 💯 

Fixed:

e92f802

jsonrpc/eth_endpoint.go
zivkovicmilos: 🤦 

Good catch 💯 

Fixed:

e92f802

PR #274, opened 2021-12-06
Change signature r,s,v values to bigInt

0xAleksaOpacic:
# Description

The Transaction object wrongly has []byte instead of big.Int for the v,r,s values.
The reason why that is wrong is because a byte slice is completely valid to start with a leading 0 when encoded to hex, but that's not the case with any numbers, e.g. big.Int.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/types/transaction.go
+++ b/types/transaction.go
@@ -14,9 +14,9 @@ type Transaction struct {
 	To       *Address
 	Value    *big.Int
 	Input    []byte
-	V        []byte
-	R        []byte
-	S        []byte
+	V        big.Int
+	R        big.Int
+	S        big.Int
 	Hash     Hash
 	From     Address
 
@@ -51,10 +51,8 @@ func (t *Transaction) Copy() *Transaction {
 	tt.Value = new(big.Int)
 	tt.Value.Set(t.Value)
 
-	tt.R = make([]byte, len(t.R))
-	copy(tt.R[:], t.R[:])
-	tt.S = make([]byte, len(t.S))
-	copy(tt.S[:], t.S[:])
+	tt.R = *big.NewInt(0).SetBits(t.R.Bits())
+	tt.S = *big.NewInt(0).SetBits(t.S.Bits())
 
 	tt.Input = make([]byte, len(t.Input))
 	copy(tt.Input[:], t.Input[:])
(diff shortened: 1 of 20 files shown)


types/transaction.go
```diff
@@ -14,9 +14,9 @@ type Transaction struct {
 	To       *Address
 	Value    *big.Int
 	Input    []byte
-	V        []byte
-	R        []byte
-	S        []byte
+	V        big.Int
```
zivkovicmilos: Shouldn't this be a `*big.Int`?

zivkovicmilos (changes requested):
Thank you for the PR 🙏 

Please update and fill out the PR description so we know the context of this PR.
I'm curious to know why you didn't always use a `*big.Int`, but went ahead with de-referencing it every time?

types/transaction.go
0xAleksaOpacic: It makes more sense to be *big.Int. But I was unsure how to refactor the sender logic then.
The Previous type for these attributes was []byte and if we create Transaction object without them they would still get assigned as null values. These attributes (V, R, S) we use to encode signature (even if they are null values).
If we use pointers we would not be able to encode signature because the default value is nill. That's why I used reference because null values are the same. So I do not change sender/signature logic. 

The solution that was implemented before switching to reference was putting guard before encodeSignature to define attributes if they do not exist. 
```
	if tx.S == nil {
		tx.S = big.NewInt(0)
	}
```
But I find this solution messy.

The second thing I wanted to try was changing overall logic but was not comfortable doing it, not knowing the code base well enough.
If we need both R and S attributes for signing signature, inside ValidateSignatureValues we could check do these attributes exists and return error if not 
```
func ValidateSignatureValues(v byte, r, s *big.Int) bool {
	// TODO: ECDSA malleability

	if r == nil || s == nil {
		return false
	}

```

0xAleksaOpacic: I've changed these attributes to pointers.

inderashah007 (approved):

zivkovicmilos (approved):
Looks good 💯 

Kourin1996 (approved):
Looks good to me 👍 
Thank you for the PR.

dbrajovic (approved):

brkomir (approved):
Looks good now!

lazartravica (approved):

PR #284, opened 2021-12-09
Add support for EIP-1898

### LINKED ISSUE #253: eth_call does not accept EIP-1898 'blockHash' param
### Description

We noticed some RPC errors come up while making calls to `eth_call` using the EIP-1898 `blockHash` param (see https://eips.ethereum.org/EIPS/eip-1898).

It looks like the current unmarshaling logic fails on trying to validate the `params` arary as a result of this new field. 

Note that Geth 1.9.6 implements this proposal in https://github.com/ethereum/go-ethereum/pull/19491.

### Your environment

* OS and version `Ubuntu 20`
* version of the Polygon SDK `0a30e47 `

### Steps to reproduce

1. Make eth_call request containing `blockHash` param:
```
{"jsonrpc":"2.0","method":"eth_call","params":[{"data":"0x313ce567","gas":"0x2faf080","to":"0xfb70fca25af612114672118941fe7ec0b0734658"},{"blockHash":"0xeb3f1429c8868cf7d4d3db207fd9466fa31072928c59ac00f7079343fb085e07"}],"id":277408}`
```
2. Observe that the dispatcher cannot properly unmarshal the req.Params, responding with a `-32602 Invalid Params` error. 
3. Calls to `eth_call` without this new param object still work:
```
{"method":"eth_call","params":[{"to":"0xbec64eaab6fa7ca0184b9f264c7f80318cc04f34","data":"0x70a082310000000000000000000000002079a994e9bbc88496df03a59a0cfa2c9e2af473"},"latest"],"id":122723,"jsonrpc":"2.0"}
```

### Expected behaviour

JSON-RPC methods should support this and all variations of request bodies. This would allow for more flexible integrations with the SDK and therefore better user adoption.

### Actual behaviour

eth_call as well as the following methods return an error response when including the blockHash param as a second object within the params array:

- eth_getBalance
- eth_getStorageAt
- eth_getTransactionCount
- eth_getCode
- eth_getProof

0xpanoramix:
# Description

This PR adds support for the EIP 1898, meaning that a set of JSON-RPCs methods will now support a new query parameter : `blockHash`.

Here is how you can use this parameter with curl :
```
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x1010101010101010101010101010101010101010", {"blockHash":"0xe0ee62fd4a39a6988e24df0b406b90af71932e1b01d5561400a8eab943a33d68"}],"id":1}' localhost:8545
```

The implementation is backward compatible, meaning this query also works :
```
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x1010101010101010101010101010101010101010", "latest"],"id":1}' localhost:8545
```

It closes #253.

# Changes include

- [X] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

The following JSON-RPC methods are affected :

- getBalance
- getTransactionCount
- getCode
- getStorageAt
- call

# Checklist

- [X] I have tested this code


zivkovicmilos (approved):
The PR looks great 💯 

Thank you for documenting the code in tricky places, helps us with maintaining it down the line 🙏

jsonrpc/codec.go
```diff
@@ -102,6 +103,77 @@ const (
...
 
+type BlockNumberOrHash struct {
+	BlockNumber *BlockNumber `json:"blockNumber,omitempty"`
+	BlockHash   *types.Hash  `json:"blockHash,omitempty"`
+}
+
+// Unmarshal will try to extract the filter's data.
+// Here are the possible input formats :
+//
+// 1 - "latest", "pending" or "earliest"	- self-explaining keywords
```
0xAleksaOpacic, line 113: we do not support pending blocks because they are used only for POW. And our platform does not support the earliest headers yet. I will create an issue for this

jsonrpc/codec.go
```diff
@@ -102,6 +103,77 @@ const (
...
+	}
+
+	err = json.Unmarshal(data, &placeholder)
+	if err != nil {
+		var keyword string
+		err = json.Unmarshal(data, &keyword)
+		if err == nil {
+			// Try to extract keyword
+			switch keyword {
+			case "pending":
```
0xAleksaOpacic: we do not support pending blocks because they are used only for POW

jsonrpc/codec.go
0xAleksaOpacic: https://polygon-sdk.myjetbrains.com/youtrack/issue/PSDK-284

jsonrpc/codec.go
```diff
@@ -102,6 +103,77 @@ const (
...
+	}
+
+	// Try to extract object
+	bnh.BlockNumber = placeholder.BlockNumber
+	bnh.BlockHash = placeholder.BlockHash
+
+	if bnh.BlockNumber != nil && bnh.BlockHash != nil {
+		return fmt.Errorf("cannot get account balance using both block number hash")
+	}
+
```
Kourin1996, line 139: There is a possibility that both `bnh.BlockNumber` and `bnh.BlockHash` is `nil`. In this case, panic occurs when getting its value. Actually bnh.BlockNumber which may be nill is read in `GetTransactionCount`. 

```go
if bnh.BlockHash != nil {
                 // safe to access to bnh.BlockHash
		block, ok := e.d.store.GetBlockByHash(*bnh.BlockHash, false)
		if !ok {
			return nil, fmt.Errorf("could not find block referenced by the hash %s", bnh.BlockHash.String())
		}
                ...
	} else {
                 // unsafe to access to bnh.BlockNumber!
		nonce, err = e.d.getNextNonce(address, *bnh.BlockNumber)
	}
```

By the way, such values can be got by passing unexpected value like:

```json
{
	"jsonrpc":"2.0",
	"method":"eth_getTransactionCount",
	"params":[
		"0xBa8CAE91c858120E9699f9b9936418C6bd3ecCC1",
        {"hoge": "0x1"}
	],
	"id":1
}
```

So my suggestion is that we need to return error when both BlockHash and BlockNumber is nil

Kourin1996 (changes requested):

jsonrpc/codec.go
```diff
@@ -102,6 +103,77 @@ const (
...
+				// Try to extract hex number
+				s, ok := (*input).(string)
+				if !ok {
+					return fmt.Errorf("input cannot be converted to string")
+				}
+				if len(s) < 3 || !strings.HasPrefix(s, "0x") {
+					return fmt.Errorf("invalid hexadecimal number provided for block number")
+				}
+				number, err := strconv.ParseInt(s[2:], 16, 64)
+				fmt.Println(number, err)
```
Kourin1996: you can remove this debug code?

jsonrpc/eth_endpoint.go
```diff
@@ -577,30 +637,81 @@ func (e *Eth) GetBalance(
...
+		if !ok {
+			return nil, fmt.Errorf("could not find block referenced by the hash %s", bnh.BlockHash.String())
+		}
+
+		header := block.Header
+		nonce, err = e.d.getNextNonce(address, BlockNumber(header.Number))
+		if err != nil {
+			if errors.As(err, &ErrStateNotFound) {
+				return argUintPtr(0), nil
+			}
```
Kourin1996: You should be able to use `errors.Is` instead of `errors.As`

```go
	return argUintPtr(0), nil
}
```

Kourin1996 (approved):
LGTM. Thank you for the fix 👍

jsonrpc/codec.go
0xpanoramix: Good catch ! I would like to remove these non-supported options in a new PR, to keep the single responsability principle for this one.

0xAleksaOpacic: Agreed we can do it in new PR

0xAleksaOpacic (approved):

jsonrpc/codec.go
```diff
@@ -102,6 +103,78 @@ const (
...
+}
+
+// Unmarshal will try to extract the filter's data.
+// Here are the possible input formats :
+//
+// 1 - "latest", "pending" or "earliest"	- self-explaining keywords
+// 2 - "0x2"								- block number #2 (EIP-1898 backward compatible)
+// 3 - {blockNumber:	"0x2"}				- EIP-1898 compliant block number #2
+// 4 - {blockHash:		"0xe0e..."}			- EIP-1898 compliant block hash 0xe0e...
+func (bnh *BlockNumberOrHash) Unmarshal(input *interface{}) error {
```
ferranbt: Please add a unit test to test this.

jsonrpc/eth_endpoint.go
```diff
@@ -550,19 +590,39 @@ func (e *Eth) GetLogs(filterOptions *LogFilter) (interface{}, error) {
...
 
-// GetBalance returns the account's balance at the referenced block
-func (e *Eth) GetBalance(
-	address types.Address,
-	number *BlockNumber,
-) (interface{}, error) {
-	if number == nil {
-		number, _ = createBlockNumberPointer("latest")
+// GetBalance returns the account's balance at the referenced block.
+func (e *Eth) GetBalance(address types.Address, filter interface{}) (interface{}, error) {
```
ferranbt: This applies to all the endpoints that implement this EIP:
- Avoid using an interface in the param, it is easier if the dispatcher can do the type uncasting instead.
- The code for all the functions is exactly the same, do not copy paste it every time because that is going to break for sure, please abstract that into a function ```getHeaderFromBlockNumberOrHash()```.
- Is there any unit test coverage of this?

e2e/eip_1898_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/0xPolygon/polygon-sdk/jsonrpc"
+	"github.com/0xPolygon/polygon-sdk/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/umbracle/go-web3"
+	"math/big"
+	"testing"
+	"time"
+)
+
+func Test_GetBalance(t *testing.T) {
```
ferranbt: E2E test are meant to test things in integration (consensus + txpool + syncer) but I think the JsonRPC server can be tested in unit tests.

jsonrpc/codec.go
```diff
@@ -102,6 +103,78 @@ const (
...
+
+	data, err := json.Marshal(*input)
+	if err != nil {
+		return fmt.Errorf("failed to serialize input: %v", err)
+	}
+
+	err = json.Unmarshal(data, &placeholder)
+	if err != nil {
+		var keyword string
+		err = json.Unmarshal(data, &keyword)
```
ferranbt: Aren't you repeating here what it is already done in ```stringToBlockNumber```?

jsonrpc/codec.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-sdk/pull/284/commits/14c3e134c8727de6332cda429dff4f1e8527a0c6

jsonrpc/eth_endpoint.go
0xpanoramix: Thanks for the advices 💯 
I've made updates according to them in https://github.com/0xPolygon/polygon-sdk/pull/284/commits/6810ace3fb31792d7a8dcf4a084464e83961e769

jsonrpc/codec.go
0xpanoramix: Good catch 👍 Updated in https://github.com/0xPolygon/polygon-sdk/pull/284/commits/0d3704c7e71429ee06a365926544f6018716d771

jsonrpc/eth_endpoint_test.go
```diff
@@ -649,8 +663,8 @@ func TestEth_State_GetStorageAt(t *testing.T) {
 				}
 			}
 			dispatcher := newTestDispatcher(hclog.NewNullLogger(), store)
-			blockNumber, _ := createBlockNumberPointer(tt.blockNumber)
-			res, err := dispatcher.endpoints.Eth.GetStorageAt(tt.address, tt.index, blockNumber)
+			// blockNumber, _ := createBlockNumberPointer(tt.blockNumber)
```
zivkovicmilos: Please clean up this comment

e2e/eip_1898_test.go
zivkovicmilos: @PtitLuca 

I agree with Ferran, there's no reason this should be an integration test, since you're just testing out different queries for endpoints.

Please make it a unit test with a testing table that covers these cases 🙏

jsonrpc/eth_endpoint_test.go
0xpanoramix: Nice catch ! Removed in https://github.com/0xPolygon/polygon-sdk/pull/284/commits/78906462560351619f9a53584b93f252d2283ccc.

e2e/eip_1898_test.go
zivkovicmilos: @PtitLuca fixed it in df1f861 👍

PR #287, opened 2021-12-13
TxPool refactor

dbrajovic:
# Description

This PR addresses the various issues that appeared in the old implementation of the pool. The refactor introduces a new flow - handling enqueuing a transaction and promoting it asynchronously. 

The pool separates the transactions by their sender's  address, processing them within the sender's `account`. The `account` struct is essential in ensuring the transactions are processed correctly, but is also what enables the pool to handle various transactions concurrently and separately among all registered accounts.
 
When all the transactions ready for execution need to be processed, the pool has an intermediary queue (`executables`)
which contains the first-in-line transactions from all accounts, sorted by max gas price. (See `Prepare()`, `Next()`, `Pop()`, `Drop()`, `Demote()` for more details) 

All other functionalities are explained in greated depth within code documentation. 

# Changes include

# Checklist

- [x] I have tested this code


consensus/dev/dev.go
```diff
@@ -103,43 +103,46 @@ type transitionInterface interface {
...
 
-		txns = append(txns, txn)
+		successful = append(successful, tx)
 	}
 
-	// we return recoverable txns that were popped from the txpool after the above for loop breaks,
-	// since we don't want to return the tx to the pool just to pop the same txn in the next loop iteration
-	for _, retFunc := range returnTxnFuncs {
-		retFunc()
+	d.logger.Info("picked out txns from pool", "num", len(successful), "remaining", len(recoverables))
```
brkomir: This log was previously indicating how many transactions are remaining in the pool, not how many are recoverable. Replace len(recoverables) with len(recoverables) + txpool.length

consensus/ibft/ibft.go
```diff
@@ -457,43 +459,46 @@ type transitionInterface interface {
...
 
-		txns = append(txns, txn)
+		successful = append(successful, tx)
 	}
 
-	// we return recoverable txns that were popped from the txpool after the above for loop breaks,
-	// since we don't want to return the tx to the pool just to pop the same txn in the next loop iteration
-	for _, retFunc := range returnTxnFuncs {
-		retFunc()
+	i.logger.Info("picked out txns from pool", "num", len(successful), "remaining", len(recoverables))
```
brkomir: same as for dev consensus

consensus/ibft/ibft.go
```diff
@@ -468,43 +471,45 @@ type transitionInterface interface {
...
+		if tx == nil {
 			break
 		}
 
-		if txn.ExceedsBlockGasLimit(gasLimit) {
+		if tx.ExceedsBlockGasLimit(gasLimit) {
 			i.logger.Error(fmt.Sprintf("failed to write transaction: %v", state.ErrBlockLimitExceeded))
-			i.txpool.DecreaseAccountNonce(txn)
+			i.txpool.Drop()
+			// i.txpool.RollbackNonce(tx) // unrecoverable tx
```
brkomir: remove comment

consensus/ibft/ibft_test.go
```diff
@@ -657,38 +656,60 @@ func (s *mockSyncer) Broadcast(b *types.Block) {
...
-	p.resetWithHeaderCalled = true
-	p.resetWithHeadersParam = headers
+func (p *mockTxPool) LockPromoted(write bool) {
+}
+
+func (p *mockTxPool) UnlockPromoted() {
+}
+
+func (p *mockTxPool) Recover(tx *types.Transaction) {
+	p.transactions = append(p.transactions, tx)
```
brkomir: Since locking the promoted queue is critical for the process of writing transactions to a block, you can use these mock methods to test that `writeTransactions` method in ibft.go actually calls lock/unlock properly.
For example, add bool fields to mock struct, locked and unlocked, set them to true when the methods are called and include assertions in the unit test for `writeTransactions`

txpool/mock.go
```diff
@@ -0,0 +1,56 @@
+package txpool
```
brkomir: Since this file is used only to hold mocks, rename it to "_test.go"

txpool/queue.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"container/heap"
+
+	"github.com/0xPolygon/polygon-sdk/types"
+)
+
+/* Queue implementations required by the heap interface */
+
+type transactions []*types.Transaction
```
brkomir: Why create a named type and don't do anything with it?
You should create named types to provide methods that extend the functionality of the source type and it looks to me like the only thing you wanted to achieve here is to have a convenient name for a slice of transaction pointers.

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
 
-		slotsToRemove -= txSlots
+	return nil
+}
+
+// handleGossipTx handles receiving transactions
+// gossiped by the network.
+func (p *TxPool) handleGossipTx(obj interface{}) {
+	if !p.sealing {
+		return
```
brkomir: I do not understand this part and I'm not sure if I'm missing something... But here we go :)
If the transaction is received by gossip and we are not sealing node, we just return here, meaning the tx will not end up in the pool. Does this mean that we will not gossip about this transaction to our peers?
If my understanding is correct, we should not gossip a transaction only if it is already present in our tx pool, regardless of the sealing status...

txpool/txpool.go
```diff
@@ -192,576 +193,219 @@ func NewTxPool(
...
-// GetPendingTx returns the transaction by hash in the TxPool (pending txn) [Thread-safe]
-func (t *TxPool) GetPendingTx(txHash types.Hash) (*types.Transaction, bool) {
-	t.txnLookupMapLock.RLock()
-	defer t.txnLookupMapLock.RUnlock()
+// Drop is called within ibft for any transaction
+// deemed unrecoverable during writing to the state.
+// This call ensures that any subsequent transactions
+// must not be processed before the unrecoverable one
+// is re-sent again for that account.
+func (p *TxPool) Drop() {
```
brkomir: I think `Drop` should do gauge decrease, as in `Pop`.
`p.gauge.decrease(slotsRequired(tx))`

txpool/txpool.go
```diff
@@ -192,576 +193,219 @@ func NewTxPool(
...
-		lowestNonceTx := a.accountQueue.txs.Peek()
-
-		if lowestNonceTx == nil || // There is nothing in the account specific queue
-			lowestNonceTx.Nonce >= a.accountQueue.nextNonce { // The lowest nonce tx is valid
-			// All good
-			break
+// Start runs the pool's main loop in the background.
+// On each request received, the appropriate handler
+// is invoked in a separate goroutine.
+func (p *TxPool) Start() error {
```
brkomir: why do you return `error` here, this function can't possibly fail

txpool/txpool.go
```diff
@@ -192,576 +193,219 @@ func NewTxPool(
...
 
-// GetPendingTx returns the transaction by hash in the TxPool (pending txn) [Thread-safe]
-func (t *TxPool) GetPendingTx(txHash types.Hash) (*types.Transaction, bool) {
-	t.txnLookupMapLock.RLock()
-	defer t.txnLookupMapLock.RUnlock()
+// Drop is called within ibft for any transaction
+// deemed unrecoverable during writing to the state.
+// This call ensures that any subsequent transactions
+// must not be processed before the unrecoverable one
+// is re-sent again for that account.
```
brkomir: Please improve on comments here, it is not relevant for the function definition to be described who is calling it and when. This comment should state that the method removes the first transaction from the pending queue and resets the next nonce for that account.

txpool/txpool.go
```diff
@@ -192,576 +193,219 @@ func NewTxPool(
...
-func (t *TxPool) deleteTxFromLookup(txHash types.Hash) {
-	t.txnLookupMapLock.Lock()
-	defer t.txnLookupMapLock.Unlock()
-
-	delete(t.txnLookupMap, txHash)
-}
+// Demote is called within ibft for all transactions
+// that are valid but couldn't be written to the state
+// at the given time (recoverable). Issues an addRequest to the pool
+// indicating a transaction is returning to it.
```
brkomir: Similar stuff as for the Drop function, `Demote` removes the transaction from the pending queue but doesn't eject it from the pool and does not manipulate the next nonce for the account.

txpool/txpool.go
```diff
@@ -192,576 +193,219 @@ func NewTxPool(
...
+	p.addReqCh <- addRequest{tx: tx, demoted: true}
 }
 
-// ResetWithHeader does basic txpool housekeeping after a block write
-func (t *TxPool) ResetWithHeaders(headers ...*types.Header) {
-	evnt := &blockchain.Event{
+// ResetWithHeaders is called from within ibft when the node
+// has received a new block from a peer. The pool needs to align
+// its own state with the new one so it can correctly process
+// further incoming transactions.
```
brkomir: Again, it is not relevant who is calling the method but why.

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
+	// send request [BLOCKING]
+	p.addReqCh <- addRequest{tx: tx, demoted: false}
 
-		slotsToRemove -= txSlots
+	return nil
+}
+
+// handleGossipTx handles receiving transactions
+// gossiped by the network.
+func (p *TxPool) handleGossipTx(obj interface{}) {
```
brkomir: this method is uncovered by unit tests

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
 	}
 
-	delete(t.index, tx.Hash)
-	heap.Remove(t.heap, item.index)
+	return tx, true
+}
 
-	return true
+// Map of expected nonces for all accounts (known by the pool).
+type nonceMap struct {
```
brkomir: It is unclear to me as a first-time reader of this code what values are you trying to match here.
I concluded that this struct is used to map the account address to the next nonce for that account.
So if you agree, something like `addressToNextNonce` would give the reader a better understanding of the purpose of this map.

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
-func (t *txPriceHeap) Push(tx *types.Transaction) error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+func (m *nonceMap) store(addr types.Address, nonce uint64) {
+	m.Store(addr, nonce)
+}
 
-	price := new(big.Int).Set(tx.GasPrice)
+// Thread safe map of all accounts registered by the pool.
+type accountsMap struct {
```
brkomir: Similar stuff as for the "nonceMap", what is mapped to what?

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
+
+// account is a thread-safe wrapper object
+// around an account queue (enqueued transactions).
+//
+// Accounts are only ever initalized once and stored
+// atomically in the accountsMap.
+//
+// All methods assume the approriate lock is held
+// and at the right time (context).
+type account struct {
```
brkomir: types like this one, which doesn't depend on the logic of the transaction pool struct but the other way around (they are used by txpool), can and should reside in their own files. There are several of them here, so I will name them all. We would greatly reduce the amount of code in the txpool.go file and make it much easier to read and understand.

txpool/txpool.go
```diff
@@ -823,466 +463,712 @@ func (t *TxPool) validateTx(
...
-	}
-	return &localAccounts{
-		accounts: accounts,
-		mutex:    sync.RWMutex{},
+// Returns valid and pruned transactions after filtering the queue.
+// Only PrunePromoted calls this method.
+//
+// Valid - transactions not considered stale.
+//
+// Pruned - transactions considered as stale by the nonceMao.
```
brkomir: typo "nonceMao"

txpool/txpool.go
```diff
@@ -25,12 +24,13 @@ const (
...
 )
 
 var (
 	ErrIntrinsicGas        = errors.New("intrinsic gas too low")
 	ErrNegativeValue       = errors.New("negative value")
-	ErrNonEncryptedTxn     = errors.New("non-encrypted transaction")
+	ErrNonEncryptedTx      = errors.New("non-encrypted transaction")
 	ErrInvalidSender       = errors.New("invalid sender")
 	ErrTxPoolOverflow      = errors.New("txpool is full")
 	ErrUnderpriced         = errors.New("transaction underpriced")
```
brkomir, line 33: I see `ErrUnderpriced` is unused, is this intended?

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1232 @@
...
+	slots    uint64
 }
 
-const validGasLimit uint64 = 100000
-
-func TestAddingTransaction(t *testing.T) {
-	senderPriv, _ := tests.GenerateKeyAndAddr(t)
-	_, receiverAddr := tests.GenerateKeyAndAddr(t)
-
+func TestAddTxErrors(t *testing.T) {
```
brkomir, line 99: I have noticed you are missing cases for `ErrNonceTooLow` and `ErrInsufficientFunds`

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1232 @@
...
+				}()
+				go pool.handleAddRequest(<-pool.addReqCh)
+				<-pool.promoteReqCh
+			case ErrInvalidAccountState:
+				pool.store = faultyMockStore{}
+			case ErrOversizedData:
+				data := make([]byte, 989898)
+				_, err := rand.Read(data)
+				assert.NoError(t, err)
+				tc.tx.Input = data
```
brkomir: This is not the way to go, switch case in the body of the unit test isn't something that is considered a good practice. As I see here, you have this switch case for the sole purpose of the conforming structure of the test to be a table test. I suggest you do a common setup method for these tests and make each case here a separate test (`t.Run(...)`) with its own setup logic and description of what is being tested. My opinion is that this would greatly improve readability of the code

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1232 @@
...
+	addr8  = types.Address{0x8}
+	addr9  = types.Address{0x9}
+	addr10 = types.Address{0x10}
+)
 
-func (m *mockStore) GetNonce(types.Hash, types.Address) uint64 {
-	return 0
-}
+// returns a new valid tx of slots size with the given nonce
+func newDummyTx(addr types.Address, nonce, slots uint64) *types.Transaction {
```
brkomir: Dummy is a name used for a test double that does not do anything except for filling up the "empty space", please remove it from the method name since it's misleading because the method doesn't return a dummy transaction but a real one. For reference on how to name test doubles check this [link](https://jesusvalerareales.medium.com/testing-with-test-doubles-7c3abb9eb3f2#:~:text=The%20five%20types%20of%20Test,the%20class%20is%20being%20used.)

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1232 @@
...
+
+		{ // set up prestate
+			pool.createAccountOnce(addr)
+			assert.Equal(t, uint64(0), pool.accounts.from(addr).length())
+
+			pool.nextNonces.store(addr, 20)
+			nextNonce, ok := pool.nextNonces.load(addr)
+			assert.True(t, ok)
+			assert.Equal(t, uint64(20), nextNonce)
+		}
```
brkomir: Why these { ... } ?

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
+	}()
 
-	return len(wrapper.accountQueue.txs)
+	return nil
 }
 
-func (t *TxPool) AddSigner(s signer) {
-	t.signer = s
+// Stops the pool's main loop.
+func (p *TxPool) Stop() {
```
brkomir: I don't see where this method is used

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-	// It is used for grouping transactions on an account basis
-	// and for storing next state nonce information for easy lookup
-	// Account address -> processEventWrapper
-	eventWrapperMap := make(map[types.Address]*processEventWrapper)
-
-	// Grab the latest state root now that the block has been inserted
-	stateRoot := t.store.Header().StateRoot
+// Collects the latest nonces for each account containted
+// in the received event. Resets all accounts with the new nonce.
+func (p *TxPool) processEvent(event *blockchain.Event) {
```
brkomir, line 409: It seems that there are no tests covering this function

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-		)
+	p.logger.Debug("demoted transaction", "hash", tx.Hash.String())
+}
 
-		cleanupCallback(droppedPendingTx)
+// ResetWithHeaders is called from within ibft when the node
+// has received a new block from a peer. The pool needs to align
+// its own state with the new one so it can correctly process
+// further incoming transactions.
+func (p *TxPool) ResetWithHeaders(headers ...*types.Header) {
```
brkomir, line 397: Are there no tests for this method?

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+			// that is gossiped back
+			p.logger.Debug(
+				"dropping known gossiped transaction",
+				"hash", tx.Hash.String(),
+			)
 
-		txSlots := slotsRequired(pricedTx.tx)
-		if slotsToRemove < txSlots {
-			return dropped, true
+			return nil
```
brkomir: Not covered in unit tests

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
 }
 
-// newTxHeapWrapper creates a new account based tx heap
-func newTxHeapWrapper(logger hclog.Logger) *txHeapWrapper {
-	return &txHeapWrapper{
-		logger: logger,
-		txs:    txHeap{},
+// handleGossipTx handles receiving transactions
+// gossiped by the network.
+func (p *TxPool) handleGossipTx(obj interface{}) {
```
brkomir: no unit tests for this method

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+	})
+
+	return
 }
 
-func (t *txPriceHeap) Push(tx *types.Transaction) error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+// Returns all promoted and enqueued transactions (if the flag is set to true).
+func (m *accountsMap) allTxs(includeEnqueued bool) (
```
brkomir: not covered in unit tests

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
 
-func (t *txHeap) Push(x interface{}) {
-	(*t) = append((*t), x.(*types.Transaction))
+// GetTxs gets pending and queued transactions
+func (p *TxPool) GetTxs(inclQueued bool) (
+	allPromoted, allEnqueued map[types.Address][]*types.Transaction,
+) {
+	allPromoted, allEnqueued = p.accounts.allTxs(inclQueued)
+	return
 }
```
brkomir: Don't sure if i'm missing something, but I can't find where these methods are being utilized

txpool/txpool.go
```diff
@@ -25,12 +23,14 @@ const (
...
 
+// errors
 var (
 	ErrIntrinsicGas        = errors.New("intrinsic gas too low")
 	ErrNegativeValue       = errors.New("negative value")
-	ErrNonEncryptedTxn     = errors.New("non-encrypted transaction")
+	ErrNonEncryptedTx      = errors.New("non-encrypted transaction")
 	ErrInvalidSender       = errors.New("invalid sender")
 	ErrTxPoolOverflow      = errors.New("txpool is full")
 	ErrUnderpriced         = errors.New("transaction underpriced")
```
lazartravica, line 33: Unused.

txpool/txpool.go
```diff
@@ -42,15 +42,27 @@ var (
 	ErrOversizedData = errors.New("oversized data")
```
lazartravica: Do we really need this coment for ErrOversizedData?

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
 }
 
-func (t *txHeap) Pop() interface{} {
-	old := *t
-	n := len(old)
-	x := old[n-1]
-	*t = old[0 : n-1]
-	return x
+// Thread safe map of all accounts registered by the pool.
+type accountsMap struct {
```
brkomir: This one deserves its own file

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+
+	return tx, true
 }
 
-func (a *localAccounts) containsAddr(addr types.Address) bool {
-	a.mutex.RLock()
-	defer a.mutex.RUnlock()
-	return a.accounts[addr]
+// Gauge for measuring pool capacity in slots
+type slotGauge struct {
```
brkomir: This one deserves its own file

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
 	}
 
-	return t.txs[i].price.Cmp(t.txs[j].price) >= 0
+	return promoted
 }
 
-type minTxPriceHeapImpl struct {
-	txPriceHeapImplBase
+// Lookup map used to find transactions present in the pool
+type lookupMap struct {
```
brkomir: This one deserves its own file

e2e/genesis_test.go
```diff
@@ -64,6 +64,7 @@ func TestCustomBlockGasLimitPropagation(t *testing.T) {
 	srvs := framework.NewTestServers(t, 1, func(config *framework.TestServerConfig) {
 		config.SetConsensus(framework.ConsensusDev)
 		config.SetBlockLimit(blockGasLimit)
+		// config.SetDevInterval(3)
```
lazartravica: Delete comment

e2e/txpool_test.go
```diff
@@ -528,9 +405,6 @@ func TestInvalidTransactionRecover(t *testing.T) {
 	//
 	// 1. Send a transaction with gasLimit > block gas limit.
 	//		-> The transaction should not be applied, and the nonce should not be incremented.
-	//
```
lazartravica, line 539: If we want to test only the case when the tx's gasLimit is larger than the block gas limit, we can change the name of this test and get rid of the test table since we only have a single test.

jsonrpc/txpool_endpoint.go
```diff
@@ -63,18 +63,30 @@ func toTxPoolTransaction(t *types.Transaction) *txpoolTransaction {
...
 	pendingTxs, queuedTxs := t.d.store.GetTxs(true)
+
+	// collect pending
 	pendingRpcTxns := make(map[types.Address]map[uint64]*txpoolTransaction)
-	for address, nonces := range pendingTxs {
-		pendingRpcTxns[address] = make(map[uint64]*txpoolTransaction)
-		for nonce, tx := range nonces {
-			pendingRpcTxns[address][nonce] = toTxPoolTransaction(tx)
+	for addr, txs := range pendingTxs {
+		pendingRpcTxns[addr] = map[uint64]*txpoolTransaction{}
```
lazartravica: You know the amount of txs for a single addr, you can use `make` with capacity to prevent memory reallocation.

jsonrpc/txpool_endpoint.go
```diff
@@ -63,18 +63,30 @@ func toTxPoolTransaction(t *types.Transaction) *txpoolTransaction {
...
+
+	// collect enqueued
 	queuedRpcTxns := make(map[types.Address]map[uint64]*txpoolTransaction)
-	for address, nonces := range queuedTxs {
-		queuedRpcTxns[address] = make(map[uint64]*txpoolTransaction)
-		for nonce, tx := range nonces {
-			queuedRpcTxns[address][nonce] = toTxPoolTransaction(tx)
+	for addr, txs := range queuedTxs {
+
+		queuedRpcTxns = make(map[types.Address]map[uint64]*txpoolTransaction)
```
lazartravica: You know the amount of txs for a single addr, you can use make with capacity to prevent memory reallocation.

jsonrpc/txpool_endpoint.go
```diff
@@ -89,23 +101,31 @@ func (t *Txpool) Content() (interface{}, error) {
...
-		for nonce, tx := range nonces {
+	for addr, txs := range pendingTxs {
+		pendingRpcTxns[addr.String()] = make(map[string]string)
+
+		for _, tx := range txs {
 			msg := fmt.Sprintf("%d wei + %d gas x %d wei", tx.Value, tx.Gas, tx.GasPrice)
-			pendingRpcTxns[address.String()][strconv.FormatUint(nonce, 10)] = msg
+			nonceStr := strconv.FormatUint(tx.Nonce, 10)
+
+			pendingRpcTxns[addr.String()][nonceStr] = msg
```
lazartravica: you can inline the fmt.Sprintf here

jsonrpc/txpool_endpoint.go
```diff
@@ -89,23 +101,31 @@ func (t *Txpool) Content() (interface{}, error) {
 // Create response for txpool_inspect request.
 // See https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_inspect.
 func (t *Txpool) Inspect() (interface{}, error) {
-
 	pendingTxs, queuedTxs := t.d.store.GetTxs(true)
+
+	// collect pending
 	pendingRpcTxns := make(map[string]map[string]string)
```
lazartravica: you can provide capacity for make

jsonrpc/txpool_endpoint.go
```diff
@@ -89,23 +101,31 @@ func (t *Txpool) Content() (interface{}, error) {
...
-	for address, nonces := range pendingTxs {
-		pendingRpcTxns[address.String()] = make(map[string]string)
-		for nonce, tx := range nonces {
+	for addr, txs := range pendingTxs {
+		pendingRpcTxns[addr.String()] = make(map[string]string)
```
lazartravica: you can provide capacity for make

jsonrpc/txpool_endpoint.go
```diff
@@ -89,23 +101,31 @@ func (t *Txpool) Content() (interface{}, error) {
...
-		for nonce, tx := range nonces {
+	for addr, txs := range queuedTxs {
+		queuedRpcTxns[addr.String()] = make(map[string]string)
+
+		for _, tx := range txs {
 			msg := fmt.Sprintf("%d wei + %d gas x %d wei", tx.Value, tx.Gas, tx.GasPrice)
-			queuedRpcTxns[address.String()][strconv.FormatUint(nonce, 10)] = msg
+			nonceStr := strconv.FormatUint(tx.Nonce, 10)
+
+			queuedRpcTxns[addr.String()][nonceStr] = msg
```
lazartravica: you can inline the fmt.Sprintf here

server/server.go
```diff
@@ -154,16 +154,15 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 		// start transaction pool
 		m.txpool, err = txpool.NewTxPool(
 			logger,
-			m.config.Seal,
-			m.config.Locals,
-			m.config.NoLocals,
-			m.config.PriceLimit,
```
lazartravica, line 162: What happened to the price limit? Price limit was not only tied to the locals, but was how we achieved gasless and min-gas networks

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-	return pendingTxs, queuedTxs
 }
 
-// Length returns the size of the valid transactions in the txpool
-func (t *TxPool) Length() uint64 {
-	return t.pendingQueue.Length()
+// Returns the highest priced transaction
+// from the executables queue.
+func (p *TxPool) Peek() *types.Transaction {
+	return p.executables.pop()
```
brkomir, line 309: This doesn't seem right, you are calling a pop inside peek. Maybe you want to call peek here and then call the `p.executables.pop()` in the `(p *TxPool) Pop`?

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+// Checks if an account exists within the map.
+func (m *accountsMap) exists(addr types.Address) bool {
+	_, ok := m.Load(addr)
+	return ok
 }
 
-func (t *txPriceHeap) Length() uint64 {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+// Collects the primaries of all accounts.
```
brkomir: Please explain in more detail what primaries represent

txpool/operator.go
```diff
@@ -10,16 +10,16 @@ import (
 )
 
 // Status implements the GRPC status endpoint. Returns the number of transactions in the pool
-func (t *TxPool) Status(ctx context.Context, req *empty.Empty) (*proto.TxnPoolStatusResp, error) {
+func (p *TxPool) Status(ctx context.Context, req *empty.Empty) (*proto.TxnPoolStatusResp, error) {
 	resp := &proto.TxnPoolStatusResp{
-		Length: t.pendingQueue.Length(),
+		Length: p.accounts.promoted(),
```
lazartravica, line 15: Are we sure this is the same metric?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
 
-	// Indicates which txpool operator commands should be implemented
+	// indicates which txpool operator commands should be implemented
 	proto.UnimplementedTxnPoolOperatorServer
-
-	metrics *Metrics
 }
 
-// NewTxPool creates a new pool for transactions
+// Returns a new pool for processing incoming transactions.
```
lazartravica: General style in Go is to prefix the comment for describign structs/methods/functions, etc. with the name, e.g. `NewTxPool returns a...`, I will only leave this comment, but it applies to the rest of the file

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+		// update counters
+		nextNonce += 1
+		promoted += 1
 	}
-}
 
-func (t maxTxPriceHeapImpl) Less(i, j int) bool {
-	if t.txs[i].from == t.txs[j].from {
-		return t.txs[i].tx.Nonce < t.txs[j].tx.Nonce
+	// only update the nonce map if the new nonce
```
brkomir: nonce map doesn't exist anymore :)

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
+	// channels on which the pool's event loop
+	// does dispatching/handling requests.
+	enqueueReqCh chan enqueueRequest
+	promoteReqCh chan promoteRequest
 
-	// Flag indicating if the current node is a sealer,
+	// shutdown channel
+	shutdownCh chan struct{}
+
+	// flag indicating if the current node is a sealer,
```
lazartravica, line 159: Why do only sealers gossip transactions?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
-	priceLimit uint64
-
-	// Notification channel used so signal added transactions to the pool
-	NotifyCh chan struct{}
+	// prometheus API
+	metrics *Metrics
 
-	// Indicates which txpool operator commands should be implemented
+	// indicates which txpool operator commands should be implemented
 	proto.UnimplementedTxnPoolOperatorServer
```
lazartravica, line 170: Why is this needed?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
+// When consensus needs to process promoted transactions,
+// the pool generates a queue of "executable" transactions. These
+// transactions are the first-in-line of some promoted queue,
+// ready to be written to the state (primaries).
 type TxPool struct {
 	logger     hclog.Logger
 	signer     signer
 	forks      chain.ForksInTime
 	store      store
 	idlePeriod time.Duration
```
brkomir: do you know what does this field represents since it is not used anywhere?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
-	// Addresses that should be treated as local
-	locals *localAccounts
-
-	// priceLimit is a lower threshold for gas price
-	priceLimit uint64
-
-	// Notification channel used so signal added transactions to the pool
-	NotifyCh chan struct{}
+	// prometheus API
+	metrics *Metrics
```
brkomir, line 167: Usage of metrics doesn't seem to be correctly implemented.

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-	for {
-		// Check if the nonce is lower than what the TxPool is expecting
-		lowestNonceTx := a.accountQueue.txs.Peek()
-
-		if lowestNonceTx == nil || // There is nothing in the account specific queue
-			lowestNonceTx.Nonce >= a.accountQueue.nextNonce { // The lowest nonce tx is valid
-			// All good
-			break
+		// check if shutdown was called
+		select {
```
lazartravica: We are only able to gracefully stop when we are doing the Start()? I don't see us checking the shutdownCh elsewhere.

txpool/txpool.go
lazartravica: I think you need to put it under the `for` below

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-	} else {
-		if err := t.addImpl(OriginGossip, txn); err != nil {
-			if !errors.Is(err, ErrAlreadyKnown) {
-				t.logger.Error("failed to add broadcasted txn", "err", err)
-			}
-		}
-	}
+// Sets the signer the pool will use
+// to check a transaction's signature.
+func (p *TxPool) AddSigner(s signer) {
```
lazartravica: Add is a verb used when we're adding something to an array, it should be `SetSigner` in this case

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
+func (p *TxPool) EnableDev() {
+	p.dev = true
 }
 
-// AddTx adds a new transaction to the pool and broadcasts it if networking is enabled
-func (t *TxPool) AddTx(tx *types.Transaction) error {
-	if err := t.addImpl(OriginAddTxn, tx); err != nil {
+// AddTx adds a new transaction to the pool (sent from json-RPC/gRPC endpoints)
+// and broadcasts it to the network (if enabled).
+func (p *TxPool) AddTx(tx *types.Transaction) error {
```
lazartravica, line 259: What do you think about renaming it to AddLocalTx?

txpool/queue.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	heap.Init(&q.queue)
+	return &q
+}
+
+// Empties the underlying queue
+// and returns the removed transactions.
+func (q *pricedQueue) clear() (
+	removed transactions,
```
lazartravica: we don't use the returned list of transactions anywhere

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
 
-	return uint64(len(t.index))
+	return primaries
 }
 
-// dropTx removes a transaction from the priced heap. [Thread-unsafe]
-func (t *txPriceHeap) dropTx(tx *types.Transaction) bool {
-	item, ok := t.index[tx.Hash]
+// Returns the account associated with the given address.
+func (m *accountsMap) from(addr types.Address) *account {
```
lazartravica: from is a bit misleading, I think renaming to `get` or `find` is more appropriate

txpool/txpool.go
lazartravica: You can move the slotGauge struct and all associated methods (& the slotsRequired fn) to a separate slot_gauge.go file

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+		a.setNonce(nextNonce)
 	}
 
-	return t.txs[i].price.Cmp(t.txs[j].price) >= 0
+	return promoted
 }
 
-type minTxPriceHeapImpl struct {
-	txPriceHeapImplBase
+// Lookup map used to find transactions present in the pool
```
lazartravica: You can move the lookupMap struct and all associated methods to a separate lookup_map.go file

txpool/txpool.go
lazartravica: P.S. you used the verb `load` for the same thing with the lookupMap

txpool/txpool.go
lazartravica: It might make sense to rename this to addGossipTx, since it's the same kind of wrapper as AddLocalTx (AddTx)

txpool/txpool.go
```diff
@@ -192,576 +200,280 @@ func NewTxPool(
...
-			pendingTxs[sortedPricedTx.from] = make(map[uint64]*types.Transaction)
-		}
-		pendingTxs[sortedPricedTx.from][sortedPricedTx.tx.Nonce] = sortedPricedTx.tx
-	}
-	t.pendingQueue.lock.Unlock()
-	if !inclQueued {
-		return pendingTxs, nil
-	}
+	// fetch primary from each account
+	primaries := p.accounts.getPrimaries()
```
lazartravica, line 292: Perhaps nextInLine or head might be more descriptive than primary?

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+	account := p.accounts.initOnce(newAddr, stateNonce)
 
-// Required method definitions for the standard golang heap package
+	return account
+}
 
-func (t *txHeap) Peek() *types.Transaction {
-	if len(*t) == 0 {
-		return nil
+/* QUERY methods */
```
lazartravica: Feel free to move all the query methods to query.go

txpool/txpool.go
lazartravica: Feel free to move all accountsMap structs & account structs methods to a separate file: account.go

lazartravica (changes requested):
Excellent job on the tests!

Please take a look at my comments.

I am only really worried about the shutdownCh, the rest is cosmetical, but will help somebody reading the txpool code in the future.

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1701 @@
...
 
-var forks = &chain.Forks{
-	Homestead: chain.NewFork(0),
-	Istanbul:  chain.NewFork(0),
-}
-
 const (
 	defaultPriceLimit uint64 = 1
 	defaultMaxSlots   uint64 = 4096
+	validGasLimit     uint64 = 10000000
```
0xAleksaOpacic, line 22: Is this arbitrary number because this is not valid in term of block gas limit validation. Default value for genesis block gas limit size is 4712388 so we need to put value that si smaller than this but bigger then intrinsicGas

brkomir (changes requested):
Overall the code is very well written and it was a pleasure discovering how our new tx pool works in detail, really great job! I have left a bunch of comments mostly regarding the code organization, not its functionality.

txpool/txpool_test.go
0xAleksaOpacic: I will resolve this in my PR which is based to this branch. https://github.com/0xPolygon/polygon-sdk/pull/327/files

txpool/operator.go
dbrajovic: Conceptually, yes. 

Previously, there was one global promoted queue - now that queue is split up into several promoted queues, each belonging to some `account`. The resulting length from `.promoted()` is the sum of all the individual lengths.

txpool/txpool.go
dbrajovic: Used [here](https://github.com/0xPolygon/polygon-sdk/blob/712adc37d0f181cf0b02b710371a999c18d37119/txpool/txpool.go#L494-L501)

txpool/txpool.go
dbrajovic: I am not entirely sure what was the original reasoning behind this, but this is leftover from the previous implementation. If there is no harm in non-sealers broadcasting transactions, I'll remove the redundant `if` check.

txpool/txpool.go
dbrajovic: Ths is required by the proto services - operator commands. (@zivkovicmilos )

txpool/txpool.go
dbrajovic: See [this](https://github.com/0xPolygon/polygon-sdk/pull/287#:~:text=Why%20do%20only%20sealers%20gossip%20transactions%3F) conversation. It's leftover logic (not entirely understandable to me either).

txpool/txpool.go
dbrajovic: I do agree that `AddLocalTx` is what the function is doing, but from the perspective of packages using this func, the notion of `local` vs `gossip` should not be of relevance. 

I vode on `AddTx` remaining the same, but changing `handleGossipTx` to `addGossipTx` since the pool is the only module that should be aware of `gossiped` transactions.

txpool/mock.go
dbrajovic: Ending a filename with `_test.go` will have go treat the file as if it had tests in it (which is not true for a bunch of files already in the codebase).

I suggest creating ` testing.go` file within the module where all the relevant mock methods/structs and helper methods would be defined.

txpool/queue.go
dbrajovic: While I admit this code is for readability purposes only, I don't think you necessarily _need_ to extend a type with methods just because it's possible. It's not an uncommon practice to rename types solely for the sake of readability.

txpool/txpool.go
dbrajovic: Not sure what the ideal name for this method is, but `getHeads` seems even more vague.

Perhaps a detailed description on what `primaries` are ?

https://github.com/0xPolygon/polygon-sdk/pull/287/commits/3d6ed7199d295213c680009c16c6ae3d533baf5b

txpool/txpool.go
dbrajovic: After consuting with @zivkovicmilos , non-sealers will never produce blocks, so there is no need for them to accept gossiped transactions. However, transactions added with `AddTx` should still be included, so the other (sealer) nodes can hear about them.

consensus/ibft/ibft_test.go
```diff
@@ -660,38 +659,49 @@ func (s *mockSyncer) Broadcast(b *types.Block) {
...
 	nonceDecreased        map[*types.Transaction]bool
 	resetWithHeaderCalled bool
 	resetWithHeadersParam []*types.Header
 }
 
-func (p *mockTxPool) ResetWithHeaders(headers ...*types.Header) {
-	p.resetWithHeaderCalled = true
-	p.resetWithHeadersParam = headers
+func (p *mockTxPool) Prepare() {
+
```
zivkovicmilos, line 689: Maybe add an explicit return here?

e2e/genesis_test.go
```diff
@@ -96,21 +97,18 @@ func TestCustomBlockGasLimitPropagation(t *testing.T) {
...
-		_, err = srv.TxnPoolOperator().AddTxn(context.Background(), &txpoolOp.AddTxnReq{
-			Raw: &any.Any{
-				Value: tx.MarshalRLP(),
-			},
-			From: types.ZeroAddress.String(),
-		})
-		if err != nil {
-			t.Fatalf("failed to add txn: %v", err)
-		}
+		_, err = srv.JSONRPC().Eth().SendRawTransaction(signedTx.MarshalRLP())
```
zivkovicmilos: Why the switch to JSON-RPC instead of using GRPC?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
-	locals []types.Address,
-	noLocals bool,
-	priceLimit uint64,
-	maxSlots uint64,
 	forks chain.ForksInTime,
 	store store,
 	grpcServer *grpc.Server,
 	network *network.Server,
 	metrics *Metrics,
+	config *Config,
```
zivkovicmilos, line 181: Can we move out this as a single param (struct)?

txpool/txpool.go
```diff
@@ -64,126 +76,122 @@ type signer interface {
...
-		locals:        newLocalAccounts(locals),
-		noLocals:      noLocals,
-		priceLimit:    priceLimit,
-		forks:         forks,
-		metrics:       metrics,
+	pool := &TxPool{
+		logger:      logger.Named("txpool"),
+		forks:       forks,
+		store:       store,
+		idlePeriod:  defaultIdlePeriod,
```
zivkovicmilos: We can remove idle period from the txpool, as it's not used anywhere

txpool/txpool.go
```diff
@@ -25,12 +23,14 @@ const (
 	defaultIdlePeriod = 1 * time.Minute
 	txSlotSize        = 32 * 1024  // 32kB
 	txMaxSize         = 128 * 1024 //128Kb
```
zivkovicmilos: What do you think about making these two configurable through flags?

txpool/txpool.go
```diff
@@ -823,466 +531,508 @@ func (t *TxPool) validateTx(
...
+//
+// Eligible transactions are all sequential in order of nonce
+// and the first one has to have nonce less (or equal) to the account's
+// nextNonce.
+func (a *account) promote() uint64 {
+	a.promoted.lock(true)
+	a.enqueued.lock(true)
+	defer func() {
+		a.promoted.unlock()
+		a.enqueued.unlock()
```
zivkovicmilos: Please reverse the order of unlocking, as this can lead to a deadlock (free the enqueued lock first, then the promoted one)

txpool/txpool.go
dbrajovic: Old leftover code.

[Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/2fc81defe752ea385ea03ec20bff10f90b33b8e4)

e2e/genesis_test.go
dbrajovic: [Deleted](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/2fc81defe752ea385ea03ec20bff10f90b33b8e4)

txpool/mock.go
brkomir: Leaving a file without "_test" suffix will have go compiler treat the file as production code, which is also not true. 
> Ending a filename with _test.go will have go treat the file as if it had tests in it (which is not true for a bunch of files already in the codebase).

I don't understand what's your point here, what are the consequences since the code in this file is actually used only in tests

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

jsonrpc/txpool_endpoint.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8b5bc7a4f7d60d84450efaa170a7fdcbb7c3bebb)

e2e/txpool_test.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/1f0de81c24641bdcebeb32a72c8fe161e2a4e1b2)

txpool/txpool.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/b06dc527263bda9daa0e7c7abb1643dce37124fc)

txpool/txpool.go
dbrajovic: You're right. I corrected the implementation.

[Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/c3f89c58ae66586d6a520c5e0625a777c0f6d4d9)

txpool/txpool.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/c3f89c58ae66586d6a520c5e0625a777c0f6d4d9)

txpool/txpool.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/c3f89c58ae66586d6a520c5e0625a777c0f6d4d9)

txpool/txpool.go
dbrajovic: I believe eth_endpoints implicitly call these methods (related to the multi-functional `blockchain` interface).

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/034e1c7dec054cece5fa24dad0deb98f77404b77)

txpool/queue.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/2340045d39b3529070b1e0c708773754ea484e40)

txpool/txpool.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/034e1c7dec054cece5fa24dad0deb98f77404b77)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: Agreed.

[Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/034e1c7dec054cece5fa24dad0deb98f77404b77)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/txpool.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/8f8c2a793b02c4744e33f3598b871ea3b1aa73e9)

txpool/queue.go
brkomir: I don't think you necessarily need to create a type with no specific characteristics/behavior just because it's possible and I would argue that the benefit of improved readability is insignificant because I don't know that `transactions` is simply a slice until go to the type definition and see nothing specific about it. Not a big deal in general, it's up to you to decide the destiny of this type :)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/355a0b10c85512a4f95eed8bd4a0910f7c377dfa)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/355a0b10c85512a4f95eed8bd4a0910f7c377dfa)

consensus/ibft/ibft_test.go
dbrajovic: IDE would mark it as a redundant return statement.

txpool/txpool.go
dbrajovic: Maybe change the `Peek()` method to `Next()`. Explanation:

To the outside caller, `Peek` really does not mutate the pool in any way. The `executables` queue is just an intermediary struct the pool uses to always have the best selected transaction available before the next call `Peek`.
Mutating this queue doesn't reflect on the pool until `Pop`, `Demote` or `Drop` are called. 

Additionally, if `peek` were called instead of `pop` here, the `pricedQueue` would have 2 identical methods called `peek` and `pop`.

txpool/txpool.go
dbrajovic: I felt like writing a unit test for this method would be overkill, since all the important logic for this call is happening in `resetAccounts`. Everything before that is just parsing the `headers` into a map of nonces (`map[type.Address]uint64`).

txpool/txpool.go
dbrajovic: Same explanation as for `ResetWithHeaders`

txpool/txpool_test.go
dbrajovic: ~~I agree that conditional branching in unit tests is not recommended, but with the `t.Run()` approach, the test would be 10 times longer with identical logic in each case except for a few lines of code. This switch case is just meant to avoid that. Everything that's common to these cases is already achieved with `newTestPool`.~~

[Refactored](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/57555e055d42c6bb4e090eb6435f731f95b471b3)
Praise the new linter (that forced the refactor).

txpool/txpool.go
dbrajovic: [Improved](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/b06dc527263bda9daa0e7c7abb1643dce37124fc)

txpool/txpool.go
dbrajovic: [Improved](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/b06dc527263bda9daa0e7c7abb1643dce37124fc)

txpool/txpool.go
dbrajovic: [Improved](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/b06dc527263bda9daa0e7c7abb1643dce37124fc)

zivkovicmilos (approved):
I've gone through the implementation again, left some comments that we should discuss 👍 

Thank you for pushing through this PR, it's a huge step forward for the PSDK 🚀 

This PR has been tried and tested multiple times, so I can certify it as working correctly and effectively ✅

txpool/txpool.go
dbrajovic: I can address this readability in a separate PR.

txpool/txpool.go
dbrajovic: I suggest a separate PR for this.

txpool/txpool_test.go
```diff
@@ -1,1125 +1,1701 @@
...
+			txs      transactions
+			newNonce uint64
+			expected result
+		}{
 			{
-				From:     addr1,
-				Gas:      validGasLimit,
-				GasPrice: big.NewInt(2),
-				Value:    big.NewInt(0),
+				name: "prune all txs",
```
brkomir: Please consider improving these test names. For example, "prune all txs from the promoted queue with nonce less than new nonce" or something similar which is more descriptive and provides more information in case when we see this test failing. This could be improved in other places but I will leave only this comment here.

txpool/txpool.go
brkomir: This is the unused one `ErrUnderpriced`

txpool/slot_gauge.go
```diff
@@ -0,0 +1,40 @@
...
+	atomic.AddUint64(&g.height, ^(slots - 1))
+}
+
+// slotsRequired calculates the number of slots required for given transaction(s).
+func slotsRequired(txs ...*types.Transaction) uint64 {
+	slots := uint64(0)
+	for _, tx := range txs {
+		slots += func(tx *types.Transaction) uint64 {
+			return (tx.Size() + txSlotSize - 1) / txSlotSize
+		}(tx)
```
Kourin1996: why function?

Kourin1996 (commented):
Looks great to me!

txpool/txpool.go
dbrajovic: This is just a parsing method. I don't think there is a real need to make it its own unit test.

txpool/txpool_test.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/bae5c576e85356ce5aa9d9963a52214d7cfbe8e7)

txpool/txpool.go
dbrajovic: [Reversed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/65dd59da6bd4b4b884127e3a5d08d5e18ae0375a)

e2e/genesis_test.go
dbrajovic: I reverted it to gRPC calls in a couple of places. 

https://github.com/0xPolygon/polygon-sdk/pull/287/commits/e18a2cbd44e9fc2dd33872da1bbe3e921261e56a

txpool/slot_gauge.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/1bf4c3f8854385fc201d8ff1d00d0fc2652ee272)

txpool/txpool.go
dbrajovic: Fixed

https://github.com/0xPolygon/polygon-sdk/pull/287/commits/35caff812f58d91df6a22d5725b3e0c8ced804c2
https://github.com/0xPolygon/polygon-sdk/pull/287/commits/3d94aa9970fbb283a6f4ff28519b0814c7515d74

server/server.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/35caff812f58d91df6a22d5725b3e0c8ced804c2)

txpool/txpool.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/b76090facfa82d40d2c261d768aa9d963b3d81eb)

txpool/txpool.go
dbrajovic: https://github.com/0xPolygon/polygon-sdk/pull/287/commits/3d6ed7199d295213c680009c16c6ae3d533baf5b

txpool/txpool.go
dbrajovic: I've changed the `Peek` method to `Next`. It's still the same thing, but I feel it takes away the confusion from seeing `peek` and `pop` in the same code. 

https://github.com/0xPolygon/polygon-sdk/pull/287/commits/6c2a55803a073c651bfbc3af19cef206ec5215ee

txpool/mock.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/36a69b206a25c717e6549dc8949c4f4b069e862b)

txpool/queue.go
dbrajovic: If the cost of not having sore eyes is one type definition lookup, I'm more that willing to pay for it.

consensus/dev/dev.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/28c0f185799eaa10ba91d130bda50803557daf4f)

consensus/ibft/ibft.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/28c0f185799eaa10ba91d130bda50803557daf4f)

txpool/txpool.go
dbrajovic: Same behaviour as the existing `ErrAlreadyKnown` test.

txpool/txpool.go
dbrajovic: I'd wait for reintroduction of `locals` feature. Right now, it's the same as adding a local tx.

txpool/txpool_test.go
dbrajovic: No particular reason, other than to indicate it's just a prep step.

txpool/txpool_test.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/57555e055d42c6bb4e090eb6435f731f95b471b3)

txpool/txpool.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/3d94aa9970fbb283a6f4ff28519b0814c7515d74)

txpool/txpool_test.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/cfde682a22bfc294a6c783b0beac700d26cdcd95)

jsonrpc/txpool_endpoint.go
lazartravica: inline as in `pendingRpcTxns[addr.String()][nonceStr] = fmt.Sprintf....`

jsonrpc/txpool_endpoint.go
lazartravica: inline as in `pendingRpcTxns[addr.String()][nonceStr] = fmt.Sprintf....`

txpool/txpool.go
lazartravica: It's good that we know the cause. In the future this might be a problem if you have a node that's only used for JSON-RPC to query the state of the txpool. Let's create an issue for that for later @dbrajovic

txpool/txpool.go
lazartravica: That's a good argument 👍

txpool/txpool.go
lazartravica: Let's simply create an issue for addressing this in the future (and investigate why do we enforce such a limit, when I think Ethereum doesn't).

lazartravica: @dbrajovic

txpool/txpool_test.go
lazartravica: https://medium.com/golangspec/scopes-in-go-a6042bb4298c

☝️ might confuse someone into thinking we wanted to re-declare a variable in the inner scope. let's get rid of it if there's no reason.

txpool/txpool.go
lazartravica: @dbrajovic let's create an issue for that 👍

txpool/queue.go
lazartravica: I'm with @brkomir on this one. We should generally adopt that approach in our code-style: no type definitions that have no special methods on them.

Providing functional abstractions is a good thing, but this is not one of those cases, as we are simply hiding from plainsight the important information about a type being a slice of pointers to transactions.

txpool/txpool.go
dbrajovic: Current behaviour is no different than adding a local transaction.

jsonrpc/txpool_endpoint.go
dbrajovic: [Inlined](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/4ebbb3feb7264636d48dec39d0916b98cc296f2a)

jsonrpc/txpool_endpoint.go
dbrajovic: [Inlined](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/4ebbb3feb7264636d48dec39d0916b98cc296f2a)

txpool/queue.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/c6177c8b9242560887009e612904cee5a248bbd6)

txpool/txpool_test.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/06335c8ed50c3f30dbea46ac2c2e7932035b90f2)

txpool/txpool.go
brkomir: > I felt like writing a unit test for this method would be overkill, since all the important logic for this call is happening in resetAccounts. Everything before that is just parsing the headers into a map of nonces (map[type.Address]uint64).

I don't agree and I don't think it's overkill to unit test an untested method that occupies like 60+ lines. Maybe if you want to postpone this create a card on our issue tracking system

txpool/txpool.go
brkomir: This is introducing another confusion on the tx pool API, what is next and what is pop. Please revert to Peek or just add comments that explain this situation

txpool/txpool.go
brkomir: I disagree with you on this one, at least since you have a check with the sealing flag at the beginning. So I'm still standing with my opinion that tests should be added here. Regardless, if there are no differences why aren't those two methods refactored into a single one then? The point is that we shouldn't add new code which is not covered by unit tests.

txpool/txpool.go
brkomir: You should be thinking about uint tests in that way, they are used to assert that functionality is correct, so if you don't have them you can't be sure if the code is working as intended. Also, if we ever change this code inside allTxs for whatever reason, we won't be sure if we have broken anything if no tests are covering this. Consider the approach with TDD, you don't write a piece of code until you have a failing test for it. After you have written the unit tests, you are free to change and restructure that code however you like and you feel confident that you haven't broken anything as long as all the tests are passing.

txpool/txpool.go
brkomir: It's not the same behavior since you do not return an error here.

lazartravica (approved):
Let's wait for other comments to be resolved and we are good to go 👍

txpool/txpool.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/a8651abb29dc10ffd471129264898bb40a4811cc)

txpool/txpool.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/17adc69c6c78d2fc983e8fc1d84be193785dad7d)

txpool/txpool.go
dbrajovic: [Reverted](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/14d55e870cc39b2b6c5c76076811fe9e96bd240c)

Kourin1996 (approved):
LGTM 👏

txpool/txpool_test.go
```diff
@@ -1,1312 +1,1898 @@
...
-type mockSigner struct{}
+/* Single account cases (unit tests) */
+
+func TestAddTxErrors(t *testing.T) {
+	t.Run("ErrNegativeValue", func(t *testing.T) {
+		pool, err := newTestPool()
+		assert.NoError(t, err)
+
+		pool.EnableDev()
+		pool.SetSigner(crypto.NewEIP155Signer(100))
```
brkomir: Since this part of the code is repeating for all of the tests inside `TestAddTxErrors`, you can create a setup func and reuse it in all of the tests where applicable. Also, why is `EnableDev` necessary here?

txpool/txpool_test.go
```diff
@@ -1,1312 +1,1898 @@
...
+	slots    uint64
 }
 
-type mockSigner struct{}
+/* Single account cases (unit tests) */
+
+func TestAddTxErrors(t *testing.T) {
+	t.Run("ErrNegativeValue", func(t *testing.T) {
+		pool, err := newTestPool()
+		assert.NoError(t, err)
```
brkomir: this assert seems unnecessary

txpool/txpool.go
dbrajovic: [Added](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/ab4a5ec0a00bba97791b0f42e1e41820222acf89)

zivkovicmilos (approved):
No need to re-request an approval 🙂 

Approving again 👍

txpool/txpool_test.go
zivkovicmilos: Why do you think it's unnecessary?
It's just following standard code style, like the rest of the SDK, since `NewTxPool` can return an error

txpool/txpool_test.go
zivkovicmilos: There is already an issue created for pruning out this useless dev mode dependency from the TxPool, so we'll tackle it in some future sprint 👍 (next one)

I agree that @dbrajovic should just create a helper func inside the test for setting up the pool with the signer, since each test case uses the same block

dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-sdk/pull/287/commits/ee344186feb31d3e48a1ecba0c548ea69517fb2f)

txpool/txpool_test.go
brkomir: This is used in every single test, so if this assertion fails all of the tests will fail. We have to be certain that instantiating the txpool (with everything mocked out) in our tests shouldn't fail because this is a part of the test setup, not assertions.

brkomir: Resolving this conversation since this is of 0 importance right now.

brkomir (approved):

munna0908 (approved):
LGTM 🚀

PR #289, opened 2021-12-14
Fix/max peers limit

munna0908:
# Description

This PR will  fix an issue in identity exchange routine which checks the maxPeer count only on inbound connection(Ideally it should check on both inbound/outbound connections). This PR also increases the default `MaxPeers` count to `50` nodes 

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/command/helper/config.go
+++ b/command/helper/config.go
@@ -66,7 +66,7 @@ func DefaultConfig() *Config {
 		BlockGasTarget: "0x0", // Special value signaling the parent gas limit should be applied
 		Network: &Network{
 			NoDiscover: false,
-			MaxPeers:   20,
+			MaxPeers:   30,
 		},
 		Telemetry: &Telemetry{},
 		Seal:      false,
--- a/command/helper/config.go
+++ b/command/helper/config.go
@@ -66,7 +66,7 @@ func DefaultConfig() *Config {
 		BlockGasTarget: "0x0", // Special value signaling the parent gas limit should be applied
 		Network: &Network{
 			NoDiscover: false,
-			MaxPeers:   30,
+			MaxPeers:   50,
 		},
 		Telemetry: &Telemetry{},
 		Seal:      false,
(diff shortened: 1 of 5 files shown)


command/helper/config.go
```diff
@@ -66,7 +66,7 @@ func DefaultConfig() *Config {
 		BlockGasTarget: "0x0", // Special value signaling the parent gas limit should be applied
 		Network: &Network{
 			NoDiscover: false,
-			MaxPeers:   20,
+			MaxPeers:   30,
```
lazartravica: Increase to 50 as discussed

network/identity.go
```diff
@@ -64,19 +64,16 @@ func (i *identity) setup() {
 			peerID := conn.RemotePeer()
 			i.srv.logger.Trace("Conn", "peer", peerID, "direction", conn.Stat().Direction)
 
-			// limit by MaxPeers on incomming requests since we already limit
-			// the outgoing requests
```
lazartravica, line 68: @munna0908 Do you have any idea why this comment is here? How do we limit outgoing requests?

lazartravica (changes requested):

network/identity.go
munna0908: We check in run dial func [link](https://github.com/0xPolygon/polygon-sdk/blob/c0756d43fe3f921a9f45eaa52210c988f55b707b/network/server.go#L283), but our connection mechanism is 2 step process, we need to check in the later step aswell.

command/helper/config.go
munna0908: Sorry, Missed it. Here is the commit [link](https://github.com/0xPolygon/polygon-sdk/pull/289/commits/cdecd53d6c7a1454f58469b80d15b124f0e4f506)

lazartravica (approved):
LGTM!

Kourin1996 (approved):
LGTM 👍

zivkovicmilos (approved):
Looks good 💯

brkomir (approved):

PR #296, opened 2021-12-17
Backup and restore function

Kourin1996:
# Description

This PR adds `backup` and `restore` command to export blockchain data from running node and create blockchain from backup data.

```bash
$ polygon-sdk backup --grpc 127.0.0.1:10000 --data backup.dat [--from 0 --to 10000]
```

To restore blockchain data from this backup data, you will specify `--restore` flag in server command

```bash
$ polygon-sdk server ... --restore backup.dat
```

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


archive/backup.go
```diff
@@ -0,0 +1,163 @@
...
+
+	closeAndRemove := func(fs *os.File) {
+		if err := fs.Close(); err != nil {
+			logger.Error("an error occurred while closing file", "err", err)
+			return
+		}
+		if err = os.Remove(outPath); err != nil {
+			logger.Error("an error occurred while removing file", "err", err)
+		}
+	}
```
brkomir: This looks awfully similar to the code on lines 66-73? Can you try and make it more uniform?
For example, have two separate functions for close and remove, and call them accordingly every time they are needed?

archive/backup.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/0xPolygon/polygon-sdk/server/proto"
+	"github.com/0xPolygon/polygon-sdk/types"
+	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func CreateBackup(conn *grpc.ClientConn, logger hclog.Logger, targetFrom uint64, targetTo *uint64, outPath string) (uint64, uint64, error) {
```
brkomir: I do not understand what is this parameter "targetFrom"? As my understanding goes, we should backup the whole chain, not part's of it, am I right?
Apart from that, "targetTo" and "targetFrom" are referring to the block numbers, maybe we could rename them to "toBlock" and "fromBlocs" for better readability.

archive/backup.go
```diff
@@ -0,0 +1,163 @@
...
+	if targetTo != nil && *targetTo < uint64(status.Current.Number) {
+		// check the existence of the block when you have targetTo
+		resp, err := clt.BlockByNumber(ctx, &proto.BlockByNumberRequest{Number: *targetTo})
+		if err == nil {
+			block := types.Block{}
+			if err := block.UnmarshalRLP(resp.Data); err == nil {
+				// can use targetTo only if the node has the block at the specific height
+				return block.Number(), block.Hash(), nil
+			}
+		}
```
brkomir: what happens in the case when `err != nil`?

archive/backup.go
```diff
@@ -0,0 +1,163 @@
...
+	logger.Info("Wrote metadata to backup", "latest", to, "hash", toHash)
+	return err
+}
+
+func processExportStream(stream proto.System_ExportClient, logger hclog.Logger, writer io.Writer, targetFrom, targetTo uint64) (*uint64, *uint64, error) {
+	var from, to *uint64
+
+	getResult := func() (*uint64, *uint64, error) {
+		if from == nil || to == nil {
+			return nil, nil, errors.New("couldn't get any blocks")
```
brkomir, line 155: This error message seems too vague, maybe we can improve it for example by adding the info that nil parameters are not supported

server/system_service.go
```diff
@@ -144,3 +148,130 @@ func (s *systemService) PeersList(
...
+	if req.To != 0 {
+		if from >= req.To {
+			return errors.New("to must be greater than from")
+		}
+		to = &req.To
+	}
+
+	canLoop := func(i uint64) bool {
+		if to == nil {
+			current := s.s.blockchain.Header()
```
brkomir: s.s.blockchain... is too cryptic to read, change the second s to what It really represents

server/system_service.go
```diff
@@ -144,3 +148,130 @@ func (s *systemService) PeersList(
...
+	return nil
+}
+
+func (w *blockStreamWriter) reset() {
+	w.buf.Reset()
+	w.pendingFrom = nil
+	w.pendingTo = nil
+}
+
+func (w *blockStreamWriter) isBuffered() bool {
```
brkomir: this does not need to be a separate method

server/system_service.go
```diff
@@ -144,3 +148,130 @@ func (s *systemService) PeersList(
...
 }
+
+// BlockByNumberRequest implements the BlockByNumberRequest operator service
+func (s *systemService) BlockByNumberRequest(
+	ctx context.Context,
+	req *proto.BlockByNumberRequest,
+) (*proto.BlockResponse, error) {
+	block, ok := s.s.blockchain.GetBlockByNumber(req.Number, true)
+	if !ok {
+		return nil, errors.New("not found")
```
brkomir: add more information to the message, for example "block number <n> not found..."

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+			return 0, 0, err
+		}
+		if uint64(n) < payloadSizeSize {
+			// couldn't load required amount of bytes
+			return 0, 0, io.EOF
+		}
+		payloadSize := new(big.Int).SetBytes(payloadSizeBytes).Int64()
+		return uint64(payloadSize), uint64(payloadSizeSize), nil
+	}
+	return 0, 0, errors.New("expected arrray but got bytes")
```
brkomir: this arrrr sounds like some pirate stuff :)

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+
+// loadRLPArray loads RLP encoded array from input to buffer
+func (b *blockStream) loadRLPArray() (uint64, error) {
+	prefix, err := b.loadRLPPrefix()
+	if errors.Is(io.EOF, err) {
+		return 0, nil
+	} else if err != nil {
+		return 0, err
+	}
+	payloadSize, payloadSizeSize, err := b.loadPrefixSize(1, prefix)
```
brkomir: I was confused reading this. You call a method "loadPrefixSize" and it returns 2 uint values (which are not clear what they are from the method description), you would expect for one of them to be prefixSize, but both variables contain "payload" in its names. Please improve on code readability here

archive/restore_test.go
```diff
@@ -0,0 +1,296 @@
...
+		newCap       int
+	}{
+		{
+			name: "should expand and change size",
+			blockstream: &blockStream{
+				buffer: make([]byte, 0, 10),
+			},
+			expectedSize: 20,
+			newLen:       20,
+			newCap:       32,
```
brkomir: This test is failing...

helper/common/common.go
```diff
@@ -71,3 +73,11 @@ func createDir(path string) error {
 
 	return nil
 }
+
+// GetTerminationSignalCh returns a channel to emit signals by ctrl + c
+func GetTerminationSignalCh() <-chan os.Signal {
+	// wait for the user to quit with ctrl-c
+	signalCh := make(chan os.Signal, 4)
```
zivkovicmilos: Why 4?

archive/backup.go
zivkovicmilos: Please document what the return values of this func are, it's not clear from the get-go

command/backup/backup.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Run implements the cli.Command interface
+func (c *BackupCommand) Run(args []string) int {
+	flags := c.Base.NewFlagSet(c.GetBaseCommand(), c.Formatter, c.GRPC)
+
+	var out, rawFrom, rawTo string
+	flags.StringVar(&out, "out", "", "")
+	flags.StringVar(&rawFrom, "from", "0", "")
+	flags.StringVar(&rawTo, "to", "", "")
```
zivkovicmilos: You use `out` here, but in the flags you define `data`

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+	"math/big"
+	"os"
+
+	"github.com/0xPolygon/polygon-sdk/helper/common"
+	"github.com/0xPolygon/polygon-sdk/types"
+)
+
+var (
+	// Size of blocks to pass to WriteBlocks
+	chunkSize = 50
```
zivkovicmilos: Is there a possibility to customize this value to something higher?

50 blocks seems small, especially considering the time it takes to restore chains that have 100 000+ blocks

zivkovicmilos: It takes ~4mins to restore 100 000 blocks from 0 if we're using this `50` block batch value

zivkovicmilos (changes requested):
Overall it seems great 💯 

Thank you for also resolving some QOL stuff along the way 🙏 

I've tested out the backup procedure on a large chain (~450 000 blocks) and it took ~27s to back it up.
Screenshot 2021-12-26 at 14 13 30

The restore procedure is much, much slower, taking around ~4mins to restore 100 000 blocks.

Please check out my comments for noted issues.

Additionally, please provide a relevant docs PR since this adds new functionality 🙏 
The docs PR should modify the existing document `How to backup / restore` to include a guide on how to use this functionality, as well as the modifications to the CLI commands.

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if _, _, err = importBlocks(chain, blockStream); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// import blocks scans all blocks from stream and write them to chain
+func importBlocks(chain blockchainInterface, blockStream *blockStream) (uint64, uint64, error) {
```
brkomir: clarify what the return values represent. Also, this method is not covered in any unit tests

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func RestoreChain(chain blockchainInterface, filePath string) error {
+	fp, err := os.Open(filePath)
+	if err != nil {
+		return err
+	}
+	blockStream := newBlockStream(fp)
+
+	if _, _, err = importBlocks(chain, blockStream); err != nil {
```
brkomir: why do we need an unexported method which's return values are unused?

archive/restore.go
```diff
@@ -0,0 +1,262 @@
...
+		default:
+		}
+	}
+
+	return firstBlock.Number(), lastBlockNumber, nil
+}
+
+// consumeCommonBlocks consumes blocks in blockstream to latest block in chain or different hash
+// returns the first block to be written into chain
+func consumeCommonBlocks(chain blockchainInterface, blockStream *blockStream, shutdownCh <-chan os.Signal) (*types.Block, error) {
```
brkomir: this method is completely uncovered in unit tests

consensus/dev/dev.go
```diff
@@ -59,6 +59,11 @@ func Factory(
 	return d, nil
 }
 
+// Initialize initializes the consensus
```
brkomir, line 63: Does this mean we don't support restore for the dev mode? If so, don't forget to document this!

brkomir (changes requested):
Can you please provide more information/requirements for this feature? I am a bit confused why do we have a starting block number as the parameter for creating the backup, as I understood we should back up the chain from the genesis block onwards, maybe I'm missing something here...

archive/backup.go
Kourin1996: If a node gets an error here, it means the node doesn't have the block with the specific height. In this case, we will fetch blocks to its latest. Then return the value at the end of the function.

archive/backup.go
Kourin1996: I'm afraid I didn't get it. `from` and `to` here is stored the beginning and end height of the fetched blocks. So they are not parameters user or caller gives. meaning you couldn't get any blocks from the node when either is nil.

consensus/dev/dev.go
Kourin1996: I'm afraid I didn't get it. As you see the same method in IBFT, consensus doesn't restore from archive data in `Initialize`. restore is called from `NewServer` in `server/server.go` to be called always when you start node.

archive/backup.go
Kourin1996: Thanks. I added at 1f700bf72068dda1e41c5f022f0c54b95d9def40

archive/backup.go
Kourin1996: Thanks. I fixed at 1f700bf72068dda1e41c5f022f0c54b95d9def40

archive/restore.go
Kourin1996: I removed them from return value at 1f700bf72068dda1e41c5f022f0c54b95d9def40

archive/restore.go
Kourin1996: I removed return values at 1f700bf72068dda1e41c5f022f0c54b95d9def40

archive/restore.go
Kourin1996: `payloadSize` means the length of bytes containing each element in array while `payloadSizeSize` means the size of bytes to store `payloadSize`. The header of RLP array would be a byte or multiple bytes. But here no reason to distinguish between first byte and additional payloadSizeSize and merged them and renamed to `headerSize`

archive/restore.go
Kourin1996: Fixed at 1f700bf72068dda1e41c5f022f0c54b95d9def40 😂

command/backup/backup.go
Kourin1996: Thank you! I fixed at 1f700bf72068dda1e41c5f022f0c54b95d9def40

server/system_service.go
Kourin1996: Renamed field name at 1f700bf72068dda1e41c5f022f0c54b95d9def40

server/system_service.go
Kourin1996: I inlined this process at 1f700bf72068dda1e41c5f022f0c54b95d9def40

archive/restore_test.go
Kourin1996: Fixed at bb89be9bc57a60d895d281abfc2e2552832ad627. Shouldn't have checked new capacity because behaviour of `appends` is not same always

archive/backup.go
Kourin1996: `targetFrom` and `targetTo` are the range user prefer to get. If user specify them, then will get blocks from 0 to latest as default

Kourin1996: Renamed at a99128e47b09c01fa0f66dd2d0b3b9c857a69775

archive/restore.go
Kourin1996: Added at 7ddff6276f8210ccd2075ce6a89f789c034412a1

Kourin1996:
@zivkovicmilos @brkomir Thank you for your reviews! I fixed the codes and reply to some comments. I appreciate your help

ferranbt:
Hi, I am not sure if I follow how does this work and which is the workflow here. You can backup all the blocks (in a range) to some data file and then you can restore all the blocks to a running client? A couple of questions here:

- What happens if the backup file gets too big? Are we going to put logic to split the output in different files? Then we need to include a metadata file to organize the backup files.
- How fast can we move blocks to restore? I have done this for Ethereum mainnet and full blocks with transactions over Grpc is not as fast as it seems even with some hard code optimizations (i.e. streams of raw bytes instead of sending blocks).
- Are we restoring data into a running client? Is the syncing process being stopped? When is it enabled again? What happens if you restore a backup from a range that already existed?

Kourin1996:
> Hi, I am not sure if I follow how does this work and which is the workflow here. You can backup all the blocks (in a range) to some data file and then you can restore all the blocks to a running client? A couple of questions here:

Hello, @ferranbt. Basically, this PR contains two functions: 
* Fetching blocks in the range (0 to latest as default) from running node via gRPC and creating binary file (metadata + RLP encoded blocks)
* Restoring this generated binary on the beginning of the server if the file path is given. This restoring is done before consensus, gRPC, JSON-RPC and network layer begin.

> What happens if the backup file gets too big? Are we going to put logic to split the output in different files? Then we need to include a metadata file to organize the backup files.

* Currently, this PR doesn't contain such feature of saving/restoring to/from multiple files.

> How fast can we move blocks to restore? I have done this for Ethereum mainnet and full blocks with transactions over Grpc is not as fast as it seems even with some hard code optimizations (i.e. streams of raw bytes instead of sending blocks).

I tried with 10000 blocks without/with transactions. The result is as follows:

Screenshot 2021-12-30 at 13 21 19

Please note that PolygonSDK doesn't have the way to optimizing syncing like fast-sync in geth as you may know. Which means restoring is done by processing blocks one by one.

> * Are we restoring data into a running client? Is the syncing process being stopped? When is it enabled again? What happens if you restore a backup from a range that already existed?

Restoring is done before consensus begins. Currently, this PR doesn't contain the function to restore blocks to the running node. As you said, I think blocking is necessary in case of restoring in running node. If the node has chains already, node restore only if the latest hash in metadata of binary file doesn't equal to the hash at the same height in the chain.

Thank you for asking. I would appreciate if you could share some good suggestions (if you had) :)

consensus/dev/dev.go
Kourin1996: @brkomir I confirmed backup and restore worked with dev mode.

mrwillis:
Just playing devils advocate here. We essentially do the same process by just shutting down the SDK safely, and then simply zipping the `blockchain` and `trie` folders. It's very fast and just limited by compression and movement of the file. What is the purpose of this feature?

Kourin1996:
> Just playing devils advocate here. We essentially do the same process by just shutting down the SDK safely, and then simply zipping the `blockchain` and `trie` folders. It's very fast and just limited by compression and movement of the file. What is the purpose of this feature?

@mrwillis As you said here, you need to stop node once to create backup and access to the instance as well. With this approach, you don't verify blocks.

archive/backup.go
brkomir: I think I didn't understand the code well enough, will resolve the conversation.

archive/restore.go
brkomir: `loadPrefixSize` method in its name contains "prefix", so I relate it to some operation that will result in a prefix value as the return parameter, but here it returns two values that describe payload encoding instead. What I am trying to say is that the name of the method is misleading

brkomir (approved):
The code looks ok 👍 
Please just address the remaining conversations.

zivkovicmilos (approved):
Thank you for resolving all of the conversations, and for providing these stats 🙏 

Looks great 🚀

lazartravica (approved):
LGTM!

archive/restore.go
lazartravica: @Kourin1996 please address these comments

server/system_service.go
lazartravica: @Kourin1996 please address these comments

PR #299, opened 2021-12-20
Transactions received during the bulk sync process remain stuck forever in txpool

### LINKED ISSUE #285: Syncing validators getting remote txs stuck in queued
### Description

When testing tx spamming scripts on our testnet, we've observed non-fully-synced validators to reach a state where we see many txs stuck in `queued` (accountQueue of the txpool). Even once fully synced, these txs remain stuck in queued.

As the txpool cleanup only occurs when a validator has reached CommitState (when ResetWithHeader() gets called), my guess is that the queued txs can't all be processed once the validator reaches CommitState.

Edit: It seems that `pruneAccountTx()` is never called while syncing and the last reported state for the validator we were testing on seemed to be `RoundChangeState` -- we did assume that as it is syncing it should be in SyncState and hence the callback passed in to `WatchSyncWithPeer` should have handled the pruning.

### Your environment

* OS and version `Ubuntu 20`
* version of the Polygon SDK `3941ec8 `

### Steps to reproduce

1. Create 8-validator testnet. Ensure 7/8 are fully synced and 1/8 is syncing from block 0.
2. Run script that spams txs (hundreds) against a fully synced node
3. After the script has finished, observe that all nodes except the syncing validator all return an empty `queued` when calling `txpool_content`
4. Observe that even when the syncing node is fully synced, the queued txs remain requiring a restart

### Expected behaviour

We should not see any stuck queued txs, especially after the node is fully synced. 

### Actual behaviour

We see many queued txs on the syncing node.

### Proposed solution

Perhaps validators could avoid adding newly-gossiped txs if in SyncState.

brkomir:
# Description

This PR solves the issue described in the title and detected in GH issue #285 

# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (approved):
Looks good to me 👍

zivkovicmilos (approved):
Looks good 💯

dbrajovic (approved):
Will integrate refactor changes with the new `ResetWithHeaders` function.

LGTM 👍

melekes (approved):
👍

PR #302, opened 2021-12-20
Added network metrics to list currently connected peers

dankostiuk:
# Description

Followed the prometheus metrics convention discussed in https://github.com/0xPolygon/polygon-sdk/pull/216 to add a new `metrics.go` to track network-related metrics. I added the ability to track currently-connected peers by setting the gauge value each time we hit addPeer or delPeer.

# Changes include

# Checklist

- [x] I have tested this code


zivkovicmilos (changes requested):
Looks great 💯 

Please resolve the test panics (you probably haven't initialized the metrics struct somewhere in the tests) 👍

zivkovicmilos:
@munna0908 

It would be great if we could update the prometheus metrics on the docs repo to include this change as well

zivkovicmilos (approved):
Looks great 💯 

Thank you for resolving the failing tests, and for the contribution of course 🙏

network/server.go
```diff
@@ -370,6 +374,7 @@ func (s *Server) addPeer(id peer.ID) {
 	s.peers[id] = p
 
 	s.emitEvent(id, PeerConnected)
+	s.metrics.Peers.Set(float64(len(s.peers)))
```
munna0908, line 396: @dankostiuk I would suggest to acquire peersLock before reading the length. @zivkovicmilos Whats your take ?

munna0908 (changes requested):

network/server.go
zivkovicmilos: You're right 💯 

@dankostiuk please call s.numPeers() since `s.peers` requires concurrent access.

dankostiuk: Just made that change but it looks like addPeer/delPeer were already acquiring `s.peersLock.Lock()` - unless I am missing something..

munna0908: @dankostiuk  Yeah you're right, I missed it. Please revert back

munna0908 (approved):
LGTM 🚀

Kourin1996 (approved):
Looks good. Thank you for the PR 🙏

lazartravica (approved):

zivkovicmilos:
Hey @dankostiuk,

Please resolve the merge conflicts and pull the latest changes from develop, so we can merge out this PR 🚀

PR #293, opened 2021-12-16
Add support for the eth_syncing endpoint

zivkovicmilos:
# Description

This PR adds support for querying the bulk sync status of a node using the `eth_syncing` endpoint.

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [ ] I have updated the README and other relevant documents (guides...)
- [x] I have added sufficient documentation both in code, as well as in the READMEs

PR #306, opened 2021-12-21
Add progress of restore in eth_syncing

Kourin1996:
# Description

This PR adds the progression of restore feature in `eth_syncing` json rpc endpoint.

Awaiting PR #293, #296 before merging.
The branch of this PR forked from latest develop (e61ea129740d9b510fc3d6a9a2bb52826a93f9ae) and merged the changes of the both PRs first.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


archive/backup.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return 0, 0, err
+	}
+
+	writeMetadata(fs, logger, to, toHash)
+	resFrom, resTo, err := processExportStream(stream, logger, fs, targetFrom, to)
+
+	fs.Close()
+	if err != nil {
+		os.Remove(outPath)
```
melekes: unhandled error? or you don't care?

archive/backup.go
```diff
@@ -0,0 +1,145 @@
...
+	return uint64(status.Current.Number), types.StringToHash(status.Current.Hash), nil
+}
+
+func writeMetadata(writer io.Writer, logger hclog.Logger, to uint64, toHash types.Hash) error {
+	metadata := Metadata{
+		Latest:     to,
+		LatestHash: toHash,
+	}
+	_, err := writer.Write(metadata.MarshalRLP())
+	logger.Info("Wrote metadata to backup", "latest", to, "hash", toHash)
```
melekes: if err != nil { // success logging } ?

archive/backup.go
```diff
@@ -0,0 +1,145 @@
...
+	return err
+}
+
+func processExportStream(stream proto.System_ExportClient, logger hclog.Logger, writer io.Writer, targetFrom, targetTo uint64) (*uint64, *uint64, error) {
+	var from, to *uint64
+
+	getResult := func() (*uint64, *uint64, error) {
```
melekes, line 155: how do you justify this function? wouldn't it be better to inline this since a) it's only called in one place b) that way a developer doesn't need to jump 50 lines up

archive/backup.go
Kourin1996: Hello, Thank you for your kind suggestion. That was my fault and I should have cared about this as you said. I've added an error handling at this line at this commit d4e7bc76ce1ae4ae5c5f7c616c2fe64340109a51.

archive/backup.go
Kourin1996: Hello, Thank you for your kind suggestion. I've added an error handling and have the logger output only when error doesn't occur at this commit d4e7bc76ce1ae4ae5c5f7c616c2fe64340109a51.

archive/backup.go
Kourin1996: Hello, Thank you for your kind advice.
With all due respect, I've not wanted to inline this function because I wanted to focus on putting the flow of creating backup and initializing file and gRPC. I see your point, but `CreateBackup` will have long codes (+100 lines totally) and it will be hard for other developers to understand.

helper/progress/chain.go
```diff
@@ -0,0 +1,125 @@
...
+		case event := <-eventCh:
+			if event.Type == blockchain.EventFork {
+				continue
+			}
+
+			if len(event.NewChain) == 0 {
+				continue
+			}
+
+			pw.UpdateCurrentProgression(event.NewChain[0].Number)
```
zivkovicmilos: I'm just noticing it now, do you think it is incorrect to assume that NewChain has a single block number?

zivkovicmilos (approved):
Looks great to me 💯 

I've left some comments on PR #296 that apply here since this is just a branch off with new features.

Please look at the comment I've left behind, we should discuss it.

Also, great work on merging out the bulk sync functionality and the backup / restore functionality 👍

helper/progress/chain.go
Kourin1996: @zivkovicmilos You're right. I fixed at a5607d0e08a12df4ef44a5949d39ceb63a134b61

lazartravica (approved):
Nicely done @Kourin1996 , what is your plan with this and the other backup PR?

PR #307, opened 2021-12-21
Don't treat the already known txn message as an error in gossip

zivkovicmilos:
# Description

This PR removes the useless `failed to add broadcasted txn err= already known` logs.

# Changes include

# Checklist

- [x] I have tested this code


brkomir (approved):

dbrajovic (approved):

txpool/txpool.go
```diff
@@ -325,7 +325,9 @@ func (t *TxPool) handleGossipTxn(obj interface{}) {
 		t.logger.Error("failed to decode broadcasted txn", "err", err)
 	} else {
 		if err := t.addImpl(OriginGossip, txn); err != nil {
-			t.logger.Error("failed to add broadcasted txn", "err", err)
+			if !errors.As(err, &ErrAlreadyKnown) {
```
Kourin1996: I don't want to recommend using `errors.As` often. `errors.As` is not safe and it throws panic if you give `nil` to 2nd args. If `addImpl` returns ErrAlreadyKnown which is defined as constant value, you can use `errors.Is` instead of `errors.As`

```go
errors.Is(err, ErrAlreadyKnown)
```

In short, `errors.Is` checks if given error (1st argument) has the specific error (2nd argument) inside while unwrapping. On the other hand, `errors.As` check if given error has the error which implements the specific interface given in 2nd argument and set the error to 2nd argument. `errors.As` uses type assertion, then it's not safe.

Kourin1996 (changes requested):

txpool/txpool.go

fa84a7b

Kourin1996 (approved):
Thank you for the fix. Looks great 👌

PR #310, opened 2021-12-22
Resolve deadlock in promotedTxnCleanup

### LINKED ISSUE #291: Node's txpool stuck when spammed with txs
### Description

We've been running stress tests against our testnet using a spamming tool we built (https://github.com/mrwillis/eth-txn-spam) and managed to halt some of our nodes when spammed with txs.

Looking in our logs, it appears that the txpool blocks the consensus state-change cycle for every node we target. I suspect that when flooded with txs, the txpool somehow blocks the ibft consensus from continuing past CommitState/ValidateState (likely the accountQueue or pendingQueue lock isn't being released) halting the consensus. 

Another thing we noticed is that while rpc requests to methods such as `eth_getBlockByNumber` return a response as expected -- sending a request to any txpool-related endpoint such as `txpool_content` will hang indefinitely, leading us to believe that a txpool reference is blocking.

We have been able to reproduce this issue using the `eth-txn-spam` tool but will try to use PSDK's loadboat to halt the chain in a similar manner - I'll update this issue with any additional info we find.

### Your environment

* OS and version `Ubuntu 20`
* version of the Polygon SDK `e7b9714 `

### Steps to reproduce

1. Create testnet with 4 validators and 1 non-sealing node
2. Using https://github.com/mrwillis/eth-txn-spam, run most basic type of test within that repo (mode=0 / regular transfer) with RANGE 200 and PARALLELISM 10 targeting the RPC url of a validator
    1. Can consistently repro when running on AWS `c4.4xlarge` Ubuntu 20 instance
3. Targeted node then gets stuck, and can no longer write blocks even after stopping the test.
4. RPC requests to txpool methods such as `txpool_content` no longer receive a response.

### Expected behaviour

Sending many requests to a node should not halt its IBFT consensus as this could lead to a chain halt.

### Actual behaviour

The targeted node's txpool gets stuck, blocking the IBFT layer - as a result the node cannot write new blocks.

zivkovicmilos:
# Description

This PR resolves a deadlock situation that can occur inside `promotedTxnCleanup`.

Scenario:
* Thread A enters `promotedTxnCleanup` and grabs the promoted queue lock
* Thread A still hasn't grabbed the gauge lock in the callback method when pruning the txn
* Thread B enters `processSlots`, grabs the gauge lock since it's free
* Thread A executes the callback method inside `promotedTxnCleanup`, which attempts to grab the lock for the gauge, but blocks since Thread B has it
* Thread B attempts to grab the lock for the pending queue when dropping a txn in `processSlots`, but fails since Thread A is holding it

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

This deadlock occurs on highly concurrent environments, found using @mrwillis and @dankostiuk's tool:
https://github.com/mrwillis/eth-txn-spam

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -583,28 +583,32 @@ func (t *TxPool) promotedTxnCleanup(
 	t.pendingQueue.lock.Lock()
 
 	// Find the txns that correspond to this account
-	droppedPendingTxs := 0
+	droppedPendingTxs := make([]*types.Transaction, 0)
 	for _, pendingQueueTxn := range t.pendingQueue.index {
 		// Check if the txn in the promoted queue matches the search criteria
 		if pendingQueueTxn.from == address && // The sender of this txn is the account we're looking for
 			pendingQueueTxn.tx.Nonce < stateNonce { // The nonce on this promoted txn is invalid
 			// Transaction found, drop it from the pending queue
 			if dropped := t.pendingQueue.dropTx(pendingQueueTxn.tx); dropped {
-				// Update the log data
-				droppedPendingTxs++
-				t.logger.Debug(
-					fmt.Sprintf(
-						"Dropping promoted txn [%s]",
-						pendingQueueTxn.tx.Hash.String(),
-					),
-				)
-
-				cleanupCallback(pendingQueueTxn.tx)
+				// Save the transaction as dropped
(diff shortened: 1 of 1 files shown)


brkomir (approved):
Great job catching this one!

Kourin1996 (approved):
Looks nice. Thank you for the investigation.

Vuksan (approved):

PR #313, opened 2021-12-24
Refactor syncer unit test to not use go routines dangerously

zivkovicmilos:
# Description

This PR fixes a unit test in the syncer package where a go routine could be writing to a closed channel after the test finishes.

# Changes include

# Checklist

- [x] I have tested this code


protocol/syncer_test.go
```diff
@@ -287,26 +287,23 @@ func TestWatchSyncWithPeer(t *testing.T) {
...
+			syncer.WatchSyncWithPeer(peer, func(b *types.Block) bool {
+				if time.Now().After(endSyncTime) {
+					// Timeout
+					return true
+				}
+				// sync until latest block
+				return b.Header.Number >= latestBlock.Header.Number
+			})
+
+			if tt.synced {
```
brkomir: This is inherently wrong, since tt.synced is something that is set in the test specification struct. Please fix the assertion logic

brkomir (changes requested):

protocol/syncer_test.go
zivkovicmilos: You're right 👍 

Fixed the asserts in the original test:

ddc047b

brkomir (approved):

Kourin1996 (approved):
Looks good :)

PR #314, opened 2021-12-24
Refactor the network tests

zivkovicmilos:
# Description

This PR refactors the tests in the networking package, fixes bugs and adds new coverage tests.

Additionally, this PR resolves all pending linting errors (3937fbc) and modifies the GH Actions workflow file for the linter (886c4d5)

Encountered issues:
* Servers were using taken ports in concurrent environments in tests
* Invalid event waits within tests
* Low timeouts set for test completion, causing tests to fail even though they succeeded

## Failed linting

Currently, the linting checks fail on GH actions because some structures are unused due to test skips.
This PR should wait for https://github.com/0xPolygon/polygon-sdk/pull/312, which contains the libp2p package updates before unskipping the tests and fixing linting errors that arise from this.

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Due to a known and fixed libp2p bug (https://github.com/libp2p/go-libp2p/issues/1090), some test have been skipped for now:
* `TestSimpleGossip`
* `TestSyncer_PeerDisconnected`

The Polygon SDK is running version `v0.12.0` of libp2p, where this issue is still present.

These tests will be un-skipped as soon as https://github.com/0xPolygon/polygon-sdk/pull/312 is merged (after the libp2p package is updated to `v0.14.0`

EDIT: Tests un-skipped ✅


consensus/ibft/snapshot_test.go
```diff
@@ -711,11 +710,6 @@ func TestSnapshot_ProcessHeaders(t *testing.T) {
 						})
 					}
 					if !resSnap.Equal(snap) {
-						fmt.Println("-- wrong result --")
-						fmt.Println(resSnap.Set)
-						fmt.Println(snap.Set)
-						fmt.Println(resSnap.Votes)
-						fmt.Println(snap.Votes)
 						t.Fatal("bad")
```
brkomir, line 713: improve failure message

e2e/transaction_test.go
```diff
@@ -246,7 +244,7 @@ func TestEthTransfer(t *testing.T) {
 
 			expectedSenderBalance := previousSenderBalance
 			if testCase.shouldSucceed {
```
brkomir, line 247: the same if clause exists on line 258, try to unify it with this one

helper/keccak/keccak.go
```diff
@@ -22,7 +22,7 @@ type Keccak struct {
 // WriteRlp writes an RLP value
 func (k *Keccak) WriteRlp(dst []byte, v *fastrlp.Value) []byte {
 	k.buf = v.MarshalTo(k.buf[:0])
-	k.Write(k.buf)
+	_, _ = k.Write(k.buf)
```
brkomir: why did you add these "_" blank identifiers? the second variable is of error type, should we handle it somehow? same applies for lines 42 and 48

helper/keccak/pool.go
```diff
@@ -32,7 +32,7 @@ func (p *Pool) Put(k *Keccak) {
 // Keccak256 hashes a src with keccak-256
 func Keccak256(dst, src []byte) []byte {
 	h := DefaultKeccakPool.Get()
-	h.Write(src)
+	_, _ = h.Write(src)
```
brkomir: If all that we wanted to achieve with this is to indicate to linter that this should not be reported as an issue, maybe the best option is to use //nolint on the line above

jsonrpc/dispatcher.go
```diff
@@ -188,7 +188,7 @@ func (d *Dispatcher) HandleWs(reqBody []byte, conn wsConn) ([]byte, error) {
 	if req.Method == "eth_subscribe" {
 		filterID, err := d.handleSubscribe(req, conn)
 		if err != nil {
-			NewRpcResponse(req.ID, "2.0", nil, err).Bytes()
+			_, _ = NewRpcResponse(req.ID, "2.0", nil, err).Bytes()
```
brkomir: potential unhandled error?

network/discovery_test.go
```diff
@@ -16,80 +16,91 @@ func discoveryConfig(c *Config) {
...
-	srv1 := CreateServer(t, nil)
+	servers, createErr := createServers(2, nil)
+	if createErr != nil {
+		t.Fatalf("Unable to create servers, %v", createErr)
+	}
+	t.Cleanup(func() {
+		for _, server := range servers {
+			assert.NoError(t, server.Close())
+		}
+	})
```
brkomir: since this func is used in several tests, consider refactoring it into a new function that will be passed to the t.cleanup method or just call it at the end of every test which requires it

brkomir (approved):
Overall looks ok! Fix the linter errors and we are good to go

consensus/ibft/snapshot_test.go
zivkovicmilos: I think this is out of the scope for this PR, since we have this horrible test error handling logic in a lot of places:
Screenshot 2021-12-27 at 16 04 37

e2e/transaction_test.go
zivkovicmilos: Fixed 👍 

0239431

helper/keccak/keccak.go
zivkovicmilos: Removed this explicit ignore for `//nolint`, which is the case in the rest of the codebase:

7ad168c

helper/keccak/pool.go
zivkovicmilos: Removed this explicit ignore for `//nolint`, which is the case in the rest of the codebase:

7ad168c

jsonrpc/dispatcher.go
zivkovicmilos: Removed this explicit ignore for `//nolint`, which is the case in the rest of the codebase:

7ad168c

network/discovery_test.go
zivkovicmilos: Fixed 👍 

760ee91

consensus/ibft/snapshot_test.go
brkomir: Don't you feel a little bit sad to leave it like this? :)

zivkovicmilos: Please see the screenshot and tell me if I need to feel sad 😞

network/testing.go
```diff
@@ -20,117 +21,250 @@ import (
...
+	if connectTimeout < joinTimeout {
+		// In case the connect timeout is smaller than the join timeout, align them
+		connectTimeout = joinTimeout
 	}
 
-	logger := hclog.NewNullLogger()
+	// The join routine should be separate
+	go func() {
+		if joinErr := source.Join(destination.AddrInfo(), joinTimeout); joinErr != nil {
+			source.logger.Error(fmt.Sprintf("Unable to join peer, %v", joinErr))
```
Kourin1996: Why don't we use `t.Error` or `t.Log` instead of source.logger to show this message?

network/testing.go
```diff
@@ -20,117 +21,250 @@ import (
...
+		joinErrorsLock.Lock()
+		joinErrors = append(joinErrors, joinErr)
+		joinErrorsLock.Unlock()
+	}
+
+	numServers := len(servers)
+	var wg sync.WaitGroup
+	for indx := 0; indx < numServers; indx++ {
+		for innerIndx := 0; innerIndx < numServers; innerIndx++ {
+			if innerIndx != indx {
```
Kourin1996: Does server[1] needs to try to connect to server[0] even though server[0] has dialed to server[1] already? You might want to change the condition like this:

```go
if innerIndx < indx {
    // Skip dialing
    continue
}

```

Kourin1996 (changes requested):

network/testing.go
zivkovicmilos: I don't want to involve the `*testing.T` in the context of these helper methods.

If the join fails, the connect will time out nonetheless in the same func

What do you think?

network/testing.go
zivkovicmilos: This is a good catch 💯 

When you initially get dialed by a node, you'll also attempt to form an outbound connection.

I've added this optimization:

2ab4d06

network/testing.go
Kourin1996: But this function is intended to be called with testing Server whose logger is NullLogger as default. This log will be shown?

zivkovicmilos: You're right 👍 

I've imagined this function as a simple helper, whose source of truth about connection should come not from the go routine that starts the join process, but from the `WaitUntilPeerConnectsTo`. Sometimes we expect this join to fail, and in that case we don't need the entire test to error out or to clutter the testing logs.

I've removed the error handling in the go routine since it's irrelevant:

4ffc1da

Kourin1996 (approved):
Looks good!

lazartravica (approved):
Flawless, great job.

Good thinking about the helper methods for bootstrapping the network.

PR #318, opened 2021-12-28
Fix issue that peer stays in routing table after the failed connection

### LINKED ISSUE #309: Peer count doesn't fully recover after network partition
### Description

We've noticed an issue where if we kill connectivity (specifically outbound connectivity) to a node, in some cases it does not fully recover to the total peer count

### Your environment

* OS and version Ubuntu 20

### Steps to reproduce

* Tell us how to reproduce this issue 
Create a 4 node cluster (all bootnodes). Randomly kill connectivity with the nodes and restore them after some time. Simplest way is to make a 4 node cloud cluster and just toggle the security groups for outbound on AWS for example. You can try some variant of this script in node:

```typescript
import EC2 from "aws-sdk/clients/ec2";

function randomIntFromInterval(min: number, max: number) {
  // min and max included
  return Math.floor(Math.random() * (max - min + 1) + min);
}

function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
  return value !== null && value !== undefined;
}

const fullyConnectedGroups = [
  "sg-a",
  "sg-b",
  "sg-c",
];
const disconnectedGroups = ["sg-a", "sg-b"];
// You can set the fullyConnectedGroups to be the security groups required for fully connected nodes, i.e., all nodes can community (full inbound/full outbound)
// You can set disconnectedGroups to be a subset (or an empty set) of the fullyConnected groups, for example, remove the one that allows for outbound access (sg-c in this case) 

export async function main() {
  const ec2 = new EC2({ region: "us-east-1" });
  var params = {
    Filters: [
      {
        Name: "tag:network",
        Values: ["toronto"],
      },
    ],
  };
  const instances = await ec2.describeInstances(params).promise();
  const instanceIds = instances
    .Reservations!.map((reservation) =>
      reservation.Instances!.map((instance) => instance.InstanceId)
    )
    .flat();
  console.log(`Instance IDs ${instanceIds}`);
  const nonEmptyIds = instanceIds.filter((id) => id !== undefined);
  while (true) {
    // Shuffle array
    const shuffled = nonEmptyIds.sort(() => 0.5 - Math.random());

    // Get sub-array of first n elements after shuffled
    let selectedIds = shuffled
      .slice(0, randomIntFromInterval(0, instanceIds.length))
      .filter(notEmpty);
    console.log("Selected IDs to disconnect ", selectedIds);
    for (const id of selectedIds) {
      console.log(`Disconnecting ${id}`);
      await ec2
        .modifyInstanceAttribute({ InstanceId: id, Groups: disconnectedGroups })
        .promise();
    }
    const reconnectWaitTime = randomIntFromInterval(30, 60);
    console.log(`Waiting ${reconnectWaitTime} seconds to reconnect nodes`);
    // Wait random amount of time from 30 to 60s to reconnect
    await new Promise((r) => setTimeout(r, reconnectWaitTime * 1000));
    for (const id of selectedIds) {
      console.log(`Reconnecting ${id}`);
      await ec2
        .modifyInstanceAttribute({
          InstanceId: id,
          Groups: fullyConnectedGroups,
        })
        .promise();
    }
    const continueWaitTime = randomIntFromInterval(1, 20);
    console.log(`Waiting ${continueWaitTime} seconds to continue`);
    await new Promise((r) => setTimeout(r, continueWaitTime * 1000));
  }
}

main().then(() => console.log("Done"));

```

### Expected behaviour

Nodes should recover to the total peer count 

### Actual behaviour

Nodes only recover to a fraction of the total peers

### Logs

See our Grafana screenshot with the new metrics from 

https://github.com/0xPolygon/polygon-sdk/pull/302

You can see the behaviour prior to the script starting, and then afterwards when connectivity is restored. This was a 4 validator node cluster with 2 additional, non-sealing nodes online. Full connectivity was restored at 11:02

image

Here is another example with the 4 validator nodes and the additional 2 non-sealing nodes offline. Full connectivity was restored at 12:15:

image

In general it looks like these 2 errors come as a pair for each node:

```
[ERROR] polygon.network: failed to dial bootnode: id=16Uiu2HAmP2bpgr3sLUUzt61bEzAL7ptdNXnvsfnwcxyrfCMFjxDf err="rpc error: code = Unavailable desc = error reading from server: stream reset"

[ERROR] polygon.consensus.syncer: failed to notify: err="rpc error: code = Unavailable desc = error reading from server: stream reset"
```

### Proposed solution

Kourin1996:
# Description

Relevant: https://github.com/0xPolygon/polygon-sdk/issues/309

This PR fixes that routing table keeps holding a peer even though node fails to connect to the peer. This issue prevents reconnection to the peer.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


zivkovicmilos:
@Kourin1996 
I've added @PtitLuca  as a reviewer as well for this PR, since he's working on a similar task

network/discovery.go
```diff
@@ -24,46 +23,66 @@ var discProto = "/disc/0.1"
...
 
+func (ps *referencePeers) getRandomPeer() *referencePeer {
+	ps.mux.RLock()
+	defer ps.mux.RUnlock()
+
+	l := len(ps.peers)
+	if l == 0 {
+		return nil
+	}
+	return ps.peers[rand.Intn(l)]
```
zivkovicmilos, line 56: Where do we seed this random number generator?

zivkovicmilos (approved):
Looks great 💯

Thank you for looking into the issue 🙏 

Pinging @PtitLuca to review as well 👍

network/discovery.go
Kourin1996: @zivkovicmilos https://github.com/0xPolygon/polygon-sdk/blob/82cf7177cb9f48c9d06fcf3ab921e28375437e58/network/discovery.go#L18-L20

zivkovicmilos: This method is never called

lazartravica: @zivkovicmilos https://golangbyexample.com/init-function-golang/

lazartravica (approved):

network/discovery_test.go
```diff
@@ -98,3 +98,177 @@ func TestDiscovery_PeerAdded(t *testing.T) {
...
+		0: defaultConfig,
+		1: defaultConfig,
+	}
+
+	servers, createErr := createServers(3, paramsMap)
+	if createErr != nil {
+		t.Fatalf("Unable to create servers, %v", createErr)
+	}
+	t.Cleanup(func() {
+		// close only servers[0] because servers[0] has closed already
```
0xpanoramix: I think you meant `[...] because server[1] has closed already`.
It seems like you're closing server[1] just below.

0xpanoramix (commented):
LGTM !
Nice work on the tests 👍. They are easy to understand and replicate !

0xpanoramix (approved):

network/discovery_test.go
Kourin1996: Good catch! Thank you, I fixed at 20ae08d9f5c30f9a4e0fea7da8db05b7cc40bfb6

PR #321, opened 2021-12-29
Add CORS configuration to JSON-RPC

0xpanoramix:
# Description

This PR adds to following :

- You can now submit a cors configuration for the JSON-RPC server using a json file
- This configuration file can be injected using a CLI option (`handler-config`)
- For now, `allowed_origins` is the only field you can configure

# Changes include

# Breaking changes

# Checklist

- [ ] I have tested this code

# Additional comments

The usage documentation PR is available [here](https://github.com/0xPolygon/polygon-sdk-docs/pull/45)


zivkovicmilos (approved):
Looks good to me 💯 

Please provide some steps on how to test this out in the PR description 🙏

jsonrpc/jsonrpc.go
```diff
@@ -32,11 +33,16 @@ func (s serverType) String() string {
 	}
 }
 
+type Handlers struct {
```
ferranbt: Instead of creating a new struct, add this to the Config field, it is meant to add these type of configurations.

server/config.go
```diff
@@ -16,27 +17,28 @@ const DefaultJSONRPCPort int = 8545
...
+	Telemetry      *Telemetry
+	Network        *network.Config
+	DataDir        string
+	Seal           bool
+	Locals         []types.Address
+	NoLocals       bool
+	PriceLimit     uint64
+	MaxSlots       uint64
 	SecretsManager *secrets.SecretsManagerConfig
+	HandlersConfig *jsonrpc.Handlers
```
ferranbt: I think at this point it is not scalable to add more fields into the base Config struct. Instead, create a JsonRPC config struct that holds all the jsonrpc configuration: bind addr and handlers. Note that this is what we do in Telemetry.

go.mod
```diff
@@ -8,6 +8,8 @@ require (
 	github.com/go-kit/kit v0.9.0
 	github.com/golang/protobuf v1.5.2
 	github.com/google/uuid v1.1.4
+	github.com/gorilla/handlers v1.5.1
```
ferranbt: You should be really mindful every time a new package is included into the codebase. Do we know how this package works? Have we vetted it in detail? Is it worth it to include two new packages just to enable Cors in the json-rpc package?

command/helper/config.go
```diff
@@ -191,6 +193,25 @@ func (c *Config) BuildConfig() (*server.Config, error) {
 		}
 	}
 
+	// JSON-RPC Handlers
+	if c.HandlerFile != "" {
+		handlers := &jsonrpc.Handlers{}
```
ferranbt: I have never seen including handlers from config file before.

Kourin1996 (commented):
I'd like to give one suggestion and raise one discussion

(1) Enabling CORS
In principle, you need to set some fields in HTTP response header to enable CORS. As Ferran said, we don't need to add new package for CORS setting at the moment. You can create a `middleware` to write headers for CORS configuration

(2) CORS settings from json
CORS settings will be set at json file and program loads this if the path is given. I'm not sure the clear reasons why we will load CORS settings from json file. Even if user prefer to load CORS settings from JSON file, we have existing flags `--config` to load whole configuration from json file already. If you didn't have discussion about this before, let's discuss.

cc: @zivkovicmilos @lazartravica @brkomir

ghost (approved):

lazartravica (changes requested):
Let's not do this by including the Gorilla handlers package. 

As CORS is mainly a clientside thing, we can easily add new headers to the HTTP server.

PR #323, opened 2021-12-29
Add support for TxPool event subscriptions

zivkovicmilos:
# Description

This PR adds support for TxPool event subscriptions.

Supported events:
* Transaction added
* Transaction enqueued
* Transaction promoted
* Transaction dropped
* Transaction demoted

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Additional docs PR: https://github.com/0xPolygon/polygon-edge-docs/pull/56


command/txpool/txpool_subscribe.go
```diff
@@ -0,0 +1,216 @@
...
+type TxPoolEventResult struct {
+	EventType txpoolProto.EventType `json:"eventType"`
+	TxHash    string                `json:"txHash"`
+}
+
+func (r *TxPoolEventResult) Output() string {
+	var buffer bytes.Buffer
+
+	// current number & hash
+	buffer.WriteString("\n[TXPOOL EVENT]\n")
```
ferranbt, line 218: Why are you using a buffer here? You can do:
```
str := ""
str += "\n[TXPOOL EVENT]\n"
...
return str
```
This is how it was done for other command outputs.

txpool/txpool.go
```diff
@@ -284,6 +290,10 @@ func (p *TxPool) Drop() {
 		"nonce", tx.Nonce,
 		"addr", tx.From.String(),
 	)
+	go p.eventManager.fireEvent(&proto.TxPoolEvent{
```
ferranbt: I would not force the user of the ```fireEvent``` api (txpool) to use ```go``` everytime. Do the async part inside the fireEvent itself.

txpool/event_manager.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+	"github.com/0xPolygon/polygon-sdk/txpool/proto"
+	"github.com/google/uuid"
+	"github.com/hashicorp/go-hclog"
+	"sync"
+)
+
+type eventManager struct {
```
ferranbt, line 13: No test for this? By experience, there are always bugs on the first iteration of an event manager. Also please consider using this https://github.com/hashicorp/go-eventlogger instead.

command/txpool/txpool_subscribe.go
zivkovicmilos: We actually use a buffer for all command outputs.

Please look at this PR which introduced it:
https://github.com/0xPolygon/polygon-sdk/pull/277

txpool/txpool.go
zivkovicmilos: You're right, the caller shouldn't need to know it needs to run the `fireEvent` asynchronously 👍 

Fixed:

b71384a

txpool/event_manager.go
zivkovicmilos: Tests are still pending for this PR, along with the docs. I'll work on this and it should be wrapped up within a few days 👍 

Thank you for recommending the event logger 🙏 

It looks like a great opportunity to create a common event framework for not just the txpool, but for all modules. I'll add it as a task to our next sprint so we have a standardized way of handling events 💯

command/txpool/txpool_subscribe.go
```diff
@@ -0,0 +1,216 @@
...
+	flags := t.Base.NewFlagSet(t.GetBaseCommand(), t.Formatter, t.GRPC)
+
+	var added bool
+	var promoted bool
+	var enqueued bool
+	var dropped bool
+	var demoted bool
+
+	flags.BoolVar(&added, "added", false, "")
+	flags.BoolVar(&promoted, "enqueued", false, "")
```
Kourin1996: Perhaps `promoted`?

txpool/event_subscription.go
```diff
@@ -0,0 +1,46 @@
+package txpool
+
+import (
+	"github.com/0xPolygon/polygon-sdk/txpool/proto"
+)
+
+type subscriptionID string
```
Kourin1996: Why string? This ID seems to be used internally, so it's ok you will use an integer instead of this. (yes, of course integer represents up to 64 bit value while UUID represents 128 bit value)

txpool/operator.go
```diff
@@ -50,7 +50,30 @@ func (p *TxPool) AddTxn(ctx context.Context, raw *proto.AddTxnReq) (*proto.AddTx
...
-func (p *TxPool) Subscribe(req *empty.Empty, stream proto.TxnPoolOperator_SubscribeServer) error {
-	// TODO
-	return nil
+func (p *TxPool) Subscribe(
+	request *proto.SubscribeRequest,
+	stream proto.TxnPoolOperator_SubscribeServer,
+) error {
+	subscription := p.eventManager.subscribe(request.Types)
+
+	teardown := func() error {
```
Kourin1996: Do you need to return error for now? If you don't return error, you can use `defer`

Kourin1996 (changes requested):
I've left small comments. I'd appreciate you could check. Thanks!

command/txpool/txpool_subscribe.go
zivkovicmilos: Copypasta 🤦 

Fixed 👍 

428d24b

txpool/operator.go
zivkovicmilos: Good catch 💯 

This is leftover from an earlier iteration of the implementation.

Fixed 👍 

167e38f

txpool/event_subscription.go
zivkovicmilos: You're right 👍 

Fixed:

7d1b6bc

Kourin1996 (approved):

dbrajovic (approved):
LGTM 👍

brkomir (changes requested):
I see that you are using this eventManager in the txpool.go but not tests are covering if the usage is correct. What I mean by that is that I can go and remove the code in this file and tests would pass regardless. Also eventManager type is completely untested.

zivkovicmilos:
@brkomir 

Coverage tests added with a52e0f8 & 6b6640b 👍

brkomir (approved):

PR #325, opened 2022-01-04
[Hotfix] Update the block write interface

zivkovicmilos:
# Description

This PR fixes the blockchain write interface to be more precise.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -579,9 +579,7 @@ func (b *Blockchain) WriteBlock(block *types.Block) error {
 		)
 	}
 
-	// TODO, the wrapper around transactions
 	if hash := buildroot.CalculateTransactionsRoot(block.Transactions); hash != block.Header.TxRoot {
-
 		return fmt.Errorf(
 			"transaction root hash mismatch: have %s, want %s",
 			hash,
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -519,43 +519,29 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
 	return nil
 }
 
-// WriteBlocks writes a batch of blocks
-func (b *Blockchain) WriteBlocks(blocks []*types.Block) error {
-	// Check the size
-	size := len(blocks)
-	if size == 0 {
-		return fmt.Errorf("the passed in block array is empty")
+// WriteBlock writes a single block
+func (b *Blockchain) WriteBlock(block *types.Block) error {
+	// Check the param
+	if block == nil {
+		return fmt.Errorf("the passed in block is empty")
 	}
 
 	// Log the information
-	if size == 1 {
-		b.logger.Info(
-			"write block",
-			"num",
-			blocks[0].Number(),
-			"parent",
-			blocks[0].ParentHash(),
-		)
-	} else {
-		b.logger.Info(
-			"write blocks",
-			"num",
-			size,
-			"from",
-			blocks[0].Number(),
-			"to",
-			blocks[size-1].Number(),
-			"parent",
-			blocks[0].ParentHash())
-	}
+	b.logger.Info(
+		"write block",
+		"num",
+		block.Number(),
+		"parent",
+		block.ParentHash(),
+	)
 
-	parent, ok := b.readHeader(blocks[0].ParentHash())
+	parent, ok := b.readHeader(block.ParentHash())
 	if !ok {
 		return fmt.Errorf(
 			"parent of %s (%d) not found: %s",
-			blocks[0].Hash().String(),
-			blocks[0].Number(),
-			blocks[0].ParentHash(),
+			block.Hash().String(),
+			block.Number(),
+			block.ParentHash(),
 		)
 	}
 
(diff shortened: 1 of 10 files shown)


0xAleksaOpacic (approved):
Looks great. Thank you for refactoring you have done!

protocol/syncer.go
```diff
@@ -242,10 +242,11 @@ func (pw *progressionWrapper) runUpdateLoop(subscription blockchain.Subscription
 
 // stopProgression stops the progression tracking
 func (pw *progressionWrapper) stopProgression() {
+	pw.stopCh <- struct{}{}
```
lazartravica, line 245: Why did you move the message sending to the top of the function?

e2e/pos_test.go
```diff
@@ -613,3 +615,171 @@ func TestPoS_StakeUnstakeWithinSameBlock(t *testing.T) {
...
+) (*ibftOp.Snapshot, error) {
+	snapshot, snapshotErr := client.GetSnapshot(ctx, &ibftOp.SnapshotReq{
+		Latest: false,
+		Number: blockNum,
+	})
+
+	return snapshot, snapshotErr
+}
+
+func getNearestNextEpochBlock(blockNum uint64, epochSize uint64) uint64 {
```
lazartravica: can be renamed to simply `getNextEpochBlock`

e2e/pos_test.go
```diff
@@ -613,3 +615,171 @@ func TestPoS_StakeUnstakeWithinSameBlock(t *testing.T) {
...
+	if epochSize > blockNum {
+		return epochSize
+	}
+
+	nearest := blockNum + epochSize/2
+	nearest = nearest - (nearest % epochSize)
+
+	if nearest < blockNum {
+		nearest += epochSize
+	}
```
lazartravica: i feel we can make this more readable.

do you think this is faster than performing mod block size to check if we're already there, and then doing integer division and multiplication by epoch size to deduce the next epoch block?

e2e/pos_test.go
```diff
@@ -613,3 +615,171 @@ func TestPoS_StakeUnstakeWithinSameBlock(t *testing.T) {
...
+
+	return nearest
+}
+
+func TestSnapshotUpdating(t *testing.T) {
+	fountainKey, fountainAddr := tests.GenerateKeyAndAddr(t)
```
lazartravica: you mean faucet?

blockchain/blockchain.go
```diff
@@ -519,43 +519,29 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
-	parent, ok := b.readHeader(blocks[0].ParentHash())
+	b.logger.Info(
+		"write block",
+		"num",
+		block.Number(),
+		"parent",
+		block.ParentHash(),
+	)
+
+	parent, ok := b.readHeader(block.ParentHash())
```
lazartravica, line 538: this is incurring a performance cost when syncing blocks. in the previous implementation we only checked if the first block's parent is present in the db. now we will check it all the time. 

can you benchmark the syncing tests with and without this line here just to see how much of a performance cost can we expect? 

do you see any simple and clean way we can skip this check when bulk syncing subsequent  blocks?

blockchain/blockchain.go
```diff
@@ -564,94 +550,89 @@ func (b *Blockchain) WriteBlocks(blocks []*types.Block) error {
...
+		)
+	}
 
-			return fmt.Errorf(
-				"transaction root hash mismatch: have %s, want %s",
-				hash,
-				block.Header.TxRoot,
-			)
-		}
+	// TODO, the wrapper around transactions
```
lazartravica: i am not sure what this TODO means, do you know?

blockchain/blockchain.go
```diff
@@ -564,94 +550,89 @@ func (b *Blockchain) WriteBlocks(blocks []*types.Block) error {
...
+		"hash", header.Hash,
+		"txns", len(block.Transactions),
+	}
+	if prevHeader, ok := b.GetHeaderByNumber(header.Number - 1); ok {
+		diff := header.Timestamp - prevHeader.Timestamp
+		logArgs = append(logArgs, "generation_time_in_sec", diff)
 	}
 
+	b.logger.Info("new block", logArgs...)
 	b.logger.Info("new head", "hash", b.Header().Hash, "number", b.Header().Number)
```
lazartravica: i don't think we need this `new head` message any more, since the previous `new block` will log all the relevant information.

lazartravica (changes requested):
Good job on the tests 💪  Take a look at my review comments.

consensus/ibft/ibft_test.go
```diff
@@ -537,7 +537,7 @@ func TestWriteTransactions(t *testing.T) {
 				{Nonce: 4, Gas: 10001}, // exceeds block gas limit
 				{Nonce: 5},             // included
 				{Nonce: 6},             // reaches gas limit - returned to pool
-				{Nonce: 7}},            // not considered - stays in pool
+				{Nonce: 7}}, // not considered - stays in pool
```
lazartravica: Format

protocol/syncer.go
zivkovicmilos: `stopProgression` shouldn't be holding a lock at the moment it's sending a stop signal to the stop channel the local runner is listening on - since it can cause a deadlock situation where the thread is holding a lock and waiting for someone to read from the `stopCh` and some other thread wants to grab the lock

e2e/pos_test.go

d2cc7ff

e2e/pos_test.go
zivkovicmilos: Yep 👍 

Fixed 👍 

c831ab2

blockchain/blockchain.go
zivkovicmilos: Agreed 💯 

Fixed 👍 

a90c517

consensus/ibft/ibft_test.go
zivkovicmilos: Fixed 👍 

55ec67f

blockchain/blockchain.go
zivkovicmilos: Honestly I have no idea, this comment was made ~3 years ago.

I've removed it since I can't see the relevance any more:

39e4783

blockchain/blockchain.go
zivkovicmilos: You're completely right to assume it adds overhead - but that's something that we can only see when bulk syncing. Every other part of the PSDK already did this call on a per-block basis as they write 1 block at a time.

I don't see an easy way of skipping this check with the `WriteBlock` functionality, as it is a critical chain sanity check, it needs to be there at one point before pushing a block to the chain.

The header will be initially fetched from the in-memory cache and if there's a cache miss, it will grab it from LevelDB, either way it's a quick lookup as it's fetching only the header and not the entire block

e2e/pos_test.go
zivkovicmilos: Simplified it by a lot 👍 


8a2ee9c

protocol/syncer.go
lazartravica: got it! thanks for the explanation 👍

lazartravica (approved):
🥳

PR #327, opened 2022-01-05
Discard transaction with invalid gas limit

### LINKED ISSUE #297: Polygon-SDK doesn't correctly report dropped txns through receipts for Metamask
### Discussed in https://github.com/0xPolygon/polygon-sdk/discussions/294

<div type='discussions-op-text'>

<sup>Originally posted by **AzulPretoBranco** December 15, 2021</sup>
### Description

During our internal testing we notice that if we submit a transaction with an invalid gas limit (this was done on purpose) to a blockchain created using Polygon-SDK the transaction is accepted by the blockchain and added to the tx pool. 

Later when the method `writeTransactions`runs the blockchain will check the gas limit and if the value is invalid the transaction will be silently discarded.

This poses a problem, because the transaction is silently discarded applications that use the blockchain (like metamask) will not be able to receive a confirmation for this transaction meaning that the transaction will be "forever pending".

This exact use case doesn't happen in an ethereum test network like the Rinkeby Test Network. The ethereum blockchain validates the gas limit before adding the transaction to the tx pool and returns an appropriate error to the application interacting with the blockchain (like metamask).

Attached is a video that shows this test using the Rinkeby Test Network and a blockchain created using the Polygon-SDK.

### Your environment

* OS and version: Windows 10 (WSL Ubuntu 20.04.3 LTS
* version of the Polygon SDK: develop branch at commit 7fe8091967102192539618fd786f226d54dd312b

### Steps to reproduce

Video that shows the use cases in the issue description https://1drv.ms/u/s!AvhuC05QeOJntn9-aiUsXaKgKqZN

### Expected behaviour

We believe the Polygon-SDK should behave similar to the ethereum blockchain and return an error when we submit an invalid transaction.

### Actual behaviour

The Polygon-SDK accepts a transaction with an invalid gas limit as a valid transaction unlike the ethereum network.

### Logs

Logs from Polygon-SDK:
```
2021-12-15T16:23:52.952Z [INFO]  polygon.consensus.ibft.acceptState: we are the proposer: block=71149
2021-12-15T16:23:52.952Z [ERROR] polygon.consensus.ibft: failed to write transaction: transaction's gas limit exceeds block gas limit
2021-12-15T16:23:52.952Z [INFO]  polygon.consensus.ibft: picked out txns from pool: num=0 remaining=0
```

### Proposed solution

Maybe do something similar to go-ethereum where they have the check in `validateTx`, screenshot from go-ethereum:
image
</div>

0xAleksaOpacic:
# Description

If a transaction has gasLimit bigger than the latest blockGasLimit it is considered invalid. 
There is no way to know for sure what will be gas limit of the block where the transaction will be placed, so we have another check when we are creating a block. 

Different error cases
1. Transaction has gasLimit > latestBlockGasLimit (it will not go into pool and it will be rejected as invalid)
2. Transaction has gasLimit < latestBlockGasLimit but when we create a block gasLimit > newBlockGasLimit (we will generate failed transaction receipt)

Testing:
1. Try to add a transaction with gasLimit 10000000001.  

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


.github/workflows/main.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - main
       - develop
+      - fix/txpool-v2
```
0xAleksaOpacic: @ZeljkoBenovic and i have added this to be able to start test when based on fix/txpool-v2

txpool/txpool_test.go
```diff
@@ -17,7 +17,8 @@ import (
 const (
 	defaultPriceLimit uint64 = 1
 	defaultMaxSlots   uint64 = 4096
-	validGasLimit     uint64 = 10000000
+	validGasLimit     uint64 = 4712350
```
0xAleksaOpacic: This value is chosen to be smaller than defaultBlockGasLimit and bigger then intrinsic so it can pass all validations

e2e/txpool_test.go
```diff
@@ -400,68 +400,6 @@ func TestTxPool_StressAddition(t *testing.T) {
 	}
 }
 
-func TestInvalidTransactionRecover(t *testing.T) {
```
0xAleksaOpacic: This test is not valid anymore. Because now transaction will never go inside txPool if gas > latestBlockGasLimit. And we have covered this check in txPool unit tests

zivkovicmilos: Please link the unit test(s) that cover this functionality in the new TxPool implementation 🙏

.github/workflows/main.yml
zivkovicmilos: This should be removed from the workflow yaml file

txpool/mock.go
```diff
@@ -26,7 +26,10 @@ func (m defaultMockStore) GetBalance(types.Hash, types.Address) (*big.Int, error
 }
 
 func (m defaultMockStore) Header() *types.Header {
-	return &types.Header{}
+	//Gas limit is default gas limit for genesis block
+	return &types.Header{
+		GasLimit: 4712388,
```
zivkovicmilos: Is it possible to add a method to the `defaultMockStore` for setting the gas limit of the latest header?

Right now only 1 test uses this value I think, but in the future more unit tests may rely on this GasLimit value to be configurable

txpool/txpool.go
```diff
@@ -29,6 +29,7 @@ const (
 // errors
 var (
 	ErrIntrinsicGas        = errors.New("intrinsic gas too low")
+	ErrBlockLimitExceeded  = errors.New("transaction's gas limit exceeds block gas limit")
```
zivkovicmilos: Please make sure the error corresponds to the same one GETH uses:
https://github.com/ethereum/go-ethereum/blob/f32feeb260054d7c90c1efd57dfec6b717717e18/core/error.go#L60

This is required since many web3 frameworks hardcode these expected values in source code

zivkovicmilos (approved):
Overall it looks good 💯 

I've outlined my concerns with this solution in our previous discussions.
I'll copy part of them here:
>Imagine you discard a transaction during validation, but it would come to be executed only much later, at which point you’d consider it to be a valid transaction.
Example is a shelved transaction, it won’t get executed right away until the nonce hole is filled, but you reject it because it didn’t fit in the temporary block limit check
It goes both ways:
>
>Reject in validateTx, risk losing a good transaction
Accept in validateTx, risk a adding a bad transaction

This solution is in no way bullet proof, but only temporary - the permanent solution is writing failed txn receipts during transaction application

PR #329, opened 2022-01-05
Optimize the loadbot

zivkovicmilos:
# Description

This PR adds additional functionality to the loadbot and speeds it up:
* Switches to using GRPC instead of JSON-RPC for transaction addtions
* Introduces 2 modes (TRANSFER - simple transfer, DEPLOY - contract deployment)
* Adds detailed error outputs (transaction info is grouped by receipt errors / addition errors) if the `--detailed` flag is specified during the run
* Adds support for gas limit / gas price estimation from the network

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Relevant docs PR: https://github.com/0xPolygon/polygon-sdk-docs/pull/53


command/loadbot/execution.go
```diff
@@ -166,28 +177,53 @@ func getInitialSenderNonce(client *jsonrpc.Client, address types.Address) (uint6
...
-		Gas:      1000000,
-		Value:    value,
-		GasPrice: big.NewInt(0x100000),
-		Nonce:    nonce,
-		V:        big.NewInt(1), // it is necessary to encode in rlp
-	}, sender.PrivateKey)
+func getAverageGasPrice(client *jsonrpc.Client) (uint64, error) {
+	gasPrice, err := client.Eth().GasPrice()
+	if err != nil {
+		return 0, fmt.Errorf("failed to query initial sender nonce: %v", err)
```
Kourin1996: you might want to change error message

command/loadbot/execution.go
```diff
@@ -196,20 +232,65 @@ func (l *Loadbot) Run() error {
...
+		return fmt.Errorf("an error has occured while creating JSON-RPC client: %v", err)
+	}
+
+	grpcClient, err := createGRPCClient(l.cfg.GRPC)
 	if err != nil {
 		return fmt.Errorf("an error has occured while creating JSON-RPC client: %v", err)
 	}
 
 	defer func(client *jsonrpc.Client) {
 		_ = shutdownClient(client)
```
Kourin1996: No longer need to define function `shutdownClient`, we can inline it

command/loadbot/loadbot_command.go
```diff
@@ -64,13 +68,49 @@ func (l *LoadbotCommand) DefineFlags() {
...
+		Description: "The GRPC endpoint used to send transactions.",
+		Arguments: []string{
+			"GRPC_ADDRESS",
+		},
+		ArgumentsOptional: false,
+		FlagOptional:      false,
+	}
+
+	l.FlagMap["mode"] = helper.FlagDescriptor{
+		Description: "The mode of operation [0, 1]. Default: 0",
```
Kourin1996: We need more explanation or changing to string enum

command/loadbot/loadbot_command.go
```diff
@@ -64,13 +68,49 @@ func (l *LoadbotCommand) DefineFlags() {
...
-		Description: "The JSON-RPC endpoint used to send transactions.",
+		Description: "The JSON-RPC endpoint used to query transactions.",
 		Arguments: []string{
 			"JSONRPC_ADDRESS",
 		},
 		ArgumentsOptional: false,
 		FlagOptional:      false,
 	}
+
+	l.FlagMap["grpc"] = helper.FlagDescriptor{
```
Kourin1996: Why not `grpc-address`, which is used in other commands

command/loadbot/execution.go
```diff
@@ -22,20 +24,25 @@ const (
...
-	MaxConns int
+	TPS           uint64
+	Sender        types.Address
+	Receiver      types.Address
+	Value         *big.Int
+	Count         uint64
+	JSONRPC       string
+	GRPC          string
+	MaxConns      int
+	GeneratorMode uint64
```
Kourin1996: Why don't you use enum?

command/loadbot/execution.go
```diff
@@ -196,20 +232,65 @@ func (l *Loadbot) Run() error {
...
+	nonce, err := getInitialSenderNonce(jsonClient, sender.Address)
 	if err != nil {
-		return fmt.Errorf("an error occured while getting initial sender nonce: %v", err)
+		return fmt.Errorf("unable to get initial sender nonce: %v", err)
 	}
 
+	gasPrice, err := getAverageGasPrice(jsonClient)
+	if err != nil {
+		return fmt.Errorf("unable to get average gas price: %v", err)
+	}
```
Kourin1996: I think we need discussion here
(1) Why can't give gas price from flags?
(2) Average Gas Price would change during execution

command/loadbot/generator/transfer_generator.go
```diff
@@ -0,0 +1,92 @@
...
+func (tg *TransferGenerator) MarkFailedTxn(failedTxn *FailedTxnInfo) {
+	tg.failedTxnsLock.Lock()
+	defer tg.failedTxnsLock.Unlock()
+
+	tg.failedTxns = append(tg.failedTxns, failedTxn)
+}
+
+func (tg *TransferGenerator) SetGasEstimate(gasEstimate uint64) {
+	tg.estimatedGas = gasEstimate
+}
```
Kourin1996: The same methods are defined in DeployGenerator, too. You might want to define base struct to be embedded in DeployGenerator and TransferGenerator

command/loadbot/generator/transfer_generator.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type TransferGenerator struct {
+	failedTxns     []*FailedTxnInfo
+	failedTxnsLock sync.RWMutex
+
+	params          *GeneratorParams
+	signer          *crypto.EIP155Signer
+	receiverAddress types.Address
+	estimatedGas    uint64
```
Kourin1996: We can standardize some fields?

command/loadbot/loadbot_command.go
```diff
@@ -205,10 +271,17 @@ type TxnBlockData struct {
...
+}
+
 type LoadbotResult struct {
-	CountData      TxnCountData      `json:"countData"`
-	TurnAroundData TxnTurnAroundData `json:"turnAroundData"`
-	BlockData      TxnBlockData      `json:"blockData"`
+	CountData         TxnCountData         `json:"countData"`
+	TurnAroundData    TxnTurnAroundData    `json:"turnAroundData"`
+	BlockData         TxnBlockData         `json:"blockData"`
+	DetailedErrorData TxnDetailedErrorData `json:"detailedErrorData"`
```
Kourin1996: You might want to change the type to pointer and add `omitempty` to tag in `DetailedErrorData` so that we can hide this field if `--detailed` is not given

command/loadbot/loadbot_command.go
```diff
@@ -282,6 +377,46 @@ func (lr *LoadbotResult) Output() string {
...
+				buffer.WriteString(fmt.Sprintf("\n\n[%s]\n", detailedError.TxHash))
+			} else {
+				buffer.WriteString("\n\n[Tx Hash Unavailable]\n")
+			}
+
+			formattedStrings := make([]string, 0)
+			formattedStrings = append(formattedStrings,
+				fmt.Sprintf("Index|%d", detailedError.Index),
+				fmt.Sprintf("Error|%s", detailedError.Error.Error.Error()),
+			)
```
Kourin1996: Why use `append` instead of slice literal?

Kourin1996 (commented):
Basically looks ok to me. I've just left small comments.

command/loadbot/execution.go
zivkovicmilos: Fixed 👍 

6aafcbc

command/loadbot/execution.go
zivkovicmilos: Fixed 👍 

8778667

command/loadbot/loadbot_command.go
zivkovicmilos: Fixed 👍 

0f82b5b

command/loadbot/execution.go
zivkovicmilos: Fixed 👍 

0f82b5b

command/loadbot/loadbot_command.go
zivkovicmilos: Fixed 👍 

fdcdbb2

command/loadbot/execution.go
zivkovicmilos: You're completely right.

We used to hardcode it before, this is just a more dynamic solution.
What do you suggest?

command/loadbot/loadbot_command.go
zivkovicmilos: Good catch 👍 

Fixed 👍 


f97a690

command/loadbot/loadbot_command.go
zivkovicmilos: Fixed 👍 

5fa50e6

command/loadbot/generator/transfer_generator.go
zivkovicmilos: You're completely right 👍 

Added a base type:

291b223

command/loadbot/generator/transfer_generator.go
zivkovicmilos: Fixed during the creation of the base generator type 👍 


291b223

command/loadbot/execution.go
Kourin1996: I don't think it's a good to fetch the average gas price every time you create new transaction because it affects thorughput.
Firstly, it's better to be able to pass gas price from flags because sometimes we want to give much enough gas price to process transactions as fast as possible.

brkomir (approved):
LGTM

command/loadbot/execution.go
zivkovicmilos: I agree 👍 

I've added the ability for you to specify both the gas limit and gas price if you want, but if you don't specify something the value will be fetched from the network:

a0efd89

ZeljkoBenovic (approved):
Looks great.
I'm slapping that like button :)

command/loadbot/loadbot_command.go
```diff
@@ -121,11 +192,35 @@ func (l *LoadbotCommand) Run(args []string) int {
...
+		return 1
+	}
+
+	// Parse the gas limit
+	bigGasLimit, gasLimitErr := types.ParseUint256orHex(&gasLimit)
+	if gasLimitErr != nil {
+		l.Formatter.OutputError(fmt.Errorf("failed to decode gas limit to value: %w", err))
+
+		return 1
+	}
```
Kourin1996: Failed to parse when `gasPrice` or `gasLimit` is empty string

command/loadbot/execution.go
```diff
@@ -211,20 +253,83 @@ func (l *Loadbot) Run() error {
...
+
+	gasPrice := l.cfg.GasPrice
+	if gasPrice == nil {
+		// No gas price specified, query the network for an estimation
+		avgGasPrice, err := getAverageGasPrice(jsonClient)
+		if err != nil {
+			return fmt.Errorf("unable to get average gas price: %w", err)
+		}
+
+		gasPrice.SetUint64(avgGasPrice)
```
Kourin1996: panic happens when `gasPrice` is nil

```suggestion
		gasPrice = new(big.Int).SetUint64(avgGasPrice)
```

command/loadbot/execution.go
```diff
@@ -211,20 +253,83 @@ func (l *Loadbot) Run() error {
...
+
+	gasLimit := l.cfg.GasLimit
+	if gasLimit == nil {
+		// No gas limit specified, query the network for an estimation
+		gasEstimate, estimateErr := estimateGas(jsonClient, exampleTxn)
+		if estimateErr != nil {
+			return fmt.Errorf("unable to get gas estimate, %w", err)
+		}
+
+		gasLimit.SetUint64(gasEstimate)
```
Kourin1996: panic happens when`gasLimit`  is nil

```suggestion
		gasLimit = new(big.Int).SetUint64(gasEstimate)
```

command/loadbot/generator/deploy_generator.go
```diff
@@ -0,0 +1,70 @@
...
...0006109578261091e565b9050919050565b60006109698261091e565b9050919050565b82818337600083830152505050565b60005b8381101561099d578082015181840152602081019050610982565b838111156109ac576000848401525b50505050565b600060028204905060018216806109ca57607f821691505b602082108114156109de576109dd610a8d565b5b50919050565b6109ed82610aeb565b810181811067ffffffffffffffff82111715610a0c57610a0b610abc565b5b80604052505050565b6000610a208261091e565b91507fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff821415610a5357610a52610a5e565b5b600182019050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052602260045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b6000601f19601f8301169050919050565b610b05816108ec565b8114610b1057600080fd5b50565b610b1c8161091e565b8114610b2757600080fd5b5056fe4368616e67696e67206772656574696e672066726f6d202725732720746f2027257327a2646970667358221220a4f650513c5038cc3c919651c8ca206cda3ae09ac470aebc8d96da7b43318d5e64736f6c634300080400334465706c6f79696e67206120477265657465722077697468206772656574696e673a"
```
0xpanoramix: Why did you decide not to give users the ability to choose the contract to be deployed (using a json file for example) ?

Kourin1996 (changes requested):

0xpanoramix (commented):
This looks great 🚀 
I just have a question concerning the `contract deployment` mode, but other than that everything is ok !

command/loadbot/execution.go
zivkovicmilos: Good catch 💯 

Fixed 👍 

6e01c9f

command/loadbot/execution.go
zivkovicmilos: Good catch 💯 

Fixed 👍 

6e01c9f

command/loadbot/loadbot_command.go
zivkovicmilos: I was testing this behavior with the wrong parse func 🤦 

Fixed 👍 

6e01c9f

command/loadbot/generator/deploy_generator.go
zivkovicmilos: I love this idea 💯 

Added this functionality:

c4e1b40

0xpanoramix (approved):

Kourin1996 (approved):
Looks good to me

PR #335, opened 2022-01-12
Remove earliest from JSON-RPC block filter

0xpanoramix:
# Description

This PR removes the `earliest` block filter from the JSON-RPC module, as it is not supported.

# Changes include

# Checklist

- [X] I have tested this code

### DIFF
--- a/jsonrpc/codec.go
+++ b/jsonrpc/codec.go
@@ -94,9 +94,8 @@ func (e *ObjectError) Error() string {
 }
 
 const (
-	PendingBlockNumber  = BlockNumber(-3)
-	LatestBlockNumber   = BlockNumber(-2)
-	EarliestBlockNumber = BlockNumber(-1)
+	PendingBlockNumber = BlockNumber(-3)
+	LatestBlockNumber  = BlockNumber(-2)
 )
 
 type BlockNumber int64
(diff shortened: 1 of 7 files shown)


jsonrpc/codec.go
```diff
@@ -109,10 +108,13 @@ type BlockNumberOrHash struct {
...
-// 2 - "0x2"								- block number #2 (EIP-1898 backward compatible)
-// 3 - {blockNumber:	"0x2"}				- EIP-1898 compliant block number #2
-// 4 - {blockHash:		"0xe0e..."}			- EIP-1898 compliant block hash 0xe0e...
+// 1 - "latest" or "pending"			- self-explaining keywords
+//
+// 2 - "0x2"							- block number #2 (EIP-1898 backward compatible)
+//
+// 3 - {blockNumber:	"0x2"}			- EIP-1898 compliant block number #2
+//
+// 4 - {blockHash:		"0xe0e..."}		- EIP-1898 compliant block hash 0xe0e...
```
zivkovicmilos, line 117: Please fix the formatting here 🙏

zivkovicmilos (changes requested):
Why do we have a PR to _**remove**_ the earliest param, instead of adding support for it?

PR #336, opened 2022-01-12
Fixed some jsonrpc methods

ZeljkoBenovic:
# Description

This PR fixes / implements 3 broken jsonrpc methods:
* eth_chainID
    * changed method from eth_chainId to eth_chainID in polygon-sdk public docs
* net_peerCount
    * impemented a proper method that returns the number of peers
* web3_clientVersion
   * fixed output to accomodate json responce
# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


server/server.go
```diff
@@ -351,11 +351,18 @@ type jsonRPCHub struct {
...
+	*network.Server
 	consensus.Consensus
 }
 
 // HELPER + WRAPPER METHODS //
 
+func (j *jsonRPCHub) GetPeers() int {
+	peers := j.Server.Peers()
+
+	return len(peers)
```
zivkovicmilos: You can inline this call:
```go
return len(j.Server.Peers())
```

version/version.go
```diff
@@ -17,3 +17,8 @@ func GetVersion() string {
 
 	return version
 }
+
+// returns jsonrpc representation of the version
+func GetVersionJsonrpc() string {
+	return "v" + Version
```
zivkovicmilos, line 23: Do we know what kind of response clients like Ganache give for this RPC call?

zivkovicmilos (approved):
Thank you for fixing up these methods, it's easy to forget about these things in the midst of all the features 🙏 

I've left a couple of questions for discussion, but everything seems solid 💯

Kourin1996 (approved):

version/version.go
lazartravica: This is up to the client. Polygon network returns this: `{"jsonrpc":"2.0","id":1,"result":"bor/v0.2.13-stable-15965124/linux-amd64/go1.17.6"}`

lazartravica (approved):

PR #339, opened 2022-01-13
Generate failed txn receipt

### LINKED ISSUE #297: Polygon-SDK doesn't correctly report dropped txns through receipts for Metamask

0xAleksaOpacic:
# Description

This PR generates failed receipt if the transaction fails once it goes inside the pool. 
This can happen because we do not have a default block gas limit, we dynamically calculate it when we create the block.
When we are validating a transaction before we put it in the pool we check for the gas limit but with the previous block. 
In some specific edge-case, it can happen that when we are putting transaction in the pool to be valid (because the previous block gas limit is big enough) but when we are putting transaction in the block to be invalid.

Steps to test it:

1. Start IBFT network
2. Send transaction with gas limit 1000000000001
3. After transaction goes into block it will return ERROR receipt

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments


.github/workflows/main.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - main
       - develop
+      - fix/txpool-v2
```
Kourin1996: Why do you need this branch?

PR #326, opened 2022-01-05
Feature/inbound outbound peerlimit

munna0908:
# Description

This PR introduces two separate flags to customise the max-inbound-peers and max-outbound-peers limit for a node. 
Default values for max-inbound-peers & max-outbound-peers are `32` and `8` respectively. 
Users can also specify the limit either by using `max-peers` or `max-inbound/outbound-peers` flag
# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

~~Removed the MaxPeers flag, so user need to specify separate limits for inbound and outbound peers~~
We have reintroduced `max-peers` to support backward compatibility. 

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [x] I have updated the README and other relevant documents (guides...)
- [ ] I have added sufficient documentation both in code, as well as in the READMEs

PR #343, opened 2022-01-14
Clean up the networking server & add bootnode peer discovery

### LINKED ISSUE #309: Peer count doesn't fully recover after network partition

munna0908:
# Description

This PR is part of a bigger refactor effort on the networking package.
Contained in this PR:
* Separated out services for discovery and identity
* Completely mockable and unit testable discovery and identity
* Cleaned up code in the base networking server
* Simplified logic for networking service initialization
* Fixed bugs in connection management
* Added unit tests for modified / added functionality

## Bootnode peer discovery

A new mechanism was added to the discovery service, where bootnodes are dialed _**temporarily**_ for their peers, and their peers are then dialed for handshaking / permanent connecting.

# Changes include

# Checklist

- [x] I have tested this code

# Additional Comments

Fixes EDGE-65

# Documentation Update
[Relevant PR](https://github.com/0xPolygon/polygon-edge-docs/pull/79)


network/server.go
```diff
@@ -94,12 +97,18 @@ type Server struct {
 	joinWatchersLock sync.Mutex
 
 	emitterPeerEvent event.Emitter
+
+	inboundConnCount int64
+
+	tempeoraryDials sync.Map
```
0xpanoramix: I think there's a typo here.

munna0908: Thanks for catch Commit [link](https://github.com/0xPolygon/polygon-sdk/pull/343/commits/62ade33b25c2f45eafb7a77384c29c18cc01a8ce)

command/server/server_command.go
```diff
@@ -143,13 +143,36 @@ func (c *ServerCommand) DefineFlags() {
...
+		FlagOptional: true,
+	}
+
+	c.FlagMap["nolocals"] = helper.FlagDescriptor{
+		Description: "Sets flag to disable price exemptions for locally submitted transactions",
+		Arguments: []string{
+			"NOLOCALS",
+		},
+		FlagOptional: true,
+	}
```
Kourin1996: `--locals` and `--nolocals` has been removed in TxPool v2 PR.

network/discovery.go
```diff
@@ -246,22 +252,93 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
 			return
+		default:
+		}
+
+		select {
+		case <-ticker1.C:
+			go d.handleDiscovery()
+		case <-ticker2.C:
+			go d.bootnodeDiscovery()
 		}
```
Kourin1996: I don't remember what we concluded. But you need a signal from `ticker1.C` or `ticker2.C` at least to finish this function if a signal is sent to closeCh when you're waiting in 2nd select.
You might want to put these select together.

```go
		select {
		case <-d.closeCh:
			return
		case <-ticker1.C:
			go d.handleDiscovery()
		case <-ticker2.C:
			go d.bootnodeDiscovery()
		}
```

network/discovery.go
```diff
@@ -246,22 +252,93 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
 
 func (d *discovery) handleDiscovery() {
 	// take a random peer and find peers
-	if target := d.peers.getRandomPeer(); target != nil {
-		if err := d.attemptToFindPeers(target.id); err != nil {
-			d.srv.logger.Error("failed to dial peer", "peer", target.id, "err", err)
+	if d.srv.numOpenSlots() > 0 {
+		if target := d.peers.getRandomPeer(); target != nil {
+			if err := d.attemptToFindPeers(target.id); err != nil {
+				d.srv.logger.Error("failed to dial peer", "peer", target.id, "err", err)
```
Kourin1996: `failed to dial peer` is correct log message?

network/discovery.go
```diff
@@ -246,22 +252,93 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+		return
+	}
+
+	if err := stream.(*rawGrpc.ClientConn).Close(); err != nil {
+		d.srv.logger.Error("Error closing grpc stream", "peer", bootNode.ID, "err", err)
+
+		return
+	}
+
+	d.srv.Disconnect(bootNode.ID, "Thank you")
```
Kourin1996: What happens if the bootNode is the peer the node had connected already when entering this function. Do you need to disconnect in this case?

network/identity.go
```diff
@@ -117,13 +128,37 @@ func (i *identity) start() error {
...
+	}
+	if _, ok := i.srv.temporaryDials.Load(peerID); ok {
+		status.TemporaryDial = true
+	}
+
+	return status
+}
+
+// checkSlots checks for the available connection slots and disconnects if slots are full
+func (i *identity) checkSlots(direction network.Direction, peerID peer.ID) (slotsFull bool) {
```
Kourin1996: `checkSlots` would confuse because will disconnect from peer inside, checking is not the main task here.

network/discovery.go
munna0908: We only dial non connected bootnode, but if a node establishes a connection during the rpc call, we shouldn't disconnect. Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/458ee921952fb132999c40f8f9c0cedc5111daf1)

network/discovery.go
munna0908: Fixed it Commit [Link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/628f23f46d02e9622b7378068dfd905180bd8a3a)

network/identity.go
munna0908: Renamed it, Commit [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/23abbb4760584a06d5fbabd156a6bb83eef8b0ee)

command/server/server_command.go
munna0908: Cleaned up commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/62ece3a00a9b352650e95900bff12c1c57516486)

network/discovery.go
munna0908: Modified it, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/43da52617507c5f7069aaa6e1f0c995f0796fd4f)

network/server.go
```diff
@@ -349,11 +358,27 @@ func (s *Server) numPeers() int64 {
...
+	size := int64(len(s.discovery.bootnodes))
+	for i := 0; i < int(size); i++ {
+		randNum, _ := rand.Int(rand.Reader, big.NewInt(size))
+		if peer := s.discovery.bootnodes[randNum.Int64()]; !s.hasPeer(peer.ID) {
+			return peer
+		}
+	}
+
+	return nil
+}
```
Kourin1996: I'm not sure if this function works or not. Is there a possibility of that `rand.Int` returns always the same value during loop? (Of course, it's rare but it's possible) In this case, this function might return nil even though there are such peers.

You might want to create slice of peers that don't connect to the node first.

munna0908: Modified it, Commit [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/b798474b4de8a2439a053b5918ef98f0dc95ed48)

command/server/server_command.go
```diff
@@ -143,8 +143,15 @@ func (c *ServerCommand) DefineFlags() {
...
-		Description: fmt.Sprintf("Sets the client's max peer count. Default: %d", helper.DefaultConfig().Network.MaxPeers),
+	c.FlagMap["max-inbound-peers"] = helper.FlagDescriptor{
+		Description: "Sets the client's maximum number of inbound peers allowded. Default: 40",
+		Arguments: []string{
+			"PEER_COUNT",
+		},
+		FlagOptional: true,
+	}
+	c.FlagMap["max-outbound-peers"] = helper.FlagDescriptor{
+		Description: "Sets the client's max no.of outbound peers allowded. Default: 10",
```
zivkovicmilos: Please try to be consistent with the descriptions.

In the description for `max-inbound-peers` you've used `maximum number of ...`, but here you abbreviate it to `max no.of`

network/proto/identity.proto
```diff
@@ -22,7 +22,9 @@ message Status {
 
     int64 chain = 3;
 
-    string genesis = 4;
+    bool temporaryDial=4;
+
+    string genesis = 5;
```
zivkovicmilos: The whole point of using proto files is that they are backwards compatible.
Changing the number of the field in proto files essentially breaks it:

Please refer to the [official proto docs](https://developers.google.com/protocol-buffers/docs/proto#assigning_field_numbers) to understand why this is important

Adding new fields should involve increasing the number, not _reusing_ the old number from another field

network/discovery.go
```diff
@@ -20,6 +22,12 @@ var discProto = "/disc/0.1"
 
 const defaultBucketSize = 20
 
+const defaultPeerReqCount = 16
+
+const peerDiscoveryInterval = 5 * time.Second
+
+const bootnodeDiscoveryInterval = 60 * time.Second
+
```
zivkovicmilos: Please group these constant inits under a single constant wrapper

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
 }
 
 func (d *discovery) run() {
+	ticker1 := time.NewTicker(peerDiscoveryInterval)
+	ticker2 := time.NewTicker(bootnodeDiscoveryInterval)
```
zivkovicmilos: These tickers should have more descriptive names, like `peerDiscoveryTicker` and `bootnodeDiscoveryTicker`

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+}
+
+// bootnodeDiscovery queries a random bootnode for new peers and adds them to the routing table
+func (d *discovery) bootnodeDiscovery() {
+	if d.srv.numOpenSlots() <= 0 {
+		return
+	}
+
+	// get a random bootnode which is not connected
+	bootNode := d.srv.getBootNode()
```
zivkovicmilos: Nitpick warning: everywhere in the codebase we use `bootnode`, but for some reason it's camel case in this method

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+	if bootNode == nil {
+		return
+	}
+
+	// isTemporaryDial maintains the dial status
+	var isTemporaryDial bool
+
+	// if atlest one bootnode is connected the dial status is temporary
```
zivkovicmilos: Spelling error `atlest`

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+	if d.bootnodeConnCount > 1 {
+		isTemporaryDial = true
+	}
+
+	if isTemporaryDial {
+		if _, loaded := d.srv.temporaryDials.LoadOrStore(bootNode.ID, true); loaded {
+			return
+		}
+	}
```
zivkovicmilos: You can simplify this entire if structure to avoid double nesting

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
 
 func (d *discovery) handleDiscovery() {
 	// take a random peer and find peers
-	if target := d.peers.getRandomPeer(); target != nil {
-		if err := d.attemptToFindPeers(target.id); err != nil {
-			d.srv.logger.Error("failed to dial peer", "peer", target.id, "err", err)
+	if d.srv.numOpenSlots() > 0 {
+		if target := d.peers.getRandomPeer(); target != nil {
+			if err := d.attemptToFindPeers(target.id); err != nil {
+				d.srv.logger.Error("failed to find new peers", "peer", target.id, "err", err)
```
zivkovicmilos: CLI error logs should start out with a capitalized word

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+		}
+	}
+
+	if len(d.srv.host.Peerstore().Addrs(bootNode.ID)) == 0 {
+		d.srv.host.Peerstore().AddAddr(bootNode.ID, bootNode.Addrs[0], peerstore.AddressTTL)
+	}
+
+	stream, err := d.srv.NewProtoStream(discProto, bootNode.ID)
+	if err != nil {
+		d.srv.logger.Error("failed to open new stream", "peer", bootNode.ID, "err", err)
```
zivkovicmilos: CLI error logs should start out with a capitalized word

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+
+		return
+	}
+
+	clt := proto.NewDiscoveryClient(stream.(*rawGrpc.ClientConn))
+
+	resp, err := clt.FindPeers(context.Background(), &proto.FindPeersReq{Count: defaultPeerReqCount})
+	if err != nil {
+		d.srv.logger.Error("find peers call failed", "peer", bootNode.ID, "err", err)
```
zivkovicmilos: CLI error logs should start out with a capitalized word

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+		return
+	}
+
+	if err := stream.(*rawGrpc.ClientConn).Close(); err != nil {
+		d.srv.logger.Error("Error closing grpc stream", "peer", bootNode.ID, "err", err)
+
+		return
+	}
+
+	if !d.srv.hasPeer(bootNode.ID) || isTemporaryDial {
```
zivkovicmilos: I don't understand this check:
* If the node is **not** a peer -> disconnect?

network/discovery.go
```diff
@@ -246,22 +264,103 @@ func (d *discovery) findPeersCall(peerID peer.ID) ([]*peer.AddrInfo, error) {
...
+	for _, node := range resp.Nodes {
+		info, err := StringToAddrInfo(node)
+		if err != nil {
+			continue
+		}
+
+		if err := d.addToTable(info); err != nil {
+			d.srv.logger.Error("Unable to add peer to routing table", "peer", bootNode.ID, "err", err)
 		}
 	}
```
zivkovicmilos: This should be extracted out to a separate method

network/server.go
```diff
@@ -377,14 +408,35 @@ func (s *Server) hasPeer(peerID peer.ID) bool {
 }
 
 func (s *Server) numOpenSlots() int64 {
```
zivkovicmilos: Please rename this method to something more clear, since it returns the number of free outbound connections

network/identity.go
```diff
@@ -117,13 +128,37 @@ func (i *identity) start() error {
 	return nil
 }
 
-func (i *identity) getStatus() *proto.Status {
-	return &proto.Status{
-		Chain: int64(i.srv.config.Chain.Params.ChainID),
+func (i *identity) getStatus(peerID peer.ID) *proto.Status {
```
zivkovicmilos: Method is called `getStatus`, but you initialize the status response here

network/identity.go
```diff
@@ -132,9 +167,11 @@ func (i *identity) handleConnected(peerID peer.ID) error {
 
 	clt := proto.NewIdentityClient(conn.(*rawGrpc.ClientConn))
 
-	status := i.getStatus()
-	resp, err := clt.Hello(context.Background(), status)
+	status := i.getStatus(peerID)
 
+	status.Metadata["peerID"] = i.srv.host.ID().Pretty()
```
zivkovicmilos: Make `peerID` a constant

network/identity.go
```diff
@@ -144,13 +181,20 @@ func (i *identity) handleConnected(peerID peer.ID) error {
...
+	if !resp.TemporaryDial && !status.TemporaryDial {
+		i.srv.addPeer(peerID, direction)
+	}
 
 	return nil
 }
 
 func (i *identity) Hello(ctx context.Context, req *proto.Status) (*proto.Status, error) {
-	return i.getStatus(), nil
+	peerID, err := peer.Decode(req.Metadata["peerID"])
```
zivkovicmilos: Make `peerID` a constant

network/identity.go
```diff
@@ -144,13 +181,20 @@ func (i *identity) handleConnected(peerID peer.ID) error {
 		return ErrInvalidChainID
 	}
 
-	i.srv.addPeer(peerID)
+	if !resp.TemporaryDial && !status.TemporaryDial {
```
zivkovicmilos: This adds a peer only if both nodes are doing a temporary dial.

Peers won't be added if one node is doing a temporary dial, and the other is doing a persistent dial

Is this the behavior we want?

cc @Kourin1996

network/server.go
```diff
@@ -377,14 +408,35 @@ func (s *Server) hasPeer(peerID peer.ID) bool {
...
+	count := atomic.LoadInt64(&s.inboundConnCount)
+	if count < 0 {
+		count = 0
+	}
+
+	return count + s.identity.pendingInboundConns()
+}
+
+func (s *Server) outboundConns() int64 {
+	return (s.numPeers() - atomic.LoadInt64(&s.inboundConnCount)) + s.identity.pendingOutboundConns()
```
zivkovicmilos: Please simplify this method

network/server_test.go
```diff
@@ -690,19 +698,48 @@ func TestMinimumBootNodeCount(t *testing.T) {
...
+
+func TestTemporaryDial(t *testing.T) {
+	defaultConfig := &CreateServerParams{
+		ConfigCallback: func(c *Config) {
+			c.NoDiscover = true
+		},
+	}
+	servers, err := createServers(2, map[int]*CreateServerParams{0: defaultConfig, 1: defaultConfig})
+
+	assert.NoError(t, err)
```
zivkovicmilos: This shouldn't be an assert, but a fatal test stopping condition

network/server_test.go
```diff
@@ -690,19 +698,48 @@ func TestMinimumBootNodeCount(t *testing.T) {
...
+	//connect server0 to server1
+	err = servers[0].host.Connect(context.Background(), *servers[1].AddrInfo())
+	assert.NoError(t, err)
+	// ensure that the connection is established
+	assert.Equal(t, network.Connected, servers[0].host.Network().Connectedness(servers[1].host.ID()))
+	// wait until identity handshake is done
+	time.Sleep(10 * time.Second)
+	// since it is temporary dial, server should not contain its peer in the peerlist
+	assert.NotContains(t, servers[0].peers, servers[1].host.ID())
+	assert.NotContains(t, servers[1].peers, servers[0].host.ID())
```
zivkovicmilos: Please clean up the test so it's more readable

network/server_test.go
```diff
@@ -690,19 +698,48 @@ func TestMinimumBootNodeCount(t *testing.T) {
...
+	})
+	//add server1 to server0 temperary dial list
+	servers[0].temporaryDials.Store(servers[1].host.ID(), true)
+	//connect server0 to server1
+	err = servers[0].host.Connect(context.Background(), *servers[1].AddrInfo())
+	assert.NoError(t, err)
+	// ensure that the connection is established
+	assert.Equal(t, network.Connected, servers[0].host.Network().Connectedness(servers[1].host.ID()))
+	// wait until identity handshake is done
+	time.Sleep(10 * time.Second)
```
zivkovicmilos: You should never, _never_, **never** do an arbitrary sleep in tests, especially unit tests.

Please fix this - find a way to precisely know if the handshake succeeded.

zivkovicmilos (changes requested):
There are some essential changes that need to be made before proceeding 👍

command/server/server_command.go
```diff
@@ -143,8 +143,15 @@ func (c *ServerCommand) DefineFlags() {
 		FlagOptional: true,
 	}
 
-	c.FlagMap["max-peers"] = helper.FlagDescriptor{
-		Description: fmt.Sprintf("Sets the client's max peer count. Default: %d", helper.DefaultConfig().Network.MaxPeers),
+	c.FlagMap["max-inbound-peers"] = helper.FlagDescriptor{
+		Description: "Sets the client's maximum number of inbound peers allowded. Default: 40",
```
Kourin1996: Is it possible to give default value from helper.DefaultConfig().Network so that we can follow up the change of default config easily?

command/server/server_command.go
Kourin1996: Is it possible to give default value from helper.DefaultConfig().Network so that we can follow up the change of default config easily?

network/discovery.go
zivkovicmilos: Resolving as the use of `LoadOrStore` complicates logic, and you rely on values set in the map to be `true` instead of checking their value when dialing

network/discovery.go
munna0908: assume a case, where node1 makes a temporary conn to node 2, in the meantime, node 2 wants to make a permanent dial, in that case we should only disconnect it is not in our peers list.

network/identity.go
munna0908: @zivkovicmilos We only want to add a peer if both the participants wants to make a persistent dial because the sole purpose of this logic is to make  temporary call and find peers. BTW  If Node-1 is initiating  a temporary dial, Node-2 can't make a persistent dial, because node-1 marks this peer as a temporary dial peer.

command/server/server_command.go
munna0908: Sure, Commit [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d23d22d3f927960379bf1336b3a8d197349c119d)

network/server_test.go
munna0908: Got it, Here is the commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/ea9b338d7b7cced2866e0ee80785c2ccd4fdf7f2)

network/server_test.go
munna0908: Done,Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/ea9b338d7b7cced2866e0ee80785c2ccd4fdf7f2)

network/server_test.go
munna0908: Done,Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/ea9b338d7b7cced2866e0ee80785c2ccd4fdf7f2)

network/identity.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d3a3b8c0751b8d49287ec786a62eec524f3faef6)

network/identity.go
munna0908: renamed it, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d3a3b8c0751b8d49287ec786a62eec524f3faef6)

network/identity.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d3a3b8c0751b8d49287ec786a62eec524f3faef6)

network/server.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d9d37fe56f78bdd5d13b4d90975882044801b0be)

network/discovery.go
munna0908: Done, Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/2c55899ddfb8387fe2fd2e4071e5d2f0e40ad832)

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/proto/identity.proto
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/a0959b87ebc593e502365bd91a3bc1b6011e7a4b)

network/discovery.go
munna0908: commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

network/discovery.go
munna0908: This is a redundant check, removed it - [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/aa333f9fcf309b004f3c1fbad82398cc7118a3b7)

command/server/server_command.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/6514771aac425fde81aa4f51c72c40e41a696a7b)

command/server/server_command.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/6514771aac425fde81aa4f51c72c40e41a696a7b)

network/testing.go
```diff
@@ -363,3 +363,24 @@ func closeTestServers(t *testing.T, servers []*Server) {
 		assert.NoError(t, server.Close())
 	}
 }
+
+func isServerConnectedTo(server *Server, ids peer.ID) bool {
+	connectCtx, cancelWait := context.WithTimeout(context.Background(), DefaultJoinTimeout)
```
Kourin1996: Why not set timeout?

network/testing.go
```diff
@@ -363,3 +363,24 @@ func closeTestServers(t *testing.T, servers []*Server) {
...
+	defer cancelWait()
+
+	// This should fail, since the server0 made a temporary dial to server1
```
Kourin1996: This comment is directing to only `TestTemporaryDial`.

network/testing.go
```diff
@@ -363,3 +363,24 @@ func closeTestServers(t *testing.T, servers []*Server) {
...
+
+	return true
+}
+
+// dialServer dials the peer with the provided address info
+func dialServer(server *Server, addrs peer.AddrInfo, isTemporary bool) error {
+	if isTemporary {
+		server.temporaryDials.Store(addrs.ID, true)
+	}
+	//connect server0 to server1
```
Kourin1996: This comment is directing to only TestTemporaryDial.

network/discovery.go
```diff
@@ -218,52 +247,114 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+		return
+	}
+
+	// isTemporaryDial maintains the dial status
+	var isTemporaryDial bool
+
+	// if atleast one bootnode is connected the dial status is temporary
+	if d.bootnodeConnCount > 1 {
+		isTemporaryDial = true
+	}
```
Kourin1996: why don't you assign value to isTemporaryDial directly?
```suggestion
         // if atleast one bootnode is connected the dial status is temporary
	isTemporaryDial := d.bootnodeConnCount > 1
```

network/discovery.go
```diff
@@ -218,52 +247,114 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+	bootnode := d.srv.getBootNode()
+	if bootnode == nil {
+		return
+	}
+
+	// isTemporaryDial maintains the dial status
+	var isTemporaryDial bool
+
+	// if atleast one bootnode is connected the dial status is temporary
+	if d.bootnodeConnCount > 1 {
```
Kourin1996: Why don't you use atomic.LoadInt32 here?

Kourin1996: The comment is saying `if atleast one bootnode is connected the dial status is temporary` but isTemporaryDial will be true if the bootnodeConnCount equals to or greater than 2.

network/discovery.go
```diff
@@ -218,52 +247,114 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+		isTemporaryDial = true
+	}
+
+	if isTemporaryDial {
+		if _, loaded := d.srv.temporaryDials.LoadOrStore(bootnode.ID, true); loaded {
+			return
 		}
```
Kourin1996: temporary dial flag won't reset if node failed to dial peer. for example, keeping this flag true if NewProtoStream returns error and you can't dial the peer forever.

network/discovery.go
```diff
@@ -218,52 +247,114 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+	if err := stream.(*rawGrpc.ClientConn).Close(); err != nil {
+		d.srv.logger.Error("Error closing grpc stream", "peer", bootnode.ID, "err", err)
+
+		return
+	}
+
+	if isTemporaryDial {
+		d.srv.temporaryDials.Delete(bootnode.ID)
+		d.srv.Disconnect(bootnode.ID, "Thank you")
+	}
```
Kourin1996: When I tried, ConnectedF was called after a node disconnected here. then node continued to connect to the peer.

Kourin1996 (changes requested):

network/discovery.go
munna0908: We dont dial peer based on the value associated with the key. we just check whether the key exsists.

network/discovery.go
munna0908: can you help me with the scenario.

network/discovery.go
munna0908: I felt like, it would be more readable &  idiomatic to separate declaration with condition.

network/testing.go
munna0908: Can you give more context, timeout for what ?

network/discovery.go
munna0908: Done,commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/b84806b54fb55fdfeb0142de3f552a199ed1a96c)

munna0908: rephrased it. [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/b84806b54fb55fdfeb0142de3f552a199ed1a96c)

network/testing.go
munna0908: Removed it, [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/3e3ac480078ba5d30f18e2153ccca91a39113538)

network/testing.go
munna0908: Removed it, [link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/3e3ac480078ba5d30f18e2153ccca91a39113538)

network/testing.go
Kourin1996: sorry, never mind. You've set timeout already

network/discovery.go
Kourin1996: I'm afraid I don't get it. If so why you're using LoadOrStore here? And when do you reset (meaning set false)?

network/discovery.go
Kourin1996: @munna0908 Please share the details of the scenario you have done during working on this PR.

Kourin1996: I tested with the following nodes:
node1: Bootnodes => node1, node2, node3
node2: Bootnodes => node1, node2
node3: Bootnodes => node3, node4
node4: Bootnodes => node3, node4

node1 won't be connecting to node3 first because node3 starts after node1 and node3 will not try to connect to node1. node1 will try to connect to node3 in bootnodeDiscovery after a while and query for its table. In this case, node1 was keeping connection after bootnodeDiscovery is done.

network/discovery.go
munna0908: Got it @Kourin1996 , Fixed it commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/a351a5c113fa8b71d16327ceeb3bbd0257cc3c1c)

network/discovery.go
munna0908: @Kourin1996 I got the issue, fixed it ([link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/677053d94aec7557bc7d7778900ad03280f4caf2))

network/discovery.go
zivkovicmilos: The correct spelling is `at least`

network/discovery.go
```diff
@@ -177,6 +177,21 @@ func (d *discovery) isBootNode(id peer.ID) bool {
 
 	return false
 }
+
+func (d *discovery) addPeersToTable(nodes []string) error {
+	for _, node := range nodes {
+		info, err := StringToAddrInfo(node)
+		if err != nil {
+			continue
```
zivkovicmilos: Why not log the error?

network/discovery.go
```diff
@@ -177,6 +177,21 @@ func (d *discovery) isBootNode(id peer.ID) bool {
 
 	return false
 }
+
+func (d *discovery) addPeersToTable(nodes []string) error {
```
zivkovicmilos: Why should we error out even if 1 peer connection is unsuccessful?

We could be trying to connect to 10 peers, 1 of them will fail - we are supposed to stop trying to connect to the remaining 10?

This doesn't make sense to me, it's not a game breaking event

cc @Kourin1996

network/identity.go
zivkovicmilos: The method is called `currentStatus`, but you initialize the status inside the method. A more appropriate name here is `constructStatus` or something similar

cc @Kourin1996

network/server_test.go
```diff
@@ -724,22 +724,23 @@ func TestTemporaryDial(t *testing.T) {
...
 	t.Cleanup(func() {
 		closeTestServers(t, servers)
 	})
-	//add server1 to server0 temperary dial list
-	servers[0].temporaryDials.Store(servers[1].host.ID(), true)
-	//connect server0 to server1
-	err = servers[0].host.Connect(context.Background(), *servers[1].AddrInfo())
+
+	//make a temporary dail to server1
+	err = dialServer(servers[0], *servers[1].AddrInfo(), true)
```
zivkovicmilos: You can inline this error

network/server.go
```diff
@@ -376,15 +407,38 @@ func (s *Server) hasPeer(peerID peer.ID) bool {
...
+	}
+
+	return count + s.identity.pendingInboundConns()
+}
+
+func (s *Server) outboundConns() int64 {
+	activeOutboundConns := s.numPeers() - atomic.LoadInt64(&s.inboundConnCount)
+
+	return activeOutboundConns + s.identity.pendingOutboundConns()
+}
```
brkomir, line 432: Honestly, I don't understand this. `s.numPeers()` returns the length of the `peers` field of type `map[peer.ID]*Peer` in the `Server` struct. Peers are identified by `ID` field, so how does this work `activeOutboundConns := s.numPeers() - atomic.LoadInt64(&s.inboundConnCount)` if one peer can have both inbound/outbound connections? `s.numPeers()` isn't the same as the number of total connections in my understanding.

brkomir: Honestly, I don't understand this. `s.numPeers()` returns the length of the `peers` field of type `map[peer.ID]*Peer` in the `Server` struct. Peers are identified by ID field, so how does this work `activeOutboundConns := s.numPeers() - atomic.LoadInt64(&s.inboundConnCount)` if one peer can have both inbound/outbound connections? s.numPeers() isn't the same as the number of total connections in my understanding.

network/discovery.go
munna0908: Commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/12c1341ba166c39ce334c47b8943697cfabdfccd)

network/server_test.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/9f5c60e3ab5d0828f364ee2b5165d6fe36b62bf1)

network/discovery.go
munna0908: Modified it, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/12c1341ba166c39ce334c47b8943697cfabdfccd)

zivkovicmilos (approved):
Looks good now 💯 

Make sure to merge out the additions from https://github.com/0xPolygon/polygon-edge/pull/326 (probably when it gets merged to develop), as we've added some features there that are removed here

network/discovery.go
```diff
@@ -218,54 +250,125 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+
+	if err := stream.(*rawGrpc.ClientConn).Close(); err != nil {
+		d.srv.logger.Error("Error closing grpc stream", "peer", bootnode.ID, "err", err)
+
+		return
+	}
+
+	if isTemporaryDial {
+		d.srv.Disconnect(bootnode.ID, "Thank you")
+	}
```
Kourin1996: You added the same code inside defer above

Kourin1996 (changes requested):

network/server.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d9d37fe56f78bdd5d13b4d90975882044801b0be)

network/discovery.go
munna0908: removed it, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d4489ac2fdfeb274d783226a8aed1da5d13b99ea)

munna0908:

munna0908:
recheck

network/identity.go
munna0908: Done @zivkovicmilos  commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/cdbb38680ac0aa125d2d7e856028ecaab55bfd46)

network/discovery.go
munna0908: Done, commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/677053d94aec7557bc7d7778900ad03280f4caf2)

brkomir (commented):
I don't understand one thing. Correct me if I'm wrong, you calculate this coefficient (MaxOutboundPeers + MaxInboundPeers) / MaxOutboundPeers = (8 + 32) / 8 = 0.2. But we can have both inbound and outbound connections to the same peer, by your calculation, you would count the same peer twice in the numerator part but only once in the denominator part of this fraction?
Also, you use the terms MaxOutboundPeers and MaxInboundPeers in the configuration part but you use the term Connections afterward.

network/server.go
lazartravica: I understand the reasoning behind `numPeers` acting as a total connection count, since in `runDial` we will not dial any peers that are already connected, but I don't see anything preventing a node from connecting to us if we are already connected to it (ofcourse, leaving aside the possibility that the peer doesn't have the same version of Polygon Edge running).

lazartravica (changes requested):
Left a comment on using numPeers and not having always good total connection count.

network/discovery.go
```diff
@@ -218,54 +250,121 @@ func (d *discovery) getStream(peerID peer.ID) (interface{}, error) {
...
+		if _, loaded := d.srv.temporaryDials.LoadOrStore(bootnode.ID, true); loaded {
+			return
+		}
+	}
+
+	defer func() {
+		if isTemporaryDial {
+			d.srv.temporaryDials.Delete(bootnode.ID)
+		}
+	}()
```
Kourin1996: Actually what is `isTemporaryDial` for? It's just set to map and removed when existing. There was disconnecting in this function, I remember

Kourin1996 (changes requested):

network/server.go
munna0908: @lazartravica if server1 is connected to server2, all the new streams opened by server 2 to server 1 are multiplexed into a single connection by libp2p.

network/discovery.go
munna0908: fixed it commit-[link](https://github.com/0xPolygon/polygon-edge/pull/343/commits/8dfcf9a3244ae1c5cd8575762de3367b7c1df72d)

Kourin1996 (approved):
LGTM!

network/discovery/discovery.go
```diff
@@ -0,0 +1,426 @@
...
+
+	closeCh chan struct{} // Channel used for stopping the DiscoveryService
+}
+
+// NewDiscoveryService creates a new instance of the discovery service
+func NewDiscoveryService(
+	server networkingServer,
+	routingTable *kb.RoutingTable,
+	logger hclog.Logger,
+	closeCh chan struct{},
```
dbrajovic: If this closeCh is being created upon calling NewDiscoveryService, perhaps it should be created inside this method and not outside it

network/discovery/discovery.go
```diff
@@ -0,0 +1,426 @@
...
+	GetPeerInfo(peerID peer.ID) *peer.AddrInfo
+
+	// GetRandomPeer fetches a random peer from the server's peer store
+	GetRandomPeer() *peer.ID
+
+	// TEMPORARY DIALING //
+
+	// FetchAndSetTemporaryDial checks if the peer connection is a temporary dial,
+	// and sets a new value accordingly
+	FetchAndSetTemporaryDial(peerID peer.ID, newValue bool) bool
```
dbrajovic: I find this naming to be a little misleading since `newValue` will be set once initially and loaded on subsequent calls. Maybe change it to `FetchOrSetTemporaryDial` ?

zivkovicmilos: Completely agree 👍 

Fixed the naming here

[d2f3bc8](https://github.com/0xPolygon/polygon-edge/pull/343/commits/d2f3bc89f9bd4607c77813a0ddd679607f31d939)

network/discovery/discovery.go
```diff
@@ -0,0 +1,427 @@
...
+	// available to all the libp2p services
+	d.baseServer.AddToPeerStore(node)
+
+	if _, err := d.routingTable.TryAddPeer(
+		node.ID,
+		false,
+		false,
+	); err != nil {
+		return err
+	}
```
dbrajovic: As we discussed, it's probably best to reverse the order of these 2 calls (if that wouldn't break anything). Because in case of a failing `TryAddPeer`, the peer will still be added to the peerstore, yet it will not be set in the routing table.

network/testing/testing.go
```diff
@@ -0,0 +1,347 @@
+package testing
```
dbrajovic, line 1: This file should be renamed to `mock_test.go` (or anything ending with a `_test`) so it's not included in the resulting binary.

network/identity/identity.go
```diff
@@ -0,0 +1,216 @@
...
+
+var (
+	ErrInvalidChainID   = errors.New("invalid chain ID")
+	ErrNotReady         = errors.New("not ready")
+	ErrNoAvailableSlots = errors.New("no available Slots")
+)
+
+// networkingServer defines the base communication interface between
+// any networking server implementation and the IdentityService
+type networkingServer interface {
```
dbrajovic, line 25: The implementations for this interface (currently in `server.go`) should be moved in a separate .go file - just to make `server.go` less cluttered.

network/discovery/discovery.go
```diff
@@ -0,0 +1,427 @@
...
+	peerDiscoveryInterval = 5 * time.Second
+
+	// bootnodeDiscoveryInterval is the interval at which
+	// random bootnodes are dialed for their peer sets
+	bootnodeDiscoveryInterval = 60 * time.Second
+)
+
+// networkingServer defines the base communication interface between
+// any networking server implementation and the DiscoveryService
+type networkingServer interface {
```
dbrajovic, line 35: The implementations for this interface (currently in `server.go`) should be moved in a separate .go file - just to make `server.go` less cluttered.

network/server.go
```diff
@@ -237,70 +179,291 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+	)
+
+	// Register the identity service protocol
+	s.registerIdentityService(identityService)
+
+	// Register the network notify bundle handlers
+	s.host.Network().Notify(identityService.GetNotifyBundle())
+
+	// Set the identity service
+	s.identity = identityService
```
dbrajovic: If we're only setting this field and the server is not doing anything with it, then it should be removed from the struct.

network/identity/identity.go
```diff
@@ -0,0 +1,216 @@
...
+	clt, clientErr := i.baseServer.NewIdentityClient(peerID)
+	if clientErr != nil {
+		return fmt.Errorf(
+			"unable to create new identity client connection, %w",
+			clientErr,
+		)
+	}
+
+	status := i.constructStatus(peerID)
+	status.Metadata[PeerID] = i.hostID.Pretty()
```
dbrajovic: This can go inside the body of `constructStatus`.

network/event/peer_event.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	PeerConnected        PeerEventType = iota // Emitted when a peer connected
+	PeerFailedToConnect                       // Emitted when a peer failed to connect
+	PeerDisconnected                          // Emitted when a peer disconnected from node
+	PeerDialCompleted                         // Emitted when a peer completed dial
+	PeerAddedToDialQueue                      // Emitted when a peer is added to dial queue
+)
+
+var peerEventToName = map[PeerEventType]string{
```
dbrajovic, line 15: I think we can get rid of this map, if we implement the `String()` method to switch on the underlying value.

network/server.go
```diff
@@ -516,117 +776,146 @@ var (
...
+		return nil
+	}
+
+	return connectionInfo.getProtocolStream(protocol)
 }
 
-func (s *Server) NewProtoStream(proto string, id peer.ID) (interface{}, error) {
-	s.protocolsLock.Lock()
-	defer s.protocolsLock.Unlock()
+// NewDiscoveryClient returns a new or existing discovery service client connection
```
dbrajovic, line 856: This could go in a separate function

network/server.go
```diff
@@ -516,117 +776,146 @@ var (
...
+	// Create a new stream connection and return it
+	protoStream, err := s.newProtoConnection(common.DiscProto, peerID)
+	if err != nil {
+		return nil, err
 	}
 
-	stream, err := s.NewStream(proto, id)
+	// Discovery protocol streams should be saved,
+	// since they are referenced later on
+	s.peersLock.Lock()
```
dbrajovic: This logic could go in a separate function

network/discovery/discovery.go
```diff
@@ -0,0 +1,427 @@
...
+	// BOOTNODE QUERIES //
+
+	// IsBootnode checks if the peer is a registered bootnode
+	IsBootnode(peerID peer.ID) bool
```
Kourin1996: Do you need `IsBootnode` for this? Looks not to be used.

network/discovery/discovery.go
```diff
@@ -0,0 +1,427 @@
...
+	// GetRandomBootnode fetches a random bootnode, if any
+	GetRandomBootnode() *peer.AddrInfo
+
+	// GetBootnodeConnCount fetches the number of bootnode connections [Thread safe]
+	GetBootnodeConnCount() int64
+
+	// PROTOCOL MANIPULATION //
+
+	// NewDiscoveryClient returns a discovery gRPC client connection
+	NewDiscoveryClient(peerID peer.ID) (proto.DiscoveryClient, error)
```
Kourin1996, line 47: Why do you get discovery client from base server despite of that DiscoveryService is defined here?
Should DiscoveryClient be defined in same package, right?

network/discovery/discovery.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	return d.routingTable.Size()
+}
+
+// RoutingTablePeers fetches the peers from the routing table
+func (d *DiscoveryService) RoutingTablePeers() []peer.ID {
+	if d.routingTable == nil {
+		return []peer.ID{}
+	}
```
Kourin1996: Is there a possibility of that d.routingTable is nil?

network/server.go
```diff
@@ -237,70 +179,291 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+				s.removePeer(conn.RemotePeer())
 			}()
 		},
 	})
 
 	return nil
 }
 
+// setupDiscovery Sets up the discovery service for the node
+func (s *Server) setupDiscovery() error {
```
Kourin1996: Why not move this process into discovery package?

network/server.go
```diff
@@ -237,70 +179,291 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+	// Create an instance of the discovery service
+	discoveryService := discovery.NewDiscoveryService(
+		s,
+		routingTable,
+		s.logger,
+		make(chan struct{}),
+	)
+
+	// Register a network event handler
+	if subscribeErr := s.SubscribeFn(discoveryService.HandleNetworkEvent); subscribeErr != nil {
```
Kourin1996: I think it's natural that Subscriber calls Subscribe method

Kourin1996:
Why not create network/e2e package?

Kourin1996 (changes requested):

network/connections/connections.go
```diff
@@ -0,0 +1,150 @@
+package connections
```
Kourin1996: Did you need to create new package for ConnectionInfo?

network/dial/dial_task.go
```diff
@@ -0,0 +1,18 @@
...
+
+	// info of the task
+	addr *peer.AddrInfo
+
+	// priority of the task (the higher the better)
+	priority uint64
+}
+
+// GetTaskInfo returns the peer information associated with the dial
+func (dt *DialTask) GetTaskInfo() *peer.AddrInfo {
```
Kourin1996: Why TaskInfo instead of Addr or AddrInfo?

Kourin1996:
@zivkovicmilos I faced a failing test when running tests locally today.

Screenshot 2022-03-28 at 9 27 16

```bash
...
?       github.com/0xPolygon/polygon-edge/licenses      [no test files]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x4abbda6]

goroutine 25932 [running]:
github.com/0xPolygon/polygon-edge/network.(*PeerConnInfo).addProtocolStream(...)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/server.go:232
github.com/0xPolygon/polygon-edge/network.(*Server).NewDiscoveryClient(0xc0001c7c20, 0xc006f0e2a0, 0x27, 0x10, 0x5e01878, 0x10, 0xc004a42090)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/server.go:873 +0x186
github.com/0xPolygon/polygon-edge/network/discovery.(*DiscoveryService).findPeersCall(0xc0089e91d0, 0xc006f0e2a0, 0x27, 0xc003904100, 0x2, 0x2, 0x0, 0x0, 0x0)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/discovery/discovery.go:240 +0x62
github.com/0xPolygon/polygon-edge/network/discovery.(*DiscoveryService).attemptToFindPeers(0xc0089e91d0, 0xc006f0e2a0, 0x27, 0xc000cb4080, 0x4fe6e20)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/discovery/discovery.go:223 +0x10a
github.com/0xPolygon/polygon-edge/network/discovery.(*DiscoveryService).regularPeerDiscovery(0xc0089e91d0)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/discovery/discovery.go:307 +0x95
created by github.com/0xPolygon/polygon-edge/network/discovery.(*DiscoveryService).startDiscovery
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/network/discovery/discovery.go:282 +0x150
FAIL    github.com/0xPolygon/polygon-edge/network       72.716s
...
```

network/identity/identity.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/0xPolygon/polygon-edge/network/proto"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const PeerID = "peerID"
+
+var (
+	ErrInvalidChainID   = errors.New("invalid chain ID")
+	ErrNotReady         = errors.New("not ready")
```
dbrajovic: This seems to be unused.

network/discovery/discovery.go
zivkovicmilos: You're right, only the discovery service interacts with this channel 👍 

Fixed:

[a051b2b](https://github.com/0xPolygon/polygon-edge/pull/343/commits/a051b2bb8c3dbe9b748bb6fb9a5f5b6060b5ee44)

network/discovery/discovery.go
zivkovicmilos: Agreed 👍 

I've added rollback support, as well as a unit test covering both scenarios (since in fact this peer store addition needs to be present before adding it to the routing table):

[25e424f](https://github.com/0xPolygon/polygon-edge/pull/343/commits/25e424fc14a3708a1704165796bbe8f41392da74)

network/testing/testing.go
zivkovicmilos: Fixed 👍 

[1b09821](https://github.com/0xPolygon/polygon-edge/pull/343/commits/1b09821386f05887470dfe532f459632012b725c)

zivkovicmilos: Actually this can't be a `_test.go` file because it needs to be imported elsewhere.

I've reverted it here:

[47460e1](https://github.com/0xPolygon/polygon-edge/pull/343/commits/47460e1008876753ddd3259a9d2d52e0605c6573)

network/identity/identity.go
zivkovicmilos: Fixed 👍 

[47460e1](https://github.com/0xPolygon/polygon-edge/pull/343/commits/47460e1008876753ddd3259a9d2d52e0605c6573)

network/discovery/discovery.go
zivkovicmilos: Fixed 👍 

[47460e1](https://github.com/0xPolygon/polygon-edge/pull/343/commits/47460e1008876753ddd3259a9d2d52e0605c6573)

network/server.go
zivkovicmilos: Good catch 💯 

Fixed:

[7cb861b](https://github.com/0xPolygon/polygon-edge/pull/343/commits/7cb861b2ac7d76fc07493c355c4f3b188483a6b3)

network/identity/identity.go
zivkovicmilos: Fixed 👍 

[f763b11](https://github.com/0xPolygon/polygon-edge/pull/343/commits/f763b1182ebfbe0896f36db25fd8be46b5df4e26)

network/event/peer_event.go
zivkovicmilos: I don't think that's wise and it's not scalable. @Kourin1996 Added this map as an easy fix if we need to add additional network events - that way we just update the map, and not add a new switch case. It also makes the code much cleaner

network/server.go
zivkovicmilos: Extracted 👍 

[633131d](https://github.com/0xPolygon/polygon-edge/pull/343/commits/633131dc63ac9f0106a71ccff5d4b8a0ab361fb5)

network/identity/identity.go
zivkovicmilos: It was leftover 🤦 

Fixed:

[61aa2f3](https://github.com/0xPolygon/polygon-edge/pull/343/commits/61aa2f325e24359e6f5b5b4a46d80c4c9823a95c)

network/discovery/discovery.go
zivkovicmilos: It's leftover 🤦 

Removed:

[1a6a81c](https://github.com/0xPolygon/polygon-edge/pull/343/commits/1a6a81c7273ce53ce699e2450bbc26f958e8de6d)

network/discovery/discovery.go
zivkovicmilos: Because it allows us to easily mock the gRPC client interface for the discovery service, and because the base networking server is in charge of keeping track of stream / peer connections. The discovery service needs a reference to an _existing_ stream if it's present, and the only way to get it is from the base networking server.

It's just about shifting the responsibility of connection management - discovery, like identity, are just services that run on top of the base networking server, and they should be kept minimal

network/discovery/discovery.go
zivkovicmilos: Good catch 💯 

This was leftover. There is no possibility for the routing table to **not** be initialized.

Fixed:

[cdda743](https://github.com/0xPolygon/polygon-edge/pull/343/commits/cdda7432b5f1e21c83e592cb72d6d72ccae46229)

network/server.go
zivkovicmilos: We can't exactly move it solely to the discovery package because it is dependent on multiple methods in the networking server, namely dial queue manipulation. To avoid having the discovery service set itself up, and to make sure the base networking server is in control of the setup, it was done in the server package. The services just need to run.

I've moved out the init logic for both the identity and discovery service to a separate file so the `server.go` is much cleaner:

[1cd1083](https://github.com/0xPolygon/polygon-edge/pull/343/commits/1cd1083ae596b178b32cd09a2e39ae90778186ac)

network/connections/connections.go
zivkovicmilos: You're right, there's no need for this to be a separate package, it's leftover from a previous implementation I did 🤦 

Moved:

[ded22dd](https://github.com/0xPolygon/polygon-edge/pull/343/commits/ded22dd6622bc8c1be529b4190f9b28ad5bcf0ce)

network/dial/dial_task.go
zivkovicmilos: Agreed 💯 

Fixed:

[ad08c14](https://github.com/0xPolygon/polygon-edge/pull/343/commits/ad08c145fc150e4f2deffcb33e7147042b6f58c0)

network/event/peer_event.go
dbrajovic: Yes, but this map is used solely for the purpose of fetching a string format (only used in `String()` method). In order to scale with this approach, you would just need to implement another `case` branch. If this map had any other use cases, then we could discuss keeping it in memory (for the entire duration of the program).

network/server.go
zivkovicmilos: I completely agree, this was done in order to not have to pass jumbo data from the networking server to the discovery service when setting up the network event handler.

We will refactor this when we tackle network events 👍

network/event/peer_event.go
zivkovicmilos: @Kourin1996 what do you think?

zivkovicmilos:
@Kourin1996 

I've narrowed down the problem:
* Peer A and Peer B are connected
* Peer B closes the connection to Peer A
* Before Peer A registers that the connection is closed (DisconnectedF callback is called (removePeer), it runs the discovery service to query new peers
* It gets a random peer (from the outdated list) A (this is the discovery service)
* It opens a connection for discovery
* It attempts to save a reference to the stream, but the connection info for Peer A is nil
* Panic 🙂 

The problem is that the discovery service _assumes_ that the connections are already there (because of identity), but that doesn't have to always be the case.

I've resolved this problem and wrote up a unit test that makes sure this sort of thing doesn't happen:

[39a87a9](https://github.com/0xPolygon/polygon-edge/pull/343/commits/39a87a972b0e59cb81bf26db6e4c2b37babccb26)

dbrajovic (approved):
LGTM, the networking code is much more readable. Let's resolve @Kourin1996 's comments and we can merge this.

network/dial/dial_task.go
```diff
@@ -0,0 +1,18 @@
+package dial
+
+import "github.com/libp2p/go-libp2p-core/peer"
+
+type DialTask struct {
+	index int
+
+	// info of the task
+	addr *peer.AddrInfo
```
0xAleksaOpacic: nitpick: var is called addr but it is addrInfo, I would call it addInfo

0xAleksaOpacic (approved):
LGTM! Thank your for all refactor you have done, code looks great.

lazartravica (approved):

network/discovery/discovery.go
```diff
@@ -0,0 +1,424 @@
...
+	// Find peers from the referenced bootnode
+	foundNodes, err := d.findPeersCall(bootnode.ID, true)
+	if err != nil {
+		d.logger.Error("Unable to execute bootnode peer discovery, %w", err)
+
+		return
+	}
+
+	// Save the peers for subsequent dialing
+	d.addPeersToTable(foundNodes)
```
Kourin1996, line 372: I faced error when server tries to connect one of bootnodes, which is not connected to server now

```
2022-03-31T15:40:47.937+0300 [ERROR] polygon.network.discovery: Unable to execute bootnode peer discovery, %w: EXTRA_VALUE_AT_END="unable to create new discovery client connection, peer disconnected before the discovery client was initialized"
```

Kourin1996 (changes requested):

network/server.go
```diff
@@ -237,70 +170,153 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+		bootnodesMap[bootnode.ID] = bootnode
+	}
+
+	// It's fine for the bootnodes field to be unprotected
+	// at this point because it is initialized once (doesn't change),
+	// and used only after this point
+	s.bootnodes = &bootnodesWrapper{
+		bootnodeArr:       bootnodesArr,
+		bootnodesMap:      bootnodesMap,
+		bootnodeConnCount: int64(len(bootnodesArr)),
```
Kourin1996: Why bootnodeConnCount has value as default? This is not the number of connecting bootnode?

Kourin1996 (changes requested):

network/dial/dial_task.go
zivkovicmilos: Agreed 👍 

[adb2138](https://github.com/0xPolygon/polygon-edge/pull/343/commits/adb2138d4318556d22e0900d20e4c3793949f3be)

network/server.go
zivkovicmilos: It's 0 by default when the server is **created**, but at this point where the value is being set, it is the number of successful bootnode connections. Please check the point of calling `setupBootnodes`, and what it does, it will be much clearer 🙏

Kourin1996: @zivkovicmilos So why you increment/decrement this count by `increaseBootnodeConnCount` when connecting/disconnecting?

Kourin1996: As far as I see this function, bootnodeConnCount seems to be number of Bootnodes in genesis.json

network/discovery/discovery.go
zivkovicmilos: You're right, temporary dials **are never added to the peer store**, since they are temporary (only for a few moments).

Fixed the Discovery client method to reflect this 👍 :

[449cd90](https://github.com/0xPolygon/polygon-edge/pull/343/commits/449cd90d57247c3b5f8501094fba2dc632af0e2a)

network/server.go
zivkovicmilos: The initial value for this counter is `0`, when the server is created.
So there are essentially 2 places where we update this counter:
1. When trying to establish the _initial_ bootnode connections (method `setupBootnodes`). We update the counter with the number of _active_ (successful) bootnode connections
2. When a peer connects / disconnects, and they were part of the bootnode set. We update the number of active bootnode connections then ([here](https://github.com/0xPolygon/polygon-edge/blob/449cd90d57247c3b5f8501094fba2dc632af0e2a/network/server_identity.go#L77) and [here](https://github.com/0xPolygon/polygon-edge/blob/449cd90d57247c3b5f8501094fba2dc632af0e2a/network/server.go#L501))

Kourin1996: Even though the server is not connecting to bootnodes, this value is not zero. That's why I'm wondering

zivkovicmilos: I just noticed the problem you mentioned 🤦 

Bootnode conn counts should be updated on connections **only**, and not on this initialization (parsing) 🤦 

Fixed it here 👍 :

[c6de5d7](https://github.com/0xPolygon/polygon-edge/pull/343/commits/c6de5d7e89bcf44ac2f52ffb4a912c82f9640052)

network/server_discovery.go
```diff
@@ -0,0 +1,243 @@
...
+func (s *Server) saveProtocolStream(
+	protocol string,
+	stream *rawGrpc.ClientConn,
+	peerID peer.ID,
+) {
+	s.peersLock.Lock()
+	defer s.peersLock.Unlock()
+
+	connectionInfo := s.peers[peerID]
+	connectionInfo.addProtocolStream(protocol, stream)
```
Kourin1996: There might be nil panic

zivkovicmilos: We don't call it from an unsafe place anyways, but I've added a sanity check for the future:

[99f82f7](https://github.com/0xPolygon/polygon-edge/pull/343/commits/99f82f7bd75bc808d1a37c0dc90c37e19c904283)

Kourin1996 (approved):
Looks good to me. Thank you for refactoring!

PR #345, opened 2022-01-14
Reject eth_sendTransaction jsonrpc call with error

ZeljkoBenovic:
# Description

Fixes EDGE-17

This PR implements error response on jsonrpc call to eth_sendTransaction method.
We're rejecting this call, as we do not support wallet management
Response example:
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32600,"message":"wallet management is not supported, any request call to the eth_sendTransaction method is not supported"}}
```

# Changes include

# Checklist

- [x] I have tested this code


e2e/websocket_test.go
```diff
@@ -117,6 +118,8 @@ func TestWS_Response(t *testing.T) {
 		assert.Equalf(t, 0, preminedAccounts[0].balance.Cmp(foundBalance), "Balances don't match")
 	})
 
+	// skiping test as we disabled eth_sendTransaction jsonrpc call
+	t.SkipNow()
```
zivkovicmilos: Why not put this test skip at the beginning of the test?

jsonrpc/eth_endpoint.go
```diff
@@ -147,18 +148,11 @@ func (e *Eth) SendRawTransaction(input string) (interface{}, error) {
...
 func (e *Eth) SendTransaction(arg *txnArgs) (interface{}, error) {
-	transaction, err := e.d.decodeTxn(arg)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := e.d.store.AddTx(transaction); err != nil {
-		return nil, err
-	}
+	errMsg := "wallet management is not supported, any request call to the eth_sendTransaction method is not supported"
```
zivkovicmilos: You can inline this error message with the return

jsonrpc/types.go
```diff
@@ -210,9 +210,9 @@ func (a argBig) MarshalText() ([]byte, error) {
 	return []byte("0x" + b.Text(16)), nil
 }
 
-func argAddrPtr(a types.Address) *types.Address {
-	return &a
-}
+//func argAddrPtr(a types.Address) *types.Address {
```
zivkovicmilos: Please remove this comment block if it's not needed

zivkovicmilos (approved):
Looks good 👍 

I've left some minor suggestions 💯

0xAleksaOpacic (commented):
Well Done! I have just one suggestion that we can discuss in sync.
@zivkovicmilos @lazartravica I do not think it is good practice to have t.SkipNow() in tests in develop. I also do not think we should delete all of these tests, we should adjust them so they are left with parts of the code where they test something. And if that other part does not make sense to leave we should delete test.

0xAleksaOpacic (approved):

e2e/websocket_test.go
ZeljkoBenovic: The first part of the test is making sure that the websocket endpoint is working properly. We can remove this test entirely but than we won't have any WS tests at all. Should we remove this test entirely ?

jsonrpc/eth_endpoint.go
ZeljkoBenovic: FIxed in commit cbc116c8211dc3ecd4685616b182a8d6db4cb005

jsonrpc/types.go
ZeljkoBenovic: Fixed in commit cbc116c8211dc3ecd4685616b182a8d6db4cb005

lazartravica (changes requested):
@ZeljkoBenovic take a look at Aleksa's comment.

We don't want to:

1. Skip tests.
2. Have commented out tests

We want to only have the tests that bring us value and that's it, remove what doesn't fit in the picture with the disabled JSON-RPC call and that's it.

ZeljkoBenovic:
Removed all skipped / unused tests with commits c6a7fa65f72e068ac822491f277f0d7bcb53fc01 and 1a8b529253b69888d6846c4dd1b394579c00d212

zivkovicmilos (changes requested):
Please revert the test deletions as we discussed 🙏

e2e/websocket_test.go
ZeljkoBenovic: Reworked all test in 47c3b97be431be3734c715bbe10123d00d9b5f59

ZeljkoBenovic:
I've restored all deleted files/tests and fixed all methods to use sendRawTransacion instead of sendTransaction. Local testing now shows no issues.

e2e/framework/testserver.go
```diff
@@ -417,11 +418,29 @@ func (t *TestServer) SendRawTx(
 	signer := crypto.NewEIP155Signer(100)
 	client := t.JSONRPC()
 
+	// if there is no from addr passed in we can find it in test server configuration
+	var blankAddr types.Address
+	if tx.From == blankAddr {
+		tx.From = t.Config.PremineAccts[0].Addr
+	}
```
zivkovicmilos: Why do we need this?

e2e/framework/testserver.go
```diff
@@ -417,11 +418,29 @@ func (t *TestServer) SendRawTx(
...
+		tx.GasPrice = big.NewInt(DefaultGasPrice)
+	}
+
+	if tx.Gas == 0 {
+		tx.Gas = DefaultGasLimit
+	}
+
+	if tx.Value == big.NewInt(0) || tx.Value == nil {
+		tx.Value = big.NewInt(0)
+	}
```
zivkovicmilos: Why the default values?

I don't think we should override values like this in a common func such as `SendRawTx`

e2e/framework/testserver.go
```diff
@@ -490,12 +509,16 @@ func (t *TestServer) WaitForReady(ctx context.Context) error {
 	return err
 }
 
-func (t *TestServer) TxnTo(ctx context.Context, address web3.Address, method string) *web3.Receipt {
+//nolint:lll
```
zivkovicmilos: Remove the `nolint`, just break up the method declaration on multiple lines so it's readable

zivkovicmilos:
Please check out the changes I've made in https://github.com/0xPolygon/polygon-edge/pull/370.

This PR will be merged out soon, so you can use the fixed up common TestServer methods in this PR

e2e/framework/testserver.go
ZeljkoBenovic: Some tests were failing with not enough funds error message. 
This makes sure that that the sender always has enough funds, even if calling function doesn't specify From field. 

Should I investigate exactly which tests are failing when this check is omitted ?

e2e/framework/testserver.go
ZeljkoBenovic: tx.Value needs to be set, as SignTx function is panicking when it tries to do tx.Copy and the tx.Value is not set. [here](https://github.com/0xPolygon/polygon-edge/blob/48ec5c46a89b5e57fc538eac9d8e57ebdb350286/types/transaction.go#L53)
 
As for tx.Gas and tx.GasPrice , these same checks are done in SendTxn function ( which we're using right now ) , and I've just implemented them here. Should we remove this ?

jsonrpc/eth_endpoint.go
```diff
@@ -210,18 +210,10 @@ func (e *Eth) SendRawTransaction(input string) (interface{}, error) {
...
-		return nil, err
-	}
-
-	if err := e.store.AddTx(transaction); err != nil {
-		return nil, err
-	}
-
-	return transaction.Hash.String(), nil
+	//nolint: lll
+	return nil, fmt.Errorf("request calls to eth_sendTransaction method are not supported, use eth_sendRawTransaction insead")
```
zivkovicmilos: This should be broken up into multiple lines and this error extracted into a standardized error with `errors.New`

e2e/framework/testserver.go
zivkovicmilos: What tests fail when you omit this?

zivkovicmilos (approved):
Looks good, please resolve the leftover convos and we're good to go 🚀

e2e/framework/testserver.go
zivkovicmilos: Resolving because it's outdated

e2e/framework/testserver.go
ZeljkoBenovic: I can't reproduce this fail anymore. Removed in 6138b95b4526bbbefa2f06a54581e344e2ebc9cd

e2e/framework/testserver.go
ZeljkoBenovic: Fixed in 6138b95b4526bbbefa2f06a54581e344e2ebc9cd

Kourin1996 (approved):

lazartravica (approved):
Confirmed it works.

Thanks!

PR #348, opened 2022-01-15
Generate receipt for failed transactions

0xAleksaOpacic:
# Description

This PR generates failed receipt if the transaction fails once it goes inside the pool.
This can happen because we do not have a default block gas limit, we dynamically calculate it when we create the block.
When we are validating a transaction before we put it in the pool we check for the gas limit but with the previous block.
In some specific edge-case, it can happen that when we are putting transaction in the pool to be valid (because the previous block gas limit is big enough) but when we are putting transaction in the block to be invalid.

Steps to test it:

Start IBFT network
Send transaction with gas limit 1000000000001
After transaction goes into block it will return ERROR receipt

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -643,7 +643,7 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 
 type transitionInterface interface {
 	Write(txn *types.Transaction) error
-	WriteFailedReceipt(txn *types.Transaction)
+	WriteFailedReceipt(txn *types.Transaction) error
 }
 
 // writeTransactions writes transactions from the txpool to the transition object
@@ -663,7 +663,14 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		}
 
 		if tx.ExceedsBlockGasLimit(gasLimit) {
-			transition.WriteFailedReceipt(tx)
+			if err := transition.WriteFailedReceipt(tx); err != nil {
+				failedTxCount++
+
+				i.txpool.Drop(tx)
+
+				continue
+			}
+
 			failedTxCount++
 
 			transactions = append(transactions, tx)
@@ -693,6 +700,7 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		transactions = append(transactions, tx)
 	}
 
+	//nolint:lll
 	i.logger.Info("executed txns", "failed ", failedTxCount, "successful", successTxCount, "remaining in pool", i.txpool.Length())
 
 	return transactions
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -651,6 +651,9 @@ type transitionInterface interface {
 func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface) []*types.Transaction {
 	var transactions []*types.Transaction
 
+	successTxCount := 0
+	failedTxCount := 0
+
 	i.txpool.Prepare()
 
 	for {
@@ -661,6 +664,8 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 
 		if tx.ExceedsBlockGasLimit(gasLimit) {
 			transition.WriteFailedReceipt(tx)
+			failedTxCount++
+
 			transactions = append(transactions, tx)
 			i.txpool.Drop(tx)
 
@@ -682,10 +687,13 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		// no errors, pop the tx from the pool
 		i.txpool.Pop(tx)
 
+		successTxCount++
+
 		transactions = append(transactions, tx)
 	}
 
-	i.logger.Info("picked out txns from pool", "num", len(transactions), "remaining", i.txpool.Length())
+	//nolint:lll
+	i.logger.Info("picked out txns from pool", "numFailedTx ", failedTxCount, "numSuccessTx ", successTxCount, "remaining", i.txpool.Length())
 
 	return transactions
 }
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -643,12 +643,13 @@ func (i *Ibft) buildBlock(snap *Snapshot, parent *types.Header) (*types.Block, e
 
 type transitionInterface interface {
 	Write(txn *types.Transaction) error
+	WriteFailedReceipt(txn *types.Transaction)
 }
 
 // writeTransactions writes transactions from the txpool to the transition object
 // and returns transactions that were included in the transition (new block)
 func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface) []*types.Transaction {
-	var successful []*types.Transaction
+	var transactions []*types.Transaction
 
 	i.txpool.Prepare()
 
@@ -659,6 +660,8 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		}
 
 		if tx.ExceedsBlockGasLimit(gasLimit) {
+			transition.WriteFailedReceipt(tx)
+			transactions = append(transactions, tx)
 			i.txpool.Drop(tx)
 
 			continue
@@ -679,12 +682,12 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		// no errors, pop the tx from the pool
 		i.txpool.Pop(tx)
 
-		successful = append(successful, tx)
+		transactions = append(transactions, tx)
 	}
 
-	i.logger.Info("picked out txns from pool", "num", len(successful), "remaining", i.txpool.Length())
+	i.logger.Info("picked out txns from pool", "num", len(transactions), "remaining", i.txpool.Length())
 
-	return successful
+	return transactions
 }
 
 // runAcceptState runs the Accept state loop
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -678,6 +678,7 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 			} else if appErr, ok := err.(*state.TransitionApplicationError); ok && appErr.IsRecoverable { // nolint:errorlint
 				i.txpool.Demote(tx)
 			} else {
+				failedTxCount++
 				i.txpool.Drop(tx)
 			}
 
@@ -692,8 +693,7 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		transactions = append(transactions, tx)
 	}
 
-	//nolint:lll
-	i.logger.Info("picked out txns from pool", "numFailedTx ", failedTxCount, "numSuccessTx ", successTxCount, "remaining", i.txpool.Length())
+	i.logger.Info("executed txns", "failed ", failedTxCount, "successful", successTxCount, "remaining in pool", i.txpool.Length())
 
 	return transactions
 }
--- a/consensus/ibft/ibft_test.go
+++ b/consensus/ibft/ibft_test.go
@@ -556,7 +556,7 @@ func TestWriteTransactions(t *testing.T) {
 			[]int{1},
 			5,
 			3,
-			2,
+			3,
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


consensus/ibft/ibft_test.go
```diff
@@ -757,6 +761,10 @@ func (t *mockTransition) Write(txn *types.Transaction) error {
 	return nil
 }
 
+func (t *mockTransition) WriteFa(txn *types.Transaction) {
```
zivkovicmilos: Please rename this method to something more descriptive

consensus/ibft/ibft.go
```diff
@@ -679,12 +682,12 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		// no errors, pop the tx from the pool
 		i.txpool.Pop(tx)
 
-		successful = append(successful, tx)
+		transactions = append(transactions, tx)
 	}
 
-	i.logger.Info("picked out txns from pool", "num", len(successful), "remaining", i.txpool.Length())
+	i.logger.Info("picked out txns from pool", "num", len(transactions), "remaining", i.txpool.Length())
```
zivkovicmilos: Please refactor this log to be clear on how many transactions failed, and how many succeeded

state/executor.go
```diff
@@ -99,7 +99,14 @@ func (e *Executor) ProcessBlock(
 	}
 
 	txn.block = block
+
 	for _, t := range block.Transactions {
+		if t.ExceedsBlockGasLimit(block.Header.GasLimit) {
```
zivkovicmilos, line 104: This is very dangerous

You are omitting a state write for transactions that you have a receipt for (you're writing the receipt, but not writing the transaction itself to the transactions trie)

Please note that Ethereum writes failed transactions to state, along with their receipts

cc @lazartravica @Kourin1996

zivkovicmilos (changes requested):
I've left some comments that need to be addressed.

I've also not seen any test that validates this functionality, please add them since they are in the scope of this PR 🙏

consensus/ibft/ibft_test.go
0xAleksaOpacic: This was leftover should have deleted it.
fixed 79a49accce1b9344e439ec8fe560eccfae0392c1

consensus/ibft/ibft.go
0xAleksaOpacic: Great catch! Fixed 79a49accce1b9344e439ec8fe560eccfae0392c1

state/executor.go
0xAleksaOpacic: I am not really sure what this means, so if we can sync on this

consensus/ibft/ibft.go
```diff
@@ -679,12 +684,13 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		// no errors, pop the tx from the pool
 		i.txpool.Pop(tx)
 
-		successful = append(successful, tx)
+		successTxCount++
+		transactions = append(transactions, tx)
 	}
 
-	i.logger.Info("picked out txns from pool", "num", len(successful), "remaining", i.txpool.Length())
+	i.logger.Info("picked out txns from pool", "numFailedTx=", failedTxCount, "numSuccessTx=", successTxCount, "remaining", i.txpool.Length())
```
Kourin1996: You don't need to add `=`

0xAleksaOpacic: fixed 8b76af4b0bc360d8713ec51fd4df7e779114d994

state/executor.go
zivkovicmilos: You're writing a failed receipt to the receipts trie, but not the transaction to the transaction trie.

You can never look up these failed transactions, even though you can look up their receipts - which is not the case with Ethereum, where you can look up both receipts and transactions (failed)

0xAleksaOpacic: @lazartravica If you can, please check this. My explanation is that we do not change how we add receipt and transactions in block and state. This PR will add failed tx and receipt to tx[] receipt[] before we build block. Everything what happens after will work as it worked before. We would create block with tx[] and receipt[], Then we are going to insert it with insertBlock()...

i have checked with eth_getTransactionByHash and eth_getTransactionReceipt. I can retrieve both tx and receipt for failed transaction which was added to a block

consensus/ibft/ibft_test.go
```diff
@@ -464,13 +464,15 @@ func TestTransition_RoundChangeState_MaxRound(t *testing.T) {
...
-		expectedIncludedTxnsCount int
+		description                 string
+		txns                        []*types.Transaction
+		recoverableTxnsIndexes      []int
+		unrecoverableTxnsIndexes    []int
+		gasLimitReachedTxnIndex     int
+		expectedTxPoolLength        int
+		expectedIncludedTxnsCount   int
+		expectedFailReceiptsWritten int
+		successReceiptsWritten      int
```
zivkovicmilos: The test case should not contain a bagillion params, it should ideally have:

```go
type testCase struct {
  name string
  params struct {
     // ...
  }
}
```

state/executor.go
zivkovicmilos: My mistake 🙏 

I've falsely assumed this is the final commit step, when in fact you write both failed and successful transactions when writing the block's body:

https://github.com/0xPolygon/polygon-edge/blob/d0dc886348bd134e35f03a2fe20930cae08c64e4/blockchain/blockchain.go#L637-L655

zivkovicmilos (approved):
Looks good  💯 

consensus/ibft/ibft.go
```diff
@@ -659,6 +663,10 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
 		}
 
 		if tx.ExceedsBlockGasLimit(gasLimit) {
+			transition.WriteFailedReceipt(tx)
+			failedTxCount++
```
brkomir, line 674: Shouldn't this counter `failedTxCount`" alse be increased here `ibft.go` line 681?
`} else {
	i.txpool.Drop(tx)
}`

consensus/ibft/ibft.go
```diff
@@ -679,12 +687,15 @@ func (i *Ibft) writeTransactions(gasLimit uint64, transition transitionInterface
...
 
-		successful = append(successful, tx)
+		successTxCount++
+
+		transactions = append(transactions, tx)
 	}
 
-	i.logger.Info("picked out txns from pool", "num", len(successful), "remaining", i.txpool.Length())
+	//nolint:lll
+	i.logger.Info("picked out txns from pool", "numFailedTx ", failedTxCount, "numSuccessTx ", successTxCount, "remaining", i.txpool.Length())
```
brkomir: Please consider renaming this log since it sounds a bit off, maybe something like this would make more sense:
`i.logger.Info("executed txns", "failed ", failedTxCount, "successful", successTxCount, "remaining in pool", i.txpool.Length())`

consensus/ibft/ibft_test.go
```diff
@@ -752,7 +773,7 @@ func (t *mockTransition) Write(txn *types.Transaction) error {
 		}
 	}
 
-	t.transactionsWritten = append(t.transactionsWritten, txn)
+	t.successReceiptsWritten = append(t.successReceiptsWritten, txn)
```
brkomir, line 786: Why is this field `successReceiptsWritten` even relevant? We are testing the `writeTransactions()` method which doesn't contain any logic for writing receipts for successful transactions.

state/executor.go
```diff
@@ -202,6 +209,23 @@ func (t *Transition) Receipts() []*types.Receipt {
...
 var emptyFrom = types.Address{}
 
+func (t *Transition) WriteFailedReceipt(txn *types.Transaction) {
+	logs := t.state.Logs()
+
+	receipt := &types.Receipt{
+		CumulativeGasUsed: t.totalGas,
+		TxHash:            txn.Hash,
+	}
+	receipt.Logs = logs
```
brkomir: Lines 213 and 219 can be a single line inside the struct initialization block.

brkomir (changes requested):

consensus/ibft/ibft.go
0xAleksaOpacic: Yes, when I think about it it should. Great cactch

consensus/ibft/ibft_test.go
0xAleksaOpacic: You are right, we test everything with only expectedFailReceiptsWritten.

consensus/ibft/ibft.go
0xAleksaOpacic: fixed 1f444282222e474d3d38b871a6d5399bc36e4957

consensus/ibft/ibft_test.go
0xAleksaOpacic: fixed 1f444282222e474d3d38b871a6d5399bc36e4957

consensus/ibft/ibft.go
0xAleksaOpacic: fixed 1f444282222e474d3d38b871a6d5399bc36e4957

consensus/ibft/ibft_test.go
0xAleksaOpacic: fixed 1f444282222e474d3d38b871a6d5399bc36e4957

state/executor.go
0xAleksaOpacic: fixed 1f444282222e474d3d38b871a6d5399bc36e4957

state/executor.go
```diff
@@ -202,6 +209,22 @@ func (t *Transition) Receipts() []*types.Receipt {
...
+		Logs:              t.state.Logs(),
+	}
+
+	receipt.LogsBloom = types.CreateBloom([]*types.Receipt{receipt})
+	receipt.SetStatus(types.ReceiptFailed)
+	t.receipts = append(t.receipts, receipt)
+
+	if msg := txn.Copy(); msg.To == nil {
+		receipt.ContractAddress = crypto.CreateAddress(msg.From, txn.Nonce)
+	}
```
Kourin1996: `from` in  failed receipt was `0x0000000000000000000000000000000000000000` when I tried locally. Is it the result you expected? And I'm afraid that I didn't understand why you needed to copy transaction here.

0xAleksaOpacic: great catch!
fixed 494f1c81bc95ad91341d66f2550c456445ffd5cd

0xAleksaOpacic: Example:
`{"jsonrpc":"2.0","id":1,"result":{"root":"0x0000000000000000000000000000000000000000000000000000000000000000","cumulativeGasUsed":"0x0","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","logs":[],"status":"0x0","transactionHash":"0x45289de9bac0b44062dbd46936b14402095e3995b2800fd06fd72cd6624bb684","transactionIndex":"0x0","blockHash":"0x718f3fd1a2e4f7498e187f69fab013e1398dbfd0cdefabbbd837e12e34b7d27e","blockNumber":"0x4db","gasUsed":"0x0","contractAddress":"0x0000000000000000000000000000000000000000","from":"0x30ea4435167Ee91f9f874b5a894F3282A956C3FF","to":"0x5383Cb489FaCa92365Bb6f9f1FB40bD032E6365d"}}`

Kourin1996 (approved):
Looks good to me 👏

state/executor.go
```diff
@@ -202,6 +211,35 @@ func (t *Transition) Receipts() []*types.Receipt {
...
+	signer := crypto.NewSigner(t.config, uint64(t.r.config.ChainID))
+
+	var err error
+	if txn.From == emptyFrom {
+		// Decrypt the from address
+		txn.From, err = signer.Sender(txn)
+		if err != nil {
+			return NewTransitionApplicationError(err, false)
+		}
+	}
```
dbrajovic: You can use short-hand if statement for error checking.

dbrajovic (approved):
Minor nitpick suggestion, otherwise LGTM

state/executor.go
0xAleksaOpacic: Fixed 865904a27847be72f1012fdcc21a68716d37081b

consensus/ibft/ibft_test.go
brkomir: I don't see this has changed, you still have this field `successReceiptsWritten`, just leave it as it was before

brkomir (approved):

PR #349, opened 2022-01-15
Discard transaction with invalid gas limit

0xAleksaOpacic:
# Description

If a transaction has gasLimit bigger than the latest blockGasLimit it is considered invalid.
There is no way to know for sure what will be gas limit of the block where the transaction will be placed, so we have another check when we are creating a block.

Different error cases

Transaction has gasLimit > latestBlockGasLimit (it will not go into pool and it will be rejected as invalid)
Transaction has gasLimit < latestBlockGasLimit but when we create a block gasLimit > newBlockGasLimit (we will generate failed transaction receipt)
Testing:

Try to add a transaction with gasLimit 10000000001.

# Changes include

# Breaking changes

# Checklist

- [x] I have tested this code

# Additional comments

### DIFF
--- a/e2e/txpool_test.go
+++ b/e2e/txpool_test.go
@@ -408,44 +408,6 @@ func TestTxPool_StressAddition(t *testing.T) {
 	}
 }
 
-func TestInvalidTransactionRecover(t *testing.T) {
-	// Test scenario :
-	// Send a transaction with gasLimit > block gas limit.
-	//		-> The transaction should not be applied, and the nonce should not be incremented.
-	senderKey, senderAddress := tests.GenerateKeyAndAddr(t)
-	_, receiverAddress := tests.GenerateKeyAndAddr(t)
-
-	server := framework.NewTestServers(t, 1, func(config *framework.TestServerConfig) {
-		config.SetConsensus(framework.ConsensusDev)
-		config.SetSeal(true)
-		config.Premine(senderAddress, framework.EthToWei(100))
-	})[0]
-	client := server.JSONRPC()
-
-	tx, err := signer.SignTx(&types.Transaction{
-		Nonce:    0,
-		GasPrice: big.NewInt(10000),
-		Gas:      5000000000,
-		To:       &receiverAddress,
-		Value:    oneEth,
-		V:        big.NewInt(1),
-		From:     senderAddress,
-	}, senderKey)
-	assert.NoError(t, err, "failed to sign transaction")
-
-	// send tx
-	_, err = server.JSONRPC().Eth().SendRawTransaction(tx.MarshalRLP())
-	assert.NoError(t, err)
-
-	balance, err := client.Eth().GetBalance(web3.Address(receiverAddress), web3.Latest)
-	assert.NoError(t, err, "failed to retrieve receiver account balance")
-	assert.Equal(t, framework.EthToWei(0).String(), balance.String())
-
-	nextNonce, err := client.Eth().GetNonce(web3.Address(tx.From), web3.Latest)
-	assert.NoError(t, err)
-	assert.Equal(t, uint64(0), nextNonce)
-}
-
 func TestTxPool_RecoverableError(t *testing.T) {
 	// Test scenario :
 	//
--- a/txpool/mock_test.go
+++ b/txpool/mock_test.go
@@ -10,12 +10,25 @@ import (
 /* MOCK */
 
 type defaultMockStore struct {
+	DefaultHeader *types.Header
+}
+
+func NewDefaultMockStore(header *types.Header) defaultMockStore {
+	var defaultHeader = &types.Header{
+		GasLimit: 4712388,
+	}
+
+	if header != nil {
+		defaultHeader = header
+	}
+
+	return defaultMockStore{
+		defaultHeader,
+	}
 }
 
 func (m defaultMockStore) Header() *types.Header {
-	return &types.Header{
-		GasLimit: 4712388,
-	}
+	return m.DefaultHeader
 }
 
 func (m defaultMockStore) GetNonce(types.Hash, types.Address) uint64 {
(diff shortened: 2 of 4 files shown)


txpool/mock_test.go
```diff
@@ -13,7 +13,9 @@ type defaultMockStore struct {
 }
 
 func (m defaultMockStore) Header() *types.Header {
-	return &types.Header{}
+	return &types.Header{
+		GasLimit: 4712388,
```
0xAleksaOpacic, line 18: @zivkovicmilos COMMENT FROM PREVIOUS PR THAT WE CLOSED
Is it possible to add a method to the defaultMockStore for setting the gas limit of the latest header?
Right now only 1 test uses this value I think, but in the future more unit tests may rely on this GasLimit value to be configurable

txpool/txpool.go
```diff
@@ -26,6 +26,7 @@ const (
 // errors
 var (
 	ErrIntrinsicGas        = errors.New("intrinsic gas too low")
+	ErrBlockLimitExceeded  = errors.New("transaction's gas limit exceeds block gas limit")
```
0xAleksaOpacic: Please make sure the error corresponds to the same one GETH uses:
https://github.com/ethereum/go-ethereum/blob/f32feeb260054d7c90c1efd57dfec6b717717e18/core/error.go#L60

This is required since many web3 frameworks hardcode these expected values in source code

e2e/txpool_test.go
```diff
@@ -408,44 +408,6 @@ func TestTxPool_StressAddition(t *testing.T) {
 	}
 }
 
-func TestInvalidTransactionRecover(t *testing.T) {
```
0xAleksaOpacic, line 411: Please link the unit test(s) that cover this functionality in the new TxPool implementation 🙏

txpool/mock_test.go
0xAleksaOpacic: Not directly, because defaultMockStore is mapped to store object, and we can call only mock methods from that interface. There are a few ways we can bypass this. One idea comes to my mind.
1. create a constructor for initialization defaultMockStore object. Which can receive header as an attribute
2. when initializing the pool, we can send header data with what we want to initialize

But this will complicate pool initialization. Will research for better solution

e2e/txpool_test.go
0xAleksaOpacic: This PR will discard every transaction if they are sent with an invalid gas limit, no transaction will pass validationTx with an invalid gas limit. 
There is no way to write unit/e2e test which will test edge-case when transaction is valid when arrives in txPool (latest block is valid) but new block is invalid (while building)

txpool/txpool.go
0xAleksaOpacic: Fixed c74ebfd5ba1cd4b095ad14db50d864d7101d459f

e2e/txpool_test.go
zivkovicmilos: Why not test adding the transaction?

You can have a few test cases for valid gas limits, and some for invalid

txpool/mock_test.go
zivkovicmilos: Can we go with solution 1?

I don't feel good about hardcoding these values, even if it's a mock store. Someone will need to untangle it in some other PR if we leave it like this

zivkovicmilos (approved):
Approving again 👍 

I've left some comments for discussion, otherwise it looks good 💯

brkomir (approved):
LGTM

txpool/mock_test.go
0xAleksaOpacic: This is one possible solution. It is a bit redundant with sending nil value where we want a default value. But go does not like optional parameters. 55eeead6ab4c2431394e0611e1e6969516e6f74f

e2e/txpool_test.go
0xAleksaOpacic: In the txpool_test.go we have a unit test that covers adding error tx (TestAddTxErrors). It covers all validation rules and because I have added a new validation rule I had to add a new test case for this rule. 

https://github.com/0xPolygon/polygon-edge/pull/349/files#diff-c6d10affe0540289b09ab01da147d827eb2c3aa0a899b5e5d5c43810821422cbR113

In my opinion, because this is mock with defined values this one test covers the validation check for ErrBlockLimitExceeded. Adding multiple test cases for valid and invalid values would just be redundant check if one value is bigger than the other.

And every other test which adds a new transaction inside this package will show that if we have a valid gas limit (tx.gasLimit < latestBlockGasLimit) will be considered valid

Kourin1996 (approved):
Looks good

dbrajovic (approved):

txpool/txpool_test.go
```diff
@@ -69,11 +70,11 @@ func newTx(addr types.Address, nonce, slots uint64) *types.Transaction {
 }
 
 // returns a new txpool with default test config
-func newTestPool() (*TxPool, error) {
+func newTestPool(header *types.Header) (*TxPool, error) {
```
brkomir, line 73: Why this parameter? Not used anywhere, you pass "nil" for every func call...

lazartravica: @Aleksao998 please address this comment

0xAleksaOpacic: @zivkovicmilos requested a change to make that we can adjust header when are creating the pool in our tests and not to have default values. I agree that this is not a pretty solution at all. But was the only one I have found. I will tag both of you in this discussion.

txpool/mock_test.go
0xAleksaOpacic: @brkomir @lazartravica Here is the discussion why i had to add header parameter

txpool/txpool_test.go
zivkovicmilos: What about adding a wrapper mock struct that embeds TxPool and has a method `SetHeader` or something like that?

0xAleksaOpacic: I am not really sure what do you think
Store interface does not have method setHeader or something like that. So what we pass to the constructor when we are creating the store. These will be value when we get them.

PR #352, opened 2022-01-17
Fixes broken server if snapshot file is faulty

ZeljkoBenovic:
# Description

If snapshots file is empty/corrupted for some reason, the server can't run, as return statements break code execution.

This PR provides better log output with problems with reading shapshots/metadata files.
Also, if there are any problems with these files, they will be deleted so that we can resync with the rest of the nodes.

# Changes include

# Checklist

- [x] I have tested this code


consensus/ibft/snapshot.go
```diff
@@ -21,7 +21,7 @@ func (i *Ibft) setupSnapshot() error {
 
 	// Read from storage
 	if i.config.Path != "" {
-		if err := i.store.loadFromPath(i.config.Path); err != nil {
+		if err := i.store.loadFromPath(i); err != nil {
```
zivkovicmilos: Please change the method to accept the path and the logger, no need to accept the entire `Ibft` reference

consensus/ibft/snapshot.go
```diff
@@ -366,21 +366,29 @@ func newSnapshotStore() *snapshotStore {
...
+func (s *snapshotStore) loadFromPath(i *Ibft) error {
 	// Load metadata
 	var meta *snapshotMetadata
-	if err := readDataStore(filepath.Join(path, "metadata"), &meta); err != nil {
-		return err
-	}
+	if err := readDataStore(filepath.Join(i.config.Path, "metadata"), &meta); err != nil {
+		// if we can't read file metadata file delete it and sync again
+		// and logout that we've had some problems
+		i.logger.Error("could not read metadata file","err",err.Error())
```
zivkovicmilos: Please change the log output from `metadata` to `snapshot store`, in the entire method

consensus/ibft/snapshot.go
```diff
@@ -366,21 +366,29 @@ func newSnapshotStore() *snapshotStore {
...
+		i.logger.Error("removing faulty metadata file...")
```
zivkovicmilos: This log should happen after you run `os.Remove` and it should go `Removed invalid snapshot store file`

consensus/ibft/snapshot.go
zivkovicmilos: Is there any chance for us to see the actual error and compare it before wiping out the snapshot store file?

zivkovicmilos (changes requested):
Please resolve the linting errors and address pending comments 🙏 

Otherwise it's good to go 👍

consensus/ibft/snapshot.go
ZeljkoBenovic: Fixed with commit b0bf7cd004339996424fc7301b460a33873b367f

consensus/ibft/snapshot.go
ZeljkoBenovic: FIxed in commit b0bf7cd004339996424fc7301b460a33873b367f
The actual error will be outputted to console logs, something similar to this:
```
 2022-01-17T14:27:09.155Z [ERROR] polygon.consensus.ibft: Could not read snapshot data: err="unexpected end of JSON input"
```

consensus/ibft/snapshot.go
ZeljkoBenovic: Fixed with commit b0bf7cd004339996424fc7301b460a33873b367f

consensus/ibft/snapshot.go
zivkovicmilos: Got it, makes sense.
The file is empty so there's nothing to unmarshal for the json module

consensus/ibft/snapshot.go
zivkovicmilos: Please change out the `loadFromPath` method not to accept a spread, but instead a single logger instance

ZeljkoBenovic: Fixed with commit 576cbe70a70e335a395994055be9715d8c609c27

Kourin1996 (approved):
Looks good to me

zivkovicmilos (approved):
Lookin' good 🚀

PR #367, opened 2022-01-18
Add real TPS to the loadbot output

ZeljkoBenovic:
# Description

This PR adds the real number of transactions per second to the loadbot output, by dividing total number of transactions by total time in seconds it took to process all transactions.

# Changes include

# Checklist

- [x] I have tested this code


mrwillis:
Just chiming in here @ZeljkoBenovic and more of a discussion. I think one should consider normalized TPS by considering the gas usage. Maybe another metric. Naturally pure ETH transfers would produce the highest TPS, but I think the loadbot has some transaction modes which take higher gas, and would produce lower TPS for the same gas limit.

brkomir (approved):

ZeljkoBenovic:
I totally agree with @mrwillis . This PR introduces output which shows the estimated TPS in regards to regular ETH transfers only. It doesn't take into consideration gas and stuff like that. This will just show estimated tps by dividing total number of transactions with total number of seconds it took to complete the run. We should definitely introduce a more precise estimation of TPS in the future.

command/loadbot/loadbot_command.go
```diff
@@ -372,9 +374,14 @@ type LoadbotResult struct {
 	TurnAroundData    TxnTurnAroundData    `json:"turnAroundData"`
 	BlockData         TxnBlockData         `json:"blockData"`
 	DetailedErrorData TxnDetailedErrorData `json:"detailedErrorData,omitempty"`
+	RealTPS           uint64               `json:"realTps"`
 }
 
 func (lr *LoadbotResult) extractExecutionData(metrics *Metrics) {
+	// calculate real transactions per second value
+	// by deviding total transactions by total time in seconds
+	//nolint:lll
```
zivkovicmilos: Please remove the nolint clause and break up the next line into several lines for readability, you can do it in Go

command/loadbot/loadbot_command.go
```diff
@@ -433,6 +440,11 @@ func (lr *LoadbotResult) Output() string {
 		fmt.Sprintf("Transactions failed|%d", lr.CountData.Failed),
 	}))
 
+	buffer.WriteString("\n\n[REAL TPS]\n")
```
zivkovicmilos: Please rename this section to something like `APPROXIMATE TPS`

zivkovicmilos (approved):
Looks good, please see the conversations 💯

command/loadbot/loadbot_command.go
```diff
@@ -433,6 +440,11 @@ func (lr *LoadbotResult) Output() string {
...
+	buffer.WriteString(helper.FormatKV([]string{
+		fmt.Sprintf("Real number of transactions per second|%d", lr.RealTPS),
```
zivkovicmilos: Please add this TPS value to the json output as well 👍

Kourin1996 (approved):
Looks good to me 👌

lazartravica (approved):

command/loadbot/loadbot_command.go
ZeljkoBenovic: Renamed f9886aed1bef4a43315d3b30f970cda6e1c9e8c2

command/loadbot/loadbot_command.go
ZeljkoBenovic: Added json output f9886aed1bef4a43315d3b30f970cda6e1c9e8c2

command/loadbot/loadbot_command.go
ZeljkoBenovic: Removed linter clause and broke that line of code to multiple lines f9886aed1bef4a43315d3b30f970cda6e1c9e8c2

PR #368, opened 2022-01-19
Add cors middleware support in JSON-RPC server

0xpanoramix:
# Description

This PR adds support for CORS by setting the headers's values based on the already supported server's config file.
Fixes EDGE-64

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments

For now, the only header you can set is `Access-Control-Allow-Origin`.


jsonrpc/jsonrpc.go
```diff
@@ -91,6 +96,24 @@ func (j *JSONRPC) setupHTTP() error {
 	return nil
 }
 
+// The middlewareFactory builds a middleware which enables CORS using the provided config.
+func middlewareFactory(config *Config) func(http.Handler) http.Handler {
+	middleware := func(next http.Handler) http.Handler {
```
zivkovicmilos: You can inline this return

0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-edge/pull/368/commits/91a80b1f9db44a76c758f6c7b04c9a5bc9c2c793

command/helper/config.go
```diff
@@ -56,6 +57,11 @@ type TxPool struct {
 	MaxSlots   uint64 `json:"max_slots"`
 }
 
+// Headers defines the HTTP response headers required to enable CORS.
+type Headers struct {
+	AccessControlAllowOrigin []string `json:"access_control_allow_origin"`
```
Kourin1996: `AccessControlAllowOrigins` would be more suitable?

Kourin1996 (commented):
Looks good to me, I've left a nitpick comment

0xpanoramix:

command/helper/config.go
0xpanoramix: Nice catch :100:, fixed in https://github.com/0xPolygon/polygon-edge/pull/368/commits/381d5af44c6de770fdebf50a1d7e90b2af538b53

Kourin1996 (changes requested):
Looks good to me. But I have realized a few things:
(1) Why don't you add flag to set Access-Control-Allow-Origin?
(2) Allow CORS for all domain here https://github.com/0xPolygon/polygon-edge/blob/5698b85ec9216214a5914eeccb9561fd44dc60ae/jsonrpc/jsonrpc.go#L209. You need to remove this.

zivkovicmilos (approved):
Please resolve the convo Yoshiki left and you're good to go 🙏 

Also, please make sure you resolve any merge conflicts 👍

0xAleksaOpacic (approved):
The code looks good. Have we tested this in the browser? Converting curl to the HTTP request.

Kourin1996 (approved):
Looks good to me. I tested with local React App and it works. Please resolve conflicts

PR #373, opened 2022-01-20
Remove traces of waitUntilDone

zivkovicmilos:
# Description

Fixes EDGE-28

This PR removes invalid test waits that rely on `waitUntilDone()`

# Changes include

# Checklist

- [x] I have tested this code


txpool/queue.go
```diff
@@ -47,7 +47,10 @@ func (q *accountQueue) unlock() {
 
 // prune removes all transactions from the queue
 // with nonce lower than given.
-func (q *accountQueue) prune(nonce uint64) (pruned []*types.Transaction) {
+func (q *accountQueue) prune(nonce uint64) (
+	pruned []*types.Transaction,
+	prunedHashes []types.Hash,
+) {
```
dbrajovic, line 53: Maybe have the hash collecting as a helper method defined elsewhere ? (just to keep code minimal)

txpool/txpool_test.go
```diff
@@ -70,17 +65,30 @@ func newTx(addr types.Address, nonce, slots uint64) *types.Transaction {
 }
 
 // returns a new txpool with default test config
-func newTestPool(header *types.Header) (*TxPool, error) {
+func newTestPool(mockStore ...store) (*TxPool, error) {
+	return newTestPoolWithSlots(defaultMaxSlots, mockStore...)
+}
```
dbrajovic, line 70: Why the variable length `store` parameter ?

txpool/txpool_test.go
```diff
@@ -1174,341 +1182,401 @@ func TestDemote(t *testing.T) {
...
 }
 
-func TestAddTx100(t *testing.T) {
-	t.Run("send 100 transactions", func(t *testing.T) {
-		pool, err := newTestPool(nil)
-		assert.NoError(t, err)
-		pool.SetSigner(&mockSigner{})
-		pool.EnableDev()
+// waitForGaugeMatch is a helper function for waiting on the gauge to reach a certain slot size
+func waitForGaugeMatch(expectedSlots uint64, pool *TxPool) error {
```
dbrajovic: Let's just double check if there is a need for this method. Slots should not be waited upon.

dbrajovic (changes requested):
LGTM, except for some minor details

txpool/queue.go
zivkovicmilos: Why would we do the additional hash collection logic apart from this main loop?

txpool/txpool_test.go
zivkovicmilos: Because we introduced the ability to specify custom mock stores that the TxPool will use in an earlier PR.

This spread trick just cleans up the interface, so not every method that calls `newTestPool` needs to pass in nil for the store (to signalize a default one should be used)

txpool/txpool_test.go
zivkovicmilos: I've checked out why I added this helper in the first place, and it turns out it was a faulty test addition that was the culprit

Fixed 👍 

df73baa

zivkovicmilos:

zivkovicmilos:
recheck

txpool/queue.go
dbrajovic: It's not intrinsically tied to the `accountQueue` - it's just collecting hashes from a slice of transactions.

zivkovicmilos: Yes - but the alternative is to create 2 separate loops for the pruned event hashes outside the `prune` method:
* One for the enqueued txs
* One for the promoted txs

dbrajovic (approved):
All good - thanks for replacing the subpar WaitUntilDone approach to unit tests.

Kourin1996 (approved):
Looks good to me 👍

dbrajovic:

dbrajovic:
recheck

lazartravica (approved):
Please resolve the merge conflicts, and we're good to merge 👌

PR #374, opened 2022-01-20
Fix faulty leave in TestRoutingTable_Disconnected

zivkovicmilos:
# Description

This PR resolves the issue in `TestRoutingTable_Disconnected`.

(runDial)
If a peer leaves it triggers a notification to the notifyCh, which restarts the dial loop
The node checks if it has open slots - it has, since it just disconnected from someone - and it reconnects to the peer

# Changes include

# Checklist

- [x] I have tested this code


Kourin1996 (approved):
LGTM

network/discovery_test.go
```diff
@@ -112,13 +112,13 @@ func TestRoutingTable_Disconnected(t *testing.T) {
 		1: defaultConfig,
 	}
 
-	servers, createErr := createServers(3, paramsMap)
+	servers, createErr := createServers(2, paramsMap)
```
lazartravica, line 115: We simply didn't need more than 2 servers for be able to do this test or is it something else?

network/discovery_test.go
```diff
@@ -112,13 +112,13 @@ func TestRoutingTable_Disconnected(t *testing.T) {
...
 	if createErr != nil {
 		t.Fatalf("Unable to create servers, %v", createErr)
 	}
 
 	t.Cleanup(func() {
-		closeTestServers(t, servers)
+		closeTestServers(t, servers[1:])
```
lazartravica, line 121: Why not closing all test servers?

lazartravica (changes requested):

network/discovery_test.go
zivkovicmilos: Yes, we needed just 2 servers to test this functionality, the `3` was leftover from some other refactor

network/discovery_test.go
zivkovicmilos: Because we close Server 0 inside the test:

https://github.com/0xPolygon/polygon-edge/pull/374/files#diff-0c20fd327ddad3dc3b8cf75536c00e3d3010460d8359f35d84d749567d4e8a15R145

Double closing it would cause a panic

lazartravica (approved):

PR #380, opened 2022-01-24
[TxPool] Exclude demoted/dropped accounts from block building

dbrajovic:
# Description

This PR addresses the issue of processing accounts that had a recoverable/unrecoverable transaction (in promoted queue) during `ibft.writeTransactions()`. 

# Changes include

# Checklist

- [ ] I have tested this code

# Additional comments

Previously, when dropping/demoting a transaction from some account, the subsequent transactions could still be processed in `ibft.writeTransactions()` for loop. This processing was unnecessary for consensus as well as dangerous since it would be entirely possible to repeatedly process the same (demoted) transactions again.


dbrajovic:

txpool/txpool.go
```diff
@@ -94,8 +94,7 @@ through their designated channels. */
 // 	to the pool (Demote). These requests have the
 // 	demoted flag set to true.
```
lazartravica: Comment no longer valid

txpool/txpool.go
```diff
@@ -346,63 +345,48 @@ func (p *TxPool) Pop(tx *types.Transaction) {
...
 
-// Drop removes the given (unrecoverable) transaction from
-// its associated promoted queue (account) and rolls
-// back the account's nextNonce.
-// Will update executables with the next primary
-// from that account (if any).
+// Drop clears the entire account associated with the given transaction
+// and reverts its next (expected) nonce.
 func (p *TxPool) Drop(tx *types.Transaction) {
+	// drop the entire account and revert nonce
```
lazartravica: comment is a bit misleading as the line below it doesn't do the dropping

txpool/txpool.go
```diff
@@ -346,63 +345,48 @@ func (p *TxPool) Pop(tx *types.Transaction) {
...
-	account.promoted.pop()
+	defer func() {
+		account.enqueued.unlock()
+		account.promoted.unlock()
+	}()
 
-	// update state
-	p.index.remove(tx)
-	p.gauge.decrease(slotsRequired(tx))
+	p.logger.Debug("dropping tx (unrecoverable)", "hash", tx.Hash.String())
```
lazartravica: i don't think we have a notion of recoverable/unrecoverable really here.

if we tomorrow decide to use the Drop method in some other workflow, this log will be very misleading

txpool/txpool_test.go
```diff
@@ -1054,119 +988,10 @@ func TestDrop(t *testing.T) {
...
-
-		// enqueue demoted
-		pool.handleEnqueueRequest(addReq)
-
-		assert.Equal(t, uint64(1), pool.gauge.read())
-		assert.Equal(t, uint64(1), pool.accounts.get(addr1).enqueued.length())
-		assert.Equal(t, uint64(0), pool.accounts.get(addr1).promoted.length())
-	})
+func TestDemote(t *testing.T) {
+	// TODO dbrajovic
```
lazartravica, line 993: Can we do this as a part of this PR?

lazartravica: We're removing a lot of tests, but not adding any :)

lazartravica (changes requested):
Looks good. I wasn't able to reproduce the original issue, can you provide steps to do so?

I would want us to add tests around this as a part of this PR

txpool/txpool.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/380/commits/7d72812f7a3a8f20680a8c36b362e21185d6f2c8)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/380#:~:text=fix%20Drop()%20method-,7a9d85a,-Merge%20state)

txpool/txpool.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/380#:~:text=fix%20Drop()%20method-,7a9d85a,-Merge%20state)

txpool/txpool_test.go
dbrajovic: The reason a lot of existing tests were removed is because the entire flow for demoted transactions is removed. Essentially, there is no need to do any handling (using Demote) for recoverable transactions:

https://github.com/0xPolygon/polygon-edge/blob/5bb6451d254dc465694f85543bc7cbaedadb3419/consensus/ibft/ibft.go#L699-L701

Since, previously, these transactions would just end up back in promote anyhow, now we're just shortcutting that by leaving the account as it is (and excluding it from further processing).

I've left TestDemote as a todo for when we eventually implement the pruning mechanism.

txpool/queue.go
```diff
@@ -62,6 +62,20 @@ func (q *accountQueue) prune(nonce uint64) (pruned []*types.Transaction) {
...
+func (q *accountQueue) clear() (removed []*types.Transaction) {
+	for {
+		tx := q.pop()
+		if tx == nil {
+			break
+		}
+
+		removed = append(removed, tx)
+	}
+
```
Kourin1996: Is it possible to clear queue field in accountQueue directly?

txpool/txpool.go
```diff
@@ -346,63 +338,46 @@ func (p *TxPool) Pop(tx *types.Transaction) {
...
+	// rollback nonce
+	nextNonce := tx.Nonce
+	account.setNonce(nextNonce)
 
-	// update metrics
-	p.metrics.PendingTxs.Add(-1)
+	// drop promoted
+	dropped := account.promoted.clear()
+	p.index.remove(dropped...)
+	p.gauge.decrease(slotsRequired(dropped...))
```
zivkovicmilos: Can we extract these into a helper method? (could be inside `Drop`, since we use it a few lines further)

txpool/txpool.go
```diff
@@ -346,63 +338,46 @@ func (p *TxPool) Pop(tx *types.Transaction) {
...
-	if tx := account.promoted.peek(); tx != nil {
-		p.executables.push(tx)
-	}
+	// drop enqueued
+	dropped = account.enqueued.clear()
+	p.index.remove(dropped...)
+	p.gauge.decrease(slotsRequired(dropped...))
 
 	p.eventManager.signalEvent(proto.EventType_DROPPED, tx.Hash)
+	p.logger.Debug("dropped account",
```
zivkovicmilos: Why is the message `dropped account`?
Shouldn't it be something like `dropped transaction`?

txpool/queue.go
dbrajovic: Actually, it can. What about this approach ?

https://github.com/0xPolygon/polygon-edge/pull/380/commits/c388fac96cebfd3ede0124227c88228f24896c6b

txpool/txpool.go
dbrajovic: Sure, just give me a clever name for that helper method.

txpool/txpool.go
dbrajovic: Because we're dropping all transactions from that account. Perhaps a more detailed output on the actions of `Drop()` ?

lazartravica: We are dropping the entire enqueued array of account's transactions here AFAIK

Kourin1996 (approved):
LGTM

txpool/txpool.go
lazartravica: clearAccountQueue, can be an anonymous function, just be wary of closures

txpool/txpool.go
dbrajovic: @lazartravica not just the enqueued transactions, but all transactions the account has (promoted + enqueued)

brkomir (approved):

txpool/txpool.go
lazartravica: @dbrajovic Got it, I'm still not at 100% with the new Txpool data structure glossary :)

 @zivkovicmilos please take a look at this thread

lazartravica (approved):

txpool/txpool.go
zivkovicmilos: Got it, maybe the message should be `dropped account transactions`

zivkovicmilos (approved):
Looks good 💯 

Please resolve the minor nitpicks 🙏

txpool/txpool.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/380/commits/8211ae5617010c1be5764e1069f43b37165ff177)

txpool/txpool.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/380/commits/8211ae5617010c1be5764e1069f43b37165ff177)

txpool/txpool.go
```diff
@@ -358,25 +367,25 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 
 	// drop promoted
 	dropped := account.promoted.clear()
-	p.index.remove(dropped...)
-	p.gauge.decrease(slotsRequired(dropped...))
+	droppedCount += len(dropped)
```
zivkovicmilos: Why not shove this increment logic also in `clearAccountQueue`?

txpool/txpool.go
zivkovicmilos: How is it unambiguous now? 😄

dbrajovic: [Unambigous](https://github.com/0xPolygon/polygon-edge/pull/380#:~:text=fix%20log-,e48473c,-edit%20clearAccountQueue)

txpool/txpool.go
dbrajovic: [Moved](https://github.com/0xPolygon/polygon-edge/pull/380/commits/6a7ed4f2a87b4f7aad4025a9e921d91552622915)

lazartravica (approved):

PR #385, opened 2022-01-25
Fix linting errors

zivkovicmilos:
# Description

This PR resolves pending linting errors introduced in `v1.44` of golangci-lint

# Changes include

# Checklist

- [x] I have tested this code


zivkovicmilos:

zivkovicmilos:
recheck

dbrajovic (approved):
LGTM, praise the linter 👍

Kourin1996 (approved):
LGTM

state/txn.go
```diff
@@ -1,3 +1,4 @@
+//nolint:forcetypeassert
```
lazartravica, line 1: Why the entire file since we have nolint statements below?

Why couldn't we properly check for success of type assertion as elsewhere?

jsonrpc/dispatcher.go
```diff
@@ -129,6 +130,7 @@ func (d *Dispatcher) handleSubscribe(req Request, conn wsConn) (string, Error) {
 
 	subscribeMethod, ok := params[0].(string)
 	if !ok {
+		//nolint:forcetypeassert
```
lazartravica: Why do we not force only for string type assertions?

lazartravica (changes requested):

state/txn.go
zivkovicmilos: This package supports 0 error handling, so there is no point in refactoring the methods to support them in any way, since they're part of the legacy codebase

jsonrpc/dispatcher.go
zivkovicmilos: 🤦 
We actually already do type assertion in the line before 👍 

Fixed 👍 
87fd58f

lazartravica (approved):

PR #386, opened 2022-01-25
[Bridge] Event tracker

dbrajovic:
# Description

Fixes EDGE-9

This PR introduces a PoC event tracker that is able to listen for specific events on the rootchain (ethereum). 

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

The issue of having two different header structures remains - polygon-edge headers have different field types than that of ethereum's header. Using the appropriate structure is necessary for the websocket client to [correctly parse](https://github.com/0xPolygon/polygon-edge/blob/fc3e288c0f511f7b170312efb20c39e30854a265/bridge/tracker/rootchain.go#L176) the subscription response.


server/server.go
```diff
@@ -188,6 +192,22 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
 		m.blockchain.SetConsensus(m.consensus)
 	}
 
+	if m.config.Tracker {
+		//	create and start event tracker
+		m.tracker, err = tracker.NewEventTracker(
+			logger,
+			6, /* default */
```
zivkovicmilos: What do you think about extracting this value to a constant we can easily modify later?

By itself, it's just a magic number 🙂

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/umbracle/go-web3"
+)
+
+const (
+	lastQueriedBlockNumber = "last-block-num" //	db key
+)
+
+//	Tracker represents an event listener that notifies
+//	for each event emitted on the rootchain (ethereum).
```
zivkovicmilos: Why is `Ethereum` here? This can theoretically be used for any blockchain network that supports the formats we're expecting

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+
+	//	load db (last processed block number)
+	db, err := tracker.loadDB()
+	if err != nil {
+		logger.Error("cannot load db", "err", err)
+
+		return nil, err
+	}
+
+	tracker.db = db
```
zivkovicmilos: This can be a single func `initDB` or something like that, that throws an error if DB initialization is not successful.

You're calling the local method of the `tracker` struct here and just using that value to set an attribute of the same `tracker` struct

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+//	2. startSubscription - goroutine responsible for handling
+//	the subscription object. This object is provided by the
+//	client's subscribeNewHeads method that initiates the subscription.
+func (t *Tracker) Start() error {
+	//	create cancellable contexts
+	ctxSubscription, cancelSubscription := context.WithCancel(context.Background())
+	t.cancelSubscription = cancelSubscription
+
+	ctxEventTracking, cancelEventTracking := context.WithCancel(context.Background())
+	t.cancelEventTracking = cancelEventTracking
```
zivkovicmilos: This can be a separate method

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	//	create newHeads subscription
+	subscription, err := t.client.subscribeNewHeads(t.headerCh)
+	if err != nil {
+		t.logger.Error("cannot subscribe to rootchain", "err", err)
+
+		return err
+	}
+
+	//	start receiving new header events
+	go t.startSubscription(ctxSubscription, subscription)
```
zivkovicmilos: This can be a separate method

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	eventCh chan []byte
+
+	// cancel funcs
+	cancelSubscription, cancelEventTracking context.CancelFunc
+
+	//	rootchain client
+	client *rootchainClient
+
+	//	db for last processed block number
+	db *leveldb.DB
```
zivkovicmilos: Why do we use LevelDB here specifically? Why not use the DB ambiguous object, that way when we add support for additional DB engines, we can just as easily swap them out

What do you think @lazartravica @Kourin1996 ?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+		//	block height less than required
+		t.logger.Debug(
+			"not enough confirmations",
+			"current", latestHeight.Uint64(),
+			"required", confirmations.Uint64())
+
+		return
+	}
+
+	/*	right bound */
```
zivkovicmilos: Nitpick: these single line comments should use `//`

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	toBlock := big.NewInt(0).Sub(latestHeight, confirmations)
+
+	/*	left bound */
```
zivkovicmilos: Nitpick: these single line comments should use `//`

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	Number       uint64        `json:"number"`
+	ParentHash   types.Hash    `json:"parentHash"`
+	ReceiptsRoot types.Hash    `json:"receiptsRoot"`
+	Sha3Uncles   types.Hash    `json:"sha3Uncles"`
+	StateRoot    types.Hash    `json:"stateRoot"`
+	Timestamp    uint64        `json:"timestamp"`
+	TxRoot       types.Hash    `json:"transactionsRoot"`
+	MixHash      types.Hash    `json:"mixHash"`
+	Hash         types.Hash    `json:"hash"`
+}
```
zivkovicmilos: What do you think about moving this to a separate file?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+func (t *Tracker) calculateRange(header *ethHeader) (from, to uint64, ok bool) {
+	//	extract block number from header field
+	//	TODO: polygon-edge header
+	latestHeight := big.NewInt(0).SetUint64(header.Number)
+
+	//	or: TODO ethereum header
+	//latestHeight, _ := types.ParseUint256orHex(&header.Number)
+
+	//	check if block number is at required depth
+	confirmations := big.NewInt(0).SetUint64(t.confirmations)
```
zivkovicmilos: Why not keep confirmations as `*big.Int`? You only initialize it once anyways, but do this conversion every time for a new header

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+
+	return lastBlock
+}
+
+//	saveLastBlock stores the number of the last block processed
+//	into the tracker's database.	.
+func (t *Tracker) saveLastBlock(blockNumber *big.Int) error {
+	//	store last processed block number
+	if err := t.db.Put(
+		[]byte(lastQueriedBlockNumber),
```
zivkovicmilos: Why not have this be a constant, instead of casting it to `[]byte` every time?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+		web3.HexToAddress(ThirdEventSC),
+	}
+
+	//	set relevant topics (defined in contracts)
+	queryFilter.Topics = [][]*web3.Hash{
+		{
+			&topicPoCEvent,
+			&topicAnotherEvent,
+		},
+	}
```
zivkovicmilos: This should be a separate method, like `setupQueryFilter` or something

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		t.logger.Error("eth_getLogs failed", "err", err)
+
+		return nil
+	}
+
+	//	overwrite checkpoint
+	if err := t.saveLastBlock(big.NewInt(0).SetUint64(toBlock)); err != nil {
+		t.logger.Error("cannot save last block number proccesed", "err", err)
+	}
```
zivkovicmilos: Why do we need to overwrite?

Please explain this section better 🙏

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+//	Number       string        `json:"number"`
+//	ParentHash   types.Hash    `json:"parentHash"`
+//	ReceiptsRoot types.Hash    `json:"receiptsRoot"`
+//	Sha3Uncles   types.Hash    `json:"sha3Uncles"`
+//	StateRoot    types.Hash    `json:"stateRoot"`
+//	Timestamp    string        `json:"timestamp"`
+//	TxRoot       types.Hash    `json:"transactionsRoot"`
+//	MixHash      types.Hash    `json:"mixHash"`
+//	Hash         types.Hash    `json:"hash"`
+//}
```
zivkovicmilos: I assume we're leaving this intentionally here?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+//	it is sent to eventCh.
+func (t *Tracker) trackHeader(header *ethHeader) {
+	//	determine range of blocks to query
+	fromBlock, toBlock, ok := t.calculateRange(header)
+	if !ok {
+		//	we are returning here because the calculated
+		//	range was already queried or the chain is not
+		//	at the desired depth.
+		return
+	}
```
zivkovicmilos: This `ok` logic shouldn't be part of `t.calculateRange`, you're effectively bundling checking logic and actual range calculation logic into a single method. What do you think about splitting them up?

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,143 @@
...
+
+	//	Polygon Edge
+	rootchainWS = "ws://127.0.0.1:10002/ws"
+)
+
+const (
+	//	Smart contract addresses
+	PoCSC          = "19DC3Af00E7f7502a2A40B7e0FeA194A86CeAA0c"
+	AnotherEventSC = "69ceed5Ff0FA5106F4Df7299C8812377394A9388"
+	ThirdEventSC   = "b22a1Cd34d39D46bB2f077bd6295c850702D8e81"
```
Kourin1996: What these contracts are for?

bridge/tracker/tracker.go
Kourin1996: Might need storage to track processed logs. But I'm not sure we need right now

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+		t.logger.Error("cannot close rootchain client", "err", err)
+
+		return err
+	}
+
+	return nil
+}
+
+//	getEventChannel returns the tracker's event channel.
+func (t *Tracker) getEventChannel() <-chan []byte {
```
Kourin1996: why private?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+
+	//	set range of blocks to query
+	queryFilter.SetFromUint64(fromBlock)
+	queryFilter.SetToUint64(toBlock)
+
+	//	set smart contract addresses
+	queryFilter.Address = []web3.Address{
+		web3.HexToAddress(PoCSC),
+		web3.HexToAddress(AnotherEventSC),
+		web3.HexToAddress(ThirdEventSC),
```
Kourin1996: We shouldn't specify the events you want to subscribe to

@zivkovicmilos How do you think?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	//	set relevant topics (defined in contracts)
+	queryFilter.Topics = [][]*web3.Hash{
+		{
+			&topicPoCEvent,
+			&topicAnotherEvent,
+		},
```
Kourin1996: topics are as well

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	return logs
+}
+
+//	notify sends the given logs to the event channel.
+func (t *Tracker) notify(logs ...*web3.Log) {
+	for _, log := range logs {
+		bytesLog, err := json.Marshal(log)
+		if err != nil {
+			t.logger.Error("cannot marshal log", "err", err)
+		}
```
Kourin1996: tbh, I don't think we need to convert to bytes because SAM Pool needs event ID

Kourin1996:
@dbrajovic @zivkovicmilos This tracker seems to be specialized for bridge. I think it's better to give the event types and some settings from outside.

zivkovicmilos:
> @dbrajovic @zivkovicmilos This tracker seems to be specialized for bridge. I think it's better to give the event types and some settings from outside.

@Kourin1996 

Completely agree with this, it shouldn't be too much of a workaround to make sure these params are dynamic 👍

bridge/tracker/tracker.go
dbrajovic: @Kourin1996 is right - we need a db to store the last processed block. Imagine the node goes down and you bring it back up again: we wouldn't want to query the same blocks again.

@zivkovicmilos you mean like a `store` interface ?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,390 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	//	create or load db
+	db, err := leveldb.OpenFile(cwd+"/event_tracker/last_block_number", nil)
+	if err != nil {
+		return nil, err
+	}
```
Kourin1996: `event_tracker` directory will be created under current directory. You might want to move this directory into data directory

zivkovicmilos: Good catch, missed this 💯

bridge/tracker/tracker.go
dbrajovic: The only reason I didn't group those 2 together is so `Start()` can return an error to the caller. Perhaps there isn't a need for this ? If so, I can move `subscribeNewHeads()` into `StartSubscription()`. It all depends on how should we handle errors in this call.

bridge/tracker/tracker.go
dbrajovic: When the tracker successfully performs a query, it needs to save the last queried block number, so that it doesn't query it again on the next call.

bridge/tracker/tracker.go
dbrajovic: Until we resolved the issue of our own header type, I'd say so, yes.

bridge/tracker/tracker.go
dbrajovic: No particular reason - depends on how we structure our `bridge` module along with the other [PR](https://github.com/0xPolygon/polygon-edge/pull/419).

server/server.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-edge/pull/386/commits/47545725885115a68fa5d69b3d823bb11c6d96df)

bridge/tracker/tracker.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/c3a1d5a0b102dd34858726ae9e6fcf38cfe62204). Although, the discussion for our expected header has yet to be resolved.

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/8cff0918ede7d6c018f0d99293dd5fa0514ad40d)

bridge/tracker/tracker.go
dbrajovic: I didn't want to further abstract the `Start()` method. That separate method would probably have to return 2 contexts and set the tracker's fields in its body - further hiding away the logic.

bridge/tracker/tracker.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/6c958ebf54e9803f1570acdc157528ceef0d606c)

bridge/tracker/tracker.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/6c958ebf54e9803f1570acdc157528ceef0d606c)

bridge/tracker/tracker.go
dbrajovic: [Moved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/44989ee14dbef25e79c23bca1793f88427a372cd).

Moved to `rootchain.go`, since it makes most sense. We can discuss how to better structure that file.

bridge/tracker/tracker.go
dbrajovic: [Changed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/dd26ae7d3aaafcd50b168d1cab05b94552718c36)

bridge/tracker/tracker.go
dbrajovic: [Redefined](https://github.com/0xPolygon/polygon-edge/pull/386/commits/768525864fdabbdbb371131efa3f9f98e7ca80a7)

bridge/tracker/tracker.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-edge/pull/386/commits/07ac0e157de64918862ae67a75644fb2aa69a1f6)

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/69983416af5816102248834326555384f28b28cb)

bridge/tracker/rootchain.go
dbrajovic: Custom contracts used for PoC. [Removed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/a97c59b221d0eb64ef62d18bbe152f443b9d8b1b)

bridge/tracker/tracker.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/a97c59b221d0eb64ef62d18bbe152f443b9d8b1b)

bridge/tracker/tracker.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/a97c59b221d0eb64ef62d18bbe152f443b9d8b1b)

bridge/tracker/tracker.go
dbrajovic: ~~The db folder should go where exactly ? As in, where should it reside in the project directory.~~

[Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/cea3677cb23fb747d347d6ea99d0ea2a9a490dcb)

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,305 @@
...
+	"math/big"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/umbracle/go-web3"
+)
+
+const (
+	//	required block depth for fetching events on the rootchain
+	BlockConfirmations = 6
```
Kourin1996: `DefaultBlockConfirmations` is more proper?

bridge/tracker/tracker.go
Kourin1996: @dbrajovic I don't think we should initialize database in this struct because it's hard to switch storage when testing or changing database.

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,305 @@
...
+	return nil
+}
+
+//	getEventChannel returns the tracker's event channel.
+func (t *Tracker) getEventChannel() <-chan []byte {
+	return t.eventCh
+}
+
+//	startSubscription handles the subscription object (provided by the rootchain client).
+func (t *Tracker) startSubscription(ctx context.Context, sub subscription) {
```
Kourin1996: `startSubscription` is not proper name...? because this function just handles error or terminate signals, not process any message. And we don't need for loop here, right?

Kourin1996 (changes requested):
I've left small comments. Can you check them?

And JSON-RPC URL and parent directory of storage are better to given from outside.

bridge/tracker/tracker.go
Kourin1996: `Bridge` is defined outside package. Need to make it public :)

bridge/tracker/tracker.go
zivkovicmilos: You're directly modifying the `t.cancelSubscription` and `t.cancelEventTracking` fields, you can move away this block of code into a method called `initContexts` or something similar, that has no return values.

You keep track of the cancel funcs, but not of the contexts. Perhaps you could define a `cancelParams` type, and have 2 fields in `Tracker` called `ctxEventTracking` and `ctxSubscription` of type `cancelParams`, and you'd just use these throughout the code, instead of passing them around as params. It's weird that you keep the cancel method saved, but not the actual context it's tied to - by saving them both you make the code a lot cleaner

bridge/tracker/tracker.go
zivkovicmilos: Why not instantiate the `subscription` channel when you create the tracker, instead of when you actually start it?

That way you have error handling during creation, and your start method doesn't do 20 things like it does now, it just starts the whole show.

bridge/tracker/tracker.go
zivkovicmilos: Of course we need storage, I'm just saying to use the DB store interface instead of a specific DB engine like LevelDB. Under the hood the implementation can be whatever we want. You're reinventing the wheel in this PR by writing additional code to manage LevelDB, when we already have support for it in the storage package.

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/22ed210d9e92dd02820edc89fd3ec5dc0e6cad92)

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/22ed210d9e92dd02820edc89fd3ec5dc0e6cad92)

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/cea3677cb23fb747d347d6ea99d0ea2a9a490dcb)

bridge/tracker/tracker.go
dbrajovic: [Exported](https://github.com/0xPolygon/polygon-edge/pull/386/commits/0617b1d891be43cb3d78dfced703771a53ad713e)

bridge/tracker/tracker.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/d510c26dc3e711512605c997347433f79a05782e)

bridge/tracker/tracker.go
dbrajovic: [Refactored](https://github.com/0xPolygon/polygon-edge/pull/386/commits/22ed210d9e92dd02820edc89fd3ec5dc0e6cad92)

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+	if tracker.client, err = newRootchainClient(config.RootchainWS); err != nil {
+		logger.Error("cannot connect to rootchain", "err", err)
+
+		return nil, err
+	}
+
+	//	subscribe for new headers
+	if tracker.sub, err = tracker.client.subscribeNewHeads(); err != nil {
+		logger.Error("cannot subscribe to rootchain", "err", err)
+	}
```
Kourin1996: Nitpick. I think better to start to subscribe after Start() is called, not in constructor

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+	sub subscription
+
+	//	rootchain client
+	client *rootchainClient
+
+	//	db for last processed block number
+	db storage.Storage
+
+	//	events that tracker will listen for
+	contracts []*contractABI
```
Kourin1996: why not `[]*abi.Event`?

Kourin1996 (changes requested):

bridge/tracker/tracker.go
dbrajovic: [Moved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/d8ba42ed44ff9eb623c01b2ff90942b5530d7ebb)

bridge/tracker/tracker.go
dbrajovic: I use the `contractABI` struct to keep both the address of the smart contract as well as (abi) events defined in it. Addresses are used in [setupQueryFilter](https://github.com/0xPolygon/polygon-edge/blob/d8ba42ed44ff9eb623c01b2ff90942b5530d7ebb/bridge/tracker/rootchain.go#L82) to narrow down the query. But if topics alone (event signatures) would be enough to match logs, perhaps smart contract addresses are redundant ?

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,236 @@
...
+	queryFilter.SetFromUint64(from.Uint64())
+	queryFilter.SetToUint64(to.Uint64())
+
+	//	set contract addresses and topics
+	for _, contract := range contracts {
+		//	append address
+		queryFilter.Address = append(queryFilter.Address, contract.address)
+
+		//	topics from all contracts must be in Topics[0]
+		queryFilter.Topics[0] = append(queryFilter.Topics[0], contract.eventIDs()...)
```
Kourin1996: throws panic because `queryFilter.Topics` is not initialized

Kourin1996 (changes requested):

bridge/tracker/rootchain.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/57f0d4d8766ce3fa266d8dce01ba1bf2564e4c31)

bridge/tracker/tracker.go
Kourin1996: No, I thought we could pass parsed event information, not new struct
We need contract address and event

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,236 @@
...
+
+const (
+	//	required block depth for fetching events on the rootchain
+	DefaultBlockConfirmations = 6
+
+	//	Ropsten testnet
+	//rootchainWS   = "wss://ropsten.infura.io/ws/v3/17eac086ff36442ebd43737400eb71ca"
+
+	//	Polygon Edge
+	rootchainWS = "ws://127.0.0.1:10002/ws"
```
Kourin1996: We do no longer need these?

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,236 @@
...
+	for _, ev := range c.events {
+		id := ev.ID()
+		IDs = append(IDs, &id)
+	}
+
+	return
+}
+
+//	loadABIs parses contracts from raw map.
+func loadABIs(abisRaw map[string][]string) (contracts []*contractABI) {
```
Kourin1996: Either's fine. But map key is better to be types.Address?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+//	NewEventTracker returns a new tracker object.
+func NewEventTracker(logger hclog.Logger, config *Config) (*Tracker, error) {
+	if config == nil {
```
Kourin1996, line 62: Actually nothing we can do if config is nil. You can return error or change `config` type to `Config` (value type) instead of using default value

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,236 @@
...
+
+//	rootchain subscription object
+type subscription struct {
+	newHeadsCh chan *ethHeader
+	errorCh    chan error
+	cancel     cancelSubCallback
+}
+
+//	newHead returns the subscription's channel for new head events.
+func (s *subscription) newHead() <-chan *ethHeader {
```
Kourin1996: `newHeadCh` is better?

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,236 @@
...
+	return s.newHeadsCh
+}
+
+//	unsubscribe cancels the subscription.
+func (s *subscription) unsubscribe() error {
+	return s.cancel()
+}
+
+//	err	returns the subscription's error channel.
+func (s *subscription) err() <-chan error {
```
Kourin1996: `errCh` is better?

server/server.go
```diff
@@ -64,6 +65,9 @@ type Server struct {
 	// transaction pool
 	txpool *txpool.TxPool
 
+	//	event tracker
+	tracker *tracker.Tracker
+
```
Kourin1996: Server doesn't have to have tracker because server will not use it directly

server/server.go
```diff
@@ -188,6 +192,21 @@ func NewServer(logger hclog.Logger, config *Config) (*Server, error) {
...
+		); err != nil {
+			return nil, err
+		}
+
+		//	start tracker's listening process
+		if err := m.tracker.Start(); err != nil {
+			return nil, err
+		}
+	}
+
```
Kourin1996: We don't need initialization of Tracker because server does no longer have it.

server/server.go
```diff
@@ -577,6 +596,12 @@ func (s *Server) Close() {
 		}
 	}
 
+	if s.tracker != nil {
+		if err := s.tracker.Stop(); err != nil {
+			s.logger.Error("failed to shutdown tracker", "err", err)
+		}
+	}
+
```
Kourin1996: We don't need termination of Tracker because server does no longer have it.

Kourin1996 (commented):
There is no problem in latest code. I've just left nitpick comment. Please check them. Thank you

bridge/tracker/rootchain.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

bridge/tracker/rootchain.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

bridge/tracker/tracker.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

bridge/tracker/rootchain.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

bridge/tracker/rootchain.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

server/server.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

server/server.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

server/server.go
dbrajovic: [Resolved](https://github.com/0xPolygon/polygon-edge/pull/386/commits/585d357ff9ff834ac33cc2fd98a060539046f5dc)

0xAleksaOpacic (approved):
Except for yoshiki's comments, the code looks good. I did not tested localy

Kourin1996 (approved):
LGTM. Thank you for fixes

server/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	Network        *network.Config
 	DataDir        string
 	Seal           bool
+	Tracker        bool
```
zivkovicmilos: What do you think about renaming this to `TrackerEnabled`, since you only store the value indicating if the tracker should be initialized in the first place?

bridge/tracker/tracker.go
```diff
@@ -0,0 +1,304 @@
...
+	lastBlockNumber, ok := t.db.ReadHeadNumber()
+	if !ok {
+		return nil
+	}
+
+	return big.NewInt(0).SetUint64(lastBlockNumber)
+}
+
+//	saveLastBlock stores the number of the last block processed
+//	into the tracker's database.	.
```
zivkovicmilos: Nitpick: all method comments should be without the punctuation at the end.

Please make sure we keep the code guidelines in check regarding this detail 🙏

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,231 @@
...
+
+//	contractABI is used to create query filter
+//	that matches events defined in a smart contract.
+type contractABI struct {
+	//	address of smart contract
+	address web3.Address
+
+	// signatures of events defined in smart contract
+	events []*abi.Event
+}
```
zivkovicmilos: What do you think about renaming this to `eventABI`, since you're storing event signatures here, and the address of the actual smart contract that emits them?
Having it be `contactractABI` is misleading, as you don't keep any ABI data concerning the contract apart from the events.

bridge/tracker/rootchain.go
```diff
@@ -0,0 +1,231 @@
...
+//	Number       string        `json:"number"`
+//	ParentHash   types.Hash    `json:"parentHash"`
+//	ReceiptsRoot types.Hash    `json:"receiptsRoot"`
+//	Sha3Uncles   types.Hash    `json:"sha3Uncles"`
+//	StateRoot    types.Hash    `json:"stateRoot"`
+//	Timestamp    string        `json:"timestamp"`
+//	TxRoot       types.Hash    `json:"transactionsRoot"`
+//	MixHash      types.Hash    `json:"mixHash"`
+//	Hash         types.Hash    `json:"hash"`
+//}
```
zivkovicmilos: Can you please explain in detail the issue with using Ethereum headers in the PR description? 🙏 

I know we discussed it internally, but it's better to have this sort of data available for easy reference

zivkovicmilos (approved):
Looks good 💯 

Thank you for simplifying the code, it's much easier to look at and understand now.

I've left a few minor comments we should discuss

server/config.go
dbrajovic: Actually, great catch - we decided to remove the tracker from the server entirely. (@Kourin1996 )
[Removed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/099bb19976e1301cba0bb62c68785e6ce8149be2)

bridge/tracker/tracker.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/c3b941e0d61fa63d21c363c9e89a23807e507e35)

bridge/tracker/rootchain.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-edge/pull/386/commits/a2f94e582faa1d97c74faa41bd203adef489fc54)

bridge/tracker/rootchain.go
zivkovicmilos: Resolving because you added the description to the PR 👍

lazartravica (changes requested):
LGTM!

But please @dbrajovic @Kourin1996 let's merge `develop` into `bridge` and this `branch` since I want to double check we're not breaking anything since the difference is falsely reported as being 49 differrent files changed

Thank you.

And also guys, amazing coments and resolutions so far on this PR! 🎉

bridge/tracker/tracker.go
dbrajovic: But if ID (event signature) is all that you need then how would you be able to construct a transaction from this ID only ?

lazartravica (approved):
LGTM all except the header issue, but let's merge this and discuss fixing the headers 👍

PR #376, opened 2022-01-21
Update predeploy staking contract bytecode

Kourin1996:
# Description

Fixes EDGE-60

This PR updates the bytecode of predeploy staking contract based on the change of https://github.com/0xPolygon/staking-contracts/pull/8

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [ ] I have updated the README and other relevant documents (guides...)
- [ ] I have added sufficient documentation both in code, as well as in the READMEs

# Additional comments

Please post additional comments in this section if you have them, otherwise delete it

PR #397, opened 2022-01-28
Update predeployed staking SC

0xpanoramix:
# Description

Fixes EDGE-6

This PR updates the staking contract bytecode and sets the default maximum number of validators to max unit32.

# Changes include

# Checklist

## Testing

### Manual tests

1. Start --pos chain with default values for min/max validators and default values for min/max should be 1/4294967295
2. When creating genesys:
     - min-validator-count >= 1
     - min-validator-count <= max-validator-count
     - max-validator-count <= max.Uint32 (4294967295)
3. Can not add new validator if in the network there are max-validator-count validators
     1. Start pos network with 4 nodes and max-validator-count 4
     2. Try to add a new validator using staking (node 5), it will fail 
     3. Unstake from one of the validators
     4. Stake from 5th node, it will be successful
 4. Switch POA -> POS, there should be default values for min/max validators

# Documentation update
https://github.com/0xPolygon/polygon-edge-docs/pull/66

# Additional comments
Staking-contract PR (https://github.com/0xPolygon/staking-contracts/pull/10)


Kourin1996:
@PtitLuca Please note you need to update bytecode after https://github.com/0xPolygon/polygon-edge/pull/376 is merged.

Kourin1996 (approved):
Looks good to me. Please add explains about this feature to doc.

0xAleksaOpacic (changes requested):
Great job. I only have found only one problem.
@lazartravica @Kourin1996 When we are creating the genesis file, there is no check what is the minimum value from --max-validator-count. So the user can place any integer. 

In the normal flow where we create 4 nodes and then genesis, this would not be a problem. All 4 nodes will be validators. Even if we add --max-validator-count 1

But if we start a network with only 2 nodes, and set --max-validator-count to 2. We would not be able to add new validators using staking. And for ibft we need to have at least 4 nodes.

I would suggest to check minimum value of --max-validator-count which would be 4

0xAleksaOpacic (approved):

helper/staking/staking.go
```diff
@@ -138,10 +143,17 @@ func PredeployStakingSC(
 		return nil, fmt.Errorf("unable to generate DefaultStatkedBalance, %w", err)
 	}
 
+	if maxValidatorCount > math.MaxUint32 {
+		maxValidatorCount = math.MaxUint32
+	} else if maxValidatorCount < 4 {
+		return nil, fmt.Errorf("maximum number of validator can not be lower than 4")
+	}
```
zivkovicmilos: This check shouldn't be here, as there shouldn't be a minimum number of nodes you need to form a PoS cluster, apart from 1.

So I suggest setting this check to `1` (and of course extracting it into a constant, we don't want magic numbers 🙂 )

command/genesis/genesis.go
```diff
@@ -153,6 +154,14 @@ func (c *GenesisCommand) DefineFlags() {
 		},
 		FlagOptional: true,
 	}
+
+	c.FlagMap["max-validator-count"] = helper.FlagDescriptor{
+		Description: "Sets the flag indicating the maximum number of validators in the validator set for PoS",
```
zivkovicmilos: Please add what is the default value / or indicate that there is no upper bound if this flag isn't set 🙏

zivkovicmilos (approved):
Looks great 💯 

Please make sure to let @mrwillis know we're implementing a part of the requested feature set in this discussion: 
https://github.com/0xPolygon/polygon-edge/discussions/288

command/genesis/genesis.go
```diff
@@ -206,6 +227,8 @@ func (c *GenesisCommand) Run(args []string) int {
 	flags.Uint64Var(&epochSize, "epoch-size", ibft.DefaultEpochSize, "")
 	flags.Uint64Var(&blockGasLimit, "block-gas-limit", helper.GenesisGasLimit, "")
 	flags.BoolVar(&isPos, "pos", false, "")
+	flags.UintVar(&minNumValidator, "min-validator-count", 1, "")
+	flags.UintVar(&maxNumValidator, "max-validator-count", math.MaxUint32, "")
```
zivkovicmilos: These default values should be extracted out, since you use them in multiple places 👍

command/helper/helper.go
```diff
@@ -355,6 +356,8 @@ func generateDevGenesis(params devGenesisParams) error {
 
 	stakingAccount, err := staking.PredeployStakingSC(
 		[]types.Address{},
+		1,
+		math.MaxUint32,
```
zivkovicmilos: Please see https://github.com/0xPolygon/polygon-edge/pull/397/files#r805781164

helper/staking/staking.go
```diff
@@ -110,18 +115,22 @@ var (
...
 )
 
 // PredeployStakingSC is a helper method for setting up the staking smart contract account,
 // using the passed in validators as prestaked validators
 func PredeployStakingSC(
 	validators []types.Address,
+	minValidatorCount uint,
+	maxValidatorCount uint,
```
zivkovicmilos: What do you think about extracting these params into a structure like `PredeployParams`?

helper/staking/staking.go
```diff
@@ -138,10 +147,20 @@ func PredeployStakingSC(
 		return nil, fmt.Errorf("unable to generate DefaultStatkedBalance, %w", err)
 	}
 
+	if minValidatorCount > maxValidatorCount {
+		return nil, fmt.Errorf("minimum number of validator can not be greater than maximum number of validator")
```
zivkovicmilos: This should be defined as a separate error with `errors.New`, and not with `fmt.Errorf`

helper/staking/staking.go
```diff
@@ -110,18 +115,22 @@ var (
...
...fffffffffffffffffffffffffffffffff82169050919050565b6000819050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b600080fd5b7f4e756d626572206f662076616c696461746f72732063616e2774206265206c6560008201527f7373207468616e204d696e696d756d52657175697265644e756d56616c69646160208201527f746f727300000000000000000000000000000000000000000000000000000000604082015250565b7f4f6e6c79207374616b65722063616e2063616c6c2066756e6374696f6e000000600082015250565b7f56616c696461746f72207365742068617320726561636865642066756c6c206360008201527f6170616369747900000000000000000000000000000000000000000000000000602082015250565b7f696e646578206f7574206f662072616e67650000000000000000000000000000600082015250565b7f4f6e6c7920454f412063616e2063616c6c2066756e6374696f6e000000000000600082015250565b6112f8816110ba565b811461130357600080fd5b50565b61130f81611114565b811461131a57600080fd5b5056fea2646970667358221220a8138d18f101e6f82542c730da59e80cf72514739559efba0f22714c63d74fa164736f6c63430008070033"
```
zivkovicmilos: This bytecode doesn't match the `deployedBytecode` from the staking SC PR ⚠️ 

Please update it 🙏

zivkovicmilos (changes requested):
Thank you for adding support for this feature 🙏 

Please see the comments I've left.

This PR is lacking proper testing for this feature, please add a unit test that verifies `PredeployStakingSC` errors, and an e2e test that verifies nodes can't stake / unstake until the set criteria for the min / max values is met.

One e2e test could be for example to set the max validator count to 4, and try to make a new node stake.

The other e2e test could be to have a min 4 node cluster, but try to unstake.

command/genesis/genesis.go
0xpanoramix: Added in https://github.com/0xPolygon/polygon-edge/pull/397/commits/fe34bdd49d2f6aae469ffc12ae67d33f89fe280f

helper/staking/staking.go
0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-edge/pull/397/commits/d27667f7f5f16452596e14659a9aa9a2f7552db0

helper/staking/staking.go
0xpanoramix: Fixed in https://github.com/0xPolygon/polygon-edge/pull/397/commits/0cd22450d81525e6724c7ecf8f4761dcd64fe0d6

helper/staking/staking.go
0xpanoramix: I agree 👍 Added in https://github.com/0xPolygon/polygon-edge/pull/397/commits/fe34bdd49d2f6aae469ffc12ae67d33f89fe280f

Kourin1996 (changes requested):
I can't confirm Predeploy sets `_minimumNumValidators`. Can you check this? `_minimumNumValidators` is zero even though I gave --min-validator-count 5 to `genesis` command

helper/staking/staking.go
zivkovicmilos: Yes, I was thinking of making this error available at the beginning of the file with other errors, like

```go
var (
    errInvalidValidatorRange = errors.new("minimum ...")
)
```

0xpanoramix: Got it, fixed in https://github.com/0xPolygon/polygon-edge/pull/397/commits/1962fcb18003c084a232b432f728b8e14cfde03e

0xpanoramix:
@Kourin1996 Got it, there was just a typo in the smart contract, here are the changes : https://github.com/0xPolygon/staking-contracts/pull/10/commits/fb5bd8e47aed486aec2dcabd40d0b1b4ecab8fb7

I was writing the values in the wrong order, that's why you did not have the right value.

Kourin1996 (approved):
It works! Looks good to me.

0xAleksaOpacic (approved):
Tested it, works great! Ping me when you add tests

consensus/ibft/pos.go
```diff
@@ -146,7 +146,7 @@ func (pos *PoSMechanism) preStateCommitHook(rawParams interface{}) error {
 	}
 
 	// Deploy Staking contract
-	contractState, err := stakingHelper.PredeployStakingSC(nil)
+	contractState, err := stakingHelper.PredeployStakingSC(nil, stakingHelper.PredeployParams{})
```
Kourin1996: @Aleksao998 You will need to fill in the fields of PredeployParams

helper/staking/staking.go
```diff
@@ -106,18 +126,20 @@ var (
...
...f82169050919050565b6000819050919050565b600063ffffffff82169050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603160045260246000fd5b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b600080fd5b7f4e756d626572206f662076616c696461746f72732063616e2774206265206c6560008201527f7373207468616e204d696e696d756d52657175697265644e756d56616c69646160208201527f746f727300000000000000000000000000000000000000000000000000000000604082015250565b7f4f6e6c79207374616b65722063616e2063616c6c2066756e6374696f6e000000600082015250565b7f56616c696461746f72207365742068617320726561636865642066756c6c206360008201527f6170616369747900000000000000000000000000000000000000000000000000602082015250565b7f696e646578206f7574206f662072616e67650000000000000000000000000000600082015250565b7f4f6e6c7920454f412063616e2063616c6c2066756e6374696f6e000000000000600082015250565b6113d281611184565b81146113dd57600080fd5b50565b6113e9816111de565b81146113f457600080fd5b5056fea26469706673582212204c2964a8ac13edacf7e87d19aef0b89ce42603271129d8494f5203b3ed954c6964736f6c63430008070033"
```
dbrajovic: Should we maybe put this bytecode in to a separate .go file ?

helper/staking/staking.go
```diff
@@ -134,12 +156,23 @@ func PredeployStakingSC(
 		return nil, fmt.Errorf("unable to generate DefaultStatkedBalance, %w", err)
 	}
 
+	fmt.Println(params.MinValidatorCount, params.MaxValidatorCount)
```
dbrajovic: debug print

0xAleksaOpacic: This is leftover, removed it 703cb2ccaac20031a5f1addc0c4df1f79506659d

helper/staking/staking.go
0xAleksaOpacic: In my opinion no, it is only used for staking inside staking.go. But if you disagree we can

lazartravica (approved):
LGTM, please resolve the conversations.

consensus/ibft/pos.go
0xAleksaOpacic: Fixed 703cb2ccaac20031a5f1addc0c4df1f79506659d

helper/staking/staking.go
0xAleksaOpacic: @zivkovicmilos Please resolve conversation

command/helper/helper.go
0xAleksaOpacic: @zivkovicmilos Please resolve conversation

command/genesis/params.go
```diff
@@ -38,7 +41,7 @@ var (
 	errValidatorsNotSpecified         = errors.New("validator information not specified")
 	errValidatorsSpecifiedIncorrectly = errors.New("validator information specified through mutually exclusive flags")
 	errUnsupportedConsensus           = errors.New("specified consensusRaw not supported")
-	errMissingBootnode                = errors.New("at least 1 bootnode is required")
+	errMissingBootnode                = errors.New("at least 1 BootNode is required")
```
dbrajovic: nitpick: revert `bootNode` back to `bootnode`

helper/staking/staking.go
```diff
@@ -10,6 +12,17 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+var (
+	MinValidatorCount = uint32(1)
+	MaxValidatorCount = uint32(math.MaxUint32)
+)
+
+var (
+	errInvalidValidatorNumber = errors.New("minimum number of validators can not be greater than" +
```
zivkovicmilos: Typo here, should be `than the...`

helper/staking/staking.go
```diff
@@ -134,12 +156,22 @@ func PredeployStakingSC(
 		return nil, fmt.Errorf("unable to generate DefaultStatkedBalance, %w", err)
 	}
 
+	//Validate min and max validators number
+	if err = validateMinMaxValidatorsNumber(params.MinValidatorCount, params.MaxValidatorCount); err != nil {
+		return nil, err
+	}
```
zivkovicmilos: This validation should not be done at this level in the predeploy pipeline.

Please consider moving this check outside this method 🙏

helper/staking/staking.go
```diff
@@ -134,12 +156,22 @@ func PredeployStakingSC(
...
+
 	// Generate the empty account storage map
 	storageMap := make(map[types.Hash]types.Hash)
 	bigTrueValue := big.NewInt(1)
 	stakedAmount := big.NewInt(0)
+	minNumValidators := big.NewInt(int64(params.MinValidatorCount))
+	maxNumValidators := big.NewInt(int64(params.MaxValidatorCount))
```
zivkovicmilos: Please rename these values to `bigMinNumValidators` and `bigMaxNumValidators`, so it follows the naming convention from the rest of the method

helper/staking/staking.go
```diff
@@ -134,12 +156,22 @@ func PredeployStakingSC(
...
 
 	for indx, validator := range validators {
+		if int64(indx) == maxNumValidators.Int64() {
+			break
+		}
```
zivkovicmilos: This check shouldn't even exist here.

Why are you breaking if there is a mismatch between the number of validators in the validator set and the max count? Why even have this condition here, you should be certain when going into `PredeployStakingSC` that `len(validators) == maxNumValidators`.

Please remove this check, and make sure that the validator set you pass into the method matches the limits size-wise

helper/staking/staking.go
```diff
@@ -173,6 +205,16 @@ func PredeployStakingSC(
...
 
+	// Set the value for the minimum and maximum number of validators
+	min := PadLeftOrTrim(minNumValidators.Bytes(), 4)
+	max := PadLeftOrTrim(maxNumValidators.Bytes(), 4)
+	storageMap[types.BytesToHash(big.NewInt(minAndMaxNumValidator).Bytes())] = types.BytesToHash(
+		append(
+			max,
+			min...,
+		),
+	)
```
zivkovicmilos: This padding, and combining later on in the append can be a separate method, as it's universal for bytes that are packed into a  single uint256

helper/staking/staking.go
```diff
@@ -181,3 +223,15 @@ func PredeployStakingSC(
...
 	return stakingAccount, nil
 }
+
+func validateMinMaxValidatorsNumber(minValidatorCount uint32, maxValidatorCount uint32) error {
+	if minValidatorCount < 1 {
+		return errInvalidMinNumValidators
+	}
+
+	if minValidatorCount > maxValidatorCount {
+		return errInvalidValidatorNumber
```
zivkovicmilos: Please consider renaming this error to `errInvalidValidatorRange`

e2e/pos_test.go
```diff
@@ -82,6 +83,119 @@ func validateValidatorSet(
...
+			key:               accounts[1].key,
+			expectedExistence: false,
+			expectedSize:      numGenesisValidators + 1,
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			err := framework.StakeAmount(tt.address, tt.key, stakeAmount, srv)
+			assert.NoError(t, err)
```
zivkovicmilos, line 160: Why don't you verify that the transaction to stake actually errored out with the appropriate SC reason?

You just send a transaction here, don't even wait for a receipt - just check the validator set

e2e/pos_test.go
```diff
@@ -82,6 +83,119 @@ func validateValidatorSet(
...
+	ibftManager.StartServers(ctx)
+
+	srv := ibftManager.GetServer(0)
+
+	client := srv.JSONRPC()
+
+	err := framework.StakeAmount(stakerAddr, stakerKey, stakeAmount, srv)
+	assert.NoError(t, err)
+	validateValidatorSet(t, stakerAddr, client, true, numGenesisValidators+1)
+}
```
zivkovicmilos: What does this test actually test, that `TestPoS_ValidatorBoundaries` doesn't already cover?

e2e/framework/config.go
```diff
@@ -46,6 +46,8 @@ type TestServerConfig struct {
 	ShowsLog                bool                 // Flag specifying if logs are shown
 	IsPos                   bool                 // Specifies the mechanism used for IBFT (PoA / PoS)
 	Signer                  *crypto.EIP155Signer // Signer used for transactions
+	MinValidatorCount       uint32
+	MaxValidatorCount       uint32
```
zivkovicmilos: Description is missing for both of these field, but present for the other 50

e2e/framework/config.go
```diff
@@ -154,3 +156,11 @@ func (t *TestServerConfig) SetShowsLog(f bool) {
...
 	t.EpochSize = epochSize
 }
+
+func (t *TestServerConfig) SetMinValidatorCount(val uint32) {
+	t.MinValidatorCount = val
+}
+
+func (t *TestServerConfig) SetMaxValidatorCount(val uint32) {
+	t.MaxValidatorCount = val
+}
```
zivkovicmilos: Please add the method descriptions

consensus/ibft/pos.go
```diff
@@ -146,7 +145,10 @@ func (pos *PoSMechanism) preStateCommitHook(rawParams interface{}) error {
 	}
 
 	// Deploy Staking contract
-	contractState, err := stakingHelper.PredeployStakingSC(nil)
+	contractState, err := stakingHelper.PredeployStakingSC(nil, stakingHelper.PredeployParams{
+		MinValidatorCount: stakingHelper.MinValidatorCount,
+		MaxValidatorCount: stakingHelper.MaxValidatorCount,
+	})
```
zivkovicmilos: Why is this not something that's settable from the `ibft switch` params?

cc @Kourin1996

zivkovicmilos: Why assume that the default values are ok for the user?

command/genesis/params.go
```diff
@@ -23,6 +24,8 @@ const (
 	epochSizeFlag           = "epoch-size"
 	blockGasLimitFlag       = "block-gas-limit"
 	posFlag                 = "pos"
+	minNumValidators        = "min-validator-count"
+	maxNumValidators        = "max-validator-count"
```
zivkovicmilos: Flags are called `<>-validator-count`, but the flag references are called `<>NumValidators`

Please consider keeping the naming in check, like `minValidatorCount` and `maxValidatorCount`

command/genesis/params.go
zivkovicmilos: +1

command/genesis/params.go
```diff
@@ -58,6 +61,9 @@ type genesisParams struct {
 	blockGasLimit uint64
 	isPos         bool
 
+	minNumValidators uint32
+	maxNumValidators uint32
```
zivkovicmilos: Why do you accept only `uint32`s for this? There is no reason to limit the user, nobody in their right mind would choose to write `1073741824` over `0x40000000`.

Why not accept a `uint32` or a hex value?

command/genesis/genesis.go
```diff
@@ -126,6 +128,18 @@ func setFlags(cmd *cobra.Command) {
...
+		minNumValidators,
+		1,
+		"the minimum number of validators in the validator set for PoS.",
+	)
+	cmd.Flags().Uint32Var(
+		&params.maxNumValidators,
+		maxNumValidators,
+		math.MaxUint32,
+		"the maximum number of validators in the validator set for PoS.",
+	)
```
zivkovicmilos: Please remove the punctuation from the flag description 🙏

Sunset17 (approved):

zivkovicmilos (changes requested):
There are some bigger things that need to be considered here.

Please see the discussions I've left, and we can take it from there 💯

consensus/ibft/pos.go
Kourin1996: I think we can append these values on switching and should load from `engine` in genesis.json. So can give min validator count and max validator count when you run `ibft switch`. And PoSHook can load these values as with contract deployment height and PoS beginning height.

Kourin1996:
Please @zivkovicmilos's comments and resolve conflict

command/genesis/params.go
0xAleksaOpacic: The code accepts both uint32 and hex values. You can use --max-validator-count 0x40000 and it will set to 262144

zivkovicmilos: Which code?

I don't see you accepting string values and parsing them:
https://github.com/0xPolygon/polygon-edge/blob/731e69945d08d40242af8d7e5f032736c093c0d5/command/genesis/genesis.go#L131-L142

0xAleksaOpacic: Cobra handles it internally. I have tried using hex as a value for any of these flags and it parses it correctly.

command/genesis/params.go
0xAleksaOpacic: Fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

helper/staking/staking.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

helper/staking/staking.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

helper/staking/staking.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

helper/staking/staking.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

@Kourin1996 suggested that we should return error if len(validators) > maxNumValidators because user is giving them as validators

helper/staking/staking.go
0xAleksaOpacic: Not sure what do you mean with this, please explain

helper/staking/staking.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

e2e/pos_test.go
0xAleksaOpacic: No sure i understand. I know that framework.StakeAmount waits for receipt. Should i just send regular transaction or modify framework.StakeAmount not to wait for receipt

e2e/framework/config.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

e2e/pos_test.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

You are right, this one should be deleted

e2e/framework/config.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

consensus/ibft/pos.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

Please take a closer look at this one

command/genesis/params.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

command/genesis/genesis.go
0xAleksaOpacic: fixed a91c866e0f6d742a0cc4af63f07196b4703ec96a

e2e/pos_test.go
zivkovicmilos: Resolving as we don't have a normal testing framework, but some clunky mechanism that just sends transactions. This fix is out of the scope of this PR

command/genesis/params.go
zivkovicmilos: Learn something new everyday 👀

e2e/framework/config.go
zivkovicmilos: Please add spacing 🙏

helper/staking/staking.go
0xAleksaOpacic: Refactor 081af15b3e97e5e7ea3c38ee6b9c619af6ce2673

e2e/framework/config.go
0xAleksaOpacic: Fixed 5a4af913579b3d51492802cfd59fb968b6865b8e

helper/staking/staking.go
0xAleksaOpacic: I think the code looks much cleaner now. Fixed 79db141a49442b1de0118c5fb7c588d0e7775918

consensus/ibft/pos.go
0xAleksaOpacic: @zivkovicmilos changed to pointers 90c1e58112a6e4841679d805bfa77c5f51eb820e

zivkovicmilos (approved):
Thank you for making the requested changes 🍻  

Please add the check that I left on a recent comment, just so we're on the safe side, and please resolve the linting errors 🙏

0xAleksaOpacic:
> I think it's wise to add a check here if the prefix path is even set, because you can overwrite it with an empty set if it isn't. `areValidatorsSetByPrefix` is a helper we can use

@zivkovicmilos Agreed, fixed fc4d34fb2de3b2131c740450404b9de557f90853

Kourin1996 (approved):
Looks good to me. IBFT Switch works. Thank you for the fix

PR #399, opened 2022-01-28
Fix up the block timestamp to be read in seconds

### LINKED ISSUE #398: Header.Timestamp as ms causes invalid response for eth_getBlockByNumber
### Description

It looks like yesterday's change (https://github.com/0xPolygon/polygon-edge/pull/384) which modified Header.Timestamp to store the block epoch as ms instead of seconds introduces a regression bug.

Our testnet uses Blockcout as our explorer which breaks as it assumes the unix timestamp to be in seconds, not milliseconds. Querying `https://polygon-rpc.com` one can see the timestamp is in seconds as well.

On our own testnet, calls to previous blocks:

```
{"id":1,"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["2705407", true]}
```
resulted in a response with:
```
"timestamp": "0x61f439fb",
```
but calls to newer blocks:
```
{"id":1,"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["2705408", true]}
```
result in a response with:
```
"timestamp": "0x17ea246ec7b",
```

### Your environment

* OS and version `Ubuntu 20`
* version of the Polygon Edge `d56573f (0.2.0)`

### Steps to reproduce

1. Pull in latest changes
2. Observe JSON-RPC calls that return block objects to now show timestamp in milliseconds instead of seconds

### Expected behaviour

JSON-RPC result.timestamp should be a hex value in seconds.

### Actual behaviour

JSON-RPC result.timestamp is hex value in milliseconds.

### Proposed solution

Ensure Header.Timestamp is saved as seconds - OR ensure any JSON-RPC call converts to seconds.

zivkovicmilos:
# Description

Fixes EDGE-380

This PR patches a bug that treats the block timestamp in milliseconds instead of seconds

# Changes include

# Checklist

- [x] I have tested this code

# Additional comments

Relevant docs PR:
https://github.com/0xPolygon/polygon-edge-docs/pull/64

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -646,7 +646,7 @@ func (b *Blockchain) WriteBlock(block *types.Block) error {
 
 	if prevHeader, ok := b.GetHeaderByNumber(header.Number - 1); ok {
 		diff := header.Timestamp - prevHeader.Timestamp
-		logArgs = append(logArgs, "generation_time_in_miliseconds", diff)
+		logArgs = append(logArgs, "generation_time_in_seconds", diff)
 	}
 
 	b.logger.Info("new block", logArgs...)
--- a/command/helper/config.go
+++ b/command/helper/config.go
@@ -35,7 +35,7 @@ type Config struct {
 	Join           string                 `json:"join_addr"`
 	Consensus      map[string]interface{} `json:"consensus"`
 	RestoreFile    string                 `json:"restore_file"`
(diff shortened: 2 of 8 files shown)


lazartravica (changes requested):
Why don't we just change the resposne format and not the entire feature of configuring block time in ms itself?

Kourin1996 (approved):

lazartravica (approved):

PR #405, opened 2022-01-31
Allow longer domain suffix for server --dns value.

akshar:
# Description

When testing with valid DNS suffixes longer than 6 chars (for E.g `.technology`) the server was throwing an error `Invalid DNS ` due to the regex restricting dns suffix between 2-6 chars. 

This PR still enforces min length of 2 chars and uncaps the max length.

# Changes include

# Checklist

- [x] I have tested this code

### DIFF
--- a/helper/flags/flags_test.go
+++ b/helper/flags/flags_test.go
@@ -91,6 +91,13 @@ func TestMultiAddrFromDns(t *testing.T) {
 			err:        true,
 			outcome:    "",
 		},
+		{
+			name:       "valid long domain suffix",
+			dnsAddress: "dns/validator-1.foo.technology",
+			port:       12345,
+			err:        false,
+			outcome:    "/dns/validator-1.foo.technology/tcp/12345",
+		},
 	}
 
 	for _, tt := range tests {
--- a/helper/flags/flags_test.go
+++ b/helper/flags/flags_test.go
@@ -1,12 +1,15 @@
 package flags
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


akshar:

0xAleksaOpacic (approved):
Thank you for doing this PR.

ZeljkoBenovic (commented):
LGTM 
Thank you for this PR 🥇

Kourin1996 (approved):
Looks good to me 👍 Thank you for your PR

dbrajovic (approved):

helper/flags/flags_test.go
```diff
@@ -91,6 +91,13 @@ func TestMultiAddrFromDns(t *testing.T) {
 			err:        true,
 			outcome:    "",
 		},
+		{
+			name:       "valid long domain suffix",
+			dnsAddress: "dns/validator-1.foo.technology",
+			port:       12345,
```
zivkovicmilos: I know it's a nitpick, but could you please extract this `12345` port value into a common value within the test? 🙏 

We use it for multiple test cases, and it should not be a magic number here.

zivkovicmilos (approved):
Thank you for opening up this PR and for this bug fix 🙏 

Looks great 💯

helper/flags/flags_test.go
akshar: https://github.com/0xPolygon/polygon-edge/pull/405/commits/a92059866c233dddf41b4a8c622fce235e960a15
Resolved here

PR #412, opened 2022-02-09
[Bridge] Signed Arbitrary Message Pool

Kourin1996:
Fixes EDGE-42

# Description

This PR adds Signed Arbitrary Message Pool

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/bridge/sam/pool.go
+++ b/bridge/sam/pool.go
@@ -79,9 +79,6 @@ func (p *pool) ConsumeMessage(hash types.Hash) {
 // knows returns the flag indicating the message is known
 func (p *pool) isMessageKnown(hash types.Hash) bool {
 	_, ok := p.messageMap.Load(hash)
-	if !ok {
-		return false
-	}
 
 	return ok
 }
@@ -137,7 +134,7 @@ func (p *pool) UpdateValidatorSet(validators []types.Address, threshold uint64) 
 	defer p.changeValidatorsLock.Unlock()
 
 	oldValidators := p.validators
-	oldThreshold := p.signatureThreshold //nolint
+	oldThreshold := p.signatureThreshold //nolint:ifshort
 
 	p.validators = validators
 	atomic.StoreUint64(&p.signatureThreshold, threshold)
@@ -228,8 +225,6 @@ func (p *pool) getMessageBody(hash types.Hash) []byte {
 
 	body, ok := raw.([]byte)
 	if !ok {
-		// should not reach
-
 		return nil
 	}
 
--- a/bridge/sam/pool.go
+++ b/bridge/sam/pool.go
@@ -134,7 +134,7 @@ func (p *pool) UpdateValidatorSet(validators []types.Address, threshold uint64) 
 	defer p.changeValidatorsLock.Unlock()
 
 	oldValidators := p.validators
-	oldThreshold := p.signatureThreshold //nolint:ifshort //nolint:nolintlint
+	oldThreshold := p.signatureThreshold //nolint:nolintlint //nolint:ifshort
 
 	p.validators = validators
 	atomic.StoreUint64(&p.signatureThreshold, threshold)
--- a/bridge/sam/pool.go
+++ b/bridge/sam/pool.go
@@ -134,7 +134,7 @@ func (p *pool) UpdateValidatorSet(validators []types.Address, threshold uint64) 
 	defer p.changeValidatorsLock.Unlock()
 
 	oldValidators := p.validators
-	oldThreshold := p.signatureThreshold //nolint:ifshort
+	oldThreshold := p.signatureThreshold //nolint:ifshort //nolint:nolintlint
 
 	p.validators = validators
 	atomic.StoreUint64(&p.signatureThreshold, threshold)
--- a/bridge/sam/pool.go
+++ b/bridge/sam/pool.go
@@ -139,16 +139,16 @@ func (p *pool) UpdateValidatorSet(validators []types.Address, threshold uint64) 
 	p.validators = validators
 	atomic.StoreUint64(&p.signatureThreshold, threshold)
 
-	var maybeDemotableHashes []types.Hash
+	var maybeDemotedHashes []types.Hash
 	if removed := common.DiffAddresses(oldValidators, validators); len(removed) > 0 {
-		maybeDemotableHashes = p.messageSignatures.RemoveSignatures(removed)
+		maybeDemotedHashes = p.messageSignatures.RemoveSignatures(removed)
 	}
 
 	if oldThreshold != threshold {
 		// we need to check all messages if threshold changes
 		p.tryToPromoteAndDemoteAll()
-	} else if len(maybeDemotableHashes) > 0 {
-		for _, hash := range maybeDemotableHashes {
+	} else if len(maybeDemotedHashes) > 0 {
+		for _, hash := range maybeDemotedHashes {
 			p.tryToDemote(hash)
 		}
 	}
@@ -350,7 +350,7 @@ func (m *messageSignaturesStore) RemoveMessage(hash types.Hash) bool {
 
 // RemoveMessage removes the signatures by given addresses from all messages
 func (m *messageSignaturesStore) RemoveSignatures(addresses []types.Address) []types.Hash {
-	maybeDemotableHashes := make([]types.Hash, 0)
+	maybeDemotedHashes := make([]types.Hash, 0)
 
 	m.RangeMessages(func(entry *signedMessageEntry) bool {
 		count := 0
@@ -362,11 +362,11 @@ func (m *messageSignaturesStore) RemoveSignatures(addresses []types.Address) []t
 		}
 
 		if count > 0 {
-			maybeDemotableHashes = append(maybeDemotableHashes, entry.Hash)
+			maybeDemotedHashes = append(maybeDemotedHashes, entry.Hash)
 		}
 
 		return true
 	})
 
-	return maybeDemotableHashes
+	return maybeDemotedHashes
 }
--- a/bridge/sam/pool.go
+++ b/bridge/sam/pool.go
@@ -1,20 +1,20 @@
 package sam
 
 import (
-	"fmt"
 	"sync"
 	"sync/atomic"
 
+	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
 type pool struct {
-	// write-lock is called only when changing validators process
+	// write-lock is called only during the validator change process
 	// otherwise read-lock is called
-	// Changing validators will occur the most rarely (once per epoch)
+	// Changing validators occurs rarely (once per epoch)
 	changeValidatorsLock sync.RWMutex
 	validators           []types.Address
-	threshold            uint64 // required number of signatures for ready
+	signatureThreshold   uint64 // required number of signatures for ready
 
 	// storage for message body
 	messageMap sync.Map // types.Hash -> []byte
@@ -27,29 +27,11 @@ type pool struct {
 	messageSignatures *messageSignaturesStore
 }
 
-// diffAddresses returns a list of the addresses that are in arr1 but not in arr2
-func diffAddresses(arr1, arr2 []types.Address) []types.Address {
-	arr2Map := make(map[types.Address]bool)
-	for _, addr := range arr2 {
-		arr2Map[addr] = true
-	}
-
-	diff := make([]types.Address, 0)
-
-	for _, addr := range arr1 {
-		if !arr2Map[addr] {
-			diff = append(diff, addr)
-		}
-	}
-
-	return diff
-}
-
 func NewPool(validators []types.Address, threshold uint64) Pool {
 	return &pool{
 		changeValidatorsLock: sync.RWMutex{},
 		validators:           validators,
-		threshold:            threshold,
+		signatureThreshold:   threshold,
 		messageMap:           sync.Map{},
 		consumedMap:          sync.Map{},
 		readyMap:             sync.Map{},
@@ -77,7 +59,7 @@ func (p *pool) AddSignature(signature *MessageSignature) {
 	defer p.changeValidatorsLock.RUnlock()
 
 	if p.hasConsumed(signature.Hash) {
-		// we do no longer put the signature if the message has been consumed
+		// we no longer put the signature if the message has been consumed
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type pool struct {
+	// write-lock is called only when changing validators process
+	// otherwise read-lock is called
+	// Changing validators will occur the most rarely (once per epoch)
```
zivkovicmilos: We can rephrase this to `write-lock is called only during the validator change process`.

Also, we can rephrase this part as well to `Changing validators occurs rarely (once per epoch)`

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	// map from message Hash -> bool for message flags
+	consumedMap sync.Map
+	readyMap    sync.Map
+
+	// signatures
+	messageSignatures *messageSignaturesStore
+}
+
+// diffAddresses returns a list of the addresses that are in arr1 but not in arr2
+func diffAddresses(arr1, arr2 []types.Address) []types.Address {
```
zivkovicmilos: Nitpick, please consider changing the naming of these `arr1` and `arr2` 🙏

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	defer p.changeValidatorsLock.RUnlock()
+
+	if p.hasConsumed(message.Hash) {
+		// we do no longer put the signature if the message has been consumed
+		return
+	}
+
+	p.setMessageBody(message.Hash, message.Body)
+	p.tryToPromote(message.Hash)
+}
```
zivkovicmilos: This block of code is exactly the same as `AddSignature`, apart from the `p.setMessageBody` here and `p.messageSignatures.Put` in `AddSignature`.

What do you think about extracting this common logic?

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+
+// Add adds new message with the signature to pool
+func (p *pool) AddSignature(signature *MessageSignature) {
+	p.changeValidatorsLock.RLock()
+	defer p.changeValidatorsLock.RUnlock()
+
+	if p.hasConsumed(signature.Hash) {
+		// we do no longer put the signature if the message has been consumed
```
zivkovicmilos: Typo here, should be `we no longer put the signature....`

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+		return
+	}
+
+	p.messageSignatures.Put(signature)
+	p.tryToPromote(signature.Hash)
+}
+
+// Consume sets the consumed flag and delete the message from pool
+func (p *pool) Consume(hash types.Hash) {
```
zivkovicmilos: Consume what?

This method name should be more descriptive

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	p.consumedMap.Store(hash, true)
+
+	p.messageSignatures.RemoveMessage(hash)
+	p.messageMap.Delete(hash)
+	p.readyMap.Delete(hash)
+}
+
+// knows returns the flag indicating the message is known
+func (p *pool) knows(hash types.Hash) bool {
```
zivkovicmilos: `knows` is an abstract method name, please consider renaming it to something like `isMessageKnown`

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// UpdateValidatorSet update validators and threshold
+// This process blocks other processes because messages would lose the signatures
+func (p *pool) UpdateValidatorSet(validators []types.Address, threshold uint64) {
+	p.changeValidatorsLock.Lock()
+	defer p.changeValidatorsLock.Unlock()
+
+	oldValidators := p.validators
+	oldThreshold := p.threshold //nolint
```
zivkovicmilos: Why the `//nolint`?

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type pool struct {
+	// write-lock is called only when changing validators process
+	// otherwise read-lock is called
+	// Changing validators will occur the most rarely (once per epoch)
+	changeValidatorsLock sync.RWMutex
+	validators           []types.Address
+	threshold            uint64 // required number of signatures for ready
```
zivkovicmilos: Please consider renaming this to `signatureThreshold`, as it's not immediately clear from the usage what it represents

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	diff := make([]types.Address, 0)
+
+	for _, addr := range arr1 {
+		if !arr2Map[addr] {
+			diff = append(diff, addr)
+		}
+	}
+
+	return diff
+}
```
zivkovicmilos: What do you think about extracting this method to the helper / common package?

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+
+// canDemote return the flag indicating it's possible to change status to pending
+func (p *pool) canDemote(hash types.Hash) bool {
+	return !p.canPromote(hash)
+}
+
+// tryToPromote checks the number of signatures and threshold and update message status to ready if need
+func (p *pool) tryToPromote(hash types.Hash) {
+	fmt.Printf("tryToPromote %v\n", hash)
```
zivkovicmilos: Leftover print

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+	if newNumSignatures < 0 {
+		return 0
+	}
+
+	return uint64(newNumSignatures)
+}
+
+// IncrementNumSignatures decrements SignatureCount and return new count
+func (e *signedMessageEntry) DecrementNumSignatures() uint64 {
+	newNumSignatures := atomic.AddInt64(&e.SignatureCount, ^int64(0))
```
zivkovicmilos: Why decrement with `^int64(0)` instead of just `-1`?

bridge/sam/pool.go
```diff
@@ -0,0 +1,399 @@
...
+
+	return uint64(count)
+}
+
+// IncrementNumSignatures increments SignatureCount and return new count
+func (e *signedMessageEntry) IncrementNumSignatures() uint64 {
+	newNumSignatures := atomic.AddInt64(&e.SignatureCount, 1)
+	if newNumSignatures < 0 {
+		return 0
+	}
```
zivkovicmilos: Is this even possible?

zivkovicmilos (approved):
Looks good to me 💯 

I've left some comments you need to address.
Please resolve the linting errors 🙏 

I wanted to raise a discussion regarding persistence in this module, as far as I can see we keep these values in memory and don't persist them anywhere. Do you think we even need to bother with persisting the info stored in the SAM pool, as the messages themselves are temporary?

cc @lazartravica

bridge/sam/pool.go
Kourin1996: Actually it's necessary. It throws error because this declaration is far from for-loop which use this variable. But need to set here

bridge/sam/pool.go
Kourin1996: `minuend` and `subtrahend`? They are not clear

bridge/sam/pool.go
Kourin1996: I don't think we can extract common logic.

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Moved at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Removed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
Kourin1996: Fixed at c92a67bf10bc73bcdf8929dca389a8e53020bf04

bridge/sam/pool.go
zivkovicmilos: Something like `setA` and `setB`

bridge/sam/pool.go
zivkovicmilos: Why not?

This entire block of code is identical:
```go
p.changeValidatorsLock.RLock()
	defer p.changeValidatorsLock.RUnlock()

	if p.hasConsumed(message.Hash) {
		// we do no longer put the signature if the message has been consumed
		return
	}
```

bridge/sam/pool.go
zivkovicmilos: Aha, got it.

What's the linter that fails? if-short?

bridge/sam/pool.go
Kourin1996: p.changeValidatorsLock.RUnlock() will be called the end of `AddMessage` and `AddSignature`. Should I create function that takes callback function? I guess it makes the flow unclear. I thought didn't need to save a few lines

bridge/sam/pool.go
Kourin1996: ifshort

bridge/sam/pool.go
zivkovicmilos: Got it, let's not bundle it, it's not breaking 👍

bridge/sam/pool.go
zivkovicmilos: It's some bug with the ifshort linter itself, you can then change the `nolint` to `nolint:ifshort` 👍

bridge/sam/pool.go
```diff
@@ -0,0 +1,377 @@
...
+		for _, hash := range maybeDemotableHashes {
+			p.tryToDemote(hash)
+		}
+	}
+}
+
+// canPromote return the flag indicating it's possible to change status to ready
+// message need to have enough signatures and be known by pool for promotion
+func (p *pool) canPromote(hash types.Hash) bool {
+	isKnown := p.isMessageKnown(hash)
```
lazartravica, line 160: Why do we need to check both whether the message is known and if it has enough signatures?

Wouldn't it be enough just to check the signatures?

lazartravica (changes requested):
LGTM! Left one minor comment I would appreciate clarification on :)

bridge/sam/pool.go
Kourin1996: To pick events up, you need to have data and enough signatures. Because expecting other validators to send hash and its signature by gossip. When validating, you can just check signature count.

bridge/sam/pool.go
```diff
@@ -0,0 +1,377 @@
...
+
+// knows returns the flag indicating the message is known
+func (p *pool) isMessageKnown(hash types.Hash) bool {
+	_, ok := p.messageMap.Load(hash)
+	if !ok {
+		return false
+	}
+
+	return ok
+}
```
dbrajovic: Why not just return `ok` ?

bridge/sam/pool.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	return body
+}
+
+// signedMessageEntry is representing the data stored in messageSignaturesStore
+type signedMessageEntry struct {
+	Hash           types.Hash
+	Signatures     sync.Map
+	SignatureCount int64
```
dbrajovic, line 243: Perhaps this field can be of type `uint64` - this would remove the `uint64()` casting of the return values in `NumSignatures`, `IncrementNumSignatures` and `DecrementNumSignatures`

bridge/sam/pool.go
Kourin1996: Good catch. Fixed at 95405e5209abfed5ccbad58db15f2550f6d04bf5

bridge/sam/pool.go
Kourin1996: Added 95405e5209abfed5ccbad58db15f2550f6d04bf5

bridge/sam/pool.go
Kourin1996: To be honest, no big difference. But I renamed 95405e5209abfed5ccbad58db15f2550f6d04bf5

zivkovicmilos (approved):

bridge/sam/pool.go
lazartravica: Thank you for the clarification, resolving this comment @Kourin1996

bridge/sam/pool_test.go
```diff
@@ -0,0 +1,617 @@
...
+		threshold      uint64
+		knownMessages  []Message
+		consumedHashes []types.Hash
+		// inputs
+		signatures []MessageSignature
+		// outputs
+		numReadyMessages int
+	}{
+		{
+			name: "shouldn't promote message if the message doesn't have enough message",
```
0xAleksaOpacic: Typo 
....doesn't have enough signatures *

bridge/sam/pool.go
```diff
@@ -0,0 +1,372 @@
...
+	p.changeValidatorsLock.Lock()
+	defer p.changeValidatorsLock.Unlock()
+
+	oldValidators := p.validators
+	oldThreshold := p.signatureThreshold //nolint:nolintlint //nolint:ifshort
+
+	p.validators = validators
+	atomic.StoreUint64(&p.signatureThreshold, threshold)
+
+	var maybeDemotableHashes []types.Hash
```
0xAleksaOpacic: typo
Demotable -> Demoted

0xAleksaOpacic (approved):
LGTM!  Thank you on a great code coverage.

bridge/sam/pool_test.go
Kourin1996: Thanks. I fixed at 66dd51ca6a17686e116cf8a384877f4fce953dee

bridge/sam/pool.go
Kourin1996: Thanks. I fixed at 66dd51ca6a17686e116cf8a384877f4fce953dee

0xAleksaOpacic (approved):

dbrajovic (approved):

PR #414, opened 2022-02-10
Allow wildcard for allow control allowed origins

thegoldenmule:
# Description

The new `--access-control-allow-origins` flag did not function correctly with wildcards. This means that tech that didn't explicitly specify an `Origin` header on requests to the JSON RPC API would be blocked. This includes standard tech like MetaMask.

# Changes include

- [ x ] Bugfix (non-breaking change that solves an issue)

## Testing

- [ x ] I have tested this code manually

### Manual tests

I was able to do various actions via MetaMask and explore the blockchain via typical `eth_` functions.

Fixes EDGE-436


thegoldenmule:

jsonrpc/jsonrpc.go
```diff
@@ -112,7 +112,7 @@ func middlewareFactory(config *Config) func(http.Handler) http.Handler {
 			origin := r.Header.Get("Origin")
 
 			for _, allowedOrigin := range config.AccessControlAllowOrigin {
-				if allowedOrigin == origin {
+				if allowedOrigin == origin || allowedOrigin == "*" {
```
Kourin1996: I would avoid loop when a wildcard is given

Kourin1996 (commented):
I appreciate your PR. I've left a small comment. Would be great if you could check it. Thank you!

jsonrpc/jsonrpc.go
STZeed: Thank you

thegoldenmule: I need to loop either way to see if `*` is in the array. However, I can optimize a tad.

thegoldenmule:
@Kourin1996 Since `config.AccessControlAllowOrigin` is an array, I still need to loop to see if `*` is among the values provided. There is also some ambiguity here in that it's not clear what _should_ happen if a user passes a wildcard AND a specific domain. I considered three general cases:

```
--access-control-allow-origins "*"
```

Here the header will be set directly to `*`.

```
--access-control-allow-origins "*,https://mydomain.com"
```

Here the header will be set to `*` regardless.

```
--access-control-allow-origins "https://mydomain.com,*"
```

Here the header will be set to `https://mydomain.com` if the origin is mydomain.com, otherwise `*`.

Kourin1996:
@thegoldenmule Thank you for the comments. That's true. In current implementation, multiple domains that contains wildcard can be given. So my comment at that line was not correct actually. We need to talk about whether we can give domains with wildcard.
cc: @zivkovicmilos @lazartravica @PtitLuca

thegoldenmule:
A simple alternative, btw, would be to skip the array and loop entirely and simply set the string directly to whatever was specified in the configuration.

zivkovicmilos:
@Kourin1996 @PtitLuca @lazartravica @thegoldenmule 

What do you think about having `*` be the default value for this specific CORS configuration (`Allow-Origins`)?

It would mitigate this additional code complication, as you'd assume if anything specific was provided for allowed origins, you'd need to check against it, otherwise it's all allowed because the wildcard (`*`) would be the default.

Kourin1996:
@zivkovicmilos Not sure setting wildcard as default is good. Wildcard will allow all web browsers to access to JSON-RPC. (Of course, you can block by different layer)

thegoldenmule:
@zivkovicmilos @Kourin1996 IMO `*` by default makes the most sense. I can't tell you what the goals of your project are-- but as I see it, an opensource blockchain server that by default doesn't work with standard web3 clients (like Metamask) doesn't make a ton of sense.

STZeed:
I was thinking more along the lines of magic SDK for contributing to
charitable organizations. For the protection of both parties involved.
Sincerity is priority number one. People don't bake projects that cant is
protected. Always open to Ideas.

thegoldenmule:
@STZeed I'm not sure I'm following but if any offense was taken, it was certainly not intended. My point is only that this project provides an open API that, by default, does not work in the larger open ecosystem. It's not CORS that is providing the security guarantees behind dapps, though I understand you might as well use every tool your disposal.

As far as ideas, I have already mentioned one that may have been passed over. Why not simply set the header explicitly to the value provided by `--access-control-allow-origins` rather than any sort of dynamic logic? This option would result in the simplest implementation, the easiest documentation to understand, and cover a large number of use cases.

Kourin1996:
@thegoldenmule @STZeed Thank you for having opinions and apologize for late reply. Basically I'm not clearly disagree with that. Perhaps there will be few cases of that we want to disable CORS. It will work with allowing CORS to any domains for now.

thegoldenmule:
@Kourin1996 Please let me know how you'd like me to proceed.

Kourin1996:
I think you will need to do the following things:
1. Set '*' as the default CORS domain
2. In case of '*', set '*' in "Access-Control-Allow-Origin HTTP/S header

@thegoldenmule PR doesn't have 1 now. Thank you

Kourin1996:
A simplest way is to add default value here
https://github.com/0xPolygon/polygon-edge/blob/8a033aa1afb191abdac04636d318f83f32511f3c/command/helper/helper.go#L479

thegoldenmule:
@Kourin1996 That file seems to have gone through a large refactor recently. I changed it in `command/server/config.go`, as it looked like the right place.

ZeljkoBenovic (approved):
LGTM 🥇 
Tested everything. It works as intended.

Just one observation: 
If there is a need to allow multiple domains with ``` --access-control-allow-origins``` flag, proper usage is setting multiple flags in same command instead of setting multiple domains separated by comma.
This will work:
```
polygon-edge server  --access-control-allow-origins https://example.com --access-control-allow-origins https://edge-docs.polygon.technology
```
This will not work:
```
polygon-edge server  --access-control-allow-origins "https://example.com,https://edge-docs.polygon.technology"
```

We will publish the docs with proper usage for this flag soon.
Just fix these linter errors, and we're good to go.

Great job 💯

Kourin1996 (approved):

zivkovicmilos (changes requested):
Looks great 💯 

Thank you for the contribution 🙏 

Please check out Zeljko's [comment](https://github.com/0xPolygon/polygon-edge/pull/414#pullrequestreview-902335772) and resolve the linting errors, after that we're good to go 🚀

thegoldenmule:
@zivkovicmilos @ZeljkoBenovic Thanks so much. I _think_ I've fixed the linter issues. I have not used `golangci-lint` before, but `make lint` seems to complete successfully locally for me.

zivkovicmilos (approved):
Thank you for resolving the linting errors, and for the contribution 🙏 

Merging this now 🚀

PR #417, opened 2022-02-14
[JSONRPC] Fix websocket subscription

dbrajovic:
Fixes EDGE-387

# Description

This PR fixes the issue of websocket subscriptions becoming unresponsive after a predefined timeout. 

# Changes include

# Checklist

## Testing

### Manual tests

Use the `websocat` tool to verify the subscription is sending responses:
1. Start a 4-node cluster
2. Connect to a node using `websocat [node_ws_url]`
3. Send the request (in `websocat`):
`{"id": 1, "method": "eth_subscribe", "params": ["newHeads"]}`
4. Observe the responses are being received (subscription is not becoming silent)


0xAleksaOpacic (approved):
Great job. Thank you manual testing instructions

jsonrpc/filter_manager_test.go
```diff
@@ -178,6 +178,29 @@ func TestFilterWebsocket(t *testing.T) {
...
 }
 
+// 	This test verifies that websocket subscriptions
+//	are not susceptible to timeouts (like filters).
+func TestWSResponsiveness(t *testing.T) {
+	store := newMockStore()
+	fm := NewFilterManager(hclog.NewNullLogger(), store)
+
+	//	set the timeout to 1s, so it starts early
+	fm.timeout = 1 * time.Second
```
zivkovicmilos: Why do you need to set the FilterManager timeout here, if you completely ignore it in the implementation (it's not going to be read anyways)?

jsonrpc/filter_manager_test.go
```diff
@@ -178,6 +178,29 @@ func TestFilterWebsocket(t *testing.T) {
...
+
+	//	create block filter (as subscription)
+	mockWS := &mockWsConn{msgCh: make(chan []byte, 1)}
+	filterID := fm.NewBlockFilter(mockWS)
+
+	//	start the manager
+	go fm.Run()
+
+	//	wait for timeout to
+	time.Sleep(2 * time.Second)
```
zivkovicmilos: I don't think this is a great way to test this functionality.

You are in complete control of the testing environment here, and you know the implementation will never set up a timeout for you, but will just handle requests if it's a WS connection.

If you want to add this test, if at all, I think you could probably refactor it to use the `WriteMessage` method on `mockWs`, just send out a huge batch of messages and make sure they get delivered 

cc @Kourin1996

zivkovicmilos (approved):
Looks good, thank you for hunting down this bug 🙏 

I've left some comments regarding the test that we should discuss 👍

jsonrpc/filter_manager_test.go
dbrajovic: The original test was written to confirm the filter (websocket) was getting uninstalled after a timeout (which is why I set it to 1 second in the test). This test was meant to confirm there **is** a problem. 

So after the fix, how do you verify that the problem is gone ? Conceptually, your batch solution is no different than the existing [TestFilterWebscoket](https://github.com/0xPolygon/polygon-edge/blob/cd02189040f33ed77f68c54d72a50bc479f061ca/jsonrpc/filter_manager_test.go#L147) - it's just a difference of 1 vs 100 responses sent. 

In that case, we should just increase the payload in that test and remove the new one altogether.

zivkovicmilos: That sounds like a plan to me. I would honestly be ok with dropping this test entirely 👍

jsonrpc/filter_manager_test.go
zivkovicmilos: Closing because discussion was continued on 

https://github.com/0xPolygon/polygon-edge/pull/417#discussion_r805748284

Kourin1996 (approved):
Looks good to me

jsonrpc/filter_manager_test.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/417/commits/354d1cabaebe78aa5615808db28328f9fede66ac)

lazartravica (approved):

PR #420, opened 2022-02-14
Fix faulty check in estimateGas

zivkovicmilos:
# Description

This PR fixes a few faulty checks inside `eth_estimateGas`, namely:
* If the transaction has a value of `0` and the account has a balance of `0`, this shouldn't be a reason to fail
* EVM application errors were not being taken into account during transaction execution

# Changes include

# Checklist

## Testing

#### Account balance fix
An easy way to test this out would be to send transactions from an account that has 0 balance, and the value of the transaction being 0.

#### EVM reverts fix
To test out the EVM revert error fix during `eth_estimateGas`, a dummy SC can be deployed with a method that always fails (for example, it has a `require` statement that is always triggered), and `eth_estimateGas` called on a transaction that calls that method - the call should fail, and not return a gas estimate value.

# Additional comments

Fixed EDGE-403

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -583,11 +583,22 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 		return false, nil
 	}
 
+	isGasError := func(err error) bool {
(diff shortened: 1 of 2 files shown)


zivkovicmilos:
Closing in favor of https://github.com/0xPolygon/polygon-edge/pull/422

PR #421, opened 2022-02-15
Fix minor test wait bug

zivkovicmilos:
# Description

This PR fixes a minor bug in `TestGetTxs`, where there wasn't a proper event wait before checking the state of the TxPool, which would cause a race when asserting. 

# Changes include

# Checklist

## Testing

# Additional Comments

Fixes EDGE-404

### DIFF
--- a/txpool/txpool_test.go
+++ b/txpool/txpool_test.go
@@ -1284,9 +1284,9 @@ func TestResetAccounts_Enqueued(t *testing.T) {
 
 		// setup prestate
 		totalTx := 0
-		totalPromoted := uint64(0)
+		expectedPromoted := uint64(0)
 		for addr, txs := range allTxs {
-			totalPromoted += expected.accounts[addr].promoted
+			expectedPromoted += expected.accounts[addr].promoted
 			for _, tx := range txs {
 				totalTx++
 				go func(tx *types.Transaction) {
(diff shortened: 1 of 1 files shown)


txpool/txpool_test.go
```diff
@@ -1897,14 +1905,24 @@ func TestGetTxs(t *testing.T) {
...
 			ctx, cancelFn := context.WithTimeout(context.Background(), time.Second*10)
 			defer cancelFn()
 
-			// All txns should get added
-			assert.Len(t, waitForEvents(ctx, subscription, totalTx), totalTx)
+			// Wait for promoted transactions
+			assert.Len(t, waitForEvents(ctx, promoteSubscription, totalPromotedTx), totalPromotedTx)
+
+			// Wait for enqueued transactions, if any are present
+			totalEnqueuedTx := totalPromotedTx - len(test.allTxs)
```
Kourin1996: Nicpick, `totalPromotedTx` is just an expected value, right?

Kourin1996 (approved):
Looks good to me. I've left small comments.

txpool/txpool_test.go
zivkovicmilos: Yes, it's the total number of transactions that should be promoted within the test case, so our `waitForEvents` call knows for how many events it should wait for

zivkovicmilos: The transaction can either be promoted or enqueued in this specific test, so there is no need to keep track of enqueued transactions when we already keep track of promoted ones - if a transaction is not promoted, then it is enqueued

dbrajovic (approved):

txpool/txpool_test.go
zivkovicmilos: Fixed, in all places 👍 

[ccb1564](https://github.com/0xPolygon/polygon-edge/pull/421/commits/ccb1564169770421faa05b77fa9eb47abc4060a8)

PR #422, opened 2022-02-15
Fix faulty check in gas estimation and provide reason for REVERT calls

zivkovicmilos:
# Description

The PR has been expanded from https://github.com/0xPolygon/polygon-edge/pull/420

This PR fixes a few faulty checks inside `eth_estimateGas`, namely:
* If the transaction has a value of `0` and the account has a balance of `0`, this shouldn't be a reason to fail
* EVM application errors were not being taken into account during transaction execution

Additionally, this PR also specifies a reason for `REVERT` calls to the EVM.

The PR also updates `go-web3` to the latest version, because it needed the latest feature to work:
https://github.com/umbracle/go-web3/pull/151

# Changes include

# Checklist

## Testing

#### Account balance fix
An easy way to test this out would be to send transactions from an account that has 0 balance, and the value of the transaction being 0.

#### EVM reverts fix
To test out the EVM revert error fix during `eth_estimateGas`, a dummy SC can be deployed with a method that always fails (for example, it has a `require` statement that is always triggered), and `eth_estimateGas` called on a transaction that calls that method - the call should fail, and not return a gas estimate value.

# Additional comments

Fixes EDGE-399
Fixes EDGE-428
Fixes EDGE-439


Kourin1996 (approved):
Looks good to me

dbrajovic (approved):

dbrajovic (approved):

jsonrpc/eth_endpoint.go
```diff
@@ -518,8 +527,11 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 	gasPriceInt := new(big.Int).Set(transaction.GasPrice)
 	valueInt := new(big.Int).Set(transaction.Value)
 
-	// If the sender address is present, recalculate the ceiling to his balance
-	if transaction.From != types.ZeroAddress && transaction.GasPrice != nil && gasPriceInt.BitLen() != 0 {
+	var availableBalance *big.Int
+
+	// If the sender address is present, figure out how much available funds
+	// are we working with
+	if transaction.From != types.ZeroAddress {
```
lazartravica, line 526: Good catch, but couldn't we make this code more efficient by checking this as early as possible as well:

```
transaction.GasPrice != nil && gasPriceInt.BitLen() != 0) || transaction.Value // this means no reading acc balance necessary
```
?

lazartravica (changes requested):
Looks good, but when paranoically testing that `Input` and `Data` changes are good, I stumbled upon a scenario in which it does not work correctly.

When sending this JSON-RPC call to 3 different nodes:
`{"method":"eth_estimateGas","params":[{"to":"0xd3CdA913deB6f67967B99D67aCDFa1712C293601","value":"0x15", "input":"0x3131"}],"id":1,"jsonrpc":"2.0"}`

I get 3 different responses:

Polygon Edge (develop):
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32600,"message":"not enough funds for transfer with given value, recoverable [true]"}}
```

☝️ not good, but at least the message is somewhat logical

Polygon Edge (this PR):
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32600,"message":"unable to apply transaction for gas cap 5242880"}}
```

⛔️ Gas estimate maxed out and hit the block gas limit, message not helpful or logical

Ganache:
```
{"id":1,"jsonrpc":"2.0","result":"0x5228"}
```

✅ Returns 21032, which is a good estimate (21000 is the exact what we can expect).


P.S. Unless gas cap means something else than block gas limit, please consider using the same terminology as we had before 👍

jsonrpc/eth_endpoint.go
```diff
@@ -535,79 +547,116 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
...
-		highEnd = types.GasCap.Uint64()
+	isGasApplyError := func(err error) bool {
+		return errors.As(err, &state.ErrNotEnoughIntrinsicGas)
 	}
 
-	gasCap = highEnd
+	isGasEVMError := func(err error) bool {
+		return errors.As(err, &runtime.ErrOutOfGas) ||
+			errors.As(err, &runtime.ErrCodeStoreOutOfGas)
+	}
```
Kourin1996: You didn't test with `errors.Is`?

jsonrpc/eth_state_test.go
```diff
@@ -580,94 +581,196 @@ func TestEth_State_GetStorageAt(t *testing.T) {
...
+		nil,
+	)
+
+	assert.Equal(t, 0, estimate)
+
+	// Make sure the EVM revert message is contained
+	assert.ErrorAs(t, estimateErr, &runtime.ErrExecutionReverted)
+
+	// Make sure the EVM revert reason is contained
+	assert.ErrorAs(t, estimateErr, &revertReason)
```
Kourin1996, line 742: You can use `ErrorIs` instead of `ErrorAs` except for this line (because revertReason is created ad-hoc)
Why you don't want to use `ErrorIs`?

jsonrpc/eth_endpoint.go
zivkovicmilos: I did, and these errors aren't caught because they are bundled from the lower levels, so `errors.Is` will never catch them

jsonrpc/eth_state_test.go
zivkovicmilos: ~Answered in https://github.com/0xPolygon/polygon-edge/pull/422#discussion_r818489040~

Kourin1996: Actually this comment is different from my previous comment. Here is about test code

zivkovicmilos:
@lazartravica 

I've looked into this scenario you've listed and I've come up with some conclusions regarding the way we are able to do gas estimations.

Firstly, the reason we dropped the `Input` field from the `eth_estimateGas` and `eth_call` params is because it's not needed ([it's not in the spec](https://eth.wiki/json-rpc/API#eth_call)) and it's definitely [not in the input params we parse from go-web3](https://github.com/umbracle/ethgo/blob/1621d9ff042b885412c7db1335f885b209349744/structs.go#L188-L195). I believe the go-web3 package actually does this bundling for us somewhere, so we don't have to worry about having both `Input` and `Data`, since only `Data` will ever get filled.

We cannot do gas estimations for queries that don't contain a `from` address, because of the way our executor layer is set up - it expects a valid address to be able to perform basic checks. Go-ethereum and other clients circumvent this, as far as I can see, by having special logic execute and checks omit if the message passed into the execution layer demands it (like for `eth_estimateGas`). These changes are out of the scope of this PR and should be tackled in the future.

I've improved the error output for the message we push out (that it's impossible to execute the transaction for the highest gas limit): 
[41c34e1](https://github.com/0xPolygon/polygon-edge/pull/422/commits/41c34e15c6e1af945205b339736e7afa66a97873)

In our context, gas cap === gas limit, so there is no difference between these two terms.

jsonrpc/eth_endpoint.go
zivkovicmilos: There's no point in checking if `transaction.GasPrice` or `transaction.Value` were set, as they will always have some value: 
* [transaction.Value](https://github.com/0xPolygon/polygon-edge/blob/41c34e15c6e1af945205b339736e7afa66a97873/jsonrpc/eth_endpoint.go#L958-L960)
* [transaction.GasPrice](https://github.com/0xPolygon/polygon-edge/blob/41c34e15c6e1af945205b339736e7afa66a97873/jsonrpc/eth_endpoint.go#L962-L964)

I'm not sure what you meant by having this check as early as possible?

jsonrpc/eth_state_test.go
zivkovicmilos: I've misread this, you're talking about the testing code 🤦 

I've changed the corresponding asserts to `assert.ErrorIs`:
[a72e6da](https://github.com/0xPolygon/polygon-edge/pull/422/commits/a72e6da81988f275d54d97e893008ffe8c6141bb)

jsonrpc/errors.go
```diff
@@ -90,3 +92,12 @@ func NewInternalError(msg string) *internalError {
 func NewSubscriptionNotFoundError(method string) *subscriptionNotFoundError {
 	return &subscriptionNotFoundError{fmt.Sprintf("subscribe method %s not found", method)}
 }
+
+func constructErrorFromRevert(result *runtime.ExecutionResult) error {
```
Aleksao998, line 96: Why are we not returning unpackErr

jsonrpc/errors.go
```diff
@@ -90,3 +92,12 @@ func NewInternalError(msg string) *internalError {
...
+	revertErrMsg, unpackErr := abi.UnpackRevertError(result.ReturnValue)
+	if unpackErr != nil {
+		return result.Err
```
0xAleksaOpacic, line 99: Just for clarification, you are calling this method only if the error is revert. If you get unpackErr what does that mean?

jsonrpc/types_test.go
```diff
@@ -90,7 +90,7 @@ func TestDecode_TxnArgs(t *testing.T) {
 		assert.NoError(t, json.Unmarshal(buffer.Bytes(), &r))
 
 		if !reflect.DeepEqual(r, c.res) {
-			t.Fatal("bad")
+			t.Fatal("Resulting data and expected values are not equal")
```
0xAleksaOpacic, line 93: Thank you for renaming this error. I've searched our code, there are multiple places with this error message which we should refactor ("bad")

jsonrpc/eth_endpoint.go
0xAleksaOpacic: ping @lazartravica to resolve this communication

jsonrpc/errors.go
zivkovicmilos: It's just [basic error handling](https://github.com/umbracle/ethgo/blob/50195055efb0377415550beca8c97b892edefdc8/abi/revert.go#L10-L23) from the `go-web3` package, this error doesn't have to be a string error, but can be bundled regardless

jsonrpc/types_test.go
zivkovicmilos: I think this is the 3rd time I've been saying this on multiple PRs, this sort of refactor is out of the scope for this PR and should be tackled in the future - I simply modified a test chunk of code I actually tested out and changed

lazartravica (approved):

jsonrpc/types_test.go
0xAleksaOpacic: agreed, we will tackle this some other time

0xAleksaOpacic (approved):
LGTM!

PR #425, opened 2022-02-18
[Feature] Adding loadbot ERC20 and ERC721 token transfer mode

### LINKED ISSUE #472: LoadBot fails with SIGSEGV
### Description

Attempt to perform stress testing using built-in tool "LoadBot" fails. 

### Your environment

* OS and version: **Fedora 35**
* version of the Polygon Edge: **v0.1.0**
* branch that causes this issue: **develop**

### Steps to reproduce
Open MetaMask, create an account. In my case the following account was created (possible to import for debugging):
```
helmet pool broom actual inflict enter bitter tourist hungry walk render mind #mnemonic
db22fc73512d3e40c57056f24b2a4fa6e8a0c8ff9bdd98d4f398fa4514fd0524              #private key
0xD45d663eC1368C04Af04BC18c3d8e19a4D5F1cDD                                    #address
```

Use the address in Genesis config as pre-mined account: [genesis_json.txt](https://github.com/0xPolygon/polygon-edge/files/8356225/genesis_json.txt) (rename the file to `genesis.json`).

Start Polygon Edge on a cloud instance:
```
polygon-edge server --data-dir ./data --chain genesis.json --libp2p 0.0.0.0:1478 --seal --nat
```

Open MetaMask again, add custom network pointing to the just launched node:
lenovo-fedora-20220326-214124-642592993

Now, we can see in MetaMask that our account indeed has pre-mined balance:
lenovo-fedora-20220326-214142-807062178

Run LoadBot following [stress testing guide](https://edge-docs.polygon.technology/docs/additional-features/stress-testing):

```bash
# Example
export LOADBOT_0xD45d663eC1368C04Af04BC18c3d8e19a4D5F1cDD=db22fc73512d3e40c57056f24b2a4fa6e8a0c8ff9bdd98d4f398fa4514fd0524

polygon-edge loadbot \
  --jsonrpc http://127.0.0.1:8545 \
  --grpc-address 127.0.0.1:9632 \
  --sender 0xD45d663eC1368C04Af04BC18c3d8e19a4D5F1cDD \
  --receiver 0xD45d663eC1368C04Af04BC18c3d8e19a4D5F1cDD \ #sending to ourselves
  --count 1000 \
  --value 0x100 \
  --tps 100
```

### Expected behaviour

Transactions get created and submitted, loading the network.

### Actual behaviour

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x14e8fe5]

goroutine 1 [running]:
github.com/0xPolygon/polygon-edge/command.(*CLIOutput).getCommandOutput(...)
        /tmp/polygon-edge/command/cli_output.go:31
github.com/0xPolygon/polygon-edge/command.(*CLIOutput).WriteOutput(0xc0004487a0)
        /tmp/polygon-edge/command/cli_output.go:23 +0xc5
panic(0x16b7ac0, 0x2bb2850)
        /usr/lib/golang/src/runtime/panic.go:965 +0x1b9
github.com/0xPolygon/polygon-edge/command/loadbot.(*Loadbot).Run(0xc0004487c0, 0x0, 0x0)
        /tmp/polygon-edge/command/loadbot/execution.go:165 +0x5ad
github.com/0xPolygon/polygon-edge/command/loadbot.runLoadbot(0xc000464be0, 0xc000464b00, 0xf, 0x0, 0x15)
        /tmp/polygon-edge/command/loadbot/loadbot_command.go:164 +0xa5
github.com/0xPolygon/polygon-edge/command/loadbot.runCommand(0xc000456f00, 0xc0001ac540, 0x0, 0xe)
        /tmp/polygon-edge/command/loadbot/loadbot_command.go:151 +0x20b
github.com/spf13/cobra.(*Command).execute(0xc000456f00, 0xc0001ac460, 0xe, 0xe, 0xc000456f00, 0xc0001ac460)
        /tmp/polygon-edge/vendor/github.com/spf13/cobra/command.go:860 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003c2a00, 0xc00045e280, 0xc00045e500, 0xc00045e780)
        /tmp/polygon-edge/vendor/github.com/spf13/cobra/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        /tmp/polygon-edge/vendor/github.com/spf13/cobra/command.go:902
github.com/0xPolygon/polygon-edge/command/root.(*RootCommand).Execute(0xc000010c68)
        /tmp/polygon-edge/command/root/root.go:58 +0x32
main.main()
        /tmp/polygon-edge/main.go:17 +0x4c
```

ZeljkoBenovic:
# Description

This PR adds the ERC20 and ERC721 token transactions feature to the loadbot.
It will deploy a new ERC20 or ERC721 SC to the network and make ERC20 or ERC721 token or mint transactions.
New ```--mode erc20``` and ```--mode erc721```  switches are implemented.
The loadbot output now shows gas related metrics ( gasLimit and gasUsed) and the derived block utilization.

Fixes EDGE-452

# Changes include

# Checklist

## Testing

### Manual tests

Pull branch and run the loadbot with new mode switch. For example:
``` go run main.go loadbot --mode erc20 --jsonrpc http://127.0.0.1:40001 --grpc-address 127.0.0.1:30001  --sender 0x228466F2C715CbEC05dEAbfAc040ce3619d7CF0B --receiver 0xca48694ebcB2548dF5030372BE4dAad694ef174e --count 50 --value 0x101 --tps 200```

# Documentation update
Updated docs https://github.com/0xPolygon/polygon-edge-docs/pull/68

# Additional comments
Before running the loadbot env vars with acc/keys need to be set:
``` export LOADBOT_<account address>=<account private key>```

### CODE REVIEW

File: command/loadbot/execution.go [hunk 1]
```diff
@@ -350,23 +398,45 @@ func (l *Loadbot) Run() error {
...
-				})
-				atomic.AddUint64(&l.metrics.FailedTransactionsCount, 1)
-
-				return
+			// run different transactions for different modes
+			if l.cfg.GeneratorMode == erc20 {
+				// Execute ERC20 Contract token transaction and report any errors
+				contractAddr := types.Address(l.metrics.ContractAddress)
+
+				txHash, err = l.executeTxn(grpcClient, "erc20Transfer", &contractAddr)
```
Kourin1996: Why not define a constant of "erc20Transfer"?

Kourin1996 (commented):
Can you add "erc20" in the description of `mode` argument in `loadbot` command?

command/loadbot/generators.go
```diff
@@ -50,3 +53,39 @@ func extractSenderAccount(address types.Address) (*Account, error) {
...
+
+		if ctrArgs, err = abi.Encode([]string{"ZEXFT", "ZEXES"}, ctrArtifact.ABI.Constructor.Inputs); err != nil {
+			return nil, nil, err
+		}
+	} else {
+		ctrArtifact = &generator.ContractArtifact{
+			Bytecode: generator.DefaultContractBytecode,
+		}
+		ctrArgs = nil
+	}
```
dbrajovic: This could be a switch case.

dbrajovic (approved):
LGTM, apart from a few interfaces/structs which should be refactored (in a separate PR).

command/loadbot/execution.go
ZeljkoBenovic: Refactored in e70281e8e0f9d4e6bb0366ad09f905f4e7cefa86

command/loadbot/generators.go
ZeljkoBenovic: Refactored in e70281e8e0f9d4e6bb0366ad09f905f4e7cefa86

command/loadbot/deploy_contract.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		// No gas limit specified, query the network for an estimation
+		gasEstimate, estimateErr := estimateGas(jsonClient, exampleTxn)
+		if estimateErr != nil {
+			return fmt.Errorf("unable to get gas estimate, %w", err)
+		}
+
+		l.generator.SetGasEstimate(gasEstimate)
+	}
```
Kourin1996: This process will be done in Loadbot:Run() method also. Do you need to get and set gasLimit here?

command/loadbot/execution.go
```diff
@@ -183,12 +222,21 @@ func (l *Loadbot) Run() error {
...
 
-	var wg sync.WaitGroup
+	var (
+		wg             sync.WaitGroup
+		receiptTimeout time.Duration
+	)
 
-	receiptTimeout := calcMaxTimeout(l.cfg.Count, l.cfg.TPS)
+	// max-wait by default is 2 min.
+	receiptTimeout = time.Duration(l.cfg.MaxWait) * time.Minute
```
Kourin1996: Why do you need to declare receiptTimeout variable without initial value although you set immediately after this line?

command/loadbot/deploy_contract.go
```diff
@@ -0,0 +1,119 @@
...
+		blockInfom, err := jsonClient.Eth().GetBlockByNumber(web3.BlockNumber(k), false)
+		if err != nil {
+			return fmt.Errorf("could not fetch block by number, %w", err)
+		}
+
+		v.GasLimit = blockInfom.GasLimit
+		v.GasUsed = blockInfom.GasUsed
+		v.Utilization = calculateBlockUtilization(v)
+		l.metrics.ContractGasMetrics.Blocks[k] = v
+	}
```
Kourin1996: I think this process is same to the one in Loadbot:deployContract. Are you able to create a common method?

command/loadbot/execution.go
```diff
@@ -253,6 +305,19 @@ func (l *Loadbot) Run() error {
...
+		if err != nil {
+			return fmt.Errorf("could not fetch block by number, %w", err)
+		}
+
+		v.GasLimit = blockInfom.GasLimit
+		v.GasUsed = blockInfom.GasUsed
+		v.Utilization = calculateBlockUtilization(v)
+		l.metrics.GasMetrics.Blocks[k] = v
+	}
+
```
Kourin1996: As I commented above, you can create a method.

command/loadbot/generator/contract_txn_generator.go
```diff
@@ -0,0 +1,71 @@
...
+	//	return token transfer tx
+	return gen.signer.SignTx(&types.Transaction{
+		From:     gen.params.SenderAddress,
+		To:       gen.contractAddress,
+		Value:    big.NewInt(0),
+		GasPrice: gen.params.GasPrice,
+		Input:    gen.encodedParams,
+		V:        big.NewInt(1), // it is necessary to encode in rlp
+	}, gen.params.SenderKey)
+}
```
Kourin1996, line 46: I don't think we need if case here because nil is set to To in if block.

command/loadbot/generator/contract_txn_generator.go
```diff
@@ -0,0 +1,71 @@
...
+	return gen.signer.SignTx(&types.Transaction{
+		From:     gen.params.SenderAddress,
+		To:       gen.contractAddress,
+		Value:    big.NewInt(0),
+		Gas:      gen.estimatedGas,
+		GasPrice: gen.params.GasPrice,
+		Nonce:    newNextNonce - 1,
+		Input:    gen.encodedParams,
+		V:        big.NewInt(1), // it is necessary to encode in rlp
+	}, gen.params.SenderKey)
```
Kourin1996, line 75: This is same. You don't need if block

Kourin1996 (changes requested):

command/loadbot/params.go
```diff
@@ -187,5 +192,49 @@ func (p *loadbotParams) generateConfig(
...
+
+	switch p.mode {
+	case erc20:
+		ctrArtifact = &generator.ContractArtifact{
+			Bytecode: ERC20BIN,
+			ABI:      abi.MustNewABI(ERC20ABI),
+		}
+
+		if ctrArgs, err = abi.Encode(
+			[]string{"4314500000", "ZexCoin", "ZEX"}, ctrArtifact.ABI.Constructor.Inputs); err != nil {
```
Kourin1996: What does `4314500000` mean? Is there a possibility that you used all tokens during transfer?

command/loadbot/deploy_contract.go
ZeljkoBenovic: After we deploy the contract, we need another estimate to figure out how much gas we'll need for method invocation / token transfer. The initial gas estimate in Loadbot:Run() method is estimating now much gas it will require to deploy the contract.

If  contract address is set generator.GetExampleTransaction() will return contract function invocation transaction, otherwise it will return contract deployment transaction.

This was the fix needed to get proper block utilization.

command/loadbot/execution.go
ZeljkoBenovic: Removed in redundant declaration 2a541d3e262bd10661885fc0305c31cd2f9eba2c

command/loadbot/execution.go
ZeljkoBenovic: Created a method in 9660f0272856259509dc5d1a41a5792398b66f0f

command/loadbot/deploy_contract.go
ZeljkoBenovic: Created a method in 9660f0272856259509dc5d1a41a5792398b66f0f

command/loadbot/generator/contract_txn_generator.go
ZeljkoBenovic: That method checks if the Contract Address is set.
If it is not it sets the Contract Bytecode in the Input field, and if it is, sets the Encoded Parameters which are needed to call SC function.

So anyway, we need to set some kind of flow check in order to set different Input field.
We could basically check if To field is nul, but the if statement would remain.
Unless you have a better idea ? Please do share 🙂

command/loadbot/params.go
ZeljkoBenovic: These are totally arbitrary values.
The ```4314500000``` is the total erc20 token supply.
It's a 5 decimal token, so the value is actually ```43145.00000```.
There is a possibility to use all tokens if someone sets ```--count```  to a high value, like ```100000``` for example.

Should we change this or a higher value or make it configurable maybe?

command/loadbot/generator/contract_txn_generator.go
Kourin1996: OK. So Input is a different field besides To

command/loadbot/helper.go
```diff
@@ -52,18 +52,51 @@ func estimateGas(client *jsonrpc.Client, txn *types.Transaction) (uint64, error)
...
+			blockData.GasUsed = blockInfom.GasUsed
+			blockData.Utilization = calculateBlockUtilization(blockData)
+			gasMetrics.Blocks[blockNum] = blockData
+
+			blockNumErr <- nil
+		}(jsonClient, gasMetrics, blockNum, blockData, blockNumErr)
+
+		if err := <-blockNumErr; err != nil {
+			return err
+		}
```
Kourin1996: Why wait for err here? Now fetching block one by one? 

If you want to run them in serial, I think no need goroutine
If you want to do them in parallel, you should wait outside loop or try to think about using errgroup

command/loadbot/deploy_contract.go
Kourin1996: OK. Understood. You might want to add new method to update gas estimate?

command/loadbot/result.go
```diff
@@ -42,11 +48,14 @@ type TxnDetailedErrorData struct {
...
-	DetailedErrorData TxnDetailedErrorData `json:"detailed_error_data,omitempty"`
-	ApproxTPS         uint64               `json:"approx_tps"`
+	CountData              TxnCountData         `json:"count_data"`
+	TurnAroundData         TxnTurnAroundData    `json:"turn_around_data"`
+	ContractTurnAroundData TxnTurnAroundData    `json:"contract_turn_around_data"`
+	BlockData              TxnBlockData         `json:"block_data"`
+	DetailedErrorData      TxnDetailedErrorData `json:"detailed_error_data,omitempty"`
+	ApproxTPS              uint64               `json:"approx_tps"`
+	ContractAddress        web3.Address         `json:"contract_address"`
+	ContractBlockData      TxnBlockData         `json:"contract_block_data"`
```
zivkovicmilos: What do you think about making these omittable in case the mode is transfer?

cc @Kourin1996

command/loadbot/result.go
```diff
@@ -78,6 +87,33 @@ func (lr *LoadbotResult) initExecutionData(metrics *Metrics) {
...
+		metrics.ContractDeploymentDuration.TotalExecTime.Seconds(),
+		durationPrecision,
+	)
+	// set contract address
+	lr.ContractAddress = metrics.ContractAddress
+	// set contract block data
+	lr.ContractBlockData = TxnBlockData{
+		BlocksRequired:       uint64(len(metrics.ContractDeploymentDuration.blockTransactions)),
+		BlockTransactionsMap: metrics.ContractDeploymentDuration.blockTransactions,
+		GasData:              metrics.ContractGasMetrics.Blocks,
```
zivkovicmilos: You're initializing these values, but they're specific to the contract deployment modes.

What do you think about moving them out to a separate func and initializing this if the correct loadbot mode is set?

command/loadbot/result.go
```diff
@@ -103,18 +139,30 @@ func (lr *LoadbotResult) initDetailedErrors(gen generator.TransactionGenerator)
...
+func (lr *LoadbotResult) writeBlockData(buffer *bytes.Buffer, blockType string) {
+	var blockData *TxnBlockData
+
+	switch blockType {
+	case "contract":
+		blockData = &lr.ContractBlockData
+
+		buffer.WriteString("\n\n[CONTRACT BLOCK DATA]\n")
+	default:
+		blockData = &lr.BlockData
```
zivkovicmilos: Instead of initializing the fields `ContractBlockData` or `BlockData`, where you either use one or the other, you can make an interface method like `GetBlockData`, where your transfer and contract modes implement appropriate result setting. That way we can avoid this switch entirely

command/loadbot/result.go
```diff
@@ -176,41 +229,50 @@ func (lr *LoadbotResult) writeErrorData(buffer bytes.Buffer) {
...
 	return buffer.String()
 }
 
-func (lr *LoadbotResult) writeLoadbotResults(buffer bytes.Buffer) {
+func (lr *LoadbotResult) writeLoadbotResults(buffer *bytes.Buffer) {
 	buffer.WriteString("\n=====[LOADBOT RUN]=====\n")
 
 	lr.writeCountData(buffer)
 	lr.writeApproximateTPSData(buffer)
+	lr.writeContractDeploymentData(buffer)
```
zivkovicmilos, line 235: Why write the deployment data if the mode is not deployment?

zivkovicmilos: Resolving since you have a conditional inside the method, nevermind

command/loadbot/params.go
```diff
@@ -58,24 +62,21 @@ type loadbotParams struct {
...
+	constructorArgs  []byte
 }
 
 func (p *loadbotParams) validateFlags() error {
-	// Validate the correct mode type
-	convMode := Mode(strings.ToLower(p.modeRaw))
-	if convMode != transfer && convMode != deploy {
+	// Set and validate the correct mode type
+	p.mode = Mode(strings.ToLower(p.modeRaw))
+	if p.mode != transfer && p.mode != deploy && p.mode != erc20 && p.mode != erc721 {
```
zivkovicmilos: This can be a separate function `isValidMode`

command/loadbot/params.go
```diff
@@ -58,24 +62,21 @@ type loadbotParams struct {
...
 		return errInvalidMode
 	}
 
 	// Validate the correct mode params
-	if convMode == deploy && p.contractPath == "" {
+	if p.mode == deploy && p.contractPath == "" {
```
zivkovicmilos: This can be a separate function `hasValidDeployParams`

command/loadbot/helper.go
```diff
@@ -52,18 +52,51 @@ func estimateGas(client *jsonrpc.Client, txn *types.Transaction) (uint64, error)
...
-// based on the transaction count and tps params
-func calcMaxTimeout(count, tps uint64) time.Duration {
-	waitTime := minReceiptWait
-	// The receipt timeout should be at max maxReceiptWait
-	// or minReceiptWait + tps / count * 100
-	// This way the wait time scales linearly for more stressful situations
-	waitFactor := time.Duration(float64(tps)/float64(count)*100) * time.Second
-
-	if waitTime+waitFactor > maxReceiptWait {
-		return maxReceiptWait
```
zivkovicmilos: Why did you remove this piece of logic?

We purposefully placed it here so the receipt wait times for transactions would be calculated dynamically on the fly, instead of relying on a single default value / passed in by the user.

Why not just expand this method to account for receipt wait times on contract deployments / calls?

command/loadbot/helper.go
zivkovicmilos: Completely agree with @Kourin1996 here 👍

command/loadbot/execution.go
```diff
@@ -64,10 +66,29 @@ type metadata struct {
...
 type Metrics struct {
 	TotalTransactionsSentCount uint64
 	FailedTransactionsCount    uint64
 	TransactionDuration        ExecDuration
+
+	// contracts
+	FailedContractTransactionsCount uint64
+	ContractDeploymentDuration      ExecDuration
+	ContractAddress                 web3.Address
+	ContractGasMetrics              *BlockGasMetrics
```
zivkovicmilos: Why not extract these values into `ContractMetrics`?

command/loadbot/execution.go
```diff
@@ -183,12 +222,18 @@ func (l *Loadbot) Run() error {
 	ticker := time.NewTicker(1 * time.Second / time.Duration(l.cfg.TPS))
 	defer ticker.Stop()
 
-	var wg sync.WaitGroup
-
-	receiptTimeout := calcMaxTimeout(l.cfg.Count, l.cfg.TPS)
+	// max-wait by default is 2 min.
+	receiptTimeout := time.Duration(l.cfg.MaxWait) * time.Minute
```
zivkovicmilos: Please see my other comment regarding these max waits

command/loadbot/execution.go
```diff
@@ -183,12 +222,18 @@ func (l *Loadbot) Run() error {
...
 
 	startTime := time.Now()
 
+	// deploy contracts
+	if err := l.deployContract(grpcClient, jsonClient, receiptTimeout); err != nil {
+		return fmt.Errorf("unable to deploy smart contract, %w", err)
+	}
```
zivkovicmilos: Looking at the code it seems like this is a necessary step (contract deployment), when in fact it's only necessary for a subset of modes.

Please consider wrapping this somehow so it's not confusing 🙏

command/loadbot/execution.go
```diff
@@ -235,6 +280,10 @@ func (l *Loadbot) Run() error {
 
 				return
 			}
+			// initialise block numbers
+			l.metrics.GasMetrics.BlockGasMutex.Lock()
+			l.metrics.GasMetrics.Blocks[receipt.BlockNumber] = GasMetrics{}
+			l.metrics.GasMetrics.BlockGasMutex.Unlock()
```
zivkovicmilos: This can be a separate method

command/loadbot/abis.go
```diff
@@ -0,0 +1,378 @@
...
...nction"},{"inputs":[],"name":"decimals","outputs":[{"internalType":"uint8","name":"","type":"uint8"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"name","outputs":[{"internalType":"string","name":"","type":"string"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"symbol","outputs":[{"internalType":"string","name":"","type":"string"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"totalSupply","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"receiver","type":"address"},{"internalType":"uint256","name":"numTokens","type":"uint256"}],"name":"transfer","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"address","name":"owner","type":"address"},{"internalType":"address","name":"buyer","type":"address"},{"internalType":"uint256","name":"numTokens","type":"uint256"}],"name":"transferFrom","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"nonpayable","type":"function"},{"stateMutability":"payable","type":"receive"}]`
```
zivkovicmilos: Consider formatting this ABI nicely like you did the one below 👍

command/loadbot/generator/base.go
```diff
@@ -1,17 +1,25 @@
...
 
+	// failed contract deployment transactions
+	failedContractTxns     []*FailedContractTxnInfo
+	failedContractTxnsLock sync.RWMutex
+
 	params       *GeneratorParams
 	signer       *crypto.EIP155Signer
 	estimatedGas uint64
+
+	contractAddress *types.Address
```
zivkovicmilos: Why does the base generator need to have this information on contract deployments?

command/loadbot/generator/erc_generators.go
```diff
@@ -0,0 +1,62 @@
...
+	buf, err := hex.DecodeString(params.ContractArtifact.Bytecode)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode bytecode, %w", err)
+	}
+
+	gen.contractBytecode = buf
+	gen.contractBytecode = append(gen.contractBytecode, params.ConstructorArgs...)
+
+	if gen.encodedParams, err = params.ContractArtifact.ABI.Methods["createNFT"].Encode(
+		[]string{"https://realy-valuable-nft-page.io"}); err != nil {
```
zivkovicmilos: Typo here `realy` -> `really`

command/loadbot/generator/generator.go
```diff
@@ -3,24 +3,30 @@ package generator
...
 )
 
 type TransactionGenerator interface {
 	GenerateTransaction() (*types.Transaction, error)
 	GetExampleTransaction() (*types.Transaction, error)
 	GetTransactionErrors() []*FailedTxnInfo
 	MarkFailedTxn(failedTxn *FailedTxnInfo)
+	MarkFailedContractTxn(failedContractTxn *FailedContractTxnInfo)
 	SetGasEstimate(gasEstimate uint64)
+	SetContractAddress(types.Address)
```
zivkovicmilos: The transaction generator shouldn't need to know about Contract transactions, that was the whole point of separating them out

zivkovicmilos (changes requested):
I've left some comments that need to be addressed.

The PR looks good, but there are a few problematic parts that worry me, namely how we merged the Transaction generator and contract / token generator logic, and why we removed dynamic wait times for transactions.

Please review the comments and we can discuss 🙏

command/loadbot/params.go
dbrajovic: I'd have the hardcoded values moved into their own (global) variables, so they can be configured from there.

command/loadbot/execution.go
dbrajovic: @ZeljkoBenovic regarding this, create a simple helper `isTokenTransferMode()` so that you only call `deployContract` if that case is true.

command/loadbot/deploy_contract.go
ZeljkoBenovic: Created new method c181cf7e57b866f8b0ccf2d9b9ca1d90f88b05ad

command/loadbot/params.go
ZeljkoBenovic: Moved constructor args to const 168f73cceef7ce3e718b2c001c66af17ecd8acb9

command/loadbot/helper.go
ZeljkoBenovic: Switched to errgroup ecf9a2e7239e8fb2d1f3d2d03060f8fc38d97208

command/loadbot/result.go
ZeljkoBenovic: Fixed in 912c140f691ad1c9fd3a01e65b4de25086526eb2

command/loadbot/result.go
ZeljkoBenovic: Moved to dedicated method fd637623002a062ce9901a3caf25a11de592d358

command/loadbot/result.go
ZeljkoBenovic: Dumped the switch ecf9a2e7239e8fb2d1f3d2d03060f8fc38d97208

command/loadbot/params.go
ZeljkoBenovic: Added `isValidMode` method a9000c81b7a864ea173e53de01191f6f649f4496

command/loadbot/params.go
ZeljkoBenovic: Added `hasValidDeployParams` a9000c81b7a864ea173e53de01191f6f649f4496

command/loadbot/helper.go
ZeljkoBenovic: Restored `calcMaxTimeout` 6e26304396b9c7250e1938be83405f412e4d7e96 and increased the minimum time to wait for receipts to 1min. 3c8fa5a904c5428e790ac7668be99c7f213e07ef .
Also, with the `--max-wait` flag the user can define how many minutes to wait for receipts, as for some modes like `erc721` and a large number of txns, the `calcMaxTimeout` timeout is too short.

command/loadbot/execution.go
ZeljkoBenovic: Moved into dedicated struct 3c8fa5a904c5428e790ac7668be99c7f213e07ef

command/loadbot/execution.go
ZeljkoBenovic: Check my other comment about max waits

command/loadbot/execution.go
ZeljkoBenovic: Added isTokenTransferMode 65207facceda48c54b7e2ec14a7f98fe46b37d00

command/loadbot/execution.go
ZeljkoBenovic: Moved to dedicated method 0d0dc3410f11c71b098c4238344680a86d0d070e

command/loadbot/abis.go
ZeljkoBenovic: Prettified in 0d0dc3410f11c71b098c4238344680a86d0d070e

command/loadbot/generator/base.go
ZeljkoBenovic: Removed from BaseGenerator in 0d0dc3410f11c71b098c4238344680a86d0d070e

command/loadbot/generator/erc_generators.go
ZeljkoBenovic: Fixed the typo in 0d0dc3410f11c71b098c4238344680a86d0d070e

command/loadbot/generator/generator.go
ZeljkoBenovic: Moved to `ContractTxnGenerator` interface that is used only on `erc` modes

command/loadbot/generator/erc_generators.go
```diff
@@ -0,0 +1,62 @@
...
+
+	buf, err := hex.DecodeString(params.ContractArtifact.Bytecode)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode bytecode, %w", err)
+	}
+
+	gen.contractBytecode = buf
+	gen.contractBytecode = append(gen.contractBytecode, params.ConstructorArgs...)
+
+	if gen.encodedParams, err = params.ContractArtifact.ABI.Methods["createNFT"].Encode(
```
Kourin1996, line 56: btw, what is `createNFT` method? It's not included in ERC721 standard. Loadbot for ERC721 is checking performance on minting?

Kourin1996 (approved):
Looks good to me. I have one question. Please check it.

command/loadbot/params.go
Kourin1996: I think they should be configurable. `4314500000` looks random value. You need to write the reason why `4314500000` is proper as a comment around constants at least.

zivkovicmilos (approved):
Thank you for resolving the discussions 🙏 

Looks good 💯

command/loadbot/generator/erc_generators.go
ZeljkoBenovic: I was using teh OpenZeppelin framework and the createNFT is just a custom method that mints tokens.
The whole thing is very simple:
```
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.7.0 <0.9.0;

import "@openzeppelin/contracts/utils/Counters.sol";
import "@openzeppelin/contracts/token/ERC721/extensions/ERC721URIStorage.sol";


contract ZexNFTs is ERC721URIStorage {

    using Counters for Counters.Counter;
    Counters.Counter private _tokenIds;
    
    //constructor for an ERC721 is a name and symbol
    constructor (string memory tokenName, string memory tokenSymbol ) ERC721(tokenName, tokenSymbol){}

    //after we mint the token we are going to return the id of the token
    function createNFT(string memory tokenURI) public returns (uint256) {

    //get number from token counter
        uint256 newNFTTokenId = _tokenIds.current();

    //safely mint token for the person that called the function
        _safeMint(msg.sender, newNFTTokenId);
    
    //set the token uri of the token id of the uri passed
        _setTokenURI(newNFTTokenId, tokenURI);
    
    //increment the counter
        _tokenIds.increment();
        
    //return the token id
        return newNFTTokenId;
    }

}
```

Should we maybe post this in our pub docs ?

Kourin1996: To be honest, it's not simulating NFT transfer, am I right? I thought that was problem.

command/loadbot/params.go
Kourin1996: @ZeljkoBenovic

command/loadbot/generator/erc_generators.go
ZeljkoBenovic: Yes you're right. This loadbot erc721 mode is only minting the tokens, its not doing any transfers. That was the test for erc721 mode, that we've agreed upon and it's what was stated in [this task](https://linear.app/trapesys/issue/EDGE-390/erc-721-performance-tests). We can easily fix this or add an additional mode to stress test the actual erc721 token transfers in the future, if needed ...

Kourin1996: @ZeljkoBenovic OK. I checked the task. You're right. Resolving

command/loadbot/params.go
ZeljkoBenovic: Don't think that anyone will want to configure this, so I think we don't need to add unnecessary flags. 
I'd say that there just has to be enough funds to run all the transactions the user defined.

I've set `5000000000` as total token supply, set the transfer value to `0.001` and I've added comments.
This should be more than enough for any loadbot scenario.

@Kourin1996 ac607d15a6adf365a17996b8a84e98fa5513dad6

0xAleksaOpacic (approved):
LGTM!

PR #428, opened 2022-02-25
[Bridge] Add State Transaction and state sync mechanism in IBFT

Kourin1996:
# Description

Please provide a detailed description of what was done in this PR

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [ ] I have updated the official documentation
- [x] I have added sufficient documentation in code

## Testing

- [x] I have tested this code with the official test suite
- [x] I have tested this code manually

### Manual tests

Please complete this section if you ran manual tests for this functionality, otherwise delete it

# Documentation update

Please link the documentation update PR in this section if it's present, otherwise delete it

# Additional comments

Please post additional comments in this section if you have them, otherwise delete it

PR #434, opened 2022-03-02
[Bridge] Add State Transaction and state sync mechanism in IBFT

Kourin1996:
Fixes EDGE-401
Reopen https://github.com/0xPolygon/polygon-edge/pull/428 because this PR has been marked as merged by some reason when I did local checkout.

# Description

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

For manual testing, you will need 2 chains (Referring to these chain as root and destination chain)
You will need to setup the 2 chains and deploy StateSyncer contract to root chain and StateReceiver to destination chain. StateSyncer must emit the event with specific interface and StateReceiver must have specific method to receive method.

```sol
pragma solidity ^0.8.7;

contract StateSyncer {
    uint256 nonce = 0;

    event StateSynced(
        uint256 indexed id,
        address indexed contractAddress,
        bytes data
    );

    function stateSync(address contractAddress, bytes memory data) public {
        emit StateSynced(nonce++, contractAddress, data);
    }
}
```

```sol
pragma solidity ^0.8.7;

contract StateReceiver {
    event StateReceived(bytes data);

    string _data;

    function onStateReceived(bytes memory data) external {
        _data = string(data);
        emit StateReceived(data);
    }

    function getState() public view returns (string memory) {
        return _data;
    }
}
```

When you start destination chain, you can give `--bridge` flag to enable Bridge like this:

```
$ polygon-edge server ... --bridge --bridge-rootchain-url ws://localhost:11002/ws --bridge-rootchain-contract 0x37422f81BBF1F14244AD30D66F16705F14e78373 --bridge-rootchain-confirmations 10
```

Once event is emitted from StateSyncer, destination chain catches this, gets consensus, then applied to StateRecever in destination chain

# Documentation update

# Additional comments


types/transaction.go
```diff
@@ -1,12 +1,45 @@
...
 
+// TransactionType represents the value between 0x00 and 0x7F to represents the type of transaction
+// TransactionType is put before transaction data in RLP encoded
+type TransactionType uint8
+
+const (
+	TxTypeLegacy TransactionType = 0x0
+	// TxTypeAccessList TransactionType = 0x1
+	// TxTypeDynamicFee TransactionType = 0x2
+	TxTypeState TransactionType = 0x7f
```
zivkovicmilos: Why are these two commented out?

Did we follow any EIP guidelines for this?

types/transaction.go
```diff
@@ -1,12 +1,45 @@
...
+)
+
+func ToTransactionType(b byte) (TransactionType, error) {
+	tt := TransactionType(b)
+	switch tt {
+	case TxTypeLegacy, TxTypeState:
+		return tt, nil
+	default:
+		return TransactionType(0), fmt.Errorf("undefined transaction type: %d", b)
```
zivkovicmilos, line 31: Why not support the remaining two?

types/transaction.go
```diff
@@ -1,12 +1,45 @@
...
+func (t TransactionType) String() string {
+	switch t {
+	case TxTypeLegacy:
+		return "TxTypeLegacy"
+	case TxTypeState:
+		return "TxTypeState"
+	default:
+		return fmt.Sprintf("TxType(%x)", byte(t))
+	}
+}
```
zivkovicmilos, line 44: Why support just these 2?

types/transaction.go
```diff
@@ -20,6 +53,8 @@ type Transaction struct {
 	Hash     Hash
 	From     Address
 
+	Type TransactionType
+
 	// Cache
 	size atomic.Value
```
zivkovicmilos, line 62: What do we even use this for?

types/transaction.go
```diff
@@ -28,16 +63,17 @@ func (t *Transaction) IsContractCreation() bool {
...
-	hash.WriteRlp(t.Hash[:0], v)
-
-	marshalArenaPool.Put(ar)
-	keccak.DefaultKeccakPool.Put(hash)
+	if _, err := hash.Write(t.MarshalRLP()); err == nil {
+		hash.Sum(t.Hash[:0])
+		keccak.DefaultKeccakPool.Put(hash)
+	}
 
 	return t
```
zivkovicmilos, line 81: If the hashing fails for whatever reason, we just return the transaction without the hash filled?

Where do we even set the transaction hash here?

types/rlp_unmarshal_storage.go
```diff
@@ -30,12 +30,34 @@ func (b *Body) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
...
+				return err
+			}
+
+			if l := len(bytes); l != 1 {
+				return fmt.Errorf("expected 1 byte transaction type, but size is %d", l)
+			}
+
+			if txType, err = ToTransactionType(bytes[0]); err != nil {
+				return err
+			}
```
zivkovicmilos: Please add additional explanation here 🙏

types/rlp_unmarshal_storage.go
```diff
@@ -58,7 +80,22 @@ func (b *Body) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
 }
 
 func (t *Transaction) UnmarshalStoreRLP(input []byte) error {
-	return UnmarshalRlp(t.UnmarshalStoreRLPFrom, input)
+	txType := TxTypeLegacy
+
+	if len(input) > 0 && input[0] <= 0x7F {
```
zivkovicmilos: What is this magic number for input? Please extract it out and make it clear

types/rlp_unmarshal_storage.go
```diff
@@ -67,46 +104,113 @@ func (t *Transaction) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value)
 		return err
 	}
 
-	if len(elems) != 2 {
-		return fmt.Errorf("expected 2 elements")
+	if len(elems) != 2 && len(elems) != 3 {
```
zivkovicmilos, line 100: What are the two / three elements?

types/rlp_unmarshal_storage.go
```diff
@@ -67,46 +104,113 @@ func (t *Transaction) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value)
...
+		return fmt.Errorf("expected 2 or 3 elements")
```
zivkovicmilos: Nitpick, but this should be a standardized error with `errors.New`

types/rlp_unmarshal_storage.go
```diff
@@ -67,46 +104,113 @@ func (t *Transaction) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value)
...
+	t.ComputeHash()
+
 	return nil
 }
 
 func (r *Receipts) UnmarshalStoreRLP(input []byte) error {
 	return UnmarshalRlp(r.UnmarshalStoreRLPFrom, input)
 }
 
+//nolint:dupl
```
zivkovicmilos: Can we extract the common areas of these methods that are duplicated, and drop this `nolint`?

types/rlp_unmarshal_storage.go
```diff
@@ -67,46 +104,113 @@ func (t *Transaction) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value)
...
 	}
 
 	return nil
 }
 
 func (r *Receipt) UnmarshalStoreRLP(input []byte) error {
-	return UnmarshalRlp(r.UnmarshalStoreRLPFrom, input)
+	txType := TxTypeLegacy
+
+	if len(input) > 0 && input[0] <= 0x7F {
```
zivkovicmilos: Again, the magic number, same as previous comment

types/rlp_marshal_storage.go
```diff
@@ -56,6 +60,7 @@ func (r Receipts) MarshalStoreRLPTo(dst []byte) []byte {
 func (r *Receipts) MarshalStoreRLPWith(a *fastrlp.Arena) *fastrlp.Value {
 	vv := a.NewArray()
 	for _, rr := range *r {
+		// TODO: Typed Transaction
```
zivkovicmilos: Leftover?

types/rlp_marshal_storage.go
```diff
@@ -69,6 +74,7 @@ func (r *Receipt) MarshalStoreRLPTo(dst []byte) []byte {
 func (r *Receipt) MarshalStoreRLPWith(a *fastrlp.Arena) *fastrlp.Value {
 	// use the hash part
 	vv := a.NewArray()
+	// TODO: Typed Transaction
```
zivkovicmilos: Leftover?

types/rlp_marshal.go
```diff
@@ -104,6 +113,10 @@ func (r *Receipt) MarshalRLP() []byte {
 }
 
 func (r *Receipt) MarshalRLPTo(dst []byte) []byte {
+	if r.IsTypedTransaction() {
+		dst = append(dst, byte(r.TransactionType))
+	}
+
```
zivkovicmilos, line 123: Is there a way to move his into `MarshalRLPTo`? What do you think?

Everywhere else in the codebase `MarshalRLPTo` calls `MarshalRLPTo` with the callback `With`

types/receipt.go
```diff
@@ -31,12 +31,17 @@ type Receipt struct {
 	GasUsed         uint64
 	ContractAddress Address
 	TxHash          Hash
+	TransactionType TransactionType
```
zivkovicmilos, line 34: Did we need to put this in the receipt? Why?

types/buildroot/buildroot.go
```diff
@@ -13,10 +13,14 @@ var arenaPool fastrlp.ArenaPool
...
 
-	res := calculateRootWithRlp(len(receipts), func(i int) *fastrlp.Value {
-		ar.Reset()
-
-		return receipts[i].MarshalRLPWith(ar)
+	//res := calculateRootWithRlp(len(receipts), func(i int) *fastrlp.Value {
+	//	ar.Reset()
+	//
+	//	return receipts[i].MarshalRLPWith(ar)
+	//})
```
zivkovicmilos: Leftover

types/buildroot/buildroot.go
```diff
@@ -28,10 +32,14 @@ func CalculateReceiptsRoot(receipts []*types.Receipt) types.Hash {
 func CalculateTransactionsRoot(transactions []*types.Transaction) types.Hash {
 	ar := arenaPool.Get()
 
-	res := calculateRootWithRlp(len(transactions), func(i int) *fastrlp.Value {
-		ar.Reset()
+	//res := calculateRootWithRlp(len(transactions), func(i int) *fastrlp.Value {
+	//	ar.Reset()
+	//
+	//	return transactions[i].MarshalRLPWith(ar)
+	//})
```
zivkovicmilos: Leftover

types/buildroot/buildroot.go
```diff
@@ -59,13 +67,13 @@ func CalculateUncleRoot(uncles []*types.Header) types.Hash {
...
-
-	return CalculateRoot(num, hF)
-}
+//func calculateRootWithRlp(num int, h func(indx int) *fastrlp.Value) types.Hash {
+//	hF := func(indx int) []byte {
+//		return h(indx).MarshalTo(nil)
+//	}
+//
+//	return CalculateRoot(num, hF)
+//}
```
zivkovicmilos: Leftover

txpool/txpool.go
```diff
@@ -467,6 +468,11 @@ func (p *TxPool) processEvent(event *blockchain.Event) {
 // validateTx ensures the transaction conforms to specific
 // constraints before entering the pool.
 func (p *TxPool) validateTx(tx *types.Transaction) error {
+	// FIXME: Accepts only legacy transaction only for now
```
zivkovicmilos, line 471: Leftover?

state/executor.go
```diff
@@ -378,16 +374,159 @@ func (t *Transition) nonceCheck(msg *types.Transaction) error {
...
+
+	if msg.To == nil || *msg.To == types.ZeroAddress {
+		return 0, NewTransitionApplicationError(
+			errors.New("to of state transaction must be specified"),
+			true,
+		)
+	}
+
+	// FIXME: unbounded gas limit for now because it's hard to estimate how much the gas is used
+	// and not decided in spec yet
```
zivkovicmilos, line 473: Can we discuss this?

state/executor.go
```diff
@@ -417,48 +556,13 @@ func NewGasLimitReachedTransitionApplicationError(err error) *GasLimitReachedTra
...
-
-	// 3. the amount of gas required is available in the block
-	if err := t.subGasPool(msg.Gas); err != nil {
-		return nil, NewGasLimitReachedTransitionApplicationError(err)
-	}
-
-	// 4. there is no overflow when calculating intrinsic gas
-	intrinsicGasCost, err := TransactionGasCost(msg, t.config.Homestead, t.config.Istanbul)
+	// First check this message satisfies all consensus rules before
+	// applying the message. The rules include these clauses
```
zivkovicmilos: The comment is legacy (has a leftover part `The rules include these clauses`)

e2e/bridge_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+		res, err := StateRecieverABI.Methods["getState"].Outputs.Decode(byteData)
+		if err != nil {
+			return err, true
+		}
+
+		resMap, ok := res.(map[string]interface{})
+		if !ok {
+			return nil, false
+		}
```
zivkovicmilos: Nitpick, this should be a separate method

e2e/framework/config.go
```diff
@@ -46,6 +46,11 @@ type TestServerConfig struct {
 	ShowsLog                bool                 // Flag specifying if logs are shown
 	IsPos                   bool                 // Specifies the mechanism used for IBFT (PoA / PoS)
 	Signer                  *crypto.EIP155Signer // Signer used for transactions
+	//
```
zivkovicmilos: Leftover

e2e/framework/config.go
```diff
@@ -46,6 +46,11 @@ type TestServerConfig struct {
...
+	UseBridge                    bool
+	BridgeRootChainURL           *string
+	BridgeRootChainContract      *string
+	BridgeRootChainConfirmations *uint
```
zivkovicmilos: Please document these settings better, as others (that are even more obvious) are documented 🙏

consensus/ibft/pos.go
```diff
@@ -110,12 +110,12 @@ func (pos *PoSMechanism) acceptStateLogHook(snapParam interface{}) error {
...
 // in order to update the validator set
 func (pos *PoSMechanism) insertBlockHook(numberParam interface{}) error {
-	headerNumber, ok := numberParam.(uint64)
+	block, ok := numberParam.(*types.Block)
 	if !ok {
 		return ErrInvalidHookParam
 	}
 
-	return pos.updateValidators(headerNumber)
+	return pos.updateValidators(block.Number())
```
zivkovicmilos, line 118: Why change this to receive a block if you aren't going to use other fields in the block (and PoA doesn't implement this)?

consensus/ibft/pos.go
```diff
@@ -126,6 +126,8 @@ func (pos *PoSMechanism) verifyBlockHook(blockParam interface{}) error {
 	}
 
 	if pos.ibft.IsLastOfEpoch(block.Number()) && len(block.Transactions) > 0 {
+		pos.ibft.logger.Error("block verification failed, block at the end of epoch has transactions")
```
zivkovicmilos, line 129: Isn't this propagated up in the logs?

consensus/ibft/ibft.go
```diff
@@ -299,6 +299,16 @@ func (i *Ibft) setupMechanism() error {
 		}
 	}
 
+	if i.bridge != nil {
+		bridgeMechanism, _ := BridgeFactory(i)
+
+		if err != nil {
+			return err
+		}
```
zivkovicmilos: You're not checking the BridgeFactory error here, but some leftover error variable

consensus/ibft/ibft.go
```diff
@@ -1287,12 +1286,32 @@ func (i *Ibft) GetBlockCreator(header *types.Header) (types.Address, error) {
 	return ecrecoverFromHeader(header)
 }
 
+// PreStateCommit a hook to be called before finalizing state transition on inserting block
+func (i *Ibft) runInsertTransactionsHook(
+	header *types.Header,
+	transition *state.Transition,
+	transactions *[]*types.Transaction,
```
zivkovicmilos, line 1291: This can probably be a struct param

consensus/ibft/hooks.go
```diff
@@ -18,6 +18,10 @@ const (
 	// PoS defines the Proof of Stake IBFT type,
 	// where the validator set it changed through staking on the Staking SC
 	PoS MechanismType = "PoS"
+
+	// Bridge defines Bridge type
+	// Bridge allows to include StateTransactions into block
```
zivkovicmilos: Typo, should be `Bridge allows the inclusion of StateTransactions into a block`

consensus/ibft/hooks.go
```diff
@@ -82,6 +86,9 @@ const (
 	// CalculateProposerHook defines what is the next proposer
 	// based on the previous
 	CalculateProposerHook = "CalculateProposerHook"
+
+	// InsertTransactionsHook defines the process to insert additional transactions
+	InsertTransactionsHook = "InsertTransactionsHook"
```
zivkovicmilos, line 91: You can add a comment marking it is only used for the Bridge mechanism:
`// BRIDGE //`

consensus/ibft/ibft.go
zivkovicmilos: Aha, you never error out in the BridgeFactory.

What error are you checking here?

consensus/ibft/bridge.go
```diff
@@ -0,0 +1,173 @@
...
+
+		b.bridge.Consume(tx)
+	}
+
+	return nil
+}
+
+func (b *BridgeMechanism) calculateSignatureThreshold(set ValidatorSet) uint64 {
+	return uint64(set.Len())
+}
```
zivkovicmilos: Why is the signature threshold the entire size of the validator set and not the majority?

This means that every validator needs to sign this message in order for the threshold to be met.

Why not use the provided `NumValid` method instead?

consensus/ibft/bridge.go
```diff
@@ -0,0 +1,173 @@
...
+	b.hookMap[AcceptStateLogHook] = b.updateValidatorsHook
+
+	// Register the VerifyBlockHook
+	b.hookMap[VerifyBlockHook] = b.verifyStateTransactionsHook
+
+	// Register the InsertTransactionsHook
+	b.hookMap[InsertTransactionsHook] = b.insertStateTransactionsHook
+
+	// Register the InsertBlockHook
+	b.hookMap[InsertBlockHook] = b.consumeStateTransactionsHook
```
zivkovicmilos, line 166: The hook is called `InsertBlockHook`, but your method is called `consumeStateTransactionsHook`

bridge/event.go
```diff
@@ -0,0 +1,51 @@
...
+
+	contractAddr, ok = event["contractAddress"].(web3.Address)
+	if !ok {
+		return nil, errors.New("failed to parse contractAddress in StateSyncedEvent")
+	}
+
+	data, ok = event["data"].([]uint8)
+	if !ok {
+		return nil, errors.New("failed to parse data in StateSyncedEvent")
+	}
```
zivkovicmilos: What do you think about extracting these errors?

bridge/bridge.go
```diff
@@ -222,3 +274,37 @@ func (b *bridge) addRemoteMessage(message *transport.SignedMessage) {
 		Signature: message.Signature,
 	})
 }
+
+func getMessageHash(msg *Message) types.Hash {
+	// return msg.ComputeHash()
```
zivkovicmilos: Leftover

bridge/bridge.go
```diff
@@ -113,15 +119,48 @@ func (b *bridge) Close() error {
...
+
+	return data, nil
+}
+
+func (b *bridge) ValidateTx(tx *types.Transaction) error {
+	hash := getTransactionHash(tx)
+
+	if !b.sampool.IsMessageKnown(hash) {
+		return fmt.Errorf("unknown state transaction, hash=%s", hash.String())
+	}
```
zivkovicmilos: This is a potential problem.

You verify block state transactions by making sure that they're in your SAM pool - when this doesn't have to be the case. 

All that matters is that the signatures are ok. If by any chance you didn't have this txn in your SAM pool, then that means you'll fail this validation check and break out of the consensus cycle

zivkovicmilos (changes requested):
Overall it looks good 💯 

I've left some comments that need to be discussed.

Thank you also for making the executor apply much cleaner, it's really noticeable and more readable 🙏 

As soon as we clear these comments up, we're good to go 🚀

consensus/ibft/hooks.go
Kourin1996: @zivkovicmilos To be honest, I don't like IBFT knows what hooks each mechanism wants because the mechanism is an extension and IBFT must not know them.

consensus/ibft/hooks.go
Kourin1996: @zivkovicmilos To be honest, I don't like IBFT knows what hooks are for each mechanism because mechanism is extension and IBFT must not know what extension is.

consensus/ibft/bridge.go
Kourin1996: @zivkovicmilos I'm not sure need to use name method name to hook name. method name describes what the method does.

txpool/txpool.go
Kourin1996: This is not leftover right now. But we need to fix when we introduce other typed transaction. We can keep it right now

Kourin1996: @zivkovicmilos

types/receipt.go
Kourin1996: @zivkovicmilos Please check [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)

types/transaction.go
Kourin1996: These two commented types are introduced into Ethereum. I'm not sure we introduce them. But should avoid to use them

Kourin1996: @zivkovicmilos

types/transaction.go
Kourin1996: @zivkovicmilos 
As I commented above, Edge doesn't have remaining types (TxTypeAccessList, TxTypeDynamicFee). We shouldn't pass such transaction types. Am I correct?

types/transaction.go
Kourin1996: Same to https://github.com/0xPolygon/polygon-edge/pull/434#discussion_r825895761

types/transaction.go
Kourin1996: What are you pointing? size? Size is used in TxPool

Kourin1996: @zivkovicmilos

types/transaction.go
Kourin1996: @zivkovicmilos 

> If the hashing fails for whatever reason, we just return the transaction without the hash filled?

Yes, we should return error also?

> Where do we even set the transaction hash here?

t.Hash

bridge/bridge.go
Kourin1996: Totally agree with that. On validating we don't need to check SAM Pool. I'll fix that

types/rlp_unmarshal_storage.go
Kourin1996: Added at b4008cc9699c1c0f45193b28975cd5fba4483a95

types/rlp_unmarshal_storage.go
Kourin1996: Fixed at 7f41c6ec7c5545c7d76b828445ba99a2cfa3f1f7

types/rlp_unmarshal_storage.go
Kourin1996: Created method in TransactionType 775e5f00b9656f90fdeded9eb408867d95b37a38

types/rlp_unmarshal_storage.go
Kourin1996: Added at b4008cc9699c1c0f45193b28975cd5fba4483a95

types/transaction.go
zivkovicmilos: Agreed, I would leave them uncommented, and add a comment above them saying that they're unsupported and a link to their respective EIPs.

Implementing this is waaaay out of the scope of this PR 👍

types/rlp_marshal.go
Kourin1996: @zivkovicmilos Which MarshalRLPTo? Receipt or Receipts?

types/transaction.go
zivkovicmilos: Agreed, resolving 👍

types/transaction.go
zivkovicmilos: cc @dbrajovic

types/buildroot/buildroot.go
Kourin1996: Removed at ba8e18ada5efdd28c2afd02109313560396183bb

types/buildroot/buildroot.go
Kourin1996: Removed at ba8e18ada5efdd28c2afd02109313560396183bb

types/buildroot/buildroot.go
Kourin1996: Removed at ba8e18ada5efdd28c2afd02109313560396183bb

types/transaction.go
zivkovicmilos: 🙂 
I know that the hash is supposed to be in `t.Hash`, I'm asking at which point we actually set this field here.

Is `ComputeHash` an old function that was already there?

state/executor.go
Kourin1996: Removed at 0e801646c7042870617798acd12e4962702465d7

types/receipt.go
zivkovicmilos: Got it. We offer semi-support for these EIPs now 👍 

We should probably add this as a task for later.

txpool/txpool.go
zivkovicmilos: What do you think about dropping the `FIXME`?

Whoever takes up the implementation of these EIPs will resolve it

consensus/ibft/hooks.go
zivkovicmilos: Completely agree, it's just nicer to know for us which hooks are implemented in which mechanism

consensus/ibft/bridge.go
zivkovicmilos: I always like to keep these consistent, but I'll give in now since it's self-explanatory

types/rlp_marshal.go
zivkovicmilos: Of the receipt. If it's too much of a hassle, then I'm fine with leaving it

e2e/bridge_test.go
Kourin1996: Added a function to get state from StateReceiver at 1755956c356860b0b10fd95865c122196138b3b3

e2e/framework/config.go
Kourin1996: Removed and added comments at c7181b25e9fc6621f7ca89c10e118f0216f71709

e2e/framework/config.go
Kourin1996: Added at c7181b25e9fc6621f7ca89c10e118f0216f71709

consensus/ibft/pos.go
Kourin1996: Bridge hook needs Block and PoS

consensus/ibft/pos.go
Kourin1996: Removed duplicate part at c983b2745fdd0a1524f3d04801340087ed0f5e1b

consensus/ibft/ibft.go
Kourin1996: Added at 59388a9e1e2f9a54baa081694c9dd5740f1b9928

consensus/ibft/ibft.go
Kourin1996: @zivkovicmilos What do you mean?

consensus/ibft/hooks.go
Kourin1996: Fixed at 4a749c116ca9d4fda4904e06e5780dc40530e00f

consensus/ibft/bridge.go
Kourin1996: Changed at ffcf9323e750de75f601b0e04adc31d5636aaf00

bridge/event.go
Kourin1996: Done b43120a359b22ff674342df90ba2bda413017770

bridge/bridge.go
Kourin1996: Removed at ba163e2457be4f7109dc92c56b439ceaa53a7ed4

bridge/bridge.go
Kourin1996: @zivkovicmilos You're right.
Fixed at d48cff7feefc0c0f7db635964e4bc6a609ee5e1f

types/rlp_marshal_storage.go
Kourin1996: Fixed at 1a6c98de3df39b62fd4a7bccb7c18757c884ab35

types/rlp_marshal_storage.go
Kourin1996: Fixed at 1a6c98de3df39b62fd4a7bccb7c18757c884ab35

types/rlp_unmarshal_storage.go
Kourin1996: Fixed 6305f0f5c5c304f6726dc1b7e292905f25bb640d

consensus/ibft/ibft.go
zivkovicmilos: I'm just suggesting you bundle these three params into a struct - I actually think you already have it somewhere

types/rlp_unmarshal_storage.go
Kourin1996: If a transaction has a type, then transaction type comes first. But no place to save transaction type for legacy transaction

Kourin1996: @zivkovicmilos

bridge/bridge.go
zivkovicmilos: I'm thinking about this problem - we still rely on the SAM pool to verify the signatures (there's no way for the signatures to be attached to the block somehow as of now).

What happens if you don't have the signatures due to a network error?

types/transaction.go
zivkovicmilos: I see we use it for slots calculation. Weird that it's in this structure, resolving as it's not part of this PR

types/rlp_unmarshal_storage.go
zivkovicmilos: Got it, thank you for clarifying 🙏

types/transaction.go
Kourin1996: @zivkovicmilos 
I added comments at 9bd7bce08465cddeab18dc9aceb476e132b9de69

consensus/ibft/ibft.go
Kourin1996: @zivkovicmilos 
Sorry, I don't understand completely. Bundling below.

dbrajovic (approved):
Apart from Milos' observations LGTM - parts of the codebase outside the module seem to fit well. Let's merge this to the base branch, so the bridge module can be taken care of internally.

0xAleksaOpacic (approved):
LGTM

bridge/bridge.go
Kourin1996: @zivkovicmilos 
If the block producer doesn't have enough signatures, then such transaction will not be added to block. If validator doesn't, won't send prepare message. But looks fine.
As you said, only approach to avoid that is seal signatures into transaction. It resolve bulk sync issue also

types/transaction.go
Kourin1996: @zivkovicmilos 

>  I'm asking at which point we actually set this field here.
hash.Sum(t.Hash[:0])

> Is ComputeHash an old function that was already there?
Yes

types/rlp_marshal.go
Kourin1996: No, it's impossible. Because need to add TransactionType before RLP(Receipt). If it's done inside Recipt, TransactionType will be included, like RLP(TransactionType || Receipt)
@zivkovicmilos

types/transaction.go
zivkovicmilos: Got it 👍 

Let's leave this for now, since it's a legacy func

types/rlp_marshal.go
zivkovicmilos: Thank you for clarifying 🙏

bridge/bridge.go
zivkovicmilos: As we discussed internally, this functionality should be part of a different PR 👍

txpool/txpool.go
zivkovicmilos: Resolving the nitpick so we can merge this out

state/executor.go
zivkovicmilos: Resolving after an internal discussion

zivkovicmilos (approved):
Thank you for resolving the pending discussions 🙏 

There are a couple of things we'll be continuing to discuss on the main bridge PR.

Looks great, let's squash and merge this 🚀

types/buildroot/buildroot.go
```diff
@@ -13,10 +13,8 @@ var arenaPool fastrlp.ArenaPool
 func CalculateReceiptsRoot(receipts []*types.Receipt) types.Hash {
 	ar := arenaPool.Get()
 
-	res := calculateRootWithRlp(len(receipts), func(i int) *fastrlp.Value {
```
lazartravica, line 16: Why did we change this, wouldn't this do more memory allocations?

lazartravica (approved):
LGTM, and excellent e2e tests.

I just want to know the reason behind changing the way we do the buildroot marshaling?

types/buildroot/buildroot.go
Kourin1996: A simple reason is that we need to add TransactionType if tx has.
According to [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718), TransactionType comes before TransactionData.

```
// Add transaction type before tx
TransactionType || rlp([nonce, gasPrice, gasLimit, ...])
```

MarshalRLPWith creates one RLP value. So if I encode TransactionType inside MarshalRLPWith, data will be like this:

```
// Add transaction type before tx 
rlp([TransactionType, nonce, gasPrice, gasLimit, ...])
```

So we need to encode Transaction first and then add TransactionType

PR #410, opened 2022-02-08
CLI refactor

zivkovicmilos:
# Description

This PR features a complete rewrite of the CLI module in Polygon Edge, the simplification of command additions and the pruning of several now useless packages:
* https://github.com/mitchellh/cli
* https://github.com/imdario/mergo

The new CLI implementation is based on Cobra

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [x] I have updated the official documentation
- [x] I have added sufficient documentation in code

## Testing

- [x] I have tested this code with the official test suite
- [x] I have tested this code manually

### Manual tests

The changes have been tested mostly through the e2e test suite that we have baked in, but also by following the official documentation and checking out different command flows.

# Documentation update

[Relevant Docs PR](https://github.com/0xPolygon/polygon-edge-docs/pull/67)

# Additional comments

Fixes EDGE-44

## CLI changes

The following flags have been renamed:

| Command | Old flag | New flag     |
|---------|----------|--------------|
| `genesis` | `chainid`  | `chain-id`    |
| `server` | `grpc`     | `grpc-address` |

### `dev`

The helpful dev command has been dropped for now, and it's reintroduction should be a part of a smaller task in the future.

## New tasks

During the refactor of the CLI module, several places for improvement have been spotted.

### Dropping the `--seal` concept from Polygon Edge

Explained in detail in [E-394](https://linear.app/trapesys/issue/EDGE-394/consider-if-we-need-the-seal-option-for-consensus)

### Adding unit tests

Explained in detail in [E-397](https://linear.app/trapesys/issue/EDGE-397/add-unit-tests-for-the-cli-module)

### Generating CLI docs

Explained in detail in [E-398](https://linear.app/trapesys/issue/EDGE-398/rely-on-cobra-to-generate-cli-documentation-for-the-docs)

PR #435, opened 2022-03-03
[Fix] Change default jsonrpc binding to all interfaces (0.0.0.0)

### LINKED ISSUE #416: gRPC and JSON RPC APIs improperly bound
### Description

Because both API servers are bound to `127.0.0.1`, neither function correctly when run inside a Docker container. Instead, they need to be bound to `0.0.0.0`.

### Your environment

* macOS Monterey, CentOS 7
* 0.2.1
* develop

### Steps to reproduce

I found this issue by trying to use `polygon-edge` with `docker-compose`. My `yml` file:

```
version: '3.9'
services:
  edge:
    image: 0xpolygon/polygon-edge:0.2.1
    ports:
      - "8545:8545"
    volumes:
      - /home/ec2-user/data:/blk
    working_dir: /blk
    command: polygon-edge server --data-dir /blk-node-1 --chain /blk/genesis.json --libp2p :10001 --seal --dev --log-level debug --access-control-allow-origins "*"
```

A simple `curl -X POST 'http://localhost:8545'` on the docker host will fail. To fix this, you can append this to the command:

```
--jsonrpc "0.0.0.0:8545"
```

### Expected behaviour

Simple `docker run 0xpolygon/polygon-edge` commands and `docker-compose` approaches should work out of the box.

### Actual behaviour

Running `polygon-edge` in a docker container defaults to a non-functional state.

### Proposed solution

The defaults in [`server_command.go`](https://github.com/0xPolygon/polygon-edge/blob/develop/command/server/server_command.go) should bind both grpc and json rpc addresses to all interfaces (`0.0.0.0`) instead of to loopback (`127.0.0.1`).

I would have made a PR, but this is a large change as `127.0.0.1` is copy/pasted throughout in comments and logs. I wanted to make sure this was the agreed upon approach before making the PR.

ZeljkoBenovic:
# Description

This PR changes the default JSON PRC and Prometheus  IP binding to all interfaces ( 0.0.0.0 ) instead of localhost/loopback ( 127.0.0.1 )
This makes much more sense when starting the chain with default settings as this APIs will usually be queried from remote hosts.

Fixed server command usage defaults. For example:
```   --max-outbound-peers int                     the client's max number of outbound peers allowed. Default: 8 (default -1)```
now displays:
```  --max-outbound-peers int                     the client's max number of outbound peers allowed. (default 8)```

# Changes include

# Checklist

## Testing

### Manual tests

Started the server with ```--jsonrpc :10001``` and without ```--jsonrpc``` flag.
Started the server with ```--prometheus :5001``` flag.
The JSON RPC API always binds to all interfaces and can be queried from remote host
The Prometheus API binds to all interfaces on defined port

# Documentation update

[Doc PR#70](https://github.com/0xPolygon/polygon-edge-docs/pull/70)

# Additional comments

Fixes EDGE-421

### CODE REVIEW

File: command/helper/config.go [hunk 1]
```diff
@@ -72,8 +72,8 @@ const defaultBlockTime uint64 = 2
 // DefaultConfig returns the default server configuration
 func DefaultConfig() *Config {
 	return &Config{
-		Chain:          "test",
-		DataDir:        "./test-chain",
+		Chain:          "polygon-edge",
+		DataDir:        "./polygon-edge-chain",
```
zivkovicmilos: You should include this in the .gitignore file as well

zivkovicmilos:
@ZeljkoBenovic 

Adding a `don't merge` label here as it's directly conflicting with PR #410

zivkovicmilos (approved):
Looks good 👍 

Please don't merge this before #410

0xAleksaOpacic (approved):

Kourin1996 (approved):

zivkovicmilos:
@ZeljkoBenovic 
Removed the `don't merge` label, you can pull the latest changes from the develop branch 👍

command/helper/config.go
ZeljkoBenovic: Added to .gitignore in 189841266f8a705f8ae512daceccb5556794089f

ZeljkoBenovic:
Merged the Cobra CLI changes, changed default ip bindings and fixed few minor things with server cli command help output

command/server/server.go
```diff
@@ -182,20 +166,14 @@ func setFlags(cmd *cobra.Command) {
...
 
 	cmd.Flags().Uint64Var(
 		&params.rawConfig.BlockTime,
 		blockTimeFlag,
 		defaultConfig.BlockTime,
-		fmt.Sprintf(
-			"minimum block time in seconds. Default: %ds",
-			defaultConfig.BlockTime,
-		),
+		"minimum block time in seconds.",
```
zivkovicmilos: Please remove punctuation at the end of flag descriptions, for all commands changed in this PR 🙏

command/helper/helper.go
```diff
@@ -219,6 +220,20 @@ func ResolveAddr(address string) (*net.TCPAddr, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse addr '%s': %w", address, err)
+	}
+
+	if addr.IP == nil {
+		addr.IP = net.ParseIP("0.0.0.0")
+	}
+
+	return addr, nil
+}
```
zivkovicmilos: Is there any chance we can reuse `ResolveAddr` (call it) here?

zivkovicmilos: The only part that's different is the IP setting

zivkovicmilos: Maybe change `ResolveAddr` to include a default IP string as a param, and drop this method

zivkovicmilos (approved):
Looks great 💯 

I've left some really minor comments 👍

command/server/server.go
ZeljkoBenovic: Fixed in 4b6856b31184d7b76196a05689f226ecda3a3b3b

command/helper/helper.go
ZeljkoBenovic: Removed this func and modified ```ResolveAddr``` in commit b7595a3bfa0d5b769834a4d821b8a19cfc57a9b6

Aleksao998: Agree, something like
```
func ResolveAddr(address string, defaultIPAddress string) (*net.TCPAddr, error) {
	addr, err := net.ResolveTCPAddr("tcp", address)

	if err != nil {
		return nil, fmt.Errorf("failed to parse addr '%s': %w", address, err)
	}

	if addr.IP == nil {
		addr.IP = net.ParseIP(defaultIPAddress)
	}

	return addr, nil
}
```

0xAleksaOpacic (approved):
LGTM!

command/helper/helper.go
```diff
@@ -30,6 +31,18 @@ func (r *ClientCloseResult) GetOutput() string {
...
+	AllInterfaces string
+}
+
+var (
+	IPBinding = IPBindings{
+		Localhost:     "127.0.0.1",
+		AllInterfaces: "0.0.0.0",
+	}
+)
+
```
zivkovicmilos: Why not make this an enum?

Doesn't need to be a global var struct, just a global enum is fine 🙂 

```go
type IPBinding string

const (
   LocalHostBinding IPBinding = "127.0.0.1",
   AllInterfacesBinding IPBinding = "0.0.0.0"
)
```

Then modify the `ResolveAddr` to take IPBinding instead of string

ZeljkoBenovic: Thank you for this suggestion 🙂 Implemented as enum b4b23dca61a8cfea98ef5fd27c470cc1abb05dc8

PR #459, opened 2022-03-14
[TxPool] Reduce lock time in resetAccounts

dbrajovic:
# Description

Fixes EDGE-434

This PR reduces the amount of time account lock(s) are held during `resetAccounts`.  

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -162,7 +162,7 @@ func (a *account) setNonce(nonce uint64) {
 	atomic.StoreUint64(&a.nextNonce, nonce)
 }
 
-func (a *account) reset(nonce uint64) (
+func (a *account) reset(nonce uint64, promoteCh chan<- promoteRequest) (
 	prunedPromoted,
 	prunedEnqueued []*types.Transaction,
 ) {
@@ -191,6 +191,15 @@ func (a *account) reset(nonce uint64) (
 
 	//	update nonce expected for this account
 	a.setNonce(nonce)
+
+	//	it is important to signal promotion while
+	//	the locks are held to ensure no other
+	//	handler will mutate the account
+	if first := a.enqueued.peek(); first != nil &&
+		first.Nonce == nonce {
+		// first enqueued tx is expected -> signal promotion
+		promoteCh <- promoteRequest{account: first.From}
+	}
 
 	return
 }
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -135,43 +135,6 @@ func (m *accountsMap) allTxs(includeEnqueued bool) (
 	return
 }
 
-// resetWithNonce updates all accounts with the new nonce and clears any stale transactions.
-// May signal a promotion request if the account is eligible after pruning.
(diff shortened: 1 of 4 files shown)


txpool/account.go
```diff
@@ -135,9 +135,46 @@ func (m *accountsMap) allTxs(includeEnqueued bool) (
...
+			prunedPromoted...,
+		)
+
+		allPrunedEnqueued = append(
+			allPrunedEnqueued,
+			prunedEnqueued...,
+		)
+
+		return true
+	})
```
Kourin1996: Why do you need to scan all accounts? I think the only thing you need to do is to scan newNonces. Number of accounts is growing, while number of newNonces keeps below a certain number.

dbrajovic: Mostly because I wanted to use the `m.Range` functionality of a `sync.Map` believing it to be faster than a regular `for` loop in this case. 

It's true that for some account there may no be an updates - in which case, no locking is occurring either. I guess the final decision should be based on what works faster.

The most important difference in this PR is that "state updating" (index, gauge, metrics) is not done while pruning an account.

Kourin1996: Sorry, I don't get it. I think sync.Map::Range won't block from other accessing to the same key.

dbrajovic: Yes, you're right - it won't block other goroutines from the same entry/

It boils down to which of these 2 is faster (pseudocode below):

`Range` approach:
```
m.Range(key, value) {
    addr = key

    if !newNonces_contains(addr)  {
        continue
    }

    resetAccount(addr)
}
```

or

regular `for` loop:
```
for addr in newNonces {
    if !account_exists(addr) {
        continue
    }

    resetAccount(addr)
}
```

The actual locking is performed in `resetAccount()`.

Kourin1996: I think scanning all accounts takes much computation cost. second approach is better

Kourin1996: @zivkovicmilos What do you think?

lazartravica: @dbrajovic let's do a benchmark about this, to be clear about the performance benefits

zivkovicmilos: I agree with @Kourin1996 on this one, purely because the number of accounts who are gonna see updates is usually going to be less than the number of all accounts currently loaded in memory - that's why I think approach 2 is better, since you're only ranging on accounts that need to be ranged upon.

The performance boost is, I believe, minimal for small data sets, you'd need to really stress it to see any kind of gains, I'm all for just going with the second approach and dropping the benchmark 👍

zivkovicmilos (approved):
Believe it or not I couldn't find anything to nitpick with the PR 🙂 

I was about to write a comment on the lock release order, but then realized go runs `defer`s in LIFO order.

Looks great, please check out @Kourin1996's comment for a minor optimization 💯

txpool/account.go
dbrajovic: Sounds good to me 👍 

[Refactored](https://github.com/0xPolygon/polygon-edge/pull/459/commits/45284cc1fd33b30f9de52ad875f5a8eca48d6a6e)

Kourin1996 (approved):
LGTM

PR #201, opened 2021-11-08
Rename difficulty to totalDifficulty

dbrajovic:
# Description

Change varaible names and function calls where `totalDifficulty` is meant by `difficulty`

# Changes include

- [x] Bugfix (non-breaking change that solves an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have tested this code
- [ ] I have updated the README and other relevant documents (guides...)
- [ ] I have added sufficient documentation both in code, as well as in the READMEs

PR #465, opened 2022-03-21
[Bridge] Checkpoint

Kourin1996:
Fix Edge-459

# Description

This PR is a base PR for Bridge Checkpoint, checked out from `bridge` branch.

TODO for Phase 1
- [x] Create bridge/checkpoint package bridge
- [x] Generate new Checkpoint at the beginning of every epoch
- [x] Create Transport for Checkpoint Gossip
- [x] Sign generated Checkpoint and gossip it
- [x] Show logs once collected 2/3+ signatures

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


bridge/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"github.com/0xPolygon/polygon-edge/bridge/checkpoint/transport"
+	"github.com/0xPolygon/polygon-edge/bridge/sam"
+	"github.com/0xPolygon/polygon-edge/network"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/hashicorp/go-hclog"
+)
+
+const (
+	EpochSize = 100
```
0xAleksaOpacic: What value we want to have for checkpoint EpochSize?
1. Some default value
2. Same as epoch size for consensus
3. Cli config to setup + default value

bridge/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (c *checkpoint) Close() error {
+	return nil
+}
+
+func (c *checkpoint) SetValidators(vals []types.Address) {
+	c.validators = vals
+}
+
+func (c *checkpoint) CreateNewCheckpoint() error {
```
0xAleksaOpacic: When are we calling this method? Is it from ibft insertBlock, on every **checkpoint new epoch**?

bridge/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	// Step1: Get Latest information stored in RootChain contract
+	currentEpoch, err := c.rootchainContract.GetCurrentHeaderBlock()
+	if err != nil {
+		return err
+	}
+
+	lastChildBlock, err := c.rootchainContract.GetLastChildBlock()
```
0xAleksaOpacic, line 79: If we know what is currentEpoch in the root chain, do we need the last-child block? 
We can use currentEpoch*epochSize + 1 for lower bound

bridge/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Step2: Determine the range of next checkpoint and get blocks from local chain
+	start, end := c.determineCheckpointRange(lastChildBlock)
+	blocks, err := c.blockchain.GetBlocks(start, end)
```
0xAleksaOpacic: Is there a chance that some epoch is not going to get inserted in the rootchain contract?
If yes, and let's say our chain is on epoch 7, and the rootchain has been last updated on epoch 3. 
Will our range be from epoch 3 to latest block, or only one epoch (3-4)?

0xAleksaOpacic (approved):
Great job, I have left you a few questions so I can understand structure better, overall looks good

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: @Kourin1996 Ping

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: @Kourin1996 Ping

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: @Kourin1996 Ping

bridge/checkpoint/checkpoint.go
Kourin1996: @Aleksao998 As discussed yesterday, must be No.3 and Epoch size can be given from CLI config or genesis.json

bridge/checkpoint/checkpoint.go
Kourin1996: @Aleksao998 on every checkpoint new epoch, which means should call when IBFT consensus begins for block #201 if next checkpoint is for blocks from 100 and 200

bridge/checkpoint/checkpoint.go
Kourin1996: @Aleksao998 I created mock for phase 1. Actually we need to know last child block because no guarantee that every checkpoint is created from fixed number of blocks

0xAleksaOpacic: Thank you for an explanation!

bridge/checkpoint/checkpoint.go
Kourin1996: Yes that's why we'll have Ack and NoAck mechanism

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: Thank you!

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: Agree, @lazartravica you said we should go for the same value (consensus epochSize = checkpoint epochSize), but I agree on this with Yoshiki. What do you think?

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: So checkpoint range should be [fromLastChildBlock, toLastBlockInPolygonEdge].

Kourin1996: Ideally

bridge/checkpoint/checkpoint.go
Kourin1996: @Aleksao998 We can use same value but default IBFT epoch size is too big

0xAleksaOpacic: That's why I agree with you that we should have separate flag.

0xAleksaOpacic (approved):
LGTM! Once we merge all 3 PR into this branch we would need to review it again.

zivkovicmilos:
Closing this PR as it's been archived.

PR #466, opened 2022-03-21
Start checkpoint process on every epoch start

0xAleksaOpacic:
# Description

Fixes EDGE-457

# Changes include

# Checklist

## Testing

### Manual tests

I tested this code manually with logging all blocks after fetching them in StartNewCheckpoint (checkpoint.go)
Steps:
1. Add logs
2. Start network with --bridge and --epoch-size 10 flag 
3. On every 10th block, all blocks from the previous epoch should get logged


zivkovicmilos (changes requested):
Please resolve the linting errors and re-request a review

consensus/ibft/ibft.go
```diff
@@ -1026,6 +1026,13 @@ func (i *Ibft) insertBlock(block *types.Block) error {
 	// broadcast the new block
 	i.syncer.Broadcast(block)
 
+	// On every epoch start create checkpoint
+	if i.IsLastOfEpoch(block.Number()) {
+		if hookErr := i.runHook(CreateCheckpointHook, block.Number(), i.epochSize); hookErr != nil {
+			i.logger.Error(fmt.Sprintf("Unable to run hook %s, %v", CreateCheckpointHook, hookErr))
+		}
```
Kourin1996: Ideally IBFT must not know about Checkpoint. Now the base is depending on extensions.

Why not use AcceptStateLogHook or InsertBlockHook? (To be honest, AcceptStateLogHook will be need to rename to common name :))

consensus/ibft/ibft.go
```diff
@@ -1026,6 +1026,13 @@ func (i *Ibft) insertBlock(block *types.Block) error {
 	// broadcast the new block
 	i.syncer.Broadcast(block)
 
+	// On every epoch start create checkpoint
+	if i.IsLastOfEpoch(block.Number()) {
```
Kourin1996: You can check this check inside hook

Kourin1996 (changes requested):

consensus/ibft/ibft.go
0xAleksaOpacic: Fixed ac388779230090c3ee8dc32f1fa312ecf14305e3

consensus/ibft/ibft.go
0xAleksaOpacic: Fixed ac388779230090c3ee8dc32f1fa312ecf14305e3

0xAleksaOpacic:
> Please resolve the linting errors and re-request a review

I fixed the Linter error from this PR, the remaining should be fixed on the main checkpoint branch

Kourin1996 (approved):
LGTM

blockchain/blockchain.go
```diff
@@ -1047,17 +1047,20 @@ func (b *Blockchain) GetBlockByNumber(blockNumber uint64, full bool) (*types.Blo
 	return b.GetBlockByHash(blockHash, full)
 }
 
-// GetBlocks returns the bunch of block
-func (b *Blockchain) GetBlocks(start, end uint64, full bool) []*types.Block {
+// GetBlocks returns the blocks between start and end block number
+func (b *Blockchain) GetBlocks(start, end uint64, full bool) ([]*types.Block, bool) {
```
zivkovicmilos: Why modify the function signature?

The ability to fetch all blocks is directly coupled with the success of this method, so there is no need to return a boolean indicating success - just return `nil`, or maybe an `error` when you fail to fetch all blocks.

cc @Kourin1996

bridge/checkpoint/checkpoint.go
```diff
@@ -117,9 +122,12 @@ func (c *checkpoint) StartNewCheckpoint(epochSize uint64) error {
 	return nil
 }
 
-func (c *checkpoint) determineCheckpointRange(lastChildBlock, epochSize uint64) (uint64, uint64) {
-	// TODO: implement
-	return lastChildBlock + 1, lastChildBlock + epochSize
+func (c *checkpoint) determineCheckpointRange(lastChildBlock uint64) (uint64, uint64) {
+	header := c.blockchain.Header()
+
+	lastBlockNumber := header.Number
```
zivkovicmilos: You can inline this

bridge/checkpoint/checkpoint.go
```diff
@@ -117,9 +122,12 @@ func (c *checkpoint) StartNewCheckpoint(epochSize uint64) error {
 	return nil
 }
 
-func (c *checkpoint) determineCheckpointRange(lastChildBlock, epochSize uint64) (uint64, uint64) {
-	// TODO: implement
-	return lastChildBlock + 1, lastChildBlock + epochSize
+func (c *checkpoint) determineCheckpointRange(lastChildBlock uint64) (uint64, uint64) {
```
zivkovicmilos: What do you think about changing the method signature to:
`func (c *checkpoint) determineCheckpointRange(lastChildBlock uint64) (start uint64, end uint64)`

and then just reference these values inside before the return? I think it's much clear that way what the return values are, especially because the method doesn't have any kind of description.

consensus/ibft/bridge.go
```diff
@@ -140,6 +140,13 @@ func (b *BridgeMechanism) consumeStateTransactionsHook(numberParam interface{})
 		b.bridge.StateSync().Consume(tx)
 	}
 
+	// On every epoch end, create the checkpoint
+	if b.ibft.IsLastOfEpoch(block.Number()) {
+		if err := b.bridge.StartNewCheckpoint(b.ibft.epochSize); err != nil {
+			return err
+		}
+	}
```
zivkovicmilos: You can move this inner if check outside, and just return if the first check fails (is not the last of the epoch), to avoid having spaghetti

zivkovicmilos: @Kourin1996 

As far as I can see, this is triggered in the `InsertBlock` hook for the Bridge mechanism, which is called during _both_ commit state and sync state. Why create checkpoints during sync state?

zivkovicmilos (approved):
Looks good 💯 

I've left small comments 🙏

consensus/ibft/bridge.go
0xAleksaOpacic: Hmm, I am not sure I agree with you on this.  I agree that we should avoid spaghetti code but in my opinion, this way is clearer that checkpoints are created only at the end of the epoch
Also, everything below the first check will be unreachable if the condition is not met. I can imagine someone extending this method and adding something on the end of the method not thinking about the condition above. 
```
	if b.ibft.IsLastOfEpoch(block.Number()) {
		return nil
	}

	if err := b.bridge.StartNewCheckpoint(b.ibft.epochSize); err != nil {
		return err
	}

        .....
```

0xAleksaOpacic: Yes, you are right, I did not know that this hook is called from sync state. 
My first idea was to create a new hook which would be called from insertBlock. Buy @Kourin1996 suggested we should stick with the one we have. 
@Kourin1996 How do you think we should handle this?

Kourin1996: Why not use `AcceptStateLogHook` or some hooks called in runAcceptState. Actually don't have to call this immediately after you insert block

blockchain/blockchain.go
0xAleksaOpacic: Thank you for the suggestion, fixed https://github.com/0xPolygon/polygon-edge/commit/b65283c9249dc562ea3597d3db2fdf69c78213fd

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: Thank you for the suggestion, fixed https://github.com/0xPolygon/polygon-edge/commit/b65283c9249dc562ea3597d3db2fdf69c78213fd

bridge/checkpoint/checkpoint.go
Aleksao998: Thank you for suggestion, fixed b65283c9249dc562ea3597d3db2fdf69c78213fd

consensus/ibft/bridge.go
0xAleksaOpacic: I see that AcceptStateLogHook is used for logging validators for POA and POS, and updating validators for Bridge. I do not think that starting checkpoint should happen in a method called updateValidatorsHook. It's not intuitive to start a checkpoint from a hook called AcceptStateLogHook. 

Also somewhere in the future, we would maybe want to use this hook in some other place and we would not be able because of checkpointing process.

Kourin1996: Yes, this is current problem. The hook Naming convention is not good because it's depending on extension. 

Buy actually I meant why not you call CreateCheckpoint in runAcceptState? I was just talking about the time when you call it.

0xAleksaOpacic: I do not see any reason why we would not call CreateCheckpoint from runAcceptState. 
We can check if the parent block was the end of an epoch. And if it was to create checkpoint.

Kourin1996: then why don't you move?

0xAleksaOpacic: We can create a checkpoint from this hook, then refactor bridge hooks in the main bridge branch?

Kourin1996: Only thing you need to care about is that latest block height must be greater than the end block height of next checkpoint. What's the difference between calling CreateCheckPoint on InsertBlockHook and hook called from runAcceptState? For me, I don't see any difference.

0xAleksaOpacic: Fixed 6004cfd2083c5f5d31fd1243b005af00c518448f. @Kourin1996 check this commit please. I have done manual testing to see if everything works

bridge/checkpoint/checkpoint.go
0xAleksaOpacic: Thank you for the suggestion, fixed https://github.com/0xPolygon/polygon-edge/commit/b65283c9249dc562ea3597d3db2fdf69c78213fd

consensus/ibft/bridge.go
zivkovicmilos: +1 on what @Kourin1996 said, everything was on point

zivkovicmilos: Why do you think this
```go
        if b.ibft.IsLastOfEpoch(block.Number()) {
		if err := b.bridge.StartNewCheckpoint(b.ibft.epochSize); err != nil {
			return err
		}
	}
```

is cleaner than this?:
```go
        // ...
        if !b.ibft.IsLastOfEpoch(block.Number()) {
		return nil
	}

	return b.bridge.StartNewCheckpoint(b.ibft.epochSize)
```

zivkovicmilos: @Aleksao998 
Why are you calling the checkpoint start from the `updateValidatorsHook`?

0xAleksaOpacic: Fixed df8a837dd41fadcdf9bd23709ed0b2e78cd65aed

0xAleksaOpacic: Fixed df8a837dd41fadcdf9bd23709ed0b2e78cd65aed

dbrajovic (approved):

consensus/ibft/bridge.go
zivkovicmilos: Closing as it's been fixed in [ec122b7](https://github.com/0xPolygon/polygon-edge/pull/466/commits/ec122b7415e1b37c8cd7c55a8cd2700a006ab1e7) and subsequent commits

PR #468, opened 2022-03-22
[Bridge] Checkpoint gossip

dbrajovic:
# Description

PR that implements checkpoint gossiping. 

Fixes EDGE-458

# Changes include

# Checklist

## Testing


bridge/checkpoint/transport/proto/transport.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package v1;
+
+option go_package = "/bridge/checkpoint/transport/proto";
+
+message SignedMessage {
```
Kourin1996: @dbrajovic I noticed I can't use same name among all protocs in project. Build will fail. Can you rename this to some other name?

Kourin1996:
@dbrajovic Please add build path for checkpoint into `make protoc` and run it.

bridge/checkpoint/transport/proto/transport.proto
dbrajovic: What's wrong with having this one ? It's encapsulated within the `checkpoint` module and represents the same concept also used in `statesync`. 

Other than that, I'm open to naming suggestions.

Kourin1996: @dbrajovic For now, you can see this error on building. Please check it

Kourin1996 (approved):
Looks good to me 👍

0xAleksaOpacic (approved):
LGTM

bridge/checkpoint/transport/transport.go
```diff
@@ -62,15 +66,36 @@ func (t *libp2pGossipTransport) Subscribe(handler func(interface{})) error {
 
 		// and call handler
 
-		// protoMessage, ok := obj.(*proto.SignedMessage)
-		// if !ok {
-		// 	t.logger.Warn("received unexpected typed message", "message", obj)
+		protoMsg, ok := obj.(*proto.CheckpointMessage)
+		if !ok {
+			t.logger.Warn("received unexpected typed message", "message", obj)
```
zivkovicmilos: Why is this a warning and not an error?

bridge/checkpoint/transport/transport.go
```diff
@@ -62,15 +66,36 @@ func (t *libp2pGossipTransport) Subscribe(handler func(interface{})) error {
 
 		// and call handler
```
zivkovicmilos: Are these TODOs leftover?

bridge/checkpoint/transport/transport.go
```diff
@@ -62,15 +66,36 @@ func (t *libp2pGossipTransport) Subscribe(handler func(interface{})) error {
...
+		var message interface{}
+
+		switch protoMsg.Type {
+		case proto.CheckpointMessage_CHECKPOINT:
+			checkpoint := ctypes.Checkpoint{}
+			if err := checkpoint.UnmarshalRLP(protoMsg.Payload); err != nil {
+				t.logger.Error("unable to unmarshal payload from message", "err", err)
+
+				return
+			}
```
zivkovicmilos: This unmarshalling can be moved below the switch, so in your case statements you only set the unmarshaller base

zivkovicmilos (approved):
Looks great 💯 

Very nice touch of using the payload and RLP marshalling it based on the type - it saves us having to define more topics for the clunky pubsub implementation 👏 

Please resolve the linting errors 🙏

bridge/checkpoint/transport/transport.go
zivkovicmilos: Resolving this as you need to create a concrete struct object after unmarshalling, and I've found no easy and cleaner way to do this

bridge/checkpoint/transport/transport.go
dbrajovic: [Changed to Error](https://github.com/0xPolygon/polygon-edge/pull/468/commits/d22b06d42c1bb9df130990e51b2785d206ce5c84)

bridge/checkpoint/transport/transport.go
dbrajovic: [Removed](https://github.com/0xPolygon/polygon-edge/pull/468/commits/bac6047ea80243c7eabe7ad4b8224fe1a4ad058e)

PR #482, opened 2022-03-30
[Blockchain/IBFT] Update snapshot after processBlock ends

dbrajovic:
# Description

This PR delays the call to `processHeaders()` until after the block has been verified and processed completely. 

Previously, snapshot updating was done during `verifyHeaders()` despite `processBlock()` potentially failing afterwards. 

# Changes include

# Checklist

## Testing

### Manual tests

Tested manually with a 4-node cluster.  Steps:

1. Spin up 4 nodes (`develop` build)
2. Sequentially restart all nodes (`fix/snapshot_update` build)
3. Nothing breaks

Tested PoA voting mechanism (manually). Steps:

1. Spin up 4 nodes (`fix/snapshot_update` build)
2. While the cluster is running, spin up a 5th node (`--seal`)
3. Cluster votes for 5th node to be included in the validator set.
4. Observe the 5th node is able to join consensus on new blocks. 

## Additional comments

Fixes EDGE-465

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -1255,12 +1255,12 @@ func (i *Ibft) VerifyHeader(parent, header *types.Header) error {
 		return err
 	}
 
-	// process the new block in order to update the snapshot
-	if err := i.processHeaders([]*types.Header{header}); err != nil {
-		return err
-	}
-
 	return nil
+}
+
+// ProcessHeaders updates the snapshot based on previously verified headers
+func (i *Ibft) ProcessHeaders(headers []*types.Header) error {
+	return i.processHeaders(headers)
 }
 
 // GetBlockCreator retrieves the block signer from the extra data field
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -53,6 +53,7 @@ type Blockchain struct {
 
 type Verifier interface {
 	VerifyHeader(parent, header *types.Header) error
+	ProcessHeaders(headers []*types.Header) error
 	GetBlockCreator(header *types.Header) (types.Address, error)
 	PreStateCommit(header *types.Header, txn *state.Transition) error
 }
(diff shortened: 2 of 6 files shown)


consensus/ibft/ibft.go
```diff
@@ -1255,14 +1255,14 @@ func (i *Ibft) VerifyHeader(parent, header *types.Header) error {
...
-	}
-
 	return nil
 }
 
+// ProcessHeaders updates the snapshot based on previously verified headers
+func (i *Ibft) ProcessHeaders(headers []*types.Header) error {
+	return i.processHeaders(headers)
+}
+
```
Kourin1996, line 1265: Now processHeaders is required for only IBFT. Why blockchain needs to call processHeaders? And do you think whether it's possible to call in IBFT? (For example, you call processHeaders after WriteBlock)

zivkovicmilos: I agree with Yoshiki 👍 

Is there any way to move out this processing logic to be dependent only on the actual IBFT implementation, and not have the `blockchain` module call it?

dbrajovic: Initially I tried going in the same direction - moving snapshot updating out of `WriteBlock`. Sadly, there are a few other modules (besides consensus) that are running this method under the assumption it also updates state. (`ProcessHeaders` was in `VerifyHeaders` before).

Making `processHeaders` as part of Ibft only would break this assumption and there also wouldn't be any way to have this logic available at callsite:

restore.go:
[importBlocks](https://github.com/0xPolygon/polygon-edge/blob/86e35dfe6a6e191fd961236784617a4fe0df76d3/archive/restore.go#L76)
syncer.go: 
[WatchSyncWithPeer](https://github.com/0xPolygon/polygon-edge/blob/86e35dfe6a6e191fd961236784617a4fe0df76d3/protocol/syncer.go#L603)
[BulkWyncWithPeer](https://github.com/0xPolygon/polygon-edge/blob/86e35dfe6a6e191fd961236784617a4fe0df76d3/protocol/syncer.go#L678)

zivkovicmilos: @Kourin1996 I think it's best if we leave it here in `WriteBlock`, since it has coupling with `ProcessBlock` - from places where `ProcessBlock` would not be accessible (syncer and restore)

zivkovicmilos (approved):

Kourin1996 (approved):
LGTM

0xAleksaOpacic (approved):

PR #484, opened 2022-03-31
[Bridge] Add checkpoint generation (Merkle)

zivkovicmilos:
# Description

This PR imports the [fastmerkle](https://github.com/Trapesys/fastmerkle) package for checkpoint hash generation, and adds support for generating the actual checkpoint.

Some parts of the checkpoint process are left to be implemented in the base PR (like proposer fetching).

# Changes include

# Checklist

## Testing


bridge/checkpoint/checkpoint.go
```diff
@@ -126,8 +129,51 @@ func (c *checkpoint) determineCheckpointRange(lastChildBlock uint64) (start uint
 }
 
 func (c *checkpoint) generateCheckpoint(blocks []*types.Block) (*ctypes.Checkpoint, error) {
+	// Generate the Merkle root for the block hashes
+	blockHashes := make([][]byte, len(blocks))
+	for i, block := range blocks {
+		blockHashes[i] = block.Hash().Bytes()
```
Kourin1996: Block Hash is correct?

FYI, https://github.com/maticnetwork/bor/blob/d687d0d2de163e980c083f313504626cb1d3bdf3/consensus/bor/api.go#L161-L166

Kourin1996 (changes requested):

bridge/checkpoint/checkpoint.go
```diff
@@ -126,8 +129,51 @@ func (c *checkpoint) determineCheckpointRange(lastChildBlock uint64) (start uint
...
+		)
+	}
+
+	// Create a placeholder checkpoint
+	return &ctypes.Checkpoint{
+		Proposer:        c.getProposer(c.getCheckpointEpoch()),
+		Start:           blocks[0].Number(),
+		End:             blocks[len(blocks)-1].Number(),
+		RootHash:        types.BytesToHash(blocksMerkleTree.RootHash()),
+		AccountRootHash: types.BytesToHash(validatorsMerkleTree.RootHash()),
```
0xAleksaOpacic, line 175: I know this is from base branch, but shouldn't this be called validatorsRootHash or something like that. Just want to hear your opinion, do not expect that you do any changes in this PR.

bridge/checkpoint/checkpoint.go
zivkovicmilos: Wait, so the Block Hash for the checkpoint is actually constructed from:
* The block number
* The generation timestamp
* The root of the txn trie
* The root of the receipts trie
?

zivkovicmilos: This is how we construct header hashes for Edge:
https://github.com/0xPolygon/polygon-edge/blob/f59d3ca12586a910b3f8744ffbefdbbaa19a9e02/consensus/ibft/header_hash.go#L27-L42

zivkovicmilos: Please confirm that we should send this bundle for the Merkle root generation:
https://github.com/0xPolygon/polygon-edge/pull/484#discussion_r839655068

0xAleksaOpacic (approved):

bridge/checkpoint/checkpoint.go
zivkovicmilos: @Kourin1996

bridge/checkpoint/checkpoint.go
Kourin1996: Yes, and this is not RLP encode.

bridge/checkpoint/checkpoint.go
Kourin1996: @Aleksao998 To be honest, I don't get what your question is. Are you asking about FieldName?

0xAleksaOpacic: Yes can we change field name?

bridge/checkpoint/checkpoint.go
zivkovicmilos: Fixed 👍 

[4e758a8](https://github.com/0xPolygon/polygon-edge/pull/484/commits/4e758a8300d1d9c8423c12c292e89226d128c876)

bridge/checkpoint/checkpoint.go
Kourin1996: Basically AccountRootHash comes from original so that we can follow and refer to original code. But renaming won't break, I think

Kourin1996 (approved):
LGTM

bridge/checkpoint/checkpoint.go
zivkovicmilos: @Aleksao998 

What do you think we should call this?

lazartravica: Use naming thats logical and understandable for you above everything else

0xAleksaOpacic: ValidatorsRootHash is logical for me or something like that. But lets change it on main checkpoint branch, and merge this one

PR #499, opened 2022-04-13
Fix secrets generate issues when the type is aws-ssm

ZeljkoBenovic:
# Description

This PR fixes some issues that were overlooked in CLI refactor.
It fixes the issue with `secrets generate` command when the type is `aws-ssm`. 
Removes some required flags, as they already have default values.
Fixes the `Extra:` field in `SecretsManagerConfig` , which was initialized with nil.

# Changes include

# Checklist

## Testing

### Manual tests

Generate secretsManagerConfig.json file by running 
`go run . secrets generate --type aws-ssm --extra region=eu-central-1,ssm-parameter-path=/polygon-edge/nodes --name node1`
and use the file to store init secrets in AWS SSM 
`go run . secrets init --config secretsManagerConfig.json`

Confirmed working by inspecting AWS SSM in web console.

### DIFF
--- a/command/secrets/generate/params.go
+++ b/command/secrets/generate/params.go
@@ -3,9 +3,10 @@ package generate
 import (
 	"errors"
 	"fmt"
+	"strings"
+
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/0xPolygon/polygon-edge/secrets"
-	"strings"
 )
 
 var (
@@ -44,9 +45,6 @@ type generateParams struct {
 
 func (p *generateParams) getRequiredFlags() []string {
 	return []string{
-		dirFlag,
-		tokenFlag,
-		serverURLFlag,
 		nameFlag,
 	}
 }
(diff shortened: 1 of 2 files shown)


command/secrets/generate/params.go
```diff
@@ -44,9 +45,6 @@ type generateParams struct {
 
 func (p *generateParams) getRequiredFlags() []string {
 	return []string{
-		dirFlag,
-		tokenFlag,
-		serverURLFlag,
```
zivkovicmilos, line 49: These flags are actually required for the base file system secrets generation.

You should add a check when validating the flags to make sure they're present if the mode is disk mode.

zivkovicmilos (approved):
Looks great 💯 

Left a minor comment 🙏

command/secrets/generate/params.go
ZeljkoBenovic: Please correct me if I'm wrong, but I don't think we use `secrets generate` command when we want to use local FS. The command we use is just `secrets init`.

zivkovicmilos: You're completely right, I've overlooked this 🤦 

Resolving

Kourin1996 (approved):

PR #500, opened 2022-04-15
Add gzip stream compression support

zivkovicmilos:
# Description

This PR adds support for libp2p stream compressions, namely it introduces 2 stream compression types:
* no compression
* gzip

# Changes include

# Checklist

## Testing

### Manual tests

Regular cluster deployment with compression support enables.
Tested multiple nodes with different compression configurations

# Documentation update

// TODO

# Additional comments

// TODO post benchmark results

### DIFF
--- a/command/server/init.go
+++ b/command/server/init.go
@@ -3,6 +3,7 @@ package server
 import (
 	"fmt"
 	"github.com/0xPolygon/polygon-edge/network/common"
+	"github.com/0xPolygon/polygon-edge/network/compress"
 	"math"
 	"net"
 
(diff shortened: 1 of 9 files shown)


ferranbt:
Is this change backward compatible with running clients?

zivkovicmilos:
Hey @ferranbt,

Yes! The idea is for this compression layer to be optional, where clients would negotiate which compression layer to use (if any). This way, if one of them supports compression, but the other one doesn't, or they support differing compression layers - they can always fall back to uncompressed streams.

The PR is still in draft because we're playing around with this idea in order to get some metrics on the savings. It will see other drastic changes until it's ready for review and mergeable - but the base idea will stay the same: allow clients to negotiate compression layers for streams, and fallback to uncompressed streams in case it's not possible to come to an agreement.

zivkovicmilos:
Closed due to archiving

PR #513, opened 2022-04-21
[IBFT] Change quorum size calculation

dbrajovic:
# Description

This PR replaces the old calculation of finding the minimum number of validators reaching quorum.

`N` - number of validators in network
`F = floor((N-1) / 3)` - number of (allowed) faulty nodes in a given network

Previously :

`quorum = 2 * F + 1`

Now:

`quorum = ceil(2/3 * N)`

# Changes include

# Checklist

## Testing

### Manual tests

### Additional comments
Fixes EDGE-514

The formula (previously used) for calculating the Quorum size, which is `2F + 1` did not scale well with the number of validators in a given network. Note the following differences:

`2F + 1`:

- `N = 1`, `F = 0` -> `Q = 1`
- `N = 2`, `F = 0` -> `Q = 1`
- `N = 3`, `F = 0` -> `Q = 1`
- `N = 4`, `F = 1` -> `Q = 3`
- `N = 5`, `F = 1` -> `Q = 3`
- `N = 6`, `F = 1` -> `Q = 3`
- `N = 7`, `F = 2` -> `Q = 5`
- `N = 8`, `F = 2` -> `Q = 5`
- `N = 9`, `F = 2` -> `Q = 5`

`ceil(2/3 * N)`:
 from [table](https://www.kaleido.io/blockchain-blog/consensus-algorithms-poa-ibft-or-raft#:~:text=Istanbul%20Byzantine%20Fault%20Tolerance%20(IBFT)) (for `N=3` the formula does not apply, so a special [if statement](https://github.com/0xPolygon/polygon-edge/blob/44c154be5602fe978d70f9f64ad40a84ccf257d6/consensus/ibft/state.go#L311) was used)

- `N = 1`, `F = 0` -> `Q = 1`
- `N = 2`, `F = 0` -> `Q = 2`
- `N = 3`, `F = 0` -> `Q = 3`
- `N = 4`, `F = 1` -> `Q = 3`
- `N = 5`, `F = 1` -> `Q = 4`
- `N = 6`, `F = 1` -> `Q = 4`
- `N = 7`, `F = 2` -> `Q = 5`
- `N = 8`, `F = 2` -> `Q = 6`
- `N = 9`, `F = 2` -> `Q = 6`


Kourin1996:
@dbrajovic Can you write down the reason so that we can look back in the future?

consensus/ibft/state.go
```diff
@@ -97,12 +98,15 @@ func (c *currentState) setState(s IbftState) {
...
-	// (the current node tallying the messages will include its own message)
-	return 2 * c.validators.MaxFaultyNodes()
+func NumValid(validators ValidatorSet) int {
+	//	if the number of validators is less than 4,
+	//	then the entire set is required
+	if validators.MaxFaultyNodes() == 0 {
+		return validators.Len()
+	}
+
+	// 	Q = ceil(2/3 * N) - (quorum optimal)
```
Kourin1996: What does `- (quorum optimal)` mean?

consensus/ibft/state.go
```diff
@@ -97,12 +98,15 @@ func (c *currentState) setState(s IbftState) {
 }
 
 // NumValid returns the number of required messages
-func (c *currentState) NumValid() int {
-	// According to the IBFT spec, the number of valid messages
-	// needs to be 2F + 1
-	// The 1 missing from this equation is accounted for elsewhere
-	// (the current node tallying the messages will include its own message)
-	return 2 * c.validators.MaxFaultyNodes()
+func NumValid(validators ValidatorSet) int {
```
zivkovicmilos: What do you think about renaming this to `QuorumSize`?

Kourin1996: Why did you change from method to function? I think currentState or ValidattorSet should have this

zivkovicmilos (approved):
Looks good 💯 

I've went through the code and the spec, along with Saltini's paper, the logic seems to be sound.
I don't like that the formula differs for `N < 4`, but 🤷‍♂️ .

consensus/ibft/state.go
Kourin1996: The reason why I thought is that you can't change calculations without rewriting the code. You just create new struct and method when you want to change calc if you define it as method

consensus/ibft/state.go
dbrajovic: Misleading comment

[Fixed](https://github.com/0xPolygon/polygon-edge/pull/513/commits/058e3c6d5fd3b6afe0b09ef981dae87a0441c95d)

consensus/ibft/state.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-edge/pull/513/commits/6e5f031b9e2f50cc8129606f41001c194b87c505)

dbrajovic: Good point 👍 

[Refactored](https://github.com/0xPolygon/polygon-edge/pull/513/commits/44c154be5602fe978d70f9f64ad40a84ccf257d6)

Kourin1996 (approved):
LGTM!

txpool/txpool.go
```diff
@@ -564,6 +564,7 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 
 	// check if already known
 	if _, ok := p.index.get(tx.Hash); ok {
+		fmt.Println("known tx", "hash", tx.Hash.String())
```
0xAleksaOpacic: why do we need this log

0xAleksaOpacic (approved):

txpool/txpool.go
dbrajovic: Debug print for the GH test runs (got what I needed)

[Removed](https://github.com/0xPolygon/polygon-edge/pull/513/commits/ce66280be29bcea7c9e73f297981df266ec3910b)

PR #525, opened 2022-04-27
[Feature] Added support for YAML/YML config file format and configurable server log output to file

ZeljkoBenovic:
# Description

This PR introduces the support for starting the server with configuration file written in YAML syntax.
Added the `example-config.yaml` file with minimal configuration as a starting point.

Additionally, this PR introduces a new server flag `--log-to` which provides the operator with an option to store all server output to a log file, instead of only having console logs.

Added the make test option to the Makefile for easy testing.

# Changes include

# Checklist

## Testing

### Manual tests

Modify per your environment and use the `example-config.yaml` file as an argument to the `server --config` flag to start the server.
When starting the server set `--log-to <file_name>` flag have all logs written in a file. With this flag set, the server command will not have any console outputs.

Fixes EDGE-39


Makefile
```diff
@@ -31,3 +31,7 @@ lint:
 .PHONY: generate-bsd-licenses
 generate-bsd-licenses:
 	./generate_dependency_licenses.sh BSD-3-Clause,BSD-2-Clause > ./licenses/bsd_licenses.json
+
+.PHONY: test
+test:
+	go test -timeout 28m ./...
```
Kourin1996, line 39: How about building and generating binary before running tests for e2e?

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
...
+	var logFile *os.File
+
+	// if there is an error output logs to console
+	logFile, err := setLogFileWriter(config.LogFilePath)
+	if err != nil {
+		return hclog.New(&hclog.LoggerOptions{
+			Name:  "polygon",
+			Level: config.LogLevel,
+		})
+	}
```
Kourin1996: I guess logger will output to console when `config.LogFilePath` is empty string. It is normal behaviour, so you need to check `config.LogFilePath` (not by error handling) and create console logger if it's empty string.

Kourin1996:
Confirmed it worked locally

Makefile
ZeljkoBenovic: Totally forgot this.
Added in 3bc0a2b555df2d3ae4a8799fcb665ca33a496f58

Thanks 🥇

server/server.go
ZeljkoBenovic: The idea here is that we fallback to console output if `config.LogFilePath` is empty string or if there are any errors when creating the specified log file.

Added the console output regarding log file in d6412ffed8122cf3b77bdbd9b128b4bb181f226e 
When writing to console, if `--log-to` flag is not set, the first line will be `2022-04-28T11:25:49.825+0200 [INFO]  polygon: Log file: fallback to console output err="no file path defined"` or it will show an error on why the log file could not be created.

Kourin1996 (approved):
LGTM

command/server/params.go
```diff
@@ -108,6 +112,10 @@ func (p *serverParams) isDNSAddressSet() bool {
 	return p.rawConfig.Network.DNSAddr != ""
 }
 
+func (p *serverParams) isLogFileLocationSet() bool {
+	return p.rawConfig.LogFilePath != ""
```
zivkovicmilos, line 117: When do you actually set this value? `p.rawConfig.LogFilePath`

As far as I can see, you only manipulate `serverParams.logFileLocation`

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
 	"trie",
 }
 
+// set log output to the designated file
```
zivkovicmilos: Comment should be in the form of:
`// setLogFileWriter sets the log output...`

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
...
+func setLogFileWriter(logFilePath string) (*os.File, error) {
+	// if file path is empty string return error
+	if logFilePath == "" {
+		return nil, fmt.Errorf("log file path not defined")
```
zivkovicmilos: This should be defined as a package local error with `errors.New`, something like `errLogPathUndefined`.

Here, you could just return the error and check for them in the `newLogger` method explicitly

zivkovicmilos: You want the second error to be fatal, since something that was configured doesn't function like it should, you shouldn't swap it later for a CLI logger because it failed

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
...
+
+	logFile, err := os.Create(logFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not create log file, %w", err)
+	}
+
+	return logFile, nil
+}
+
+// create new logger; log to file if log path is defined
```
zivkovicmilos: Comment should be in the form of:
`// newLogger creates a new logger...`

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
...
+func newLogger(config *Config) (hclog.Logger, error) {
```
zivkovicmilos: What do you think about renaming this to `newLoggerFromConfig`?

server/server.go
```diff
@@ -94,6 +128,10 @@ func NewServer(config *Config) (*Server, error) {
 		restoreProgression: progress.NewProgressionWrapper(progress.ChainSyncRestore),
 	}
 
+	if logFileErr != nil {
+		m.logger.Info("Log file", "fallback to console output err", logFileErr.Error())
+	}
+
```
zivkovicmilos: As previously discussed, this shouldn't be a soft error but a fatal one.
Also, the place of checking this should be at the moment where you're initializing the logger itself

server/server.go
```diff
@@ -79,12 +80,45 @@ var dirPaths = []string{
...
+		}), err
+	}
+
+	// write logs to specified file
+	return hclog.New(&hclog.LoggerOptions{
+		Name:   "polygon",
+		Level:  config.LogLevel,
+		Output: logFile,
+	}), nil
+}
```
zivkovicmilos: You can rewrite this method to be cleaner like so:
* Have a separate method called `newFileLogger`, which is called from here if the log file path is set (internally it should contain the logic for setLogFileWriter)
* Have a separate method called `newCLILogger`, which just returns
```go
hclog.New(&hclog.LoggerOptions{
			Name:  "polygon",
			Level: config.LogLevel,
		})
```

This way you simplify this method by a lot:
- you're only conditionally calling (and returning) `newFileLogger`
- if that doesn't succeed, just return `newCLILogger`

What do you think?

cc @dbrajovic

example-config.yaml
```diff
@@ -0,0 +1,10 @@
+chain_config: test-chain/genesis.json
+data_dir: test-chain
+jsonrpc_addr: "0.0.0.0:10001"
+grpc_addr: "0.0.0.0:20001"
+seal: true
+network:
+  libp2p_addr: "127.0.0.1:1478"
+log_to: node.log
+
+# for additional parameters check out the docs page: https://edge-docs.polygon.technology/docs/configuration/sample-config
```
zivkovicmilos: This configuration should not be on the repo but on the docs

zivkovicmilos (approved):
Love this feature 💯 

Looks great, I've left a few comments 🚀

server/server.go
dbrajovic: It can still be `newLogger`, since it's only called within `NewServer`. Following your approach on rewriting `newLogger`, it does not need to return any error (but a plain `Logger`), falling back to `cliLogger` if creating a `fileLogger` fails.

server/server.go
dbrajovic: Agreed. For example:

```
func newLogger(config *Config) (hclog.Logger, error) {
	// log to file if provided
	if fileLogger, err := newFileLogger(config.LogFilePath); err == nil {
		return fileLogger
	}


	// return default cli logger
	return newCLILogger(...)
}
```

dbrajovic (approved):
Let's resolve the minor refactor suggestions and we can merge this 👍

command/server/params.go
ZeljkoBenovic: Good catch. Fixed in fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Fixed fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Refactored the logger instantiation logic fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Fixed fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Renamed it to `newLoggerFromConfig` fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Refactored logger instantiation fe41b1e3945aa3ff7c35ff6095786ce8d717692c

server/server.go
ZeljkoBenovic: Checkout new logger logic, much simpler I think fe41b1e3945aa3ff7c35ff6095786ce8d717692c

example-config.yaml
ZeljkoBenovic: Removed example config file and added command to create default config yaml file on the fly with default configuration.   
By running `server export` the user gets new `default-config.yaml` that can be used to start the server .   
1ef09e369a8f08a202d3641e7402f04762cb4384

Also, took care of some default server command values, as they were hardcoded instead of them being pulled from default configuration.

command/server/generate_config.go
```diff
@@ -0,0 +1,45 @@
+package server
```
zivkovicmilos: This file should be in a separate package within the `server` command directory, like `export` (see commands like `txpool`)

command/server/generate_config.go
```diff
@@ -0,0 +1,45 @@
...
+
+	return configCmd
+}
+
+func runGenerateConfigCommand(cmd *cobra.Command, _ []string) {
+	outputter := command.InitializeOutputter(cmd)
+
+	if err := generateConfig(*DefaultConfig()); err != nil {
+		outputter.SetError(err)
+		outputter.WriteOutput()
```
zivkovicmilos: This should be deferred after you create the outputter, and you should return here

command/server/generate_config.go
```diff
@@ -0,0 +1,45 @@
...
+	configCmd := &cobra.Command{
+		Use:   "export",
+		Short: "export default-config.yaml file with default parameters that can be used to run the server",
+		Run:   runGenerateConfigCommand,
+	}
+
+	return configCmd
+}
+
+func runGenerateConfigCommand(cmd *cobra.Command, _ []string) {
```
zivkovicmilos: What do you think about adding some kind of export success message?

See command results for other commands on how we did this

command/server/generate_config.go
```diff
@@ -0,0 +1,45 @@
...
+	config.Network.MaxOutboundPeers = -1
+
+	data, err := yaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("could not marshal Config struct")
+	}
+
+	if err := os.WriteFile("default-config.yaml", data, os.ModePerm); err != nil {
+		return fmt.Errorf("could not create and write default-config.yaml file")
+	}
```
zivkovicmilos: Why only yaml?

command/server/server.go
```diff
@@ -24,6 +21,8 @@ func GetCommand() *cobra.Command {
 	helper.RegisterLegacyGRPCAddressFlag(serverCmd)
 	helper.RegisterJSONRPCFlag(serverCmd)
 
+	// register subcommand used to generate config file on the fly
+	serverCmd.AddCommand(getGenerateConfigCmd())
```
zivkovicmilos: This should be called from a special method. Please see how this is done with other commands 🙏

command/server/generate_config.go
ZeljkoBenovic: Moved to dedicated package 9350d7d3c963f0b39070691337bd8610bbf3da8a

command/server/generate_config.go
ZeljkoBenovic: Outputter defered and added return 9350d7d3c963f0b39070691337bd8610bbf3da8a

command/server/generate_config.go
ZeljkoBenovic: Added success output message 9350d7d3c963f0b39070691337bd8610bbf3da8a

command/server/generate_config.go
ZeljkoBenovic: Added support for exporting config file to json format in addition to yaml 9350d7d3c963f0b39070691337bd8610bbf3da8a

command/server/server.go
ZeljkoBenovic: Refactored 9350d7d3c963f0b39070691337bd8610bbf3da8a

PR #528, opened 2022-04-28
Add txHash to rlp store marshal/unmarshal

### LINKED ISSUE #542: Event logs transaction hash is null
### Description

After a transaction, when looking at the event, the transaction hash is null

### Your environment

* Unbuntu(WSL2) Docker container
* latest
* develop

### Steps to reproduce

Picking up the hash using go-ethereum;

```
for {
		select {
		case myEventSubErr := <-depositsSub.Err():
			log.Fatal("error in deposits event subscription: ", myEventSubErr)
		case myEvent := <-deposits:
			fmt.Printf(myEvent.Raw.TxHash)
		}
	}
	// output 0x0000000000000000000000000000000000000000000000000000000000000000
	
```

### Expected behaviour

I should receive a valid transaction hash

### Actual behaviour

receive a null transaction hash

0xAleksaOpacic:
# Description

Fixed EDGE-531

When subscribing to events, txHash was always 0x00

# Changes include

# Checklist

## Testing

### Manual tests

- I have added a unit test that covers marshalling/unmarshalling for receipts in storage rlp.

- If someone wants to do manuall tests:
  1. Clone repo https://github.com/Aleksao998/SubscribeToEvents-EVM
  2. Deploy contract (from repo) to polygon-edge network
  4. Change address in main.go
  5. Trigger event from remix

- Manually tested backward compatibility


jsonrpc/filter_manager.go
```diff
@@ -486,6 +489,17 @@ func (f *FilterManager) appendLogsToFilters(header *types.Header, removed bool)
...
+			// Get a block where tx was sealed
+			block, ok := f.store.GetBlockByHash(header.Hash, true)
+			if !ok {
+				f.logger.Error("could not find block in store", "hash", header.Hash.String())
+
+				return nil
+			}
+			// Extract tx Hash
+			receipt.TxHash = block.Transactions[indx].Hash
+		}
```
Kourin1996: I don't think it's efficient code because GetBlockByHash will be called several times if multiple receipts don't have TxHash. You might want to get block in the outer side of loop.

types/rlp_unmarshal_storage.go
```diff
@@ -115,8 +115,8 @@ func (r *Receipt) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) err
 		return err
 	}
 
-	if len(elems) != 3 {
-		return fmt.Errorf("expected 3 elements")
+	if len(elems) != 4 {
+		return fmt.Errorf("expected 4 elements")
 	}
```
Kourin1996: Is it possible to parse old Receipt?

Kourin1996 (changes requested):

types/rlp_unmarshal_storage.go
0xAleksaOpacic: You are right, do you see why we need this condition?

0xAleksaOpacic: We can have if condition for each item

Kourin1996: I think this condition is necessary. `Condition for each item` just checks the types of each item

Kourin1996:
@Aleksao998 Test failed in my local environment. Please check it?

```
?       github.com/0xPolygon/polygon-edge/helper/tests  [no test files]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x16899f2]

goroutine 144 [running]:
github.com/0xPolygon/polygon-edge/jsonrpc.(*mockStore).GetBlockByHash(0xc0004a4680, 0x0, 0x0, 0x0, 0x200000000000000, 0xc00012c101, 0x1, 0x1)
        <autogenerated>:1 +0x32
github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).appendLogsToFilters(0xc0000a0120, 0xc0004c9200, 0x1, 0x2, 0x0)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:494 +0x3bc
github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).processEvent(0xc0000a0120, 0xc000020b40, 0x0, 0x0)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:463 +0x149
github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).dispatchEvent(0xc0000a0120, 0xc000020b40, 0x0, 0x0)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:439 +0x39
github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).Run(0xc0000a0120)
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:290 +0x1b2
created by github.com/0xPolygon/polygon-edge/jsonrpc.TestFilterLog
        /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager_test.go:18 +0x14c
FAIL    github.com/0xPolygon/polygon-edge/jsonrpc       0.378s
?       github.com/0xPolygon/polygon-edge/licenses      [no test files]
```

jsonrpc/filter_manager.go
0xAleksaOpacic: This is only for backwards compatibility. No new receipt will have TxHash == types.ZeroHash. 
What do you think is more efficient, leaving like this and fetching only for old logs, or fetching block every time once?

0xAleksaOpacic: ping @Kourin1996

types/rlp_unmarshal_storage.go
0xAleksaOpacic: Fixed af7a1d2b5fac4f1755c8b5d764f36275387e2b1c, please check again

0xAleksaOpacic:
> @Aleksao998 Test failed in my local environment. Please check it?
> 
> ```
> ?       github.com/0xPolygon/polygon-edge/helper/tests  [no test files]
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x16899f2]
> 
> goroutine 144 [running]:
> github.com/0xPolygon/polygon-edge/jsonrpc.(*mockStore).GetBlockByHash(0xc0004a4680, 0x0, 0x0, 0x0, 0x200000000000000, 0xc00012c101, 0x1, 0x1)
>         <autogenerated>:1 +0x32
> github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).appendLogsToFilters(0xc0000a0120, 0xc0004c9200, 0x1, 0x2, 0x0)
>         /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:494 +0x3bc
> github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).processEvent(0xc0000a0120, 0xc000020b40, 0x0, 0x0)
>         /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:463 +0x149
> github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).dispatchEvent(0xc0000a0120, 0xc000020b40, 0x0, 0x0)
>         /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:439 +0x39
> github.com/0xPolygon/polygon-edge/jsonrpc.(*FilterManager).Run(0xc0000a0120)
>         /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager.go:290 +0x1b2
> created by github.com/0xPolygon/polygon-edge/jsonrpc.TestFilterLog
>         /Users/kourin/go/src/github.com/0xPolygon/polygon-edge/jsonrpc/filter_manager_test.go:18 +0x14c
> FAIL    github.com/0xPolygon/polygon-edge/jsonrpc       0.378s
> ?       github.com/0xPolygon/polygon-edge/licenses      [no test files]
> ```

Fixed 25601c8e464a2bdc9a840a7d7780c8879a949b20

jsonrpc/filter_manager.go
Kourin1996: I understood it's for backwards compatibility. But it doesn't need to fetch block multiple times, does it? Why don't you save fetched block to variable?

0xAleksaOpacic: That's a good idea will do that

0xAleksaOpacic: 842f8d7149d9a6517be061a840f4bf594ca10391 check this

jsonrpc/filter_manager.go
```diff
@@ -485,7 +488,24 @@ func (f *FilterManager) appendLogsToFilters(header *types.Header, removed bool)
...
+	if len(receipts) > 0 && receipts[0].TxHash == types.ZeroHash {
+		var ok bool
+		// Get a block where tx was sealed
+		block, ok = f.store.GetBlockByHash(header.Hash, true)
+		if !ok {
+			f.logger.Error("could not find block in store", "hash", header.Hash.String())
+
+			return nil
+		}
+	}
```
dbrajovic: Since this `block` var might only (potentially) be used in the for loop below, I figure it's just best to fetch it once unconditionally. 

Either that or have a convenient callback - no unnecessary fetching but more verbose code.

Kourin1996 (approved):
LGTM. Please resolve @dbrajovic's comment

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed 4a97908488dc836d019e1f1d002516792e782fd6

dbrajovic (approved):

types/rlp_encoding_test.go
```diff
@@ -65,8 +65,62 @@ func TestRLPMarshall_And_Unmarshall_Transaction(t *testing.T) {
...
+			unmarshalledReceipt := new(Receipt)
+			marshaledRlp := receipt.MarshalStoreRLPTo(nil)
+
+			if err := unmarshalledReceipt.UnmarshalStoreRLP(marshaledRlp); err != nil {
+				t.Fatal(err)
+			}
+
+			if !reflect.DeepEqual(receipt, unmarshalledReceipt) {
+				t.Fatal("[ERROR] Unmarshalled receipt not equal to base receipt")
+			}
```
zivkovicmilos: Please do not use the `reflect` package to make sure these receipts are the same

types/rlp_unmarshal_storage.go
```diff
@@ -115,30 +115,40 @@ func (r *Receipt) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) err
 		return err
 	}
 
-	if len(elems) != 3 {
-		return fmt.Errorf("expected 3 elements")
+	if len(elems) < 3 {
+		return fmt.Errorf("expected at least 3 elements")
```
zivkovicmilos, line 119: @lazartravica @Kourin1996 @dbrajovic 

This is a breaking change - please note that the conditional here was `!=` and not `<`.
This means that whatever you add to the `Receipts` struct additionally during marshalling now **will not get unmarshalled** on older versions of the codebase (clients) - because they expect _exactly_ 3 elements.

This unmarshalling problem is present with almost all RLP structures, in a sense that there is room for basically 0 backwards compatibility. I am facing this problem with my block validation task (I wanted to add receipts, but it broke the code for older clients).

We need to decide on how we want to treat these problems going forward:
* make this change, release a version that's a major update (non-compatible with older versions of the code). It essentially means that partners would need to update **all** nodes to the latest version in order for everything to work
* continue to write 600 lines of code for backwards compatibility for every small fix or feature we want to add, because the base architecture is wrong and needs to be changed at the core

zivkovicmilos (approved):
Looks good 💯 

Please check the comment I've left for the backwards compatibility headache 🙏

types/rlp_encoding_test.go
0xAleksaOpacic: Fixed fdf5ef8f769099bd74ba8dc0d52941f5d65ade09

types/rlp_unmarshal_storage.go
0xAleksaOpacic: I think that our best option is to release this as a major change, but before release create PR-s which fix all similar issues in the package. It will be a few days of work but will make our life easier in the long run.
Adding additional logic will just make a mess in our code which will catch up with us later

zivkovicmilos: Resolving as we continued the discussion internally

lazartravica:
@zivkovicmilos @Aleksao998 Let's plan the release accordingly with:

Is the old blockchain produced by old clients still able to be synced by the new client?

PR #533, opened 2022-05-04
[PoC] Add ParentCommittedSeal in IBFT Extra

Kourin1996:
Fix EDGE-542

# Description

**This PR is for proof of concept, don't merge to develop or use for production.**

This PR adds new field `ParentCommittedSeal` to store CommittedSeal of the parent block.
Block proposer will be responsible for setting `ParentCommittedSeal` to the proposing block and other validators will need to verify them during consensus.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


consensus/ibft/extra.go
```diff
@@ -21,28 +21,18 @@ var (
...
-	ibftExtra := &IstanbulExtra{
-		Validators:    validators,
-		Seal:          []byte{},
-		CommittedSeal: [][]byte{},
-	}
-
-	extra = ibftExtra.MarshalRLPTo(extra)
-	h.ExtraData = extra
+// initIbftExtra initializes ExtraData in Header for IBFT
+func initIbftExtra(h *types.Header, validators []types.Address, parentCommittedSeals [][]byte) error {
```
dbrajovic: Should this return an error ? `putIbftExtra` cannot return one either.

consensus/ibft/ibft.go
```diff
@@ -1233,6 +1244,19 @@ func (i *Ibft) verifyHeaderImpl(snap *Snapshot, parent, header *types.Header) er
 		return err
 	}
 
+	// verify last committed seals
+	if parent.Number >= 1 {
+		// find validators who validated last block
+		parentSnap, err := i.getSnapshot(parent.Number - 1)
```
dbrajovic: Doesn't this mean we're fetching the snapshot of the grandparent (`parent.Number - 1`) ?

consensus/ibft/extra.go
```diff
@@ -51,17 +41,19 @@ func PutIbftExtra(h *types.Header, istanbulExtra *IstanbulExtra) error {
 		extra = extra[:IstanbulExtraVanity]
 	}
 
-	data := istanbulExtra.MarshalRLPTo(nil)
-	extra = append(extra, data...)
-	h.ExtraData = extra
+	h.ExtraData = istanbulExtra.MarshalRLPTo(extra)
```
zivkovicmilos, line 44: Aren't you overriding the `extra` array here?

Why did you change this logic?

consensus/ibft/sign.go
```diff
@@ -63,6 +72,7 @@ func signSealImpl(prv *ecdsa.PrivateKey, h *types.Header, committed bool) ([]byt
 	return seal, nil
 }
 
+// writeSeal creates and sets a signatures to Seal in IBFT Extra
```
zivkovicmilos: Typo here, did you mean `a signature`?

consensus/ibft/extra_test.go
```diff
@@ -40,3 +58,135 @@ func TestExtraEncoding(t *testing.T) {
...
+
+	assert.NoError(t, initIbftExtra(header, validators, lastCommittedSeal))
+
+	return header
+}
+
+// Test Scenario
+// 1. 4 IBFT Validators create headers
+// 2. A faulty node scans the past headers and appends new committed seal
+// 3. Check each hashes of headers are wrong
```
zivkovicmilos: Typo here, I'm not sure what you wanted to say

consensus/ibft/extra_test.go
```diff
@@ -40,3 +58,135 @@ func TestExtraEncoding(t *testing.T) {
...
+		header := h.Copy()
+
+		// update parent hash & committed seal
+		if i > 0 {
+			parentHeader := faultyHeaders[i-1]
+
+			// update parent hash
+			header.ParentHash = parentHeader.Hash
+
+			// get parent committeed seal
```
zivkovicmilos: Typo here, should be `committed`

zivkovicmilos (approved):
Looks good 💯 

Love the test that perfectly illustrates this functionality 👌

consensus/ibft/extra.go
Kourin1996: Sorry, I don't get it. What's your suggestion?

consensus/ibft/ibft.go
Kourin1996: Yes, you're right. Snapshot for n-th block means a snapshot that has been updated by the block at n. We want to verify parent block, which means we need to get the snapshot for grandparent block.

consensus/ibft/extra.go
Kourin1996: Please check the implementation of `MarshalRLPTo`. It appends RLP data to given byte slice. extra is made expand or shrink before this in order to match fixed size slice.

Kourin1996: In previous code, there were 2 types code to set IBFT extra data to extra field. But I'm sure they are same.

consensus/ibft/extra_test.go
Kourin1996: Thank you! Fixed at 9192f447dc5a4e63cbceb1500a1690b6283d385d

consensus/ibft/sign.go
Kourin1996: Thank you! Fixed at 9192f447dc5a4e63cbceb1500a1690b6283d385d

consensus/ibft/extra_test.go
Kourin1996: This test checks the hash of the block mined by validators is different from the hash of the block including new committed seal by a new validator who joined later.

consensus/ibft/extra.go
zivkovicmilos: > Please check the implementation of MarshalRLPTo. It appends RLP data to given byte slice. extra is made expand or shrink before this in order to match fixed size slice.

I remember we had this exact same discussion a while back, I just wanted you to confirm it 🙂

consensus/ibft/extra_test.go
zivkovicmilos: Ah, so the last line (`99`) should be something like:
`Check if each hash of the headers is wrong`

consensus/ibft/extra.go
dbrajovic: Remove the `error` return from `initIbftExtra` and `putIbftExtra` because they can't return anything but `nil`.

dbrajovic (approved):

consensus/ibft/extra_test.go
Kourin1996: Thank you. Fixed at e752d81bb195846d7dd522621a5d8c8d309b7391

0xAleksaOpacic (approved):
LGTM! Thank you for the great test coverage even on the experimental branch

0xAleksaOpacic:
@Kourin1996 This is a breaking change, can we update the PR description 😃

consensus/ibft/consensus_backend.go
```diff
@@ -138,8 +138,15 @@ func (i *backendIBFT) buildBlock(snap *Snapshot, parent *types.Header) (*types.B
 	// set the timestamp
 	header.Timestamp = uint64(time.Now().Unix())
 
-	// we need to include in the extra field the current set of validators
-	putIbftExtraValidators(header, snap.Set)
+	var parentCommittedSeal [][]byte
+
+	if parent.Number >= 1 {
```
0xAleksaOpacic: Just a suggestion, maybe to have util method isGenesisBlock(blockNumber). It's a bit more descriptive

consensus/ibft/extra.go
```diff
@@ -21,28 +21,18 @@ var (
...
-
-	extra = ibftExtra.MarshalRLPTo(extra)
-	h.ExtraData = extra
+// initIbftExtra initializes ExtraData in Header for IBFT
+func initIbftExtra(h *types.Header, validators []types.Address, parentCommittedSeals [][]byte) {
+	putIbftExtra(h, &IstanbulExtra{
+		Validators:          validators,
+		ProposerSeal:        []byte{},
+		CommittedSeal:       [][]byte{},
+		ParentCommittedSeal: parentCommittedSeals,
```
0xAleksaOpacic, line 30: Should the naming be in plural, ParentCommittedSeals, its an array of seals. I see that this is the patern for CommittedSeals so if you want we can keep this, feel free to resolve comment :)

consensus/ibft/header_hash.go
```diff
@@ -15,15 +15,11 @@ func istanbulHeaderHash(h *types.Header) types.Hash {
 	arena := fastrlp.DefaultArenaPool.Get()
 	defer fastrlp.DefaultArenaPool.Put(arena)
 
-	// when hashing the block for signing we have to remove from
-	// the extra field the seal and committed seal items
-	extra, err := getIbftExtra(h)
-	if err != nil {
+	// clear unnecessary fields in IBFT Extra for hash calculation
+	if err := filterIbftExtraForHash(h); err != nil {
```
0xAleksaOpacic, line 19: Can we add a comment which fields are used for computing block hash (validators, PAS). Just suggestion because we have diff block hash computation then ethereum.
If you don't think it's needed you can resolve this comment :)

consensus/ibft/consensus_backend.go
Kourin1996: Added at 83e82a3e529746c8a3fc7b8437e78f6abe8136a0.

0xAleksaOpacic: Thanks

consensus/ibft/extra.go
```diff
@@ -164,7 +245,7 @@ func (i *IstanbulExtra) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) er
 	{
 		vals, err := elems[2].GetElems()
 		if err != nil {
```
0xAleksaOpacic, line 247: Personally i think we should log and error obj with err msg when err occurs.I see this in multiple places in our code

consensus/ibft/extra.go
```diff
@@ -120,6 +184,23 @@ func (i *IstanbulExtra) MarshalRLPWith(ar *fastrlp.Arena) *fastrlp.Value {
...
 
+	// ParentCommittedSeal
+	if i.ParentCommittedSeal != nil {
+		if len(i.ParentCommittedSeal) == 0 {
+			vv.Set(ar.NewNullArray())
+		} else {
+			committed := ar.NewArray()
+			for _, a := range i.ParentCommittedSeal {
+				if len(a) == 0 {
+					vv.Set(ar.NewNull())
```
0xAleksaOpacic, line 195: I don't see that ever happening, but if it does unmarshal would not work or i am missing something?
When we are doing unmarshaling for ParentCommittedSeal we are just extracting the 4th element with elems[3].GetElems and If we had multiple PAS which has len(a)==0 we would have more than 4 elements.
Please correct me if i understood something wrong :)

consensus/ibft/sign.go
```diff
@@ -31,86 +41,72 @@ func ecrecoverImpl(sig, msg []byte) (types.Address, error) {
...
-	extra, err := getIbftExtra(h)
+	seal, err := unpackProposerSealFromIbftExtra(h)
 	if err != nil {
 		return types.Address{}, err
 	}
 
-	// Calculate the header hash (keccak of RLP)
-	hash, err := calculateHeaderHash(h)
+	// get the sig
+	msg, err := calculateHeaderHash(h)
```
0xAleksaOpacic, line 50: Why do we need to have both calculateHeaderHash and istanbulHeaderHash if they do same thing?

consensus/ibft/sign.go
```diff
@@ -184,35 +172,75 @@ func verifyCommittedFields(
 		return fmt.Errorf("empty committed seals")
```
0xAleksaOpacic, line 172: we have const for this err: ErrEmptyCommittedSeals

consensus/ibft/sign.go
```diff
@@ -184,35 +172,75 @@ func verifyCommittedFields(
...
+
+		visited[addr] = struct{}{}
 	}
 
 	// Valid committed seals must be at least 2F+1
 	// 2F 	is the required number of honest validators who provided the committed seals
 	// +1	is the proposer
-	if validSeals := len(visited); validSeals < quorumSizeFn(snap.Set) {
+	if validSeals := len(visited); validSeals < quorumSizeFn(validators) {
 		return fmt.Errorf("not enough seals to seal block")
```
0xAleksaOpacic, line 244: we have a const for this err: ErrNotEnoughCommittedSeals

0xAleksaOpacic (changes requested):
I found an issue with this PR:
We have a running network (let's say 4 nodes), we switch to this branch (at 50th block) and restart nodes. Everything will work. But if we add a new node to sync we would have a problem because first 50 blocks does not have PAS (they are empty) and we would keep getting an error **failed to verify ParentCommittedSeal: empty committed seals**.

Are going to have a fork from which both PAS and BLS will be enabled, still haven't checked other BLS PR-s? If we have a PR for this please update description on this PR and add that it is breaking change

consensus/ibft/sign.go
Kourin1996: I think istanbulHeaderHash has been removed in the latest PR

consensus/ibft/sign.go
Kourin1996: Resolved in the latest PR

consensus/ibft/sign.go
Kourin1996: Resolved in the latest PR

PR #537, opened 2022-05-06
Remove concurrency from Filter e2e tests

zivkovicmilos:
# Description

This PR removes concurrency from E2E filter tests, as the assertion cannot be guaranteed (logs cannot be guaranteed to be appended in time).

The runtime of these tests is also greatly reduced (no need to send 10 transactions when 1 is enough). It's questionable if these should even be E2E tests 🤷‍♂️ .

# Changes include

# Checklist

## Testing

### DIFF
--- a/e2e/logs_test.go
+++ b/e2e/logs_test.go
@@ -3,7 +3,6 @@ package e2e
 import (
 	"context"
 	"math/big"
-	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


Kourin1996 (approved):
Looks good to me

dbrajovic (approved):

PR #538, opened 2022-05-08
Implement eth_getFilterLogs

0xAleksaOpacic:
# Description

This PR implement eth_getFilterLog JSON-RPC method.

**eth_getLogs vs. eth_getFilterLogs**
These two computations will return the same results:
1. Calling eth_getLogs with params [options]
2. Calling eth_newFilter with params [options], getting a filter id <filterId> back, then calling eth_getFilterLogs with params [filterId]

# Changes include

# Breaking changes

This PR does not introduce breaking change

# Checklist

## Testing

### Manual tests

# Documentation update

[PR](https://github.com/0xPolygon/polygon-edge-docs/pull/88)

# Additional comments

Fixes EDGE-548


jsonrpc/filter_manager.go
```diff
@@ -339,6 +345,70 @@ func (f *FilterManager) Exists(id string) bool {
...
+	filter, ok := f.filters[id]
+
+	logFilters, ok := filter.(*logFilter)
+
+	if !ok {
+		return nil, ErrFilterDoesNotExists
+	}
+
+	//getAllBlocks
+	from := uint64(0)
```
Kourin1996: You don't need to start from zero because nothing happens when block height is zero in loop

jsonrpc/filter_manager.go
```diff
@@ -339,6 +345,70 @@ func (f *FilterManager) Exists(id string) bool {
 	return ok
 }
 
+// GetFilterLogs returns an array of logs for the specified filter.
+func (f *FilterManager) GetFilterLogs(id string) ([]*Log, error) {
```
Kourin1996: I guess it will take much time to return data if the chain is long. Is there any limitation?

jsonrpc/filter_manager.go
0xAleksaOpacic: Agree thank you

jsonrpc/filter_manager.go
0xAleksaOpacic: Not that I know of, but will test this with the longest chains we have. I do not see any other option for writing this with our code archistructure

0xAleksaOpacic: Completely changed the implementation so this is no longer a problem

jsonrpc/filter_manager.go
0xAleksaOpacic: Completely changed the implementation so this is no longer a problem

dbrajovic (approved):

jsonrpc/eth_endpoint.go
```diff
@@ -664,95 +664,19 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
...
-
-		return uint64(num)
-	}
-
-	from := resolveNum(query.fromBlock)
-	to := resolveNum(query.toBlock)
-
-	if to < from {
-		return nil, fmt.Errorf("incorrect range")
+// GetFilterLogs returns an array of logs for the specified filter.
```
zivkovicmilos: Nitpick: please remove punctuation from comments

jsonrpc/filter_manager.go
```diff
@@ -293,8 +297,11 @@ func (f *FilterManager) Run() {
 
 		case <-timeoutCh:
 			// timeout for filter
-			if !f.Uninstall(filterBase.id) {
-				f.logger.Error("failed to uninstall filter", "id", filterBase.id)
+			// if filter still exists
+			if filterBase != nil {
+				if !f.Uninstall(filterBase.id) {
+					f.logger.Error("failed to uninstall filter", "id", filterBase.id)
+				}
```
zivkovicmilos: Please unwrap this spaghetti

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+			}
+		}
+
+		return nil
+	}
+
+	if query.BlockHash != nil {
+		block, ok := f.store.GetBlockByHash(*query.BlockHash, true)
+		if !ok {
+			return nil, fmt.Errorf("not found")
```
zivkovicmilos: Why is this an `Errorf` and not a standardized package error?

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+		}
+
+		return uint64(num)
+	}
+
+	from := resolveNum(query.fromBlock)
+	to := resolveNum(query.toBlock)
+
+	if to < from {
+		return nil, fmt.Errorf("incorrect range")
```
zivkovicmilos: Same as https://github.com/0xPolygon/polygon-edge/pull/538/files#r877514120

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+	for i := from; i <= to; i++ {
+		block, ok := f.store.GetBlockByNumber(i, true)
+		if !ok {
+			break
+		}
+
+		if block.Header.Number == 0 || len(block.Transactions) == 0 {
+			// do not check logs in genesis and skip if no txs
+			continue
+		}
```
zivkovicmilos: Why even fetch the block if you're gonna just continue if the block is the genesis block?

You know the number of the block before you even fetch it (i), so you should perform this check sooner (if it's genesis, continue)

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+
+		if err := parseReceipts(block); err != nil {
+			return nil, err
+		}
+
+		return result, nil
+	}
+
+	head := f.store.Header().Number
```
zivkovicmilos: Please consider renaming this to something more clear, like `latestBlockNumber`

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+
+		if num == EarliestBlockNumber {
+			num = 0
+		}
+
+		if num == LatestBlockNumber {
+			return head
+		}
+
+		return uint64(num)
```
zivkovicmilos: https://gobyexample.com/switch

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,116 @@ func (f *FilterManager) Exists(id string) bool {
...
+			// no txs in block, return empty response
+			return result, nil
+		}
+
+		if err := parseReceipts(block); err != nil {
+			return nil, err
+		}
+
+		return result, nil
+	}
```
zivkovicmilos: This can be extracted out (since it's practically the same logic you do in a range)

cc @dbrajovic

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+	t.Parallel()
+
+	blockHash := types.StringToHash("1")
+
+	// Topics we're searching for
+	topic1 := types.StringToHash("4")
+	topic2 := types.StringToHash("5")
+	topic3 := types.StringToHash("6")
+
+	var topics = [][]types.Hash{{topic1}, {topic2}, {topic3}}
```
zivkovicmilos, line 26: This can be a helper method

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+
+	testTable := []struct {
+		name           string
+		query          *LogQuery
+		shouldFail     bool
+		expectedLength int
+	}{
+		{"Found matching logs, fromBlock < toBlock",
```
zivkovicmilos: Nitpick: Please make sure that each part of this struct is on a new line

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+		{"No logs found", &LogQuery{
+			fromBlock: 4,
+			toBlock:   5,
+			Topics:    topics,
+		}, false, 0},
+		{"Invalid block range", &LogQuery{
+			fromBlock: 10,
+			toBlock:   5,
+			Topics:    topics,
+		}, true, 0},
```
zivkovicmilos: Why not check for the exact error here?

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+	}
+
+	// setup test
+	store := &mockBlockStore{}
+	store.topics = []types.Hash{topic1, topic2, topic3}
+	store.setupLogs()
```
zivkovicmilos: You can inline this

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+				},
+				{
+					Value: big.NewInt(11),
+				},
+				{
+					Value: big.NewInt(12),
+				},
+			},
+		})
+	}
```
zivkovicmilos: This can be a helper method

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+			foundLogs, logError := f.GetLogsForQuery(testCase.query)
+
+			if logError != nil && !testCase.shouldFail {
+				// If there is an error and test isn't expected to fail
+				t.Fatalf("Error: %v", logError)
+			} else if !testCase.shouldFail {
+				assert.Lenf(t, foundLogs, testCase.expectedLength, "Invalid number of logs found")
+			} else {
+				assert.Nil(t, foundLogs, "Expected first return param to be nil")
+			}
```
zivkovicmilos: Please resolve this spaghetti

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+		Addresses: []types.Address{addr1},
+		toBlock:   10,
+		fromBlock: 0,
+	}
+
+	// add log filter
+	id := m.NewLogFilter(logFilter, &MockClosedWSConnection{})
+
+	retrivedLogFilter, err := m.GetLogFilterFromID(id)
+	assert.Nil(t, err, "No Error expected")
```
zivkovicmilos: There is a [method](https://pkg.go.dev/github.com/stretchr/testify/assert#NoError) for `assert`, it's called `NoError`

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+		})
+	}
+}
+
+func Test_GetLogFilterFromID(t *testing.T) {
+	store := newMockStore()
+
+	m := NewFilterManager(hclog.NewNullLogger(), store)
+
+	go m.Run()
```
zivkovicmilos, line 143: Where do you stop the FilterManager in this test? Can you even stop it?

jsonrpc/filter_manager_test.go
```diff
@@ -11,6 +13,123 @@ import (
...
+	logFilter := &LogQuery{
+		Addresses: []types.Address{addr1},
+		toBlock:   10,
+		fromBlock: 0,
+	}
+
+	// add log filter
+	id := m.NewLogFilter(logFilter, &MockClosedWSConnection{})
+
+	retrivedLogFilter, err := m.GetLogFilterFromID(id)
```
zivkovicmilos: This can be inlined

zivkovicmilos (changes requested):
There need to be some changes made before we can give it the go-ahead 💯

jsonrpc/eth_endpoint.go
0xAleksaOpacic: Fixed 17248859ede6ca3b4ece593f3a81d5a1821c8ff9

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed 81645b5447a911cbb2a8d1ade24eb4b4e16a7a55

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed dd2769c0533c4517d248f746a48bc10cf6ecc2b0

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed dd2769c0533c4517d248f746a48bc10cf6ecc2b0

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed fa99b5d16e185f74fc9c216e6578d223e8145ad1
Fixed Linter 9d46530204c5507e2451aa0d2b09c0cc34713b36

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed e622f47de8235b9b3181f9be5cde53ad1034167d

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed 989989f60c76d903f470ea79cb58f877c391f9e3

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed 7049ef0c6f4e0773267d841fcd558493366e2a80

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Not sure what should be here a helper method

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed d7792a6b16106250d40469ce08013d60e1dc9dc3

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Not sure what should be inlined here

jsonrpc/filter_manager_test.go
0xAleksaOpacic: fixed 72e0e35fee7e00cc073fd746e5d321dbfede7af3

jsonrpc/filter_manager.go
0xAleksaOpacic: Yes, but because the one uses GetBlockByHash and the other GetBlockByNumber I would need to pass multiple params and have condition inside it. Can discuss tmrw about it

jsonrpc/filter_manager_test.go
0xAleksaOpacic: What exactly? Adding multiple elements into the store with default values? Some test may require additional fields in the Header and Transaction object.

jsonrpc/filter_manager_test.go
0xAleksaOpacic: fixed b1b8aed2cccac3f87378830e809954c97965d5cb

jsonrpc/filter_manager_test.go
0xAleksaOpacic: From what I can see in the code, this test was not designed to receive an error message. 
Fixed 99ca16124f3c6827c1587da7427bda6eae93bb29

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed a6f7a0fbb0cb51a21ae29066ef596e135076b4c8

jsonrpc/filter_manager.go
0xAleksaOpacic: Also most of these test in the jsonrpc plus the whole filter_menager should be refactored. So i would not add more utility methods if we are going to refactor it in general.

jsonrpc/filter_manager_test.go
zivkovicmilos: This should be `assert.ErrorIs` instead of equal

jsonrpc/filter_manager_test.go
zivkovicmilos: Please notice that your `shouldFail` and `expectedError` fields are tied -> if you expect an error, then it should fail; if you don't expect an error it shouldn't fail.

You can drop the `shouldFail` boolean from this test

jsonrpc/filter_manager_test.go
zivkovicmilos: 💀 

Please no timeouts 🙏 

Remove it please, I thought the FM has a `Close` method

jsonrpc/filter_manager_test.go
zivkovicmilos: `id` can be inlined, because you only use it here

jsonrpc/filter_manager_test.go
zivkovicmilos: Have a helper method like `appendBlocksToStore`, where you would add blocks (pass in an array of blocks)

jsonrpc/filter_manager_test.go
zivkovicmilos: Helper method like `generateTopics`, where you pass in the number of topics you want

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed 222d7de8b58904bbf23187d73a4a26ff3a407358

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed 530d473ab4255e17ba5c8bfe4ba2002787ad1445

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed 73fa6eafe4f57ac1755b833697633562354dfd25

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Fixed 39a2b14bfa45f79462998d98cf29d35a63e472e9

jsonrpc/filter_manager_test.go
0xAleksaOpacic: Here is the fix e44499d304fa50924cb8793f3c07d73e6b581794. If you don't think this is the way, say and we will improve it :)

zivkovicmilos (approved):

I think you can agree that it's much cleaner now and easier to manage 💯

0xAleksaOpacic:
> Thank you for making the requested changes 🙏 
> 
> 
> 
> I think you can agree that it's much cleaner now and easier to manage 💯 

Of course, sorry for not fixing this method and test before sending for PR. Thank you on a detailed review 💪🏼

dbrajovic (approved):

jsonrpc/filter_manager.go
```diff
@@ -342,6 +349,127 @@ func (f *FilterManager) Exists(id string) bool {
...
+	return logs, nil
+}
+
+func (f *FilterManager) getLogsFromBlocks(query *LogQuery) ([]*Log, error) {
+	latestBlockNumber := f.store.Header().Number
+
+	resolveNum := func(num BlockNumber) uint64 {
+		switch num {
+		case PendingBlockNumber:
+			num = LatestBlockNumber
```
Kourin1996: Why did you set -2 here. This value will be converted to uint64 when existing from this function and then underflow happens

Kourin1996 (changes requested):
I've found a small issue but looks good apart from that

jsonrpc/filter_manager.go
0xAleksaOpacic: fixed 0a5b3edce7b7382bdae0c186ebe0d5eeeb64ab6f

zivkovicmilos: @Aleksao998 

How does the commit fix this exactly? You can still cause an underflow

zivkovicmilos: Aha, nevermind, I didn't see you returned an error for pending block requests 👍

Kourin1996 (approved):
LGTM. We need to discuss log size limitation in order to prevent some DOS attack in the future

PR #544, opened 2022-05-10
[Blockchain] Simplify block verification

zivkovicmilos:
# Description

This PR aims to separate out block insertion and block verification.
There are 2 kinds of block verifications that can occur:
* Node wants to verify a previously sealed block (from a different node)
* Node wants to verify a _proposed_ block (from a different node)

It also refactors the way the blockchain module is tested.

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EDGE-469


blockchain/blockchain.go
```diff
@@ -608,82 +667,225 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
+	// Log the information
+	b.logger.Info(
+		"write block",
+		"num",
+		block.Number(),
+		"parent",
+		block.ParentHash(),
+	)
+
+	// Checks are passed, write the chain
```
dbrajovic: This comment seems redundant now.

dbrajovic (approved):
LGTM. Great job on breaking up the `WriteBlock` logic as well as adding test coverage 🙏

blockchain/blockchain.go
```diff
@@ -190,15 +226,38 @@ func NewBlockchain(
...
 
 	// Push the initial event to the stream
 	b.stream.push(&Event{})
 
 	return b, nil
 }
 
+// initCaches initializes the blockchain caches with the specified size
+func (b *Blockchain) initCaches(size int) error {
+	var cacheErr error
```
Kourin1996: If you're using same name error, you can just call it `err`, not `cacheErr`?

blockchain/blockchain.go
```diff
@@ -190,15 +226,38 @@ func NewBlockchain(
...
+
+	b.headersCache, cacheErr = lru.New(100)
```
Kourin1996: * Is `100` the meaning value?
* All cache will have same size?
* Why don't you define constant value to give meaning to magic number?

blockchain/blockchain.go
```diff
@@ -608,82 +667,225 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
+		return err
+	}
+
+	// Make sure the consensus layer verifies this block header
+	if err := b.consensus.VerifyHeader(block.Header); err != nil {
+		return fmt.Errorf("failed to verify the header: %w", err)
+	}
+
+	return nil
+}
```
Kourin1996: We know whether it works because IBFT always verifies header before verifying block. But it's anti pattern that Blockchain knows about IBFT, isn't it?

blockchain/blockchain.go
```diff
@@ -608,82 +667,225 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
+func (b *Blockchain) VerifySealedBlock(block *types.Block) error {
+	// Do the initial block verification
+	if err := b.verifyBlock(block); err != nil {
+		return err
+	}
+
+	// Make sure the consensus layer verifies this block header
+	if err := b.consensus.VerifyHeader(block.Header); err != nil {
+		return fmt.Errorf("failed to verify the header: %w", err)
+	}
```
Kourin1996: Do you have any specific reason to verify header before block?

blockchain/blockchain.go
```diff
@@ -608,82 +667,225 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
+	}); err != nil {
+		return fmt.Errorf("unable to verify block execution result, %w", err)
+	}
+
+	return nil
+}
+
+// verifyBlockResult verifies that the block transaction execution result
+// matches up to the expected values
+func (br *BlockResult) verifyBlockResult(result *referenceBlockResult) error {
```
Kourin1996: Why you can't give block directly? referenceBlockResult is helping us not to misunderstand code? Defining new DTO makes your brain create new context.

Kourin1996 (changes requested):
Just a few comments

consensus/ibft/ibft.go
```diff
@@ -847,6 +845,14 @@ func (i *Ibft) runAcceptState() { // start new round
...
+				i.logger.Error("block header verification failed", "err", err)
+				i.handleStateErr(errBlockVerificationFailed)
+
+				continue
+			}
+
+			// Verify other block params
+			if err := i.blockchain.VerifyProposedBlock(block); err != nil {
 				i.logger.Error("block verification failed", "err", err)
 				i.handleStateErr(errBlockVerificationFailed)
```
Kourin1996: Can you give me the context why you need to verify block during AcceptState? Is this the improvement that is different from original IBFT flow, right?

Kourin1996 (changes requested):

blockchain/blockchain.go
zivkovicmilos: Removed 👍 

[a7c2300](https://github.com/0xPolygon/polygon-edge/pull/544/commits/a7c23001b14dda29acbba26ac50f327326cc54dd)

blockchain/blockchain.go
zivkovicmilos: Renamed 👍 

I've also noticed that `size` wasn't used 🤦 

[fa47659](https://github.com/0xPolygon/polygon-edge/pull/544/commits/fa47659a7d2f92dc5b82db47e942f0bc82433e50)

blockchain/blockchain.go
zivkovicmilos: Made the number a constant. The `100` value was just chosen at random at one point, so I stuck to it for the newly defined cache.

[87386b7](https://github.com/0xPolygon/polygon-edge/pull/544/commits/87386b77d8c016608d42018a9f5fbfc1b3052171)

blockchain/blockchain.go
zivkovicmilos: Yes - I completely agree with you.

It's part of the technical debt that is accumulating, that the `blockchain` module needs to double check with the consensus module on the validity of a block 🤦‍♂️ 

Currently, there is no way to circumvent this. If you open up this `VerifyHeader` method, you'll notice that it checks for the committed seals - something you don't have when verifying the block from the point of consensus `AcceptState` - that's why block verification is split up into these 2 methods (`VerifySealedBlock` and `VerifyProposedBlock`)

blockchain/blockchain.go
zivkovicmilos: Great catch - header verification by the consensus is less costly than `verifyBlock`.

Fixed 👍 

[80954f3](https://github.com/0xPolygon/polygon-edge/pull/544/commits/80954f376a31326bce369c54ca9a3ba07c818792)

blockchain/blockchain.go
zivkovicmilos: You're not dragging a block, per se, only the result of the block execution:

https://github.com/0xPolygon/polygon-edge/blob/80954f376a31326bce369c54ca9a3ba07c818792/blockchain/blockchain.go#L869-L873

consensus/ibft/ibft.go
zivkovicmilos: This is actually the original IBFT spec flow.

The flow we have currently on the `develop` branch, is that you **don't actually verify a block** before committing to it - you're essentially doing consensus on top of a header filed (check out `verifyHeaderImpl` that was the only check in `AcceptState`).
The logic was that if the block is invalid, you'll error out to round change state, which is insane if you consider that you just went through a consensus round for an unverified block.

Calling the block verification here makes sure that the proposed block is actually valid, and that every part of the consensus process is based on a **verified** block.

Commit state is just writing the block down, no need to verify it again since you're sure it's valid.

Kourin1996 (commented):
Just a comment. Blockchain struct will be no longer safe because it's possible to insert block with incorrect parent hash due to wrong way to use Blockchain. (Please correct me if I'm misreading)

My suggestion is to add checking ParentHash in WriteBlock at least.

zivkovicmilos:
@Kourin1996 

You're right, but it comes down to the semantics of `WriteBlock`. Everywhere in the codebase it's been used with the assumption that it will also verify in addition to writing.

After I've decoupled these concepts, I needed to add a `VerifySealedBlock` from every module that was expecting the verification (I believe only `restore` and `syncer`).

@dbrajovic what do you think about @Kourin1996's suggestion?

blockchain/blockchain.go
Kourin1996: I'm talking about `referenceBlockResult`. All fields come from Block. Am I right?

https://github.com/0xPolygon/polygon-edge/blob/80954f376a31326bce369c54ca9a3ba07c818792/blockchain/blockchain.go#L801-L806

blockchain/blockchain.go
Kourin1996: We need to think about this. If we assume that we don't have access from JSON-RPC, we don't need to fetch past blocks except for parent in usual case. so 100 is too large size, I think. I don't think this is suitable value.

We can skip this discussion for now. But keep in mind.

blockchain/blockchain.go
Kourin1996: At least, the naming of `VerifySealedBlock` and `VerifyProposedBlock` depends on IBFT consensus completely. We need to think about renaming these method.

Kourin1996:
> You're right, but it comes down to the semantics of WriteBlock. Everywhere in the codebase it's been used with the assumption that it will also verify in addition to writing.

@zivkovicmilos I don't think this is the verification. It's like a nil check before inserting.

blockchain/blockchain.go
```diff
@@ -39,6 +54,17 @@ type Blockchain struct {
...
 	difficultyCache *lru.Cache // LRU cache for the difficulty
 
+	// We need to keep track of block receipts between the verification phase
+	// and the insertion phase of a new block coming in. To avoid having to
+	// execute the transactions twice, we save the receipts from the initial execution
+	// in a cache, so we can grab it later when inserting the block.
+	// This is of course not an optimal solution - a better one would be to add
+	// the receipts to the proposed block (like we do with Transactions and Uncles), but
+	// that is currently not possible because it would break backwards compatibility due to
+	// insane conditionals in the RLP unmarshal methods for the Block structure, which prevent
```
0xAleksaOpacic, line 64: Now that we have fixed that funky condition in RLP, can we add receipts to proposed block? Anyways we should refactor this comment because now we can extedns fields in RLP (at least all clients with latest develop would not have a problem)

0xAleksaOpacic (approved):
LGTM! I have manually tested branch with old chain for backwards compactibility. It works

blockchain/blockchain.go
zivkovicmilos: What do you propose we name these?

blockchain/blockchain.go
zivkovicmilos: Agree we should tackle this in the future 👍 

This should be a separate PR and card that tackles the performance implications of having `100`, having it less / more

blockchain/blockchain.go
zivkovicmilos: Ah, you're right 💯 

Fixed 👍 

[18f3f87](https://github.com/0xPolygon/polygon-edge/pull/544/commits/18f3f8704482b7c4abd48650b712dc5efc680ad2)

blockchain/blockchain.go
zivkovicmilos: I would leave the caches for now, it would mess up the flow that's in place and just introduce backtracking for a PR that's already ready.

This PR needs to be viewed in the context it was created in, the fact that it wasn't reviewed in time is a different problem

zivkovicmilos:
@Kourin1996 

I've added a notice to the `WriteBlock` method:

[a74f79a](https://github.com/0xPolygon/polygon-edge/pull/544/commits/a74f79a9ad363149f171a8558ccbd75ff6416c89)

People who need to call `WriteBlock`, will not expect verification. If they want verification, they can call the verification method before committing the block locally

blockchain/blockchain.go
0xAleksaOpacic: Agree, works for me

dbrajovic:
cc @zivkovicmilos @Kourin1996 

Judging by all the places where `WriteBlock()` is used, there should be no side effects despite taking the block verification outside of the method. At each call site,I backtracked the stack to see if verification is still ensured after the call to `WriteBlock` - this does seem to be case. 

As for the naming, it is anti-pattern, but so is the blockchain module doing any verification in the first place. This should be tackled in a separate PR, imo.

consensus/ibft/ibft.go
Kourin1996: @lazartravica I'd like you to confirm it. I may overlook something

blockchain/blockchain.go
Kourin1996: It's much cleaner to call 2 methods in consensus rather than calling specific method of Blockchain.

Kourin1996:
@zivkovicmilos Could you check whether node check each CommittedSeals signature before inserting block?

blockchain/blockchain.go
```diff
@@ -608,82 +657,219 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
+	txn, err := b.executor.ProcessBlock(parent.StateRoot, block, blockCreator)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := b.consensus.PreStateCommit(header, txn); err != nil {
+		return nil, err
 	}
 
+	_, root := txn.Commit()
```
Kourin1996, line 848: `executeBlockTransactions` is called from `verifyBlockBody`. If timeout happens and block proposer changes in IBFT, there will be time wrong state is applied for a while. Am I correct? I think will not affect transaction process but view function might return wrong result.

Kourin1996 (changes requested):

blockchain/blockchain.go
zivkovicmilos: I'm not sure I understand what you mean, we already call 2 methods in consensus:
1. During `AcceptState` - > `VerifyProposedBlock`
2. During `CommitState` -> `WriteBlock`

consensus/ibft/ibft.go
zivkovicmilos: We have a Gitbook doc that details and links why this verification needs to be done in `AcceptState` - it's not really a question whether it should be there, but a correction of bad module design.

zivkovicmilos:
@Kourin1996 

I've gone into a detailed comparison between checks done on develop and checks done on this branch - the only thing that I found missing was the CommittedSeals check before inserting a block, as it was previously contained inside `WriteBlock`.

I've fixed this in:

[92f8113](https://github.com/0xPolygon/polygon-edge/pull/544/commits/92f8113c573706c184508938168b3e0cbd8980f8)

blockchain/blockchain.go
Kourin1996: The problem is that Blockchain has 2 methods to verify block. It completely depends on consensus. The method names breaks blockchain module. The name must explain what this method does. If you can't, you need to remove `VerifyProposedBlock` and call all steps in `VerifyProposedBlock` from consensus directly.

zivkovicmilos: I don't think adding these 2 methods breaks the blockchain module - it was broken the moment we introduced and tied blockchain functionality to the consensus layer:
* [PreStateCommits](https://github.com/0xPolygon/polygon-edge/blob/ddc34803e0ce1f28fc1999023cc02a9248ddd6e4/blockchain/blockchain.go#L844-L846)
* [Snapshot updating](https://github.com/0xPolygon/polygon-edge/blob/ddc34803e0ce1f28fc1999023cc02a9248ddd6e4/blockchain/blockchain.go#L898-L900)
* [Header verification](https://github.com/0xPolygon/polygon-edge/blob/ddc34803e0ce1f28fc1999023cc02a9248ddd6e4/blockchain/blockchain.go#L672-L674)
* [Block creator fetches](https://github.com/0xPolygon/polygon-edge/blob/ddc34803e0ce1f28fc1999023cc02a9248ddd6e4/blockchain/blockchain.go#L834-L837)

Let's not pretend that all of these methods are not IBFT specific and endanger the independence of the blockchain module.

We're discussing if having 2 methods breaks some kind of blockchain independence, when it was broken to begin with, before we added this separation of concern.

The consensus module cannot do the block verification internally because it would require a much bigger separation of concern (primarily of the executor and blockchain module), and additional exporting of blockchain specific methods for this verification to actually happen within consensus.

The best we can do, as of this task and the scope of this task, is to rename these methods to something more meaningful. What names do you suggest?

Kourin1996: I think the examples you're raising are different from this. 

Sure they have been added because consensus needed, but they are just delegation and it means giving some limited ability to inject call. I don't think this is same.

btw, I agree with your opinion. All I can do right now is to rename method.

Kourin1996: How about `VerifySealedBlock` -> `VerifyFinalizedBlock` and `VerifyProposedBlock` -> `VerifyPotentialBlock`? (There may be better name than the second)

zivkovicmilos: That sounds reasonable to me 👍 

Renamed:

[dc73808](https://github.com/0xPolygon/polygon-edge/pull/544/commits/dc73808a50978b603b72e409e9c4b1a19dd4cfb2)

Kourin1996 (approved):

PR #567, opened 2022-05-30
Add additional checks for data handling

zivkovicmilos:
# Description

This PR adds some minor checks for situations where malformed data can cause client panics.

# Changes include

# Checklist

## Testing

### Manual tests

Tested with the scripts provided by the audit crew.

# Additional comments

Fixes EDGE-577
Fixes EDGE-580

### DIFF
--- a/protocol/service_v1.go
+++ b/protocol/service_v1.go
@@ -29,6 +29,12 @@ type rlpObject interface {
 	UnmarshalRLP(input []byte) error
 }
 
+var (
+	errMalformedNotifyRequest = errors.New("malformed notify request")
+	errMalformedNotifyBody    = errors.New("malformed notify body")
+	errMalformedNotifyStatus  = errors.New("malformed notify status")
+)
+
 func (s *serviceV1) Notify(ctx context.Context, req *proto.NotifyReq) (*empty.Empty, error) {
 	var id peer.ID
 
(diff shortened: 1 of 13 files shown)


protocol/service_v1.go
```diff
@@ -55,6 +66,26 @@ func (s *serviceV1) Notify(ctx context.Context, req *proto.NotifyReq) (*empty.Em
 	return &empty.Empty{}, nil
 }
 
+// verifyNotifyRequest verifies the notify request to the peer
+func verifyNotifyRequest(request *proto.NotifyReq) error {
+	// Make sure the request is formed
+	if request == nil {
+		return errMalformedNotifyRequest
+	}
```
Kourin1996, line 74: Is there possibility of that request is nil?

Kourin1996 (approved):
LGTM. Have you faced nil panic on gRPC?

dbrajovic (approved):

0xAleksaOpacic (approved):

protocol/service_v1.go
zivkovicmilos: This is actually a good question 👀 

I believe it can in fact be `nil`, since it's a param that the peer passes in (`req` can be potentially `nil`):
https://github.com/0xPolygon/polygon-edge/blob/f1566f2c0c4529b4aff85c3756a9d240ed8a045a/protocol/syncer.go#L338

Kourin1996 (approved):

PR #570, opened 2022-05-30
Implement removing from filter by ws

### LINKED ISSUE #639: Recurring Error messages `Unable to process flush` / `Unable to write WS message` still exist
### Description

The issue I originally opened in https://github.com/0xPolygon/polygon-edge/issues/486 still exists. 

We are seeing the following errors when a client connected by WS closes prematurely: 

```
[ERROR] polygon.filter: Unable to process flush, writev tcp 127.0.0.1:5000->127.0.0.1:52646: use of closed network connection
```
```
[ERROR] polygon.jsonrpc: Unable to write WS message, writev tcp 127.0.0.1:5000->127.0.0.1:52646: use of closed network connection
```

These errors only go away on node restart.

---

The fix for my original issue seems to only address `websocket: close sent` messages as mentioned in the PR below:
https://github.com/0xPolygon/polygon-edge/pull/487

### Your environment

* OS and version `Ubuntu 20`
* version of the Polygon Edge `0.4.1`

### Steps to reproduce

1. Connect a client directly to a node + subscribe to a WS connected (e.g. Blockscout explorer)
2. Terminate the client process
3. Observe recurring `Unable to process flush` on the host node

### Expected behaviour

I would only expect an initial error indicating an issue with the WS connection, rather than seeing the same error persist requiring a node restart.

### Actual behaviour

The same error persists requiring a node restart.

0xAleksaOpacic:
# Description

This PR fixes the issue with closing ws connection. 

Fixes EDGE-584

# Changes include

# Checklist

## Testing

### Manual tests

Subscribe to newBlockHeadEvent (https://github.com/Aleksao998/SubscribeToNewHeadEvent) and close the connection, you should not see errors (on develop), on this branch no error logs should appear
```
2022-05-30T14:59:00.022+0200 [ERROR] polygon.jsonrpc: Unable to write WS message, writev tcp 127.0.0.1:10002->127.0.0.1:64893: use of closed network connection
2022-05-30T14:59:00.022+0200 [ERROR] polygon.filter: Unable to process flush, writev tcp 127.0.0.1:10002->127.0.0.1:64893: use of closed network connection
```


jsonrpc/filter_manager.go
```diff
@@ -390,6 +390,25 @@ func (f *FilterManager) removeFilterByID(id string) bool {
 	return true
 }
 
+// removeFilterByWs removes the filter with given WS, unsafe against race condition
+func (f *FilterManager) RemoveFilterByWs(ws wsConn) bool {
+	for _, filter := range f.filters {
```
Aleksao998, line 395: This could be method getFilterByWs

jsonrpc/jsonrpc.go
```diff
@@ -209,6 +210,11 @@ func (j *JSONRPC) handleWs(w http.ResponseWriter, req *http.Request) {
 				j.logger.Info("Closing WS connection with error")
 			}
 
+			ok := j.dispatcher.RemoveFilterByWs(wrapConn)
+			if !ok {
+				j.logger.Error("Unable to remove filter with specific WS")
+			}
+
```
zivkovicmilos: This can be inlined

jsonrpc/jsonrpc.go
```diff
@@ -40,6 +40,7 @@ type JSONRPC struct {
 }
 
 type dispatcher interface {
+	RemoveFilterByWs(conn wsConn) bool
```
zivkovicmilos: Why do you need to have feedback data from this method?

jsonrpc/filter_manager.go
```diff
@@ -390,6 +390,25 @@ func (f *FilterManager) removeFilterByID(id string) bool {
...
+		}
+
+		delete(f.filters, filter.getFilterBase().id)
+
+		if removed := f.timeouts.removeFilter(filter.getFilterBase()); removed {
+			f.emitSignalToUpdateCh()
+		}
+
+		return true
+	}
```
zivkovicmilos: This can be optimized - iterating this way is really not performant.

What do you think about adding support for some kind of WS connection ID system? Then, when someone disconnects, you can just delete the connection from the FM with that ID.

jsonrpc/jsonrpc.go
0xAleksaOpacic: fixed dfa9cf9b1ee51761e111616bca079e4d33220df0

jsonrpc/jsonrpc.go
0xAleksaOpacic: fixed dfa9cf9b1ee51761e111616bca079e4d33220df0

jsonrpc/filter_manager.go
0xAleksaOpacic: Fixed dd60f2d672f94444aa9a46fc732c33a46ae61a30 + 865b2c3aed0c45e48c16f21980f4b44c10b319a3

Kourin1996 (approved):
LGTM

dbrajovic (commented):
Tbh, I don't find it great that now you have to manage (add/remove) the same object in two places: `filters` and `filterByWsConn`. But I guess that's due to the technical debt pf this package. 

Just curious: was there any chance to expand the `wsConn` interface with like an `ID() string` method, so that you could easily remove the filter from the `filters` map when calling `dispatcher`'s `RemoveFilterByWs(conn wsConn)` ?

0xAleksaOpacic:
> Tbh, I don't find it great that now you have to manage (add/remove) the same object in two places: `filters` and `filterByWsConn`. But I guess that's due to the technical debt pf this package.
> 
> Just curious: was there any chance to expand the `wsConn` interface with like an `ID() string` method, so that you could easily remove the filter from the `filters` map when calling `dispatcher`'s `RemoveFilterByWs(conn wsConn)` ?

Not sure I understand the second part. 
`filters  map[string]filter`
So filter map is a map where the key is filterID and the value is filter. This bug is due to when we are closing WS connection we are not removing this filter. So I had to have a way of finding the filter by wsConn and then removing it from the FILTER map. 
Not sure if wsConn should know about filters. It's a basic web socket connection. FilterManager should handle everything related to the filters

dbrajovic:
> Not sure if wsConn should know about filters. It's a basic web socket connection. FilterManager should handle everything > related to the filters

Yes, exactly - you needed a way to remove a certain filter just by having its `wsConn`. But because you can't extract the ID based on `wsConn`, you keep another record of the same object. 

The same filter object can be found in 2 places;

[1st](https://github.com/0xPolygon/polygon-edge/blob/756512e55a142f3cfbb6f7a580aeed9f15013899/jsonrpc/filter_manager.go#L254)
[2nd](https://github.com/0xPolygon/polygon-edge/blob/756512e55a142f3cfbb6f7a580aeed9f15013899/jsonrpc/filter_manager.go#L258)

This just adds to the technical debt of this package (which will be tackled in a separate issue).

Approving now. 👍

dbrajovic (approved):

jsonrpc/filter_manager.go
zivkovicmilos: I'm not sure if you should use an interface here as the map key, even though the language spec allows it:
>As mentioned earlier, map keys may be of any type that is comparable. The [language spec](https://go.dev/ref/spec#Comparison_operators) defines this precisely, but in short, comparable types are boolean, numeric, string, pointer, channel, and interface types, and structs or arrays that contain only those types. Notably absent from the list are slices, maps, and functions; these types cannot be compared using ==, and may not be used as map keys.

@Kourin1996 ?

zivkovicmilos: @Aleksao998 

I like @dbrajovic's suggestion on having this interface give you an ID back:

https://github.com/0xPolygon/polygon-edge/pull/570#pullrequestreview-1039944856

zivkovicmilos (approved):
Looks good now, thank you for resolving the nitpicks 🙏 

This filter manager is trash and god willing we'll get a chance to take it out soon, but until then this will work.

Please see the comments I've left you 🙏 

Looks good 💯

jsonrpc/filter_manager.go
0xAleksaOpacic: Really love when we are playing and optimising PR-s especially these ones where we are sure that no big refactors are needed af9f7ff6d4b01815b6e7ea3f1aba80c4026dd1bb) :)

zivkovicmilos (approved):

jsonrpc/filter_manager.go
```diff
@@ -459,18 +479,29 @@ func (f *FilterManager) GetLogsForQuery(query *LogQuery) ([]*Log, error) {
...
+	f.RLock()
+	defer f.RUnlock()
+
+	filter, exists := f.filters[filterID]
+
+	if !exists {
+		return nil
+	}
+
+	return filter
```
Kourin1996: filter is interface type and map returns nil if the value doesn't exist in the map. then I don't think we need nil check here.

jsonrpc/filter_manager_test.go
```diff
@@ -292,6 +292,24 @@ func TestFilterTimeout(t *testing.T) {
...
 
+func TestRemoveFilterByWebsocket(t *testing.T) {
+	store := newMockStore()
+
+	mock := &mockWsConn{
+		msgCh: make(chan []byte, 1),
+	}
+
+	m := NewFilterManager(hclog.NewNullLogger(), store)
+	go m.Run()
```
Kourin1996, line 326: Please close after the test

Kourin1996 (changes requested):
LGTM

jsonrpc/filter_manager.go
zivkovicmilos: You're actually right 👍 

Fixed: 
[12f591a](https://github.com/0xPolygon/polygon-edge/pull/570/commits/12f591a234b3db3ba72e4894e1c6b077b3b916d7)

jsonrpc/filter_manager_test.go
zivkovicmilos: Fixed, for all tests:

[246f04e](https://github.com/0xPolygon/polygon-edge/pull/570/commits/246f04eeb3f24e18b3807fe59dd54343f7fc7441)

Kourin1996 (approved):
LGTM

PR #571, opened 2022-05-31
[TxPool] Prune stale enqueued txs

dbrajovic:
# Description

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -143,8 +143,8 @@ func (m *accountsMap) allTxs(includeEnqueued bool) (
 	return
 }
 
-func (m *accountsMap) pruneStaleEnqueuedTxs() uint64 {
-	total := uint64(0)
+func (m *accountsMap) pruneStaleEnqueuedTxs() []*types.Transaction {
+	pruned := make([]*types.Transaction, 0)
 
 	m.Range(func(_, value interface{}) bool {
 		account, _ := value.(*account)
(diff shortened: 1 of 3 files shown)


txpool/account.go
```diff
@@ -3,10 +3,15 @@ package txpool
 import (
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+var (
+	maxAccountInactivity = 3 * time.Hour
```
0xAleksaOpacic, line 12: Why have we decided for 3h?

txpool/account.go
```diff
@@ -135,6 +143,28 @@ func (m *accountsMap) allTxs(includeEnqueued bool) (
 	return
 }
 
+func (m *accountsMap) pruneStaleEnqueuedTxs() []*types.Transaction {
+	pruned := make([]*types.Transaction, 0)
+
+	m.Range(func(_, value interface{}) bool {
+		account, _ := value.(*account)
```
0xAleksaOpacic, line 150: Do we need to add check for casting? Something like this https://github.com/0xPolygon/polygon-edge/pull/567/commits/9c4a3a823cd481af109e807c409586feae5f1dc2#diff-155513a5d2e83af1befd35c715a88ec4e88f1ea6b3eb5eb95ead0dd73ef76276R680

txpool/account.go
DarianShawn: The default value is coming from `Ethereum`, I guess.

Would be better to shorten it, like `30min` since the block generation time is `1/6` or so of the `Ethereum`.

PR #574, opened 2022-06-01
[PoC] BLS Aggregate Signature in IBFT ComittedSeal

Kourin1996:
Fix EDGE-550

# Description

This PR adds BLS feature which enables aggregate CommittedSeal signatures.
To use BLS, you will need to start BLS mode like this

```
$ polygon-edge secrets init ... --bls # display BLS Public Key as well
[SECRETS INIT]
Public key (address) = <VALIDATOR_ADDR>
BLS Public key       = <VALIDATOR_BLS_PUBLIC_KEY>
Node ID              = <LIBP2P_NODE ID>

$ polygon-edge genesis --consensus ibft --bls \
--ibft-validator <VALIDATOR_ADDR>:<VALIDATOR_BLS_PUBLIC_KEY> ...

$ polygon-edge server --bls ...
```

With BLS, validator uses 2 signing algorithm ECDSA for Seal and BLS for CommittedSeal because BLS doesn't provide ecrecover, which returns address from the signature.
Currently BLS secrete key is generated from ECDSA private key but each validators should know all validator's BLS public keys to verify CommittedSeal. For that reason, you need to give BLS pubkeys when generating genesis.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


zivkovicmilos:
@Kourin1996 

Can you merge out the `develop branch` into whatever is the target branch here?

It's hard to follow along changes that are already present on `develop`

command/genesis/params.go
```diff
@@ -197,21 +232,30 @@ func (p *genesisParams) initValidatorSet() error {
 }
 
 func (p *genesisParams) isValidatorNumberValid() bool {
-	return uint64(len(p.ibftValidators)) <= p.maxNumValidators
+	return p.ibftValidators == nil || uint64(p.ibftValidators.Len()) <= p.maxNumValidators
```
zivkovicmilos, line 234: How is it correct for this check to return ✅ if the `p.ibftValidators` is `nil`?

command/genesis/params.go
```diff
@@ -319,3 +369,69 @@ func (p *genesisParams) getResult() command.CommandResult {
...
+		vals = append(vals, validators.BLSValidator{
+			Address:   types.BytesToAddress(addrBytes),
+			BLSPubKey: pubKeyBytes,
+		})
+	}
+
+	newValSet := validators.BLSValidatorSet(vals)
+
+	return &newValSet, nil
+}
```
zivkovicmilos: What do you think about moving these methods out do some better package, like `validators`?

This is not logic that is specific for the genesis params

command/genesis/params.go
```diff
@@ -319,3 +369,69 @@ func (p *genesisParams) getResult() command.CommandResult {
...
+	}
+
+	for _, v := range values {
+		bytes, err := hex.DecodeString(strings.TrimPrefix(v, "0x"))
+		if err != nil {
+			return nil, err
+		}
+
+		addrs = append(addrs, types.BytesToAddress(bytes))
+	}
```
zivkovicmilos: This method (and the analog `parseBLSValidators`) does way too much:
* It parses a validator set if it's passed in, and uses it as the base validator set
* It parses the validator set in the form of addresses from the params

What do you think about splitting / simplifying this logic up?

cc @dbrajovic

command/genesis/utils.go
```diff
@@ -119,8 +122,47 @@ func getValidatorsFromPrefixPath(prefix string) ([]types.Address, error) {
...
+
+		vals[idx] = validators.BLSValidator{
+			Address:   crypto.PubKeyToAddress(&key.PublicKey),
+			BLSPubKey: pubkey,
+		}
+	}
+
+	valSet := validators.BLSValidatorSet(vals)
+
+	return &valSet, nil
```
zivkovicmilos: What do you think about moving this out to a more appropriate place?

command/secrets/init/params.go
```diff
@@ -44,10 +50,18 @@ func (ip *initParams) validateFlags() error {
 		return errInvalidParams
 	}
 
+	if ip.rawKeyType == "" {
+		return errInvalidParams
```
zivkovicmilos: This error is not appropriate for the key type, as it's related to directory initializations.

Please use a more appropriate error (create a new one) 🙏

command/server/server.go
```diff
@@ -197,6 +198,14 @@ func setFlags(cmd *cobra.Command) {
 		"write all logs to the file at specified location instead of writing them to console",
 	)
 
+	cmd.Flags().BoolVar(
+		&params.useBls,
+		blsFlag,
+		false,
+		// TODO
```
zivkovicmilos: Leftover TODO

server/server.go
```diff
@@ -405,6 +405,7 @@ func (s *Server) setupConsensus() error {
 			Metrics:        s.serverMetrics.consensus,
 			SecretsManager: s.secretsManager,
 			BlockTime:      s.config.BlockTime,
+			BLS:            s.config.BLS,
```
zivkovicmilos, line 409: Why are we passing it as part of the `server` command configuration?

Can't we use it as part of the consensus `config` in the `genesis.json`?

That way `server` doesn't have to worry about whether BLS is enabled or not

secrets/helper/helper.go
```diff
@@ -79,22 +79,37 @@ func SetupGCPSSM(
...
 
+	address := crypto.PubKeyToAddress(&validatorKey.PublicKey)
+
 	// Write the validator private key to the secrets manager storage
 	if setErr := secretsManager.SetSecret(
 		secrets.ValidatorKey,
 		validatorKeyEncoded,
 	); setErr != nil {
-		return nil, setErr
+		return types.ZeroAddress, setErr
```
zivkovicmilos, line 79: Why `types.ZeroAddress` and not `nil`?

secrets/helper/helper.go
```diff
@@ -79,22 +79,37 @@ func SetupGCPSSM(
...
+
+func GetBLSPubkeyFromValidatorKey(secretsManager secrets.SecretsManager) ([]byte, error) {
+	ecdsaKey, err := crypto.ReadConsensusKey(secretsManager)
+	if err != nil {
+		return nil, err
+	}
+
+	pubkeyBytes, err := crypto.ECDSAToBLSPubkey(ecdsaKey)
+	if err != nil {
+		return nil, err
```
zivkovicmilos: Why is this in the `secrets` package, when it only contains crypto functionality?

e2e/ibft_test.go
```diff
@@ -30,6 +32,7 @@ func TestIbft_Transfer(t *testing.T) {
 		func(i int, config *framework.TestServerConfig) {
 			config.Premine(senderAddr, framework.EthToWei(10))
 			config.SetSeal(true)
+			config.SetShowsLog(i == 0)
```
zivkovicmilos: This is leftover

crypto/crypto.go
```diff
@@ -299,3 +316,29 @@ func ReadConsensusKey(manager secrets.SecretsManager) (*ecdsa.PrivateKey, error)
...
+
+	_, sk, err := blsPop.KeygenWithSeed(keyBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return sk, nil
+}
+
+func ECDSAToBLSPubkey(ecdsaKey *ecdsa.PrivateKey) ([]byte, error) {
```
zivkovicmilos: Nitpick: Maybe we should call this method something different, as it doesn't really output the Public key, but the marshalled public key.

What do you think about `ECDSAToBLSPublicMarshalled`?

cc @dbrajovic

consensus/ibft/ibft.go
```diff
@@ -32,6 +31,7 @@ var (
 	ErrInvalidHookParam     = errors.New("invalid IBFT hook param passed in")
 	ErrInvalidMechanismType = errors.New("invalid consensus mechanism type in params")
 	ErrMissingMechanismType = errors.New("missing consensus mechanism type in params")
+	ErrSignerNotFound       = errors.New("not found signer in validator set")
```
zivkovicmilos: Typo here, should be `signer not found in validator set`

consensus/ibft/ibft.go
```diff
@@ -210,12 +222,11 @@ func (i *Ibft) Start() error {
 		proto.RegisterIbftOperatorServer(i.Grpc, i.operator)
 	}
 
-	// Set up the node's validator key
-	if err := i.createKey(); err != nil {
-		return err
-	}
+	i.msgQueue = newMsgQueue()
+	i.closeCh = make(chan struct{})
+	i.updateCh = make(chan struct{})
```
zivkovicmilos: Just curious, why init it here and now? 👀

zivkovicmilos: Ah, you moved it from `createKey` which is now deprecated 👍

consensus/ibft/ibft.go
```diff
@@ -998,27 +963,18 @@ func (i *Ibft) updateMetrics(block *types.Block) {
...
-					commit.From,
-				),
-			)
+	committedSeals := make(map[types.Address][]byte)
 
-			continue
+	for addr, commit := range i.state.committed {
+		committedSeal, err := hex.DecodeHex(commit.Seal)
+		if err != nil {
+			return fmt.Errorf("failed to decode committed seal from %s: %w", addr, err)
```
zivkovicmilos: Do we want this to be fatal?

You dropped the `continue` from this conditional. This means that if you have `1` invalid committed seal, you will go into round change

consensus/ibft/ibft.go
```diff
@@ -1244,9 +1200,9 @@ func (i *Ibft) setState(s IbftState) {
 }
 
 // forceTimeout sets the forceTimeoutCh flag to true
-func (i *Ibft) forceTimeout() {
-	i.forceTimeoutCh = true
-}
+// func (i *Ibft) forceTimeout() {
+// 	i.forceTimeoutCh = true
+// }
```
zivkovicmilos: Leftover

consensus/ibft/ibft.go
```diff
@@ -1290,7 +1246,14 @@ func (i *Ibft) verifyHeaderImpl(snap *Snapshot, parent, header *types.Header) er
...
+		err = i.signer.VerifyParentCommittedSeal(
+			parentSnap.Set,
+			parent,
+			header,
+			i.quorumSize(header.Number),
+		)
+
+		if err != nil {
 			return fmt.Errorf("failed to verify ParentCommittedSeal: %w", err)
 		}
```
zivkovicmilos: You can inline this

consensus/ibft/ibft.go
```diff
@@ -1389,6 +1352,18 @@ func (i *Ibft) Close() error {
...
+func (i *Ibft) SetHeaderHash() {
+	types.HeaderHash = func(h *types.Header) types.Hash {
+		hash, err := i.signer.CalculateHeaderHash(h)
+		if err != nil {
+			return types.ZeroHash
+		}
+
+		return hash
+	}
+}
```
zivkovicmilos: Is there any way we can mitigate setting this **global** callback to something specific?

Usually this is a red flag for bad module design, can we evaluate where we need this `types.HeaderHash`?

cc @dbrajovic

zivkovicmilos: I know we were setting this global variable previously as well, I'm just wondering if we can go away with this shenanigans 👀

consensus/ibft/pos.go
```diff
@@ -204,13 +206,21 @@ func (pos *PoSMechanism) ShouldWriteTransactions(blockNumber uint64) bool {
...
 	}
 
-	return staking.QueryValidators(transition, pos.ibft.validatorKeyAddr)
+	valAddrs, err := staking.QueryValidators(transition, pos.ibft.signer.Address())
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: PoS contract doesn't hold BLS PubKeys
+	valSet := validators.ECDSAValidatorSet(valAddrs)
```
zivkovicmilos: Leftover TODO?

command/server/server.go
Kourin1996: This is not leftover but we need to update an interface of contract and the its implementation.

secrets/helper/helper.go
Kourin1996: No reason. When this method return error, first return value will be zero value and will not be used. Or do you want caller to check first returning value is nil?

crypto/crypto.go
Kourin1996: Not `ECDSAToMarshalledBLSPublic`?

consensus/ibft/pos.go
Kourin1996: I commented in other your review

consensus/ibft/ibft.go
Kourin1996: HashCalculation is used widely and I know global variable is not proper

Kourin1996: We need to figure out which modules are using this

Kourin1996: But hash calculation completely depends on consensus module due to extra field

consensus/ibft/snapshot_test.go
```diff
@@ -50,36 +52,42 @@ func (t *testerAccount) Address() types.Address {
...
-	h, _ = writeSeal(t.priv, h)
+func (t *testerAccount) sign(h *types.Header) (*types.Header, error) {
+	signer := signer.NewECDSASignerFromKey(t.priv)
 
-	return h
+	return signer.WriteSeal(h)
 }
 
 type testerAccountPool struct {
+	t        *testing.T
```
zivkovicmilos, line 64: I honestly don't think we should be dragging a reference to `testing.T` around in the helper methods.

Why not just return errors for methods that can error out?

consensus/ibft/signer/signer.go
```diff
@@ -0,0 +1,43 @@
...
+	ErrInvalidValidatorSet        = errors.New("invalid validator set type")
+	ErrInvalidBLSSignature        = errors.New("invalid BLS signature")
+	ErrNilParentHeader            = errors.New("parent header is nil")
+)
+
+// Signer is responsible for signing for blocks and messages in IBFT
+type Signer interface {
+	Address() types.Address
+	InitIBFTExtra(header, parent *types.Header, set validators.ValidatorSet) error
+	GetIBFTExtra(*types.Header) (*IstanbulExtra, error)
```
zivkovicmilos, line 30: I love the idea of having a separate Signer.

Now we can easily plug and play in tests since the `Signer` is an interface.

Do you think there is any way we can separate out this IBFT specific logic from the signer?
If not right now, just curious what you think on separating it out in the future.

consensus/ibft/signer/helper.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return err
+	}
+
+	msg.From = addr.String()
+
+	return nil
+}
+
+// TODO: move to IBFT?
```
zivkovicmilos: Leftover?

consensus/ibft/signer/extra_test.go
```diff
@@ -0,0 +1,199 @@
+package signer
```
zivkovicmilos, line 1: This entire file is commented out

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,362 @@
...
+		return nil, err
+	}
+
+	err = s.packFieldIntoIbftExtra(header, func(ie *IstanbulExtra) {
+		ie.Seal = seal
+	})
+
+	if err != nil {
+		return nil, err
+	}
```
zivkovicmilos: This can be inlined

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,362 @@
...
+		return ErrInvalidCommittedSealType
+	}
+
+	validatorSet, ok := rawSet.(*validators.ECDSAValidatorSet)
+	if !ok {
+		return ErrInvalidValidatorSet
+	}
+
+	// get the message that needs to be signed
+	// this not signing! just removing the fields that should be signed
```
zivkovicmilos: Deprecated comment?

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,362 @@
...
+		return ErrInvalidCommittedSealType
+	}
+
+	validatorSet, ok := rawParentSet.(*validators.ECDSAValidatorSet)
+	if !ok {
+		return ErrInvalidValidatorSet
+	}
+
+	// get the message that needs to be signed
+	// this not signing! just removing the fields that should be signed
```
zivkovicmilos: Deprecated comment?

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,362 @@
...
+
+	putIbftExtra(h, extra)
+
+	return nil
+}
+
+func (s *ECDSASigner) filterHeaderForHash(header *types.Header) (*types.Header, error) {
+	// This will effectively remove the Seal and Committed Seal fields,
+	// while keeping proposer vanity and validator set
+	// because extra.Validators, extra.ParentCommittedSeal is what we got from `h` in the first place.
```
zivkovicmilos: Deprecated comment

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,362 @@
...
+		if !validators.Includes(addr) {
+			return ErrNonValidatorCommittedSeal
+		}
+
+		visited[addr] = struct{}{}
+	}
+
+	// Valid committed seals must be at least 2F+1
+	// 	2F 	is the required number of honest validators who provided the committed seals
+	// 	+1	is the proposer
```
zivkovicmilos: This comment is no longer valid

cc @dbrajovic

consensus/ibft/signer/bls.go
```diff
@@ -0,0 +1,462 @@
...
+}
+
+func (s *BLSSigner) Address() types.Address {
+	return s.address
+}
+
+func (s *BLSSigner) InitIBFTExtra(header, parent *types.Header, set validators.ValidatorSet) error {
+	var parentCommittedSeal Sealer
+
+	if header.Number > 1 {
```
zivkovicmilos: Why is this check even required?

consensus/ibft/signer/bls.go
```diff
@@ -0,0 +1,462 @@
...
+		return ErrInvalidCommittedSealType
+	}
+
+	validatorSet, ok := rawSet.(*validators.BLSValidatorSet)
+	if !ok {
+		return ErrInvalidValidatorSet
+	}
+
+	// get the message that needs to be signed
+	// this not signing! just removing the fields that should be signed
```
zivkovicmilos: Deprecated?

Same as with `ecdsa.go`, for the remaining comments

consensus/ibft/signer/bls.go
```diff
@@ -0,0 +1,462 @@
...
+
+// 	sigBytes, err := sig.MarshalBinary()
+// 	if err != nil {
+// 		return err
+// 	}
+
+// 	msg.Signature = hex.EncodeToHex(sigBytes)
+
+// 	return nil
+// }
```
zivkovicmilos: Leftover?

consensus/ibft/validators/bls.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Add adds a new address to the validator set
+func (v *BLSValidatorSet) Add(addr types.Address) {
+	// TODO: doesn't support right now
+}
+
+// Del removes an address from the validator set
+func (v *BLSValidatorSet) Del(target types.Address) {
+	// TODO: doesn't support right now
+}
```
zivkovicmilos: Does this mean that the set can't ever be changed from the start?

consensus/ibft/signer/bls.go
Kourin1996: Because genesis block doesn't have CommittedSeal

consensus/ibft/validators/bls.go
Kourin1996: Do you agree with changing type of Miner in Header from Address to []byte?
@zivkovicmilos

Kourin1996:
Clean up Signer in https://github.com/0xPolygon/polygon-edge/pull/574/commits/b2a4844949c1d111d6bc3ea959291e44fc47cdcb

ECDSASigner and BLSSigner have redundant codes, so I renamed them to ECDSAKeyManager, BLSManager and added new Signer. Now Signer defines the flow and ECDSASigner and BLSSigner defines the implementation

command/genesis/params.go
Kourin1996: Good idea. Moved at b8b41249444ab5cda3c156f251c6a9ced84bb5f1

command/genesis/params.go
Kourin1996: In b8b41249444ab5cda3c156f251c6a9ced84bb5f1, I remove appending from this function

command/genesis/utils.go
Kourin1996: Good catch. Moved in 54c8009a3f1e8d324502565a9aff5bf57d3697b0

command/secrets/init/params.go
Kourin1996: Added new error in 5c1ee77068408fb15fdddbcf3a0e77d14f46abc8

secrets/helper/helper.go
Kourin1996: Moved in b12f4bdc7253b1bbb514a566e2e1bb25b2602b1d

e2e/ibft_test.go
Kourin1996: Removed in 1cb50b199da5a3e25ef7293593079cef6bf454b5

crypto/crypto.go
Kourin1996: Renamed in ab0ac775cd9fc1db493b913b2cf8d0239da4638a

consensus/ibft/ibft.go
Kourin1996: Fixed in b5b0944399257ea29d9b1210cc16dcab4344aeea

consensus/ibft/ibft.go
Kourin1996: Followed latest code in d9af3b51e19bf7d8ab606f4a549fbaaf68a19113

consensus/ibft/ibft.go
Kourin1996: Fixed in a90c04b73f85d87689c1701e24cc87b156a06331

consensus/ibft/ibft.go
Kourin1996: Fixed in c04ef82713935e921b1f320d7788a70dde7cf307

consensus/ibft/signer/helper.go
Kourin1996: Removed in c04ef82713935e921b1f320d7788a70dde7cf307

consensus/ibft/signer/ecdsa.go
Kourin1996: Fixed in c04ef82713935e921b1f320d7788a70dde7cf307

consensus/ibft/signer/ecdsa.go
Kourin1996: Removed

consensus/ibft/signer/ecdsa.go
Kourin1996: Removed

consensus/ibft/signer/ecdsa.go
Kourin1996: No, but I moved below c04ef82713935e921b1f320d7788a70dde7cf307

consensus/ibft/signer/bls.go
Kourin1996: Removed

consensus/ibft/signer/bls.go
Kourin1996: Removed

consensus/ibft/signer/extra_test.go
Kourin1996: Uncommented in 4ba1feaf1bdaa053ece5b8a4c3ffd7ecd4a19362

Kourin1996:
@zivkovicmilos Please resolve comments

secrets/helper/helper.go
dbrajovic: Also, you can't use `nil` when the return type is `types.Address`

consensus/ibft/snapshot.go
```diff
@@ -254,7 +256,72 @@ type Snapshot struct {
...
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	s.Number = raw.Number
+	s.Hash = raw.Hash
+	s.Votes = raw.Votes
+
+	if len(raw.Set) > 0 {
+		switch raw.Set[0].(type) {
```
dbrajovic, line 350: Maybe we could extract this switch logic into a separate helper ?

consensus/ibft/ibft.go
```diff
@@ -998,12 +954,11 @@
...
 
 func (i *Ibft) insertBlock(block *types.Block) error {
-	committedSeals := make([][]byte, 0)
+	committedSeals := make(map[types.Address][]byte)
 
-	for _, commit := range i.state.committed {
-		// no need to check the format of seal here because writeCommittedSeals will check
-		committedSeal, decodeErr := hex.DecodeHex(commit.Seal)
-		if decodeErr != nil {
+	for addr, commit := range i.state.committed {
```
dbrajovic: Readability nitpick: make this a separate method (extractSeals)

consensus/ibft/ibft.go
dbrajovic: Given how widespread usage of `Header`'s `ComputeHash()` method is, I'd say this is best we can do right now. 

A good direction to think about this is how many modules would actually need the hash of a given `Header`. If that list isn't too long, then maybe this implementation could be more specific.

consensus/ibft/validators/bls.go
```diff
@@ -0,0 +1,303 @@
...
+	return val
+}
+
+func (v *BLSValidator) UnmarshalRLPFrom(p *fastrlp.Parser, vv *fastrlp.Value) error {
+	vals, err := vv.GetElems()
+	if err != nil {
+		return fmt.Errorf("mismatch of RLP type for CommittedSeal, expected list but found %s", vv.Type())
+	}
+
+	{
```
dbrajovic: Nitpick: maybe the wrapping scope braces can be removed ?

consensus/ibft/validators/bls.go
```diff
@@ -0,0 +1,303 @@
...
+func (v *BLSValidatorSet) UnmarshalRLPFrom(p *fastrlp.Parser, vv *fastrlp.Value) error {
+	vals, err := vv.GetElems()
+	if err != nil {
+		return fmt.Errorf("mismatch of RLP type for ECDSAValidatorSet, expected list but found %s", vv.Type())
+	}
+
+	valSet := make([]BLSValidator, len(vals))
+
+	for index := range vals {
+		vals := vals
```
dbrajovic: Why the identity assignment ?

consensus/ibft/signer/bls.go
zivkovicmilos: Of course 🤦‍♂️

consensus/ibft/validators/bls.go
zivkovicmilos: Honestly, I'm completely fine with whatever gives us greater flexibility. If it's `[]byte` than I'm cool with that 🙂 

cc @dbrajovic @Aleksao998

consensus/ibft/signer/ecdsa.go
```diff
@@ -0,0 +1,182 @@
...
+	committedSeal *SerializedSeal,
+	msg []byte,
+	validators validators.ECDSAValidatorSet,
+) (int, error) {
+	numSeals := len(*committedSeal)
+	if numSeals == 0 {
+		return 0, ErrEmptyCommittedSeals
+	}
+
+	visited := map[types.Address]struct{}{}
```
dbrajovic: What about a map of booleans (set) ?

consensus/ibft/pos.go
```diff
@@ -204,13 +206,21 @@ func (pos *PoSMechanism) ShouldWriteTransactions(blockNumber uint64) bool {
...
 		return nil, err
 	}
 
-	return staking.QueryValidators(transition, pos.ibft.validatorKeyAddr)
+	valAddrs, err := staking.QueryValidators(transition, pos.ibft.signer.Address())
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: PoS contract doesn't hold BLS PubKeys
```
dbrajovic: Is this a reference for a future feature ?

dbrajovic (changes requested):
Good job on encapsulating the `signer` logic from the rest of ibft. On that note, what do you think about the following:

Since the `signer` package defines an interface which it implements with `KeyManager` under the hood (`SignerImpl`), maybe this is a good opportunity for custom constructors ? (e.g. `NewBLSSigner`, `NewECDSASigner`). The constructors would still return a `Signer` type, preserving the used api. 

This would add great flexibility should there be a need for another implementation. (I realize it might be a hassle to change this now, just interested in your opinion).

cc @zivkovicmilos

Kourin1996:
@dbrajovic I'm not sure. Custom contract is different feature. At least signer will not depend on its implementation. Signer should focus on only signing and validators will be managed by different module

dbrajovic:
> At least signer will not depend on its implementation.


@Kourin1996 
Exactly. `ibft` package sees just the `Signer` interface, in which case the `InitSigner` can be simplified to:

```
func InitSigner(secretManager secrets.SecretsManager, isBLS bool) (signer.Signer, error) {
	if isBLS {
		return NewBLSSigner(secretManager)
	}
	
	return NewECDSASigner(secretManager)
}
````

**Comment by lazartravica** [comment]
@Kourin1996 we can move this to draft until the backwards compatibility is clear.

**Comment by zivkovicmilos** [comment]
@Kourin1996 

Please pull in the latest develop and resolve any conflicts 🙏

File: server/server.go (hunk 7 above)
Comment (Kourin1996) [reply server/server.go:409]: Resolving because this code has changed in the next PR https://github.com/0xPolygon/polygon-edge/pull/649.
Basically IBFT Hook module takes this configurations

File: consensus/ibft/validators/bls.go (hunk 31 above)
Comment (Kourin1996) [reply consensus/ibft/validators/bls.go]: Resolving because this stuff is done in this PR https://github.com/0xPolygon/polygon-edge/pull/649

File: consensus/ibft/snapshot.go (hunk 32 above)
Comment (Kourin1996) [reply consensus/ibft/snapshot.go:350]: Resolving this because this issue is resolved in the next PR https://github.com/0xPolygon/polygon-edge/pull/649

File: consensus/ibft/pos.go (hunk 37 above)
Comment (Kourin1996) [reply consensus/ibft/pos.go]: Resolving because this issue is fixed in the next PR https://github.com/0xPolygon/polygon-edge/pull/649

File: consensus/ibft/signer/ecdsa.go (hunk 36 above)
Comment (Kourin1996) [reply consensus/ibft/signer/ecdsa.go]: Good idea. Fixed at 52379b35c037308d4ff8e5b77253cacc292214f1

File: consensus/ibft/validators/bls.go (hunk 35 above)
Comment (Kourin1996) [reply consensus/ibft/validators/bls.go]: Resolving because this code has been removed already

File: consensus/ibft/validators/bls.go (hunk 34 above)
Comment (Kourin1996) [reply consensus/ibft/validators/bls.go]: Resolving because this code has changed already

File: validators/types.go [hunk 38]
```diff
@@ -0,0 +1,96 @@
...
+	case ECDSAValidatorType, BLSValidatorType:
+		*t = x
+
+		return nil
+	default:
+		return fmt.Errorf("invalid validator type: %s", s)
+	}
+}
+
+func NewValidatorFromType(t ValidatorType) Validator {
```
dbrajovic, line 38: Unused ctor

validators/types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	switch x {
+	case ECDSAValidatorType, BLSValidatorType:
+		*t = x
+
+		return nil
+	default:
+		return fmt.Errorf("invalid validator type: %s", s)
+	}
+}
```
dbrajovic, line 36: Nitpick: replace `FromString` method with `ValidatorTypeFromString` free func.

```
func ValidatorTypeFromString(s string) (ValidatorType, error) {
	vt := ValidatorType(s)

	switch vt {
	case ECDSAValidatorType, BLSValidatorType:
		return vt, nil
	default:
		return "", errors.New("asd")
	}
}
```

types/types.go
```diff
@@ -150,6 +150,10 @@ func BytesToAddress(b []byte) Address {
 	return a
 }
 
+func AddressToBytes(a Address) []byte {
```
dbrajovic, line 153: Unused

validators/types.go
Kourin1996: Removed in the next PR, a85e43ec5b5186573f08192d7be09d99242748b4

validators/types.go
Kourin1996: Can you tell me what the difference is?

types/types.go
Kourin1996: Removed in the next PR, a42d46541be2da9a1ad487aec790e151ecd06fad

PR #575, opened 2022-06-03
[TxPool] Fix duplicate transactions in account

dbrajovic:
# Description

This PR fixes the issue of two identical transactions ending up back-to-back inside an account. This bug is the true source of the numerous `failed to apply tx: incorrect nonce` errors.

The second transaction is a gossip of the first, but because indexing happens in `handleEnqueueRequest` both of them complete the preceding call to `addTx` bypassing the (now deprecated) check for `ErrAlreadyKnown`. 

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EDGE-585

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -594,22 +594,6 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 		return ErrAlreadyKnown
 	}
 
-	//// check if already known
-	//if _, ok := p.index.get(tx.Hash); ok {
-	//	if origin == gossip {
-	//		// silently drop known tx
-	//		// that is gossiped back
-	//		p.logger.Debug(
-	//			"dropping known gossiped transaction",
-	//			"hash", tx.Hash.String(),
-	//		)
-	//
-	//		return nil
-	//	} else {
-	//		return ErrAlreadyKnown
-	//	}
-	//}
-
 	// initialize account for this address once
 	if !p.accounts.exists(tx.From) {
 		p.createAccountOnce(tx.From)
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -589,21 +589,26 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 
 	tx.ComputeHash()
 
-	// check if already known
-	if _, ok := p.index.get(tx.Hash); ok {
-		if origin == gossip {
-			// silently drop known tx
-			// that is gossiped back
-			p.logger.Debug(
-				"dropping known gossiped transaction",
-				"hash", tx.Hash.String(),
-			)
-
-			return nil
-		} else {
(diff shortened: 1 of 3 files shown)


txpool/txpool.go
```diff
@@ -693,14 +682,20 @@ func (p *TxPool) addGossipTx(obj interface{}) {
 
 	// decode tx
 	if err := tx.UnmarshalRLP(raw.Raw.Value); err != nil {
-		p.logger.Error("failed to decode broadcasted tx", "err", err)
+		p.logger.Error("failed to decode broadcast tx", "err", err)
```
zivkovicmilos, line 685: Slomiće te sistem

cc @Aleksao998

zivkovicmilos (approved):
`txpool: failed to apply tx: incorrect nonce`, I banish you to the shadow realm 🚀

txpool/txpool.go
0xAleksaOpacic: Hahahhaahh well educated man 😂

0xAleksaOpacic: Should I tell him or will you?

txpool/txpool.go
```diff
@@ -589,20 +589,9 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
...
-			)
-
-			return nil
-		} else {
-			return ErrAlreadyKnown
-		}
+	//	add to index
+	if ok := p.index.add(tx); !ok {
+		return ErrAlreadyKnown
 	}
```
Kourin1996, line 595: I think you can move this check above so that we can skip transaction validation twice. What do you think?
cc: @zivkovicmilos

Kourin1996 (approved):
LGTM

0xAleksaOpacic (approved):

txpool/txpool.go
dbrajovic: Sounds good, as long as the transaction is removed from the index map, if triggering overflow and validate errors. One thing to note about the hash used for the index - I remember the old implementation having a comment on  the hash being safe to compute after all `validateTx` checks have passed.

Not sure what it meant at the time, but if it's okay to compute the hash of the transaction before validation checks, I'll move the indexing up top of the call stack. (cc @zivkovicmilos )

zivkovicmilos: @dbrajovic 

The comment was referring to the fact that we were initializing the `From` field from the `V`, `R`, `S` sig fields.
That way computing the hash would be unique, as the `From` field would be initialized

dbrajovic: ~~So it's perfectly fine to compute the hash before validation ?~~ 

It turns out we can't really compute the hash before validation is done. The `From` field in the transaction needs to be extracted before we do the computing. 


@Kourin1996 
In other words, it's only safe to compute the hash **after** validation checks pass. I cannot move the indexing to the top of the call.

PR #581, opened 2022-06-06
Unlock proposed block after bulk sync

Kourin1996:
Fix EDGE-573

# Description

This PR fixes the issue that a validator will not unlock the proposed block if it's left at past height for some reason.
And this PR adds safe check to the correctness of proposing block height because validator can propose block with wrong height by setting wrong view in preprepare message.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -730,6 +730,7 @@ func (i *Ibft) runAcceptState() { // start new round
 	parent := i.blockchain.Header()
 	number := parent.Number + 1
 
+	// Check whether current sequence matches the next height
 	switch {
 	case number > i.state.view.Sequence:
 		// Node has synced already but state needs to update
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -853,7 +853,7 @@ func (i *Ibft) runAcceptState() { // start new round
 
 		if block.Number() != i.state.view.Sequence {
 			i.logger.Error("sequence not correct", "block", block.Number, "sequence", i.state.view.Sequence)
-			i.setState(SyncState)
+			i.handleStateErr(errIncorrectBlockHeight)
 
 			return
 		}
@@ -1066,9 +1066,10 @@ func (i *Ibft) insertBlock(block *types.Block) error {
 }
 
 var (
-	errIncorrectBlockLocked    = fmt.Errorf("block locked is incorrect")
-	errBlockVerificationFailed = fmt.Errorf("block verification failed")
-	errFailedToInsertBlock     = fmt.Errorf("failed to insert block")
+	errIncorrectBlockLocked    = errors.New("block locked is incorrect")
+	errIncorrectBlockHeight    = errors.New("proposed block number is incorrect")
+	errBlockVerificationFailed = errors.New("block verification failed")
+	errFailedToInsertBlock     = errors.New("failed to insert block")
 )
 
 func (i *Ibft) handleStateErr(err error) {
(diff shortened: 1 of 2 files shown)


consensus/ibft/ibft.go
```diff
@@ -734,9 +730,22 @@ func (i *Ibft) runAcceptState() { // start new round
...
+			i.state.unlock()
+		}
+
+		i.startNewSequence(number)
+	case number < i.state.view.Sequence:
+		// For some reason, sequence exceeds current block height
+		// Needs to wait until other validators catch up
+
+		i.logger.Warn("sequence exceeds latest block height", "latest", parent.Number, "sequence", i.state.view.Sequence)
+		i.handleStateErr(errIncorrectSequence)
```
zivkovicmilos: I'm trying to figure out if this is even possible with the added verification flow.

By possible, I mean your latest block header from storage being `!=` from the next block that should be in the sequence. We update the sequence during `Commit` state and `Sync` state. If the flow is correct for both, then this can never be possible.

What do you think?

cc @dbrajovic

consensus/ibft/ibft.go
```diff
@@ -847,6 +856,13 @@ func (i *Ibft) runAcceptState() { // start new round
 			return
 		}
 
+		if block.Number() != i.state.view.Sequence {
+			i.logger.Error("sequence not correct", "block", block.Number, "sequence", i.state.view.Sequence)
+			i.handleStateErr(errIncorrectBlockHeight)
+
+			return
+		}
```
zivkovicmilos, line 867: This check is already handled in the `blockchain` module:

https://github.com/0xPolygon/polygon-edge/blob/dad1ed048b7e0d7c54b9235b0663c1cc1eb6258c/blockchain/blockchain.go#L737-L746

zivkovicmilos (changes requested):
I've left some comments that need discussing 🙏 

Please resolve the linting errors when you get the chance

consensus/ibft/ibft.go
Kourin1996: I fixed and move catch up process into the end of Sync process

consensus/ibft/ibft.go
Kourin1996: Hey, please don't talk about other module 😂. I don't like removing required checks because other module does.

consensus/ibft/ibft.go
zivkovicmilos: You're talking about commits:
* [9756882](https://github.com/0xPolygon/polygon-edge/pull/581/commits/975688215a644c51e4873dc4c9e5f471688166e5)
* [a80dca3](https://github.com/0xPolygon/polygon-edge/pull/581/commits/a80dca3919f2e510ff097c5eab53b08d686e7ff0)
?

consensus/ibft/ibft.go
zivkovicmilos: Think of the nanoseconds we're wasting by doing this double checking 😢

Kourin1996: If you want to reduce nano seconds time, I recommend you to write all code in main function. or we have still a lot of things to optimize. This is not a waste of time

consensus/ibft/ibft.go
Kourin1996: Yes

consensus/ibft/ibft.go
Kourin1996: I meant robustness, extensibility and readability sacrificed performance a little bit. But this is acceptable.

zivkovicmilos (approved):
Thank you for simplifying the code 🙏 

Looks great 💯

dbrajovic (approved):

PR #590, opened 2022-06-13
[Fix] Improve eth_gasPrice responce when --price-limit flag is set

ZeljkoBenovic:
# Description

Currently, `eth_gasPrice` does not take into account if `--price-limit` flag has been set by the chain operator.   
 Even if `--price-limit` flag is set to some value ( for example `100000` ), `eth_gasPrice` will return `0`.    

This PR fixes the issue. Now `eth_gasPrice` returns at minimum the value set by `--price-limit` flag.

# Changes include

# Checklist

## Testing

### Manual tests
Spin up a chain using this PR code, but run the server with `--price-limit` set to `100000` for example.    
Run curl targeting  `eth_gasPrice` endpoint. For example:     
`curl  localhost:8545 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_gasPrice","params":[],"id":1}'`     
     
You should get the value set in `--price-limit` flag ( `100000` ) from the API.

# Additional comments

Fixes EDGE-538


zivkovicmilos (changes requested):
This is not how `--price-limit`  should be affecting other endpoints, it's a purely node-based check for txpool validations and nothing more.

`eth_estimateGas`  returns the gas estimate for the transaction, regardless of the price limit. Making it depend on the price limit effectively breaks the endpoint, as you'll get differing results from other clients which is something we should really try to avoid.

`eth_gasPrice`  returns the average gas price for the running chain - if it was 0 it returns 0, I don't see a need to have it return anything other than what it currently is, otherwise the endpoint should've been called `eth_averageGasPriceOrMinimum` . The only optimization the avg. gas price mechanism needs is historical data (write to DB the previous average and count, and on node bootup load them into memory)

ZeljkoBenovic:
Removed `eth_estimateGas` "fixes" 5881a09b5a01821d88d485d43b9e18379c9e2ebd.

Kourin1996 (approved):
LGTM

0xAleksaOpacic (changes requested):
LGTM, thank you for this PR! 
Can we please add PR for the doc where we explain the change in the logic of JSON-RPC gasPrice method https://edge-docs.polygon.technology/docs/get-started/json-rpc-commands#eth_gasprice

zivkovicmilos (approved):
To the JSON-RPC module:

Get ready, we're coming for you 🙂 

Looks good @ZeljkoBenovic 💯

0xAleksaOpacic (approved):

Kourin1996 (approved):

PR #591, opened 2022-06-15
Syncer V2

Kourin1996:
# Description

Implementation of the new syncing protocol (Lightning Sync)

# Changes include

# Breaking changes

Yes this is a braking change

# Checklist

## Testing

### Manual tests

We have done a regression test for this module
Some of the test cases:

1. Syncing new node and continuing in watchSync state
2. Syncing multiple nodes at the same time with 1 validator
3. Stopping the network while in bulkSync and restarting to see will it continue
4. From watchSync to validator to watchSync

# Documentation update

Will link when i open a PR

# Additional comments

Fixes EDGE-592, EDGE-593, EDGE-594, EDGE-595


consensus/ibft/ibft.go
```diff
@@ -490,8 +482,7 @@ func (i *Ibft) runSyncState() {
 
 	for i.isState(SyncState) {
 		// try to sync with the best-suited peer
-		p := i.syncer.BestPeer()
-		if p == nil {
+		if !i.syncer.HasSyncPeer() {
```
zivkovicmilos, line 489: The naming here should be different - you're effectively looking for a peer that has a better chain than you do (bigger height) to jump out of RoundChangeState or to know if you should attempt syncing with them.

What do you think about `HigherChainExists`?

cc @dbrajovic

consensus/ibft/ibft.go
```diff
@@ -536,7 +527,7 @@ func (i *Ibft) runSyncState() {
 			// The snapshot store is currently updated for PoA inside the ProcessHeadersHook
 			callInsertBlockHook(newBlock.Number())
 
-			i.syncer.Broadcast(newBlock)
+			i.syncer.UpdateLocalLatestBlock(newBlock)
```
zivkovicmilos: Why is this a call here, and not something that is done asynchronously in the syncer itself?

It's in the spec we've shared.

All you need to do is tap into blockchain events and just fire away a pubsub message when you get a new block added event

network/server_identity.go
```diff
@@ -130,3 +134,10 @@ func (s *Server) registerIdentityService(identityService *identity.IdentityServi
 
 	s.RegisterProtocol(common.IdentityProto, grpcStream)
 }
+
+func (s *Server) GetPeerDistance(peerID peer.ID) *big.Int {
```
zivkovicmilos, line 138: Please add an explanation as to what distance this method actually gives you (from the kademlia table)

syncer/peers/nofork/proto/syncer.proto
```diff
@@ -0,0 +1,37 @@
...
+
+option go_package = "/syncer/peers/nofork/proto";
+
+import "google/protobuf/empty.proto";
+
+service SyncPeer {
+  rpc GetBlocks(GetBlocksRequest) returns (stream Block);
+  rpc GetBlock(GetBlockRequest) returns (Block);
+  rpc GetStatus(google.protobuf.Empty) returns (Status);
+  rpc NotifyStatus(Status) returns (google.protobuf.Empty);
```
zivkovicmilos: Why are you diverging from the spec we've agreed upon?

In the spec we've specifically decided to eliminate this status fetching and explicit notify RPC calls.
Why did you reintroduce them back, as they add an insane network and flow overhead?

syncer/syncer.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const (
+	LoggerName = "syncer"
+)
+
+type Syncer interface {
+	Start()
+	GetSyncProgression() *progress.Progression
+	BestPeer() *peers.BestPeer
```
zivkovicmilos: Why is this an exported method? Who needs to know about the best peer apart from the actual syncing protocol?

syncer/syncer.go
```diff
@@ -0,0 +1,141 @@
...
+	// Write block to chain
+	WriteBlock(*types.Block) error
+}
+
+type Progression interface {
+	StartProgression(startingBlock uint64, subscription blockchain.Subscription)
+	UpdateHighestProgression(highestBlock uint64)
+	GetProgression() *progress.Progression
+	StopProgression()
+}
```
zivkovicmilos: Please notice that you are defining an interface that other modules will use for interacting with the syncer (`Syncer` interface in this package), but **are also defining interfaces you require from other modules such as blockchain and progression**.

Please look into keeping these interfaces where they belong 🙏 

cc @dbrajovic

syncer/syncer.go
```diff
@@ -0,0 +1,141 @@
...
+
+		if err := s.blockchain.WriteBlock(block); err != nil {
+			return fmt.Errorf("failed to write block while bulk syncing: %w", err)
+		}
+
+		newBlockCallback(block)
+	}
+
+	return nil
+}
```
zivkovicmilos: Why are you diverging from the agreed upon spec?

You are not handling any of the cases where the sending node chooses to drop the stream connection with the current node, as we've outlined and discussed extensively in the specification.

Please add these cases 🙏

syncer/peers/errors.go
```diff
@@ -0,0 +1,8 @@
+package peers
+
+import "errors"
+
+var (
+	ErrorHeaderNotFound = errors.New("header not found")
+	ErrorBlockNotFound  = errors.New("block not found")
+)
```
zivkovicmilos: Why are these errors defined here?

syncer/peers/heap.go
```diff
@@ -0,0 +1,114 @@
...
+// Case 2: Update peer info and reorder if it exists already
+func (h *PeerHeap) Put(peer Peer) {
+	h.RWMutex.Lock()
+	defer h.RWMutex.Unlock()
+
+	index, ok := h.lookupMap[peer.ID()]
+	if ok {
+		// exists already
+		h.peers[index] = peer
+		heap.Fix(h, index)
```
zivkovicmilos: Why are you re-adding a peer if they are already in the heap? What kind of information can change that would trigger a reorder?

syncer/peers/nofork/service.go
```diff
@@ -0,0 +1,131 @@
...
+type SyncPeerService struct {
+	proto.UnimplementedSyncPeerServer
+
+	logger     hclog.Logger
+	blockchain Blockchain
+	network    Network
+
+	id           string // node ID
+	peerStatusCh chan *NoForkPeer
+}
```
zivkovicmilos: What is the point of redefining this here in a separate package?

You need to drag around references to the blockchain and network modules, and you fall into a situation where your parent class has a reference to you, but both of you have these interface references as well 🤷‍♂️

syncer/peers/nofork/service.go
```diff
@@ -0,0 +1,131 @@
...
+	req *proto.Status,
+) (*empty.Empty, error) {
+	s.peerStatusCh <- &NoForkPeer{
+		id:       req.Id,
+		Number:   req.Number,
+		Distance: s.network.GetPeerDistance(peer.ID(req.Id)),
+	}
+
+	return &empty.Empty{}, nil
+}
```
zivkovicmilos: Completely useless and unnecessary. These are obsolete in the spec

syncer/peers/nofork/types.go
```diff
@@ -0,0 +1,36 @@
+package nofork
```
zivkovicmilos: Why are these defined in the `nofork` package?

syncer/peers/nofork/manager.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+const (
+	SyncPeersManagerLoggerName = "sync-peers-manager"
+)
+
+type SyncPeersManager struct {
+	logger     hclog.Logger
+	network    Network
+	blockchain Blockchain
```
zivkovicmilos: Please note that by far:
* the outer syncer obj has references to these modules
* the manager obj has references to these modules
* the service server obj has references to these modules

And they all reference each other in some way 🤷‍♂️ 

cc @dbrajovic

syncer/peers/nofork/manager.go
```diff
@@ -0,0 +1,241 @@
...
+) peers.SyncPeers {
+	return &SyncPeersManager{
+		logger:     logger.Named(SyncPeersManagerLoggerName),
+		network:    network,
+		blockchain: blockchain,
+		server: NewSyncPeerService(
+			logger,
+			blockchain,
+			network,
+		),
```
zivkovicmilos: Does this look right to you?

cc @dbrajovic

syncer/peers/nofork/manager.go
```diff
@@ -0,0 +1,241 @@
...
+
+			continue
+		}
+
+		clt.NotifyStatus(context.Background(), &proto.Status{
+			Id:     m.id,
+			Number: latest.Number,
+		})
+	}
+}
```
zivkovicmilos: Why are you diverging from the agreed upon spec?

syncer/peers/nofork/manager.go
```diff
@@ -0,0 +1,241 @@
...
+func (m *SyncPeersManager) newSyncPeerClient(peerID peer.ID) (proto.SyncPeerClient, error) {
+	stream, err := m.network.NewStream(peers.SyncerProto, peerID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open a stream, err %w", err)
+	}
+
+	conn := grpc.WrapClient(stream)
+
+	return proto.NewSyncPeerClient(conn), nil
+}
```
zivkovicmilos: Why are you diverging from the agreed upon spec?

In the spec, we explicitly agreed that you **ignore** messages from non-connected peers.

Here, you're instantiating a new stream towards a peer, when you can't even guarantee that they're a connected and verified peer

zivkovicmilos (changes requested):
I really don't like that you've ignored parts of the spec that we've agreed upon. We dropped some functionality and flows for a reason, but you chose to ignore that and reintroduced them back (pubsub is completely missing, Notify and Broadcast are back, no agreed upon error handling...).

Why are you supporting forking? It complicates the code exponentially, and makes it harder to read and work with, especially implement `WatchSync` that @Aleksao998 is working on. I especially don't like the complexity of the `nofork` package - I strongly believe we should drop it completely.

I don't see us introducing a consensus mechanism that supports forking in the near future, or even at all at this point - we can always add that functionality when we need it in a year or more, or ever. 

There are also 0 tests in this PR, which is a shame since the whole point of us basing Lighting Sync on gRPC streams and RPC method calls, and simplifying the flow, was that it made it much easier to test and verify.

consensus/ibft/ibft.go
Kourin1996: `HigherChainExists` is unclear, I think. It's hard to see where the chain is?
@zivkovicmilos

consensus/ibft/ibft.go
Kourin1996: I was going to use Blockchain subscription in syncer. This is the left code, I guess

syncer/peers/nofork/proto/syncer.proto
Kourin1996: I confirmed about Gossip with this morning and I've not worked on this yet.
I will let you know once I pushed this change

syncer/syncer.go
Kourin1996: @zivkovicmilos 
I don't think this is duplicate definition. for example, we need to write mock that implements all methods defined in the interface in blockchain package when using interface of blockchain defined in blockchain package. Basically the use of blockchain depends on behaviour of the module.

Kourin1996: Of course, no need to define the interface if it is not subset of the original

syncer/syncer.go
Kourin1996: The client will receive error and process will terminate. Or do you need an additional condition to terminate stream in sender side?

syncer/peers/heap.go
Kourin1996: (Latest Block Height, Distance) is used for ordering heap. Latest Block Height is not constant value, we need to update this when receiving new status

Kourin1996: @zivkovicmilos

syncer/peers/nofork/manager.go
Kourin1996: @zivkovicmilos I don't see what you wanted to say. What is your concern?

syncer/peers/nofork/manager.go
Kourin1996: @zivkovicmilos I'm working on changing from gRPC to Gossip now

syncer/peers/nofork/manager.go
Kourin1996: With gRPC, node should not communicate with non-connected peer. This problem occurs when I change the code to use Gossip
@zivkovicmilos

Kourin1996:
@zivkovicmilos 

Firstly, please note this PR is not ready for merge right now. It will be ready in the next few days. But we're still discussing about something about spec. It was completely my mistake not to mark it as draft. 

> I really don't like that you've ignored parts of the spec that we've agreed upon. We dropped some functionality and flows for a reason, but you chose to ignore that and reintroduced them back (pubsub is completely missing, Notify and Broadcast are back, no agreed upon error handling...).

There are several things we're discussing. I confirmed the thing about Gossip this morning. I've not worked on this change yet.

> Why are you supporting forking? It complicates the code exponentially, and makes it harder to read and work with, especially implement WatchSync that @Aleksao998 is working on. I especially don't like the complexity of the nofork package - I strongly believe we should drop it completely.
I don't see us introducing a consensus mechanism that supports forking in the near future, or even at all at this point - we can always add that functionality when we need it in a year or more, or ever.

This syncer doesn't support forking right now. But I don't want to sacrifice modularity. Those who wants to implement new consensus themselves, they need to implement syncer completely. But actually the flow of syncer will be same. The common process should be defined in the core of syncer and the part that may depend on consensus must be separated from the core.

> There are also 0 tests in this PR, which is a shame since the whole point of us basing Lighting Sync on gRPC streams and RPC method calls, and simplifying the flow, was that it made it much easier to test and verify.

I'l try to keep the existing tests as much as possible. But some of them are not proper because they depend on network layer

syncer/peers/nofork/manager.go
Kourin1996: @zivkovicmilos 
Fixed 17fd0aa3cb63d501b009ab85867dee90fe26ae8d

syncer/peers/nofork/manager.go
Kourin1996: @zivkovicmilos 
Fixed 17fd0aa3cb63d501b009ab85867dee90fe26ae8d

syncer/peers/nofork/manager.go
Kourin1996: the way to decide best peer depends on consensus, so there is no problem

consensus/ibft/ibft.go
Kourin1996: Removed 65ccd2e2d1c75c12d64044b32b5540368b29b770

syncer/peers/nofork/proto/syncer.proto
Kourin1996: @zivkovicmilos 
Removed 17fd0aa3cb63d501b009ab85867dee90fe26ae8d

syncer/peers/errors.go
Kourin1996: What do you mean?

Kourin1996: @zivkovicmilos

syncer/peers/nofork/service.go
Kourin1996: I don't see what you're trying to say. The struct will have the references it needs

syncer/peers/nofork/types.go
Kourin1996: To keep extensibility, New syncer will not work with other consensus

syncer/proto/syncer.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package v1;
```
0xAleksaOpacic, line 3: If we want to use message Status, we need to use a different package. Networks proto are in v1 package and already have message Status

syncer/peers.go
```diff
@@ -0,0 +1,71 @@
...
+	var bestPeer *NoForkPeer
+
+	m.Range(func(key, value interface{}) bool {
+		peer, ok := value.(*NoForkPeer)
+		if !ok {
+			// shouldn't reach here
+			return false
+		}
+
+		if skipMap != nil && skipMap[bestPeer.ID] {
```
0xAleksaOpacic: I think it needs to be skipMap[**peer.id**]

syncer/syncer.go
```diff
@@ -0,0 +1,187 @@
...
+
+	var lastReceivedNumber uint64
+
+	for {
+		select {
+		case block, ok := <-blockCh:
+			if !ok {
+				return lastReceivedNumber, nil
+			}
+
```
0xAleksaOpacic, line 291: we should not sync genesys block, so if block.Number === 0 continue

syncer/syncer.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		// Set the target height
+		s.syncProgression.UpdateHighestProgression(bestPeer.Number)
+
+		lastNumber, err := s.bulkSyncWithPeer(bestPeer.ID, newBlockCallback)
+		if err != nil {
+			s.logger.Warn("failed to complete bulk sync with peer, try to next one", "peer ID", bestPeer.ID)
+		}
+
```
0xAleksaOpacic, line 198: We need to update localLatest with lastNumber

Kourin1996: I added 074b472668bee39e3645502cb6763a7e2ebbffa6

syncer/syncer.go
Kourin1996: Syncer will try to sync from the next of latest, but I added safe check here

Kourin1996: 074b472668bee39e3645502cb6763a7e2ebbffa6

syncer/peers.go
Kourin1996: Thank you. I fixed 074b472668bee39e3645502cb6763a7e2ebbffa6

syncer/proto/syncer.proto
Kourin1996: What do you mean? Do you want to define Status in another proto?

0xAleksaOpacic: All our proto files in polygon-edge repo are inside the same package (**package v1**). In one package you can not have a message with the same name. **identity.proto** already has a message named **Status**. 

We have added in syncer.proto **message Status**. If you try to run it you will get an error that you cannot have the same message name inside one package.

I am suggesting just to move syncer.proto in another package and then we can have same message names in multiple files

syncer/syncer.go
```diff
@@ -165,6 +172,11 @@ func (s *syncer) bulkSyncWithPeer(peerID string, newBlockCallback func(*types.Bl
 	for {
 		select {
 		case block, ok := <-blockCh:
+			// safe check
+			if block.Number() == 0 {
+				continue
+			}
+
 			if !ok {
```
0xAleksaOpacic: if !ok needs to be before if block.Number == 0, or we can get null pointer error

syncer/proto/syncer.proto
Kourin1996: I understood now. Fixed ef8b18e795c3af1692bc5c5d49c764dc23d64c3a

syncer/syncer.go
Kourin1996: Good catch.
Fixed 67e14fffdf70d5f699d1d272d2f37906497bce5f

syncer/client.go
```diff
@@ -69,7 +69,28 @@ func (m *syncPeerClient) Close() {
...
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), defaultTimeoutForStatus)
+	defer cancel()
+
+	status, err := clt.GetStatus(timeoutCtx, &emptypb.Empty{})
+	if err != nil {
+		return nil, err
+	}
+
+	return &NoForkPeer{
+		ID:       peerID.String(),
```
0xAleksaOpacic: Not sure why, but this syntax was not working, looks like peerId.String return string which is used only for output I had to change to 
ID:       string(peerID),
to make it work, will investigate this further, just wanted to give you info

Kourin1996: Fixed 5b4810311e5b20ac5e2437b0ec48b85bdb4c8402

txpool/txpool.go
```diff
@@ -659,7 +659,7 @@ func (p *TxPool) handlePromoteRequest(req promoteRequest) {
 
 // addGossipTx handles receiving transactions
 // gossiped by the network.
-func (p *TxPool) addGossipTx(obj interface{}) {
+func (p *TxPool) addGossipTx(obj interface{}, _ string) {
```
0xAleksaOpacic: we need to change _ string -> _ peer.ID (after 5b4810311e5b20ac5e2437b0ec48b85bdb4c8402)

Kourin1996: Fixed a03539daf2596aa9955f09933c0d04821e816e07

0xAleksaOpacic: Thank you!

0xAleksaOpacic:
When i run sincer i get spam of warnings on validator nodes.
How to reproduce it:
1. start  a network with 4 validator nodes
2. start 5-th node, syncing will start
3. check logs on validator nodes
```
2022-06-21T11:10:02.510+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.512+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.514+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.516+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.518+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.520+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.522+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.524+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
2022-06-21T11:10:02.526+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
```

0xAleksaOpacic:
> When i run sincer i get spam of warnings on validator nodes. How to reproduce it:
> 
> 1. start  a network with 4 validator nodes
> 2. start 5-th node, syncing will start
> 3. check logs on validator nodes
> 
> ```
> 2022-06-21T11:10:02.510+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.512+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.514+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.516+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.518+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.520+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.522+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.524+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> 2022-06-21T11:10:02.526+0200 [WARN]  polygon.consensus.sync-peers-manager: received status from non-connected peer, ignore: id=16Uiu2HAm5dThqDuZLqsgjYNVoWibx1iAKguBEVxh9YJogQtRbJbJ
> ```

@Kourin1996 When you have time can you please check this

syncer/proto/syncer.proto
```diff
@@ -0,0 +1,27 @@
...
+
+message Block {
+  // RLP Encoded Block Data
+  bytes block = 1;
+}
+
+message SyncPeerStatus {
+  // Latest block height
+  uint64 number = 1;
+}
```
zivkovicmilos: Please add more in-code docs for these proto concepts 🙏

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+	peerConnectionUpdateCh chan *event.PeerEvent
+}
+
+func NewSyncPeerClient(
+	logger hclog.Logger,
+	network Network,
+	blockchain Blockchain,
+) SyncPeerClient {
+	return &syncPeerClient{
+		logger:                 logger.Named(SyncPeersManagerLoggerName),
```
zivkovicmilos: You are creating a `SyncPeerClient`, but the logger is named `sync-peers-manager`

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
+	var (
+		ps            = m.network.Peers()
+		syncPeers     = make([]*NoForkPeer, 0, len(ps))
+		syncPeersLock = sync.Mutex{}
+		wg            sync.WaitGroup
+	)
+
+	for _, p := range ps {
+		p := p
```
zivkovicmilos, line 123: Why don't you just pass this in as a param for the go func, instead of redefining it here?

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+		Number:   status.Number,
+		Distance: m.network.GetPeerDistance(peerID),
+	}, nil
+}
+
+func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
+	var (
+		ps            = m.network.Peers()
+		syncPeers     = make([]*NoForkPeer, 0, len(ps))
+		syncPeersLock = sync.Mutex{}
```
zivkovicmilos: I don't think you need to define it explicitly here, you can only state the type like you did with wg:

`syncPeersLock sync.Mutex`

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	return &NoForkPeer{
+		ID:       peerID,
+		Number:   status.Number,
+		Distance: m.network.GetPeerDistance(peerID),
+	}, nil
+}
+
+func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
```
zivkovicmilos, line 114: Please add an explanation for this method

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+	return syncPeers
+}
+
+func (m *syncPeerClient) GetPeerStatusUpdateCh() <-chan *NoForkPeer {
+	return m.peerStatusUpdateCh
+}
+
+func (m *syncPeerClient) GetPeerConnectionUpdateEventCh() <-chan *event.PeerEvent {
+	return m.peerConnectionUpdateCh
+}
```
zivkovicmilos, line 158: What do you think about naming this `GetPeerEventCh`?

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+
+func (m *syncPeerClient) startNewBlockProcess() {
+	m.subscription = m.blockchain.SubscribeEvents()
+
+	for event := range m.subscription.GetEventCh() {
+		if event.Type == blockchain.EventHead && len(event.NewChain) > 0 {
+			latest := m.blockchain.Header()
+			if latest == nil {
+				return
+			}
```
zivkovicmilos: Can't you just extract the latest header number from the event?

Also, I don't think you need the conditional check (when is this even possible?)

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+		m.logger.Error("failed to subscribe", "err", err)
+
+		return
+	}
+
+	for e := range peerEventCh {
+		switch e.Type {
+		case event.PeerConnected, event.PeerDisconnected:
+			m.peerConnectionUpdateCh <- e
+		}
```
zivkovicmilos: Do you even need a switch here?
This can be simplified

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+}
+
+func (m *syncPeerClient) CloseStream(peerID peer.ID) error {
+	return m.network.CloseProtocolStream(SyncerProto, peerID)
+}
+
+func (m *syncPeerClient) GetBlocks(
```
zivkovicmilos, line 243: What do you think about naming this `GetBlocksFromPeer` to avoid confusion with the gRPC method `GetBlocks`?

syncer/client.go
```diff
@@ -0,0 +1,277 @@
...
+			if err != nil {
+				break
+			}
+
+			block, err := fromProto(protoBlock)
+			if err != nil {
+				m.logger.Warn("failed to decode a block from peer", "peerID", peerID, "err", err)
+
+				break
+			}
```
zivkovicmilos: What happens when the peer hangs in sending you blocks?

You should rely on a timeout here somewhere

syncer/peers.go
```diff
@@ -0,0 +1,77 @@
...
+// BestPeer returns the top of heap
+func (m *PeerMap) BestPeer(skipMap map[peer.ID]bool) *NoForkPeer {
+	var bestPeer *NoForkPeer
+
+	m.Range(func(key, value interface{}) bool {
+		peer, ok := value.(*NoForkPeer)
+		if !ok {
+			// shouldn't reach here
+			return false
+		}
```
zivkovicmilos: I think you can just assume that the cast is ok

syncer/service_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	ctx := context.Background()
+	conn, err := grpc.DialContext(ctx, "bufnet",
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithContextDialer(
+			func(ctx context.Context, address string) (net.Conn, error) {
+				return lis.Dial()
+			},
+		),
+	)
```
zivkovicmilos, line 42: Please note that you are not "mocking" anything here, you're literally instantiating a gRPC client.

Why couldn't you just write a mock client implementation, and tested it the same way we test `discovery` and `identity` mocks?

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
...
+
+	"github.com/0xPolygon/polygon-edge/helper/progress"
+	"github.com/0xPolygon/polygon-edge/network/event"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/hashicorp/go-hclog"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const (
+	LoggerName  = "syncer"
```
zivkovicmilos: Why is this exported?

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
...
+	SyncerProto = "/syncer/0.2"
+)
+
+var (
+	errTimeout = errors.New("timeout")
```
zivkovicmilos: Please expand on what kind of timeout this is

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
+package syncer
```
zivkovicmilos, line 1: Please add in-code docs for methods in this file 🙏

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	return nil
+}
+
+func (s *syncer) WatchSync(ctx context.Context, callback func(*types.Block) bool) error {
+	localLatest := s.blockchain.Header().Number
+	skipList := make(map[peer.ID]bool)
+
+	// Loop until context is canceled
```
zivkovicmilos: Typo here, should be `cancelled`

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
...
+	for {
+		//Wait for a new event to arrive
+		<-s.newStatusCh
```
zivkovicmilos, line 219: Where do you close this channel?

syncer/syncer.go
```diff
@@ -0,0 +1,272 @@
...
+
+		if shouldTerminate {
+			break
+		}
+	}
+
+	return nil
+}
+
+func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.Block) bool) (uint64, bool, error) {
```
zivkovicmilos, line 262: Can we not have this method return 50 different things? 🙃

syncer/types.go
```diff
@@ -0,0 +1,79 @@
...
+	GetPeerDistance(peer.ID) *big.Int
+	// NewProtoConnection opens up a new stream on the set protocol to the peer,
+	// and returns a reference to the connection
+	NewProtoConnection(protocol string, peerID peer.ID) (*rawGrpc.ClientConn, error)
+	// NewTopic Creates New Topic for gossip
+	NewTopic(protoID string, obj proto.Message) (*network.Topic, error)
+	IsConnected(peerID peer.ID) bool
+	CloseProtocolStream(protocol string, peerID peer.ID) error
+	SaveProtocolStream(protocol string, stream *rawGrpc.ClientConn, peerID peer.ID)
+}
```
zivkovicmilos: Some methods are documented, some aren't

syncer/types.go
```diff
@@ -0,0 +1,79 @@
...
+type SyncPeerClient interface {
+	Start() error
+	Close()
+	GetPeerStatus(id peer.ID) (*NoForkPeer, error)
+	GetConnectedPeerStatuses() []*NoForkPeer
+	GetBlocks(context.Context, peer.ID, uint64) (<-chan *types.Block, error)
+	GetPeerStatusUpdateCh() <-chan *NoForkPeer
+	GetPeerConnectionUpdateEventCh() <-chan *event.PeerEvent
+	CloseStream(peerID peer.ID) error
+}
```
zivkovicmilos: Nothing is documented

vendor/modules.txt
```diff
@@ -930,6 +930,7 @@ google.golang.org/grpc/serviceconfig
 google.golang.org/grpc/stats
 google.golang.org/grpc/status
 google.golang.org/grpc/tap
+google.golang.org/grpc/test/bufconn
```
zivkovicmilos, line 933: You do not need this import to test out the gRPC client / server functionality. You can just mock them instead

zivkovicmilos (approved):
Overall, the PR is good 👍 

The performance gains we have are massive, so kudos for that 🍻 

The only thing that bugged me about the PR was that we, instead of mocking the clients / servers, wrote complex code to literally instantiate real connections inside the test and have them connect asynchronously - I think we could've improved there.

Great feature 💯

syncer/client.go
Kourin1996: I don't know the difference. Is it faster? Using local variables to write mode code.

syncer/service_test.go
Kourin1996: I don't see what you're trying to say. Do you know `bufnet`? Or do you want to create mock of gRPC?

Kourin1996: I don't know what's the benefit of mocking here

syncer/syncer.go
Kourin1996: I don't understand. Please tell us correct info

vendor/modules.txt
Kourin1996: Why you don't like this approach?

syncer/service_test.go
Kourin1996: If we define it as mock, what do we test? We test mock returns predefined value? Is this test?

Kourin1996: The objective of the tests in this file is to check (1) server can get info by correct way (2) gRPC encoding/decoding is done correctly

syncer/client.go
Kourin1996: We can't set constant timeout here because it's impossible to expect how long it takes to finish sending all blocks. Then we need to start new goroutine and can reset timer if new block reached, otherwise you can call cancel context

Kourin1996: What do you think?

zivkovicmilos: That's exactly what I was thinking - have a base timeout for a block to reach you during bulk sync. What this timeout should be, is a different question 🙂 

@Aleksao998

syncer/service_test.go
zivkovicmilos: Yes, depends on what you're mocking:

* If you're mocking a server, then your client code is "real"
* If you're mocking a client, then your server code is "real"

This approach you took doesn't take up time (it's only a few seconds), but I just wanted to note we could've done this without `bufnet` and real gRPC connections

syncer/syncer.go
zivkovicmilos: `bulkSyncWithPeer` returns 3 different types, a `uint64`, a `bool` and an `error`

syncer/client.go
Kourin1996: Sorry, I forgot to rename. Fixed at 85c8b684c6bbbff7f3e604b870668149a2a1bd0a

syncer/client.go
Kourin1996: Fixed at 1d536ebdd4fb293d0943f325289ca950e84da5c2

syncer/client.go
Kourin1996: It's little bit difficult. What does `PeerEvent` means because `Peer` here doesn't mean the one in network layer

syncer/client.go
Kourin1996: This is a `SyncPeerClient`. So `GetBlocks` means to get blocks from peer 😂

syncer/peers.go
Kourin1996: I know. But I don't like adding `//nolint:forcetypeassert` 😂 That's what I don't like in Golang

syncer/syncer.go
Kourin1996: Thanks. But do I need to tell this to Google and AWS team...?

Screenshot 2022-07-04 at 16 27 35

Kourin1996: @zivkovicmilos

syncer/syncer.go
Kourin1996: I don't think it's wrong. But we need to consider defining struct when returning 4+ values

Kourin1996: cc: @Aleksao998

syncer/client.go
Kourin1996: Fixed for you at d75dc06974a15bbf55ba2bc69f8173b95e531db2

syncer/peers.go
Kourin1996: Fixed at d75dc06974a15bbf55ba2bc69f8173b95e531db2

syncer/syncer.go
Kourin1996: Fixed at d75dc06974a15bbf55ba2bc69f8173b95e531db2

syncer/syncer.go
Kourin1996: Fixed at d75dc06974a15bbf55ba2bc69f8173b95e531db2

syncer/syncer.go
Kourin1996: Fixed at d75dc06974a15bbf55ba2bc69f8173b95e531db2

syncer/syncer.go
0xAleksaOpacic: One L or two? While both canceled and cancelled are acceptable for the past tense of cancel, the version with one L is more common in American English, while the version with two L's is more common in British English.

@Kourin1996 Milos was always American in his heart 😂

syncer/client.go
Kourin1996: Added 02c0803ee5da57408a89be70c80a1871f62adc70

syncer/types.go
Kourin1996: Added 02c0803ee5da57408a89be70c80a1871f62adc70

syncer/types.go
Kourin1996: Added 02c0803ee5da57408a89be70c80a1871f62adc70

syncer/client.go
zivkovicmilos: Found it weird, is all :)

syncer/client.go
zivkovicmilos: Yes, but this is not the server implementation of the gRPC method, but a clientside method that exists independently.
Anybody who's searching for the implementation of GetBlocks (the gRPC method) will stumble upon this one

syncer/peers.go
zivkovicmilos: To trick the linter you can do:

`obj, _ := myCast.(castType)`

The `_` will fool it

syncer/syncer.go
zivkovicmilos: Ah, so with one `l` it's the American spelling 🙂, ignore

syncer/syncer.go
zivkovicmilos: Where do you call `Syncer.Close()`?

syncer/client.go
Kourin1996: Yes, but this struct is responsible for only interacting with peer...

syncer/client.go
Kourin1996: https://github.com/0xPolygon/polygon-edge/pull/591/commits/cb1ec52b9e95ad1eed924b8535b979cb73ff4bb8
I added

syncer/peers.go
Kourin1996: Fixed de1f1cd96b55bee48e6a1cd812f97f0e9e3108ae

syncer/service_test.go
Kourin1996: Then what is our conclusion. Do you just want me to rename?

syncer/syncer.go
Kourin1996: Added b51dc66694731673b5516b5e78c4cf8037e56cb1

syncer/syncer.go
Kourin1996: from ibft, you can check in latest code

vendor/modules.txt
0xAleksaOpacic: ping @zivkovicmilos

syncer/syncer.go
0xAleksaOpacic: I also prefer having max 2 attributes as the return value, but here was either this or duplicating the same method. Another solution was to define struct as Yoshiki said but do not think it is necessary.
@zivkovicmilos Can you please suggest how you think we should refactor this

syncer/client.go
0xAleksaOpacic: I like this discussion hahaha both of you have a point. Personally, I am ok leaving it as it is. But I really like the way of thinking about how people will search for this.

syncer/peers.go
0xAleksaOpacic: ping @zivkovicmilos

syncer/service_test.go
0xAleksaOpacic: ping @zivkovicmilos

syncer/syncer.go
0xAleksaOpacic: @zivkovicmilos It will be called if when we are closing ibft consensus https://github.com/0xPolygon/polygon-edge/blob/feature/sync-v2/consensus/ibft/ibft.go#L1375

syncer/client.go
0xAleksaOpacic: I see that this was fixed in this commit b51dc66694731673b5516b5e78c4cf8037e56cb1

zivkovicmilos: This is just a method description commit, I don't see where this was changed?

0xAleksaOpacic: My mistake wrong commit hash 02c0803ee5da57408a89be70c80a1871f62adc70

syncer/client.go
zivkovicmilos: @Kourin1996 

You defined this method **as a part of** `SyncPeerClient`, where you pass in the **peer you want to get blocks from**, as part of the arguments. It makes sense to be called `GetBlocks` on the server side of this interaction (which it is), not so much on the SyncPeer**Client** side

syncer/client.go
0xAleksaOpacic: https://github.com/0xPolygon/polygon-edge/commit/02c0803ee5da57408a89be70c80a1871f62adc70#diff-8a2b17ad6dffb987fc1f8ede0d788f1cd430169b5840d37d2810154172f9680dL182

syncer/peers.go
zivkovicmilos: This commit hash does not point to the fix

syncer/service_test.go
zivkovicmilos: @Kourin1996 

No need, just wanted to have this discussion 🙂

syncer/syncer.go
zivkovicmilos: I think the cleanest way, if we can't break up the method itself, is to introduce a return struct.

@dbrajovic what do you think?

syncer/proto/syncer.proto
zivkovicmilos: ping @Kourin1996

syncer/peers.go
Kourin1996: https://github.com/0xPolygon/polygon-edge/pull/591/commits/03dc08ad39a048560211355b3335b53e63589de6

syncer/client.go
Kourin1996: This is different things. What is the special reason to use different name on server side and client side?

Kourin1996: As you said, this method take peerID to specify the peer client fetch data from

syncer/proto/syncer.proto
Kourin1996: @zivkovicmilos 3a189f27dc13901076086fd4eb4b49bb1ef153b6

syncer/client.go
Kourin1996: @Aleksao998 cc

syncer/syncer.go
Kourin1996: To be honest, I don't know why to return struct instead of tuple here. It gives more complexity. There are some methods that return 3 values in other place... it's fuzzy

Kourin1996: The problem is that we didn't have threshold. But in my opinion, returning 3 values is acceptable.

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+	SyncerProto      = "/syncer/0.2"
+)
+
+var (
+	errTimeout = errors.New("timeout awaiting block from peer")
+)
+
+// XXX: Don't use this syncer for the consensus that may cause fork.
+// This syncer doesn't assume fork. Consensus may be broken.
+// TODO: Add extensibility for fork before merge
```
dbrajovic, line 26: I take it we won't be having consensuses that are not instant finality in a long time. Can this TODO be removed ? (and everything related to forks)

syncer/peers.go
```diff
@@ -0,0 +1,73 @@
...
+	return peerMap
+}
+
+func (m *PeerMap) PutPeers(peers []*NoForkPeer) {
+	for _, p := range peers {
+		m.Put(p)
+	}
+}
+
+func (m *PeerMap) Put(peer *NoForkPeer) {
```
dbrajovic: You can make this a variadic argument, like: `Put(peers ...*NoForkPeer)`, use the for loop like in `PutPeers` and have a single function for putting peers into a peer map.

syncer/types.go
```diff
@@ -0,0 +1,104 @@
...
+	StartProgression(startingBlock uint64, subscription blockchain.Subscription)
+	// UpdateHighestProgression updates highest block number
+	UpdateHighestProgression(highestBlock uint64)
+	// GetProgression returns Progression
+	GetProgression() *progress.Progression
+	// StopProgression finishes progression
+	StopProgression()
+}
+
+type SyncPeerService interface {
```
dbrajovic, line 81: Where's the other half of this interface (Close method) ?
Is it not needed at all ?

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+
+// GetSyncProgression returns progression
+func (s *syncer) GetSyncProgression() *progress.Progression {
+	return s.syncProgression.GetProgression()
+}
+
+// HasSyncPeer returns whether syncer has the peer to syncs blocks
+// return false if syncer has no peer whose latest block height doesn't exceed local height
+func (s *syncer) HasSyncPeer() bool {
+	if s.peerMap == nil {
```
dbrajovic: This check seems redundant - you initialize the map upon constructing the syncer. Also, the field itself does not have to be a pointer (`*PeerMap`).

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+		return false
+	}
+
+	bestPeer := s.peerMap.BestPeer(nil)
+	header := s.blockchain.Header()
+
+	return bestPeer != nil && bestPeer.Number > header.Number
+}
+
+func (s *syncer) BulkSync(ctx context.Context, newBlockCallback func(*types.Block) bool) error {
```
dbrajovic: Argument `ctx` is unused (same as in WatchSync)

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		// fetch block from the peer
+		lastNumber, shouldTerminate, err := s.bulkSyncWithPeer(bestPeer.ID, callback)
+		if err != nil {
+			s.logger.Warn("Error ", err)
+			s.logger.Warn("failed to complete bulk sync with peer, try to next one", "peer ID", bestPeer.ID)
+		}
+
+		if err != nil || lastNumber < bestPeer.Number {
```
dbrajovic: Considering there is an || (or) in this if statement, maybe the `err != nil` is redundant? It looks like the entire point is to ignore (skip) the peer if it returned a lesser height, regardless of the error from `bulkSyncWithPeer`.

syncer/client.go
0xAleksaOpacic: @Kourin1996 Both of you are right, but I am leaning toward @zivkovicmilos naming, will change it :)

syncer/syncer.go
0xAleksaOpacic: @zivkovicmilos I don't see the way to break this method and not to duplicate code, also i agree with @Kourin1996 that in this case 3 values are acceptable. If you are strongly not against it i would leave it as it is :)

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+// Close terminates goroutine processes
+func (s *syncer) Close() {
+	close(s.newStatusCh)
+}
+
+// initializePeerMap fetches peer statuses and initializes map
+func (s *syncer) initializePeerMap() {
```
dbrajovic, line 90: I think you can nicely break the responsibility of this method to have more code clarity. The main goal of this method is to be spawned as a background thread (looping on for), in which case the map initialization can be moved to after the `s.syncPeerService.Start()` statement in the syncer's `Start()` method. Instead of `initializePeerMap` this method can be called `listenForPeerEvents` (or something else).

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+	peerStatuses := s.syncPeerClient.GetConnectedPeerStatuses()
+	s.peerMap.PutPeers(peerStatuses)
+
+	for peerStatus := range s.syncPeerClient.GetPeerStatusUpdateCh() {
+		s.peerMap.Put(peerStatus)
+
+		select {
```
dbrajovic: This notifying select case is used in other places is well. You could make it a separate method of syncer. (called `notify` for example)

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+		case s.newStatusCh <- struct{}{}:
+		default:
+		}
+	}
+}
+
+// startPeerConnectionEventProcess processes peer connection change events
+func (s *syncer) startPeerConnectionEventProcess() {
```
dbrajovic, line 103: Naming suggestion: `handlePeerConnections`

syncer/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+	for e := range s.syncPeerClient.GetPeerConnectionUpdateEventCh() {
+		peerID := e.PeerID
+
+		switch e.Type {
+		case event.PeerConnected:
+			go func() {
```
dbrajovic: Encapsulating this func into a separate method of syncer would improve readability.

syncer/syncer.go
Kourin1996: Actually this comment is not for us, but someone who wants to extend Edge and use it

Kourin1996: @dbrajovic

Kourin1996: I think module must not depend on the implementation of the module. Unfortunately syncer v2 completely doesn't work with the consensus that doesn't guarantee immediate finality. This is breaking modularity completely. Initially syncer v2 had such an extensibility but stopped it because it made the code complex.

Kourin1996: For that reason, I've left such comments

Kourin1996: Comments will be for people with various aspects, future engineer at Polygon Edge, user, and third-party who wants extend Polygon Edge.

syncer/peers.go
Kourin1996: Good suggestion. Fixed at 812f2d68d6d0cdc6297b55bbc5f1328e5b05f3ee

syncer/types.go
Kourin1996: I don't think module has to have Close method always. but fixed to close gRPC stream at 27eb6ab4dc3c3477bae54459f2b64e6240bc6a91

syncer/syncer.go
Kourin1996: Good catch, fixed at 693b4791eb446e76c4732b468375c4059a20b816

syncer/syncer.go
Kourin1996: your saying sounds correct. 
What do you think? @Aleksao998

syncer/syncer.go
Kourin1996: Good catch. Removed 9ce8ea8c3a1e4ea6da6c889896fb3af1afc1e587

syncer/syncer.go
Kourin1996: Good suggestion
Fixed
8a0176d4d4165062629ea1cd2857ec1fec765a7e
4683c852ede4be183ef6201a009cd7766c3636cc

syncer/syncer.go
Kourin1996: Fixed 8a0176d4d4165062629ea1cd2857ec1fec765a7e

syncer/syncer.go
Kourin1996: I think this name sounds a handler. This method starts new process

syncer/syncer.go
Kourin1996: Fixed a8a80091220950df443c73e27f902051f5fd4c93

syncer/syncer.go
0xAleksaOpacic: Agree with @Kourin1996, definition of TODO tag is

_In general, a TODO is written to tell future maintainers about something important: something that should be added, or should be changed, or should not be forgotten._

@dbrajovic  But also we need to rephrase this because we will merge the syncer module without adding support for the fork.
(maybe remove **before merge** from the sentence)

syncer/syncer.go
0xAleksaOpacic: @dbrajovic, Make sense, you are right thanks

0xAleksaOpacic: fixed 2140ffc9bac11ba227bff5c0c932539ed4fb8549

dbrajovic (approved):

0xAleksaOpacic (approved):

syncer/client.go
0xAleksaOpacic: @Kourin1996 @zivkovicmilos Please don't resolve comments just by looking at the code. This fix introduced an issue where every few logs we are going to get and error log
```
2022-07-12T11:32:50.033+0200 [INFO]  polygon.blockchain: new block: number=247 hash=0x489a639febe493ce9ee25baf9ce96cc7cf9b0eb1703234381499374f30ad53f0 txns=0 generation_time_in_seconds=2
2022-07-12T11:32:52.019+0200 [ERROR] polygon.consensus.sync-peer-client: failed to get block from gRPC stream: peer=16Uiu2HAkvNbq28Ep41byugeMDQZCWtK8YkKCubaJgGTwbdYVYanH err=EOF
2022-07-12T11:32:52.021+0200 [INFO]  polygon.blockchain: write block: num=248 parent=0x489a639febe493ce9ee25baf9ce96cc7cf9b0eb1703234381499374f30ad53f0

```

0xAleksaOpacic: The issue was when we are doing watchSync, we are opening the stream and requesting all blocks that we do not have. Once we get all blocks we would get EOF response. All responses we are catching as err and we were logging this as error event though it is expected behavior
Fixed: 26daf07b2011c1cd3e2748531d2cae02238a0b7c
Fix linter: d65d1c8153d392ebf662664c079a94a8444f8e96

PR #598, opened 2022-06-20
Feature/sync v2 watch sync implementation

0xAleksaOpacic:
# Description

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


consensus/ibft/ibft.go
```diff
@@ -503,10 +503,11 @@ func (i *Ibft) runSyncState() {
 
 			continue
 		}
-
-		if err := i.syncer.BulkSync(context.Background(), func(newBlock *types.Block) {
+		if err := i.syncer.BulkSync(context.Background(), func(newBlock *types.Block) bool {
 			callInsertBlockHook(newBlock.Number())
 			i.txpool.ResetWithHeaders(newBlock.Header)
+
+			return false
```
Kourin1996, line 510: Callback shouldn't be given in BulkSync because it always returns false

syncer/syncer.go
```diff
@@ -51,6 +54,7 @@ func NewSyncer(
 		syncPeerService: NewSyncPeerService(network, blockchain),
 		syncPeerClient:  NewSyncPeerClient(logger, network, blockchain),
 		blockTimeout:    blockTimeout,
+		newStatus:       make(chan struct{}),
```
Kourin1996: Please add `Ch` in suffix of the name

syncer/syncer.go
```diff
@@ -73,7 +77,18 @@ func (s *syncer) initializePeerMap() {
 	s.peerMap.PutPeers(peerStatuses)
 
 	for peerStatus := range s.syncPeerClient.GetPeerStatusUpdateCh() {
+		bestPeer := s.peerMap.BestPeer(nil)
+		peerBlockNum := peerStatus.Number
```
Kourin1996: Please note current peerMap scans all peers when calling BestPeer. I'm not sure whether comparing between current best peer and new status is necessary here

syncer/syncer.go
```diff
@@ -163,31 +177,61 @@ func (s *syncer) BulkSync(ctx context.Context, newBlockCallback func(*types.Bloc
...
+			break
+		}
+
+		// if the bestPeer does not have a new block continue
+		if bestPeer.Number <= localLatest {
+			continue
+		}
 
 		// fetch block from the peer
+		lastNumber, isValidator, err := s.bulkSyncWithPeer(bestPeer.ID, callback)
```
Kourin1996: `isValidator` is not proper name because there is no relationship between syncer and validator.

Kourin1996: I think you can't use bulkSyncWithPeer because peer will terminate stream when reaching latest block.

syncer/syncer.go
```diff
@@ -163,31 +177,61 @@ func (s *syncer) BulkSync(ctx context.Context, newBlockCallback func(*types.Bloc
...
+	localLatest := s.blockchain.Header().Number
+	skipList := make(map[peer.ID]bool)
+
 	// Loop until context is canceled
 	for {
+		select {
+		case <-s.newStatus:
+		case <-time.After(s.blockTimeout):
+			return errTimeout
+		}
```
Kourin1996: Why don't you put this in the end of the loop?

syncer/syncer.go
```diff
@@ -163,31 +177,61 @@ func (s *syncer) BulkSync(ctx context.Context, newBlockCallback func(*types.Bloc
...
 		}
 	}
+
+	return nil
 }
 
-func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.Block)) (uint64, error) {
+func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.Block) bool) (uint64, bool, error) {
 	localLatest := s.blockchain.Header().Number
+	isValidator := false
```
Kourin1996: `isValidator` is not right name in syncer package

Kourin1996 (commented):
Just comments

syncer/syncer.go
```diff
@@ -73,7 +77,18 @@ func (s *syncer) initializePeerMap() {
 	s.peerMap.PutPeers(peerStatuses)
 
 	for peerStatus := range s.syncPeerClient.GetPeerStatusUpdateCh() {
+		bestPeer := s.peerMap.BestPeer(nil)
+		peerBlockNum := peerStatus.Number
+
 		s.peerMap.Put(peerStatus)
+
+		// send a signal to watchSync if a new block arrives
+		if peerBlockNum > bestPeer.Number {
```
Kourin1996: I think bestPeer might be nil sometimes. (For instance, case of that node doesn't connect to any peer at first)

syncer/syncer.go
0xAleksaOpacic: Agree, fixed 3866ac73c466ae445be4ade0cfedb447a3d52066

syncer/syncer.go
0xAleksaOpacic: Per spec, we should find the best peer on every new block and open a new connection to fetch all blocks. So i think this is fine. Do you agree?

syncer/syncer.go
0xAleksaOpacic: This event is triggered when any peer sends a new event. If a peer has sent a new event that means that we have best peer inserted.
I added a guard just to be sure 380d48e091ae9f9daf8545f1e96847cf941e7937

syncer/syncer.go
0xAleksaOpacic: My idea was to call watchSync only when we are sure that we have a new block to sync. That's why I've fetched bestPeer before putting a new peerStatus into the map.
Here 2 scenarios that can happen:
1. Best peer has a bigger blockNum + distance than a new status (we do not want to sync, because we know that we are synced with our bestPeer from the previous round)
2. peerStatus is better than bestPeer which means that peerStatus is a new best and we want to sync so we re sending the signal.

Yes we could trigger watchSync on every event and check inside watchSync do we need to sync with bestPeer, but then when a new block arrives, we would sync with first which sends us the event, and will go in watchSync for all other connected validator nodes, just to leave after the condition. I am fine with this also if you think is better code structure

syncer/syncer.go
0xAleksaOpacic: Fixed, what do you think about this naming 3d198079c12bfe45e2b7f54e4aae7b9fddbea315

syncer/syncer.go
0xAleksaOpacic: Fixed, what do you think about this naming 3d198079c12bfe45e2b7f54e4aae7b9fddbea315

syncer/syncer.go
0xAleksaOpacic: Fixed 217daa8125c7873e78c3e48c79105faa0cabd806

syncer/syncer.go
0xAleksaOpacic: I've fixed this yesterday, but now when i am thinking about it again, it seems it's better to revert it to the beginning of the loop (because of this line https://github.com/0xPolygon/polygon-edge/pull/598/files#diff-e725a03bd475dc4ea28028613ede1fd88a8cc8f881f0a1788a189e766f191830R198) 
If we do not get a new block we would just be in the loop where we fetch bestPeer check block and continue.

syncer/syncer.go
```diff
@@ -74,6 +78,11 @@ func (s *syncer) initializePeerMap() {
 
 	for peerStatus := range s.syncPeerClient.GetPeerStatusUpdateCh() {
 		s.peerMap.Put(peerStatus)
+
+		select {
+		case s.newStatusCh <- struct{}{}:
+		default:
+		}
```
Kourin1996, line 85: I think you can emit outside loop

Kourin1996 (changes requested):

syncer/syncer.go
0xAleksaOpacic: I think we can not put it outsthe ide loop. Loop will be trigger on every new status, so once we put status in peerMap we need to emit an event.

PR #617, opened 2022-06-30
Support custom contract address for PoS validators and block rewards for PoA + PoS

dankostiuk:
# Description

This PR introduces a new IBFT switch flag `--custom-contract-address`. The idea is that when present, the PoS `InsertBlockHook` will update the snapshot's validator set calling `validators()` from the custom contract address (as opposed to the predeployed Staking.sol at 0x1001).

Additionally, for both PoA and PoS, `InsertBlockHook` now also supports updating the snapshot with the current custom block rewards amount in wei by calling `getBlockReward()` on the custom contract address. The idea here is that an admin can dynamically change the blockReward amount (setBlockReward(uint256)) on their SC which will set the amount that will be rewarded for every block proposed for the duration of the current epoch and subsequent epochs until they set this value to another amount (or 0 to essentially disable block rewards).

The `PreStateCommitHook` for both PoA and PoS will then pay out the proposer the current block reward amount defined in the snapshot.

I've also updated the snapshot output so that validator owners can query the current block reward amount in wei:
```
/main ibft snapshot --grpc-address localhost:10000
```

```
[IBFT SNAPSHOT]
Block = 88500
Hash  = 0xe0c9e8fcea5c96e38f73637d0d739cea15f5f29a1085b6f243081d0c77aee5a6

[VOTES]
No votes found

[VALIDATORS]
ADDRESS
0x30b3eCF398eEB21D038Ba6a9221BD40255B4fE04
0xF016C0F8854bcE260261b41150173FB071c9E3F6
0x74615232119ab0eBA508BdBc910FAF7bdD8196d6
0xa24ef1E3C22D2Ab3F20D75f809Fc2D3cf47eF42c

[BLOCK REWARD]
10000000000000000 wei
```

---

I've created this PR as a draft as I still need to add more unit testing, however in the meantime I'd appreciate any feedback or suggestions as I think this change will be beneficial to any edge developers looking to support validator rewards bonuses while running new, low-traffic chains. Until a chain has a lot of traffic, rewards solely based on gas will probably not incentivize users enough to run their own validators for new chains.

# Changes include

# Checklist

## Testing

### Manual tests

- [x] PoA -> PoA w/ custom contract
- [x] PoA -> PoS w/ custom contract
- [x] PoS -> PoS w/ custom contract

With epoch set to 100 blocks, we can observe a validator's balance change over time:

- set block rewards to 0.1 SX per block on SC prior to 1st epoch (block 6062300 at 14:30)
- set block rewards to 0 SX per block on SC prior to 3rd epoch (6062500, ~14:36:30)
- set block rewards to 0.2 SX per block on SC block prior to 4th epoch (6062600 , ~14.39:80)

Screen Shot 2022-07-07 at 2 41 29 PM


mrwillis:
@Kourin1996 @zivkovicmilos 

Curious if we could get your thoughts on this and if it's sound. We're currently testing some transitions between validator selection transitions (PoA -> PoS, etc) and making sure we don't have any consensus failures

PR #624, opened 2022-07-06
Added new server config flag `--bootnode-only-sync` to ignore non-bootnodes in BestPeer()

dankostiuk:
# Description

When testing potential breaking changes on our testnets, we occasionally must stop all nodes and revert to a recent backup in order to revert the blockchain state to what it was prior to the breaking change. This process is fairly straightforward while dealing with a chain entirely operated by us. 

The issue lies when other peers or validators are connected to our mainnet chain which we do not have control over. As a precaution, if a new version of edge containing a breaking change was introduced on mainnet, it would be difficult to perform any kind of revert as our bootnodes would always sync to a peer that could have a higher blockNumber than the backup we want to revert to.

This PR introduces a new flag `--bootnode-only-sync` that we could require all validators to add after clearing their data-dirs to facilitate the unlikely event of having to reset to an earlier state. Additionally we would also add this flag to our bootnodes.

# Changes include

# Checklist

## Testing

### Manual tests

1. Spin up 4-node testnet with all 4 nodes as bootnodes
2. Add a 5th node to the testnet and give enough time to sync up
3. Clear the data (`rm -rf data/blockchain/ data/trie/ data/consensus/snapshots data/consensus/metadata`) of node 1.
4. Restart node 1 with `./main server --bootnode-only-sync` and observe that the non-bootnode peer is ignored when determining best peer within BestPeer()

# Documentation update

Will create a documentation PR once this PR is accepted.


mrwillis:
The PR just blocks syncing from any peer but the bootnodes.

Kourin1996:
@dankostiuk Hello, Thank you for a PR! We've refactored syncer recently. Could you fetch and merge latest `develop` if you've not done? Sorry for the inconvenience.

network/server.go
```diff
@@ -523,6 +523,17 @@ func (s *Server) updateBootnodeConnCount(peerID peer.ID, delta int64) {
...
+// ShouldIgnoreSyncToPeer Check if client should ignore this peer when determining best peer to sync to
+func (s *Server) ShouldIgnoreSyncToPeer(peerID peer.ID) bool {
+	// If the peer is not a bootnode and the client has BootnodeOnlySync enabled
+	ignoreSync := s.config.BootnodeOnlySync && !s.bootnodes.isBootnode(peerID)
+	if ignoreSync {
+		s.logger.Debug("avoid syncing to non-bootnode peer", "id", peerID.String())
+	}
+
+	return ignoreSync
+}
```
Kourin1996, line 535: Right now only syncer syncs some data between peers. I think network layer shouldn't have such a function. Network layer just provides a method to return a peer is bootnode or not.

syncer/syncer.go
```diff
@@ -209,10 +211,24 @@ func (s *syncer) BulkSync(newBlockCallback func(*types.Block) bool) error {
 	return nil
 }
 
+// makeSkipList initializes list of peers to skip
+func (s *syncer) makeSkipList() map[peer.ID]bool {
+	skipList := make(map[peer.ID]bool)
+
+	s.peerMap.Range(func(key, value interface{}) bool {
+		peer, _ := value.(*NoForkPeer)
+		skipList[peer.ID] = s.network.ShouldIgnoreSyncToPeer(peer.ID)
```
Kourin1996, line 220: Is it possible to rewrite this code by calling a method to given peer is bootnode or not?

Kourin1996 (changes requested):

lazartravica (changes requested):
Hello Dan,

I'm very sorry but I would need to close this Pull Request, as this would introduce a new class of "privileged" nodes (in this case bootnodes) that can be allowed to send blocks.

I understand the issue with coordinating breaking changes and upgrades, but this needs to be solved on the organizational side, not on the side of changing the syncing logic.

PR #638, opened 2022-07-12
Add blockRangeLimit and batchRequestLimit JSON-RPC flags to help prevent node DDoS

dankostiuk:
# Description

This PR introduces 2 new server flags that help prevent node DDoS:
`--json-rpc-block-range-limit` and `--json-rpc-batch-request-limit`

### Block Range Limit

During testing, we observed that a couple JSON-RPC requests that include block ranges could allow an attacker to use up all the resource on a particular node if they specify large ranges, essentially DDoS-ing that node. For example, the following request could take down any polygon-edge node running on a chain that consists of many blocks:
```
{"jsonrpc":"2.0","id":1,"method":"eth_getLogs","params":[{"fromBlock":"0","toBlock":"latest"}]}
```

To prevent such an attack, this PR enforces a default block range limit of 1000, optionally giving the node operator the ability to override this by specifying a value for `--json-rpc-block-range-limit` on their node's server command.

### Batch Request Limit

In addition to the above, since we currently don't have a limit for the allowed size of a batch JSON-RPC request body, attackers could spam RPC nodes with requests that contain thousands of batched requests, for example:
```
[
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 0}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 1}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 2}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 3}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 4}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 5}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 6}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 7}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 8}, 
	{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 9},
        ...
        {"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params": ["latest", true], "id": 99999},
]
```

Part of this PR also enforces a default batch request size of 20, optionally giving the node operator the ability to override this by specifying a value for `--json-rpc-batch-request-limit` on their node's server command.

# Changes include

# Checklist

## Testing

### Manual tests

1. Run this branch on a node
2. Attempt to query with `{"jsonrpc":"2.0","id":1,"method":"eth_getLogs","params":[{"fromBlock":"0","toBlock":"1002"}]}` (this exceeds the default value of 1000). Observe error message in response.
3. Attempt to query with batch JSON-RPC request containing 25 entries (this exceeds the default of 20). Observe error message in response.
4. Add `--json-rpc-block-range-limit 1100` and `--json-rpc-batch-request-limit 30` to the server command on the same node and restart the node.
5. Attempt both queries from step 2. and 3. and observe that the node now responds with success responses.

# Documentation update

Will create a separate documentation PR once this PR is approved.


command/server/server.go
```diff
@@ -204,6 +204,21 @@ func setFlags(cmd *cobra.Command) {
 		"the CORS header indicating whether any JSON-RPC response can be shared with the specified origin",
 	)
 
+	cmd.Flags().Uint64Var(
+		&params.batchLengthLimit,
+		batchRequestLimitFlag,
+		20,
```
Kourin1996: It'd be great if you could define this value in constant 🙏

command/server/server.go
```diff
@@ -204,6 +204,21 @@ func setFlags(cmd *cobra.Command) {
...
+		"the max length to be considered when handling json-rpc batch requests",
+	)
+
+	//nolint:lll
+	cmd.Flags().Uint64Var(
+		&params.blockRangeLimit,
+		blockRangeLimitFlag,
+		1000,
```
Kourin1996: This is also

jsonrpc/dispatcher.go
```diff
@@ -39,23 +39,26 @@ type endpoints struct {
...
+	filterManager    *FilterManager
+	endpoints        endpoints
+	chainID          uint64
+	priceLimit       uint64
+	batchLengthLimit uint64
 }
 
-func newDispatcher(logger hclog.Logger, store JSONRPCStore, chainID uint64, priceLimit uint64) *Dispatcher {
+func newDispatcher(logger hclog.Logger, store JSONRPCStore, chainID uint64, priceLimit uint64,
+	batchLengthLimit uint64, blockRangeLimit uint64) *Dispatcher {
```
Kourin1996: Could you define an argument per line if you need to define them in 2+ lines like this:

```suggestion
func newDispatcher(
        logger hclog.Logger,
        store JSONRPCStore,
        chainID uint64,
        priceLimit uint64,
	batchLengthLimit uint64,
	blockRangeLimit uint64,
) *Dispatcher {
```

jsonrpc/dispatcher_test.go
```diff
@@ -266,25 +266,86 @@ func TestDispatcherFuncDecode(t *testing.T) {
...
+		return res
+	}
 
-	jsonerr := &ObjectError{Code: -32602, Message: "Invalid Params"}
+	cases := []struct {
+		id          uint8
+		desc        string
+		dispatcher  *Dispatcher
+		reqBody     []byte
+		expectError bool
```
Kourin1996: how about giving an error to be expected to return instead of flag?

jsonrpc/dispatcher_test.go
```diff
@@ -266,25 +266,86 @@ func TestDispatcherFuncDecode(t *testing.T) {
...
+	for _, c := range cases {
+		res := handle(c.dispatcher, c.reqBody)
+
+		if c.expectError {
+			var resp ErrorResponse
+
+			assert.NoError(t, expectBatchJSONResult(res, &resp))
+
+			switch c.id {
+			case 3:
```
Kourin1996: I think using switch-case by test id is not best practice. All information that needs test can be given in `cases` data

Kourin1996 (changes requested):
Hello, Thank you for PR! The code looks good and it works completely. I've left some minor issues. It'd be cool if you could fix them

jsonrpc/dispatcher_test.go
```diff
@@ -266,25 +266,86 @@ func TestDispatcherFuncDecode(t *testing.T) {
...
+			var resp []SuccessResponse
+			assert.NoError(t, expectBatchJSONResult(res, &resp))
+
+			if c.name == "leading-whitespace" {
+				assert.Len(t, resp, 4)
+				jsonerr := &ObjectError{Code: -32602, Message: "Invalid Params"}
+				assert.Equal(t, resp[0].Error, jsonerr)
+				assert.Nil(t, resp[1].Error)
+				assert.Nil(t, resp[2].Error)
+				assert.Nil(t, resp[3].Error)
```
Kourin1996: Why not define these errors in cases?

jsonrpc/dispatcher_test.go
```diff
@@ -266,25 +266,86 @@ func TestDispatcherFuncDecode(t *testing.T) {
...
+			} else if c.name == "valid-batch-req" {
+				assert.Len(t, resp, 6)
+				assert.Nil(t, resp[0].Error)
+				assert.Nil(t, resp[1].Error)
+				assert.Nil(t, resp[2].Error)
+				assert.Nil(t, resp[3].Error)
+				assert.Nil(t, resp[4].Error)
+				assert.Nil(t, resp[5].Error)
```
Kourin1996: this is also

Kourin1996 (changes requested):

jsonrpc/dispatcher_test.go
Kourin1996: Let me know if it's impossible

Kourin1996 (approved):

command/server/server.go
```diff
@@ -204,6 +204,21 @@ func setFlags(cmd *cobra.Command) {
...
+		jsonRPCBatchRequestLimitFlag,
+		defaultConfig.JSONRPCBatchRequestLimit,
+		"the max length to be considered when handling json-rpc batch requests",
+	)
+
+	//nolint:lll
+	cmd.Flags().Uint64Var(
+		&params.jsonRPCBlockRangeLimit,
+		jsonRPCBlockRangeLimitFlag,
+		defaultConfig.JSONRPCBatchRequestLimit,
```
zivkovicmilos: Did you mean `defaultConfig.JSONRPCBlockRangeLimit` here?

zivkovicmilos (approved):
Looks great 💯 

Thank you for the contribution, as always 🙏 

Please make sure to pull in the latest `develop` changes, and to check out the comment I've left.
We're good to go with this PR as soon as that's resolved 🚀

PR #642, opened 2022-07-14
Add custom marshalling and unmarshalling for block Header

0xAleksaOpacic:
# Description

Polygon Edge Header cannot be unmarshalled with go-etherem ethClient. This PR implements custom json marshal and unrmashal for Header so it is compatible with other EVM networks

# Changes include

# Breaking changes

I don't think this is a breaking change
I've manually tested:
started a network on the latest develop branch, switching to this branch, adding a new node to bulkSync, waiting in watchSync, voting to become validator, watching consensus 

# Checklist

## Testing

### Manual tests

I've created go-ethereum ethClient script to test this feature. If you run this script on develop branch you will see bunch of errors
```

package main

import (
	"context"
	"fmt"
	"log"

	"github.com/ethereum/go-ethereum/core/types"
	"github.com/ethereum/go-ethereum/ethclient"
)

func main() {
	client, err := ethclient.Dial("ws://localhost:10002/ws")
	if err != nil {
		log.Fatal(err)
	}

	headers := make(chan *types.Header)
	sub, err := client.SubscribeNewHead(context.Background(), headers)
	if err != nil {
		log.Fatal(err)
	}

	for {
		select {
		case err := <-sub.Err():
			log.Fatal(err)
		case header := <-headers:
			fmt.Println(header)

		}
	}
}
```

# Additional comments

Fixes EDGE-440


helper/hex/hex.go
```diff
@@ -52,6 +52,17 @@ func EncodeUint64(i uint64) string {
 	return string(strconv.AppendUint(enc, i, 16))
 }
 
+// DecodeUint64 decodes a hex string with 0x prefix to uint64
+func DecodeUint64(hexStr string) uint64 {
+	// remove 0x suffix if found in the input string
+	cleaned := strings.Replace(hexStr, "0x", "", -1)
```
Kourin1996: `TrimPrefix` is better?

helper/hex/hex.go
```diff
@@ -52,6 +52,17 @@ func EncodeUint64(i uint64) string {
...
+
+	// base 16 for hexadecimal
+	result, _ := strconv.ParseUint(cleaned, 16, 64)
```
Kourin1996: Why not check error?

0xAleksaOpacic: My mistake we should check the error 
fixed 406fe2c54dacc8d6d08496f8a2829cc85b2e95e1 + ff727c450cd58043583531e0ac4fa9bf96d81d72
I've also added a test to confirm that encode, decode, and MustDecode works, just wasn't sure how to reproduce error. If you have an idea please help :)

helper/hex/hex.go
0xAleksaOpacic: Do you mean TrimPrefix is better than cleaned? 
Maybe slicedHex or formattedHex, trimmedHex? What do you think

Kourin1996: No, I meant strings package had `TrimPrefix` function

helper/hex/hex.go
```diff
@@ -52,6 +52,35 @@ func EncodeUint64(i uint64) string {
...
+	// remove 0x suffix if found in the input string
+	cleaned := strings.Replace(hexStr, "0x", "", -1)
+
+	// base 16 for hexadecimal
+	result, err := strconv.ParseUint(cleaned, 16, 64)
+	if err != nil {
+		panic(fmt.Errorf("could not decode hex: %w", err))
+	}
+
+	return result
```
Kourin1996: Why not call DecodeUint64? I recommend you not to implement the same code twice

types/header_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Fatalf("Unable marshall header,  %v", err)
+	}
+
+	unmarshallHeader := Header{}
+	assert.NoError(t, json.Unmarshal(headerJSON, &unmarshallHeader))
+
+	if !reflect.DeepEqual(unmarshallHeader, header) {
+		t.Fatal("Resulting data and expected values are not equal")
+	}
+}
```
Kourin1996: Can you add test to check unmarshalling from json?

Kourin1996 (changes requested):

helper/hex/hex.go
Aleksao998: fied 36cd3c09d6ee00272a1ce9454d16bb685e06ce8f

0xAleksaOpacic: Thanks agree 204fec563dfe954c6af8e159b5f50495fd4631aa

helper/hex/hex.go
0xAleksaOpacic: thanks agree, fixed 204fec563dfe954c6af8e159b5f50495fd4631aa

scalalang2:
related #643

scalalang2:
Another test code here.

```golang
func TestBlockByHash(t *testing.T) {
	edgeNet := "https://rpc-edgenet.polygon.technology"
	// https://explorer-edgenet.polygon.technology/block/1320009/transactions
	blockHash := "0xd4f333b237f56436f6e3074b22a1e0dada6dff4ee574778fa5dcdcebe118323d"
	client, err := ethclient.DialContext(context.Background(), edgeNet)
	if err != nil {
		t.Errorf("failed to connect to the network")
	}

	block, err := client.BlockByHash(context.Background(), ethcommon.HexToHash(blockHash))
	if err != nil {
		t.Errorf("failed to fetch block")
	}

	if block.Hash().Hex() != blockHash {
		t.Errorf("received block is not eqaul, query: %s, got: %s", blockHash, block.Hash().Hex())
	}
}
```

0xAleksaOpacic:
> related #643
> 
> If this PR is merged, Can I expect it would works well without modifying my codes written with ethClient?

Hi,
I need to investigate what is the cause of this issue. If it is something related with this PR it will be included if not i will open separate PR :)

helper/hex/hex.go
```diff
@@ -52,6 +52,31 @@ func EncodeUint64(i uint64) string {
 	return string(strconv.AppendUint(enc, i, 16))
 }
 
+// DecodeUint64 decodes a hex string with 0x prefix to uint64
+func DecodeUint64(hexStr string) (uint64, error) {
+	// remove 0x suffix if found in the input string
+	cleaned := strings.TrimPrefix(hexStr, "0x")
+
+	// base 16 for hexadecimal
+	result, err := strconv.ParseUint(cleaned, 16, 64)
```
zivkovicmilos: This can be inlined

helper/hex/hex.go
```diff
@@ -52,6 +52,31 @@ func EncodeUint64(i uint64) string {
...
+	if err != nil {
+		return 0, err
+	}
+
+	return result, nil
+}
+
+// MustDecodeUint64 decodes a hex string with 0x prefix to uint64
+// Returns panic if there is an error
+func MustDecodeUint64(hexStr string) uint64 {
```
zivkovicmilos: Please avoid adding methods that contain explicit panics

helper/hex/hex_test.go
```diff
@@ -0,0 +1,42 @@
+package hex
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestEncodingUint64(t *testing.T) {
```
zivkovicmilos: Why not have this test simply generate a random number array (with 0 included), and test it out that way?

helper/hex/hex_test.go
```diff
@@ -0,0 +1,42 @@
...
+	testTable := []struct {
+		name       string
+		value      uint64
+		shouldFail bool
```
zivkovicmilos: Why have this in the test if none of your test cases fail?

helper/hex/hex_test.go
```diff
@@ -0,0 +1,42 @@
...
+			decodedValue, err := DecodeUint64(encodedValue)
+			if err != nil && !testCase.shouldFail {
+				t.Fatalf("Unable to decode uint64, %v", err)
+			}
+
+			assert.Equal(t, testCase.value, decodedValue)
+
+			decodedValue = MustDecodeUint64(encodedValue)
+
+			assert.Equal(t, testCase.value, decodedValue)
```
zivkovicmilos: This is redundant, since `MustDecodeUint64` is redundant

types/header.go
```diff
@@ -77,6 +176,15 @@ func (n Nonce) MarshalText() ([]byte, error) {
 	return []byte(n.String()), nil
 }
 
+func (n *Nonce) UnmarshalText(input []byte) error {
+	input = input[2:]
```
zivkovicmilos: What is this magic number `2`?

zivkovicmilos: Aha, the hex prefix.

Please make this clear

helper/hex/hex.go
0xAleksaOpacic: I've added this method following the standard from this package. As you can see other decode methods (https://github.com/0xPolygon/polygon-edge/blob/develop/helper/hex/hex.go#L38) have MustDecode versions used mainly for tests. Agree we can remove this only because we don't want to risk that someone will use this i code and trigger panic.
fixed: a50aef5ac33eb44c76d87c8367a20a1158c146fe

helper/hex/hex_test.go
0xAleksaOpacic: fixed a50aef5ac33eb44c76d87c8367a20a1158c146fe

types/header.go
```diff
@@ -77,6 +176,15 @@ func (n Nonce) MarshalText() ([]byte, error) {
 	return []byte(n.String()), nil
 }
 
+func (n *Nonce) UnmarshalText(input []byte) error {
```
zivkovicmilos, line 173: Why does this method assume that `Nonce` will be hex encoded?

types/header.go
```diff
@@ -30,6 +32,103 @@ type Header struct {
 	Hash         Hash    `json:"hash"`
 }
 
+// HeaderJson represents a block header used for json calls
+type HeaderJSON struct {
```
zivkovicmilos: Why does this need to be a separate struct for JSON?

zivkovicmilos: Why not define a simple JSON marshal method for the existing struct?

zivkovicmilos: I see that you use it in the marshal method for the old `Header` type.

Why not define it locally, why is it exported?

zivkovicmilos (changes requested):
Left some comments that need resolving 🙏 

Please look at @Kourin1996's comments as well.

Overall it looks good 👍

helper/hex/hex.go
0xAleksaOpacic: fixed 21948606924f88571f8f05e9df02242792ceac8c

types/header.go
0xAleksaOpacic: Fixed 06d747cdb9a1b7d260f36baf70ae71ce32455c9d. Do you think we should use this PR to change in other packages as well

zivkovicmilos: Why is this an exported method?

0xAleksaOpacic: Because every unmarshal method in our code is exported. I don't see a reason why it shouldn't be

0xAleksaOpacic: Also i am positive that marshaling and unmarshaling methods needs to be exported

helper/hex/hex_test.go
0xAleksaOpacic: fixed 8ef3965ab7c423466a31342f16a9715bd53826b6

helper/hex/hex_test.go
0xAleksaOpacic: fixed 8ef3965ab7c423466a31342f16a9715bd53826b6

0xAleksaOpacic: fixed lint 70e04dfc8686d02178eb0868d3bf78e3f7d54f4f

helper/hex/hex_test.go
0xAleksaOpacic: fixed lint 70e04dfc8686d02178eb0868d3bf78e3f7d54f4f

types/header.go
0xAleksaOpacic: because I am using it for both UnmarshalJSON and MarshalJSON I didn't want to have it locally in both methods. What do you think of making it non-exportable

helper/hex/hex_test.go
zivkovicmilos: Hahah I didn't mean literally generate 10 random numbers and paste them here, I mean define a local function that generates however many random numbers you request, and use that array as the input for the test

cc @dbrajovic

types/header.go
zivkovicmilos: I meant why is `RemoveHexPrefixFromByteArray` exported

types/header.go
zivkovicmilos: It should not be exported outside of the scope you're using it in

helper/hex/hex_test.go
0xAleksaOpacic: What am I getting with that test, what random number will show me that something is not working, why would I have the necessary generation if it doesn't test anything?  
This subset of numbers contains min, max, and few in between which covers all test cases. 
I don't like implementing anything with a random generator in the test. I need to be sure that the test covers all cases if not then the test is not correct. 
If you guys want this we can add it. Don't see any value in adding a generator method locally. Maybe if we add a global util function that we can use from all unit tests it will make sense :)
cc @Kourin1996

types/header.go
0xAleksaOpacic: I saw that we have pattern [2:] in another packages also in the future maybe i am going to need this somewhere outside of this package, that's why i left it exported

types/header.go
0xAleksaOpacic: agreed 4816d822d636264251cac988f4c23f57c2841c34

types/header.go
0xAleksaOpacic: Agree i can not be sure, but to be honest not sure how would i write this different, please add suggestion :)

helper/hex/hex_test.go
Kourin1996: I agree with @Aleksao998's opinion. In my mind, all inputs should be defined statically. Should think about random test separately.

types/header_test.go
0xAleksaOpacic: fixed 24f329c560da5a8abf82615c553ae4508bd5925f. please check if i should add more test cases :)

0xAleksaOpacic: 24f329c560da5a8abf82615c553ae4508bd5925f

Kourin1996 (approved):
LGTM. No new comments from my side. Please resolve the discussions raised by @zivkovicmilos

zivkovicmilos:
Hey @scalalang2.

Confirming that this PR doesn't fix your issue - it's probably a mismatch in RLP encoding between Edge and the `go-ethereum` client. We'll look into this in a separate PR 🙏

zivkovicmilos (approved):

helper/hex/hex_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	for _, value := range generateRandomNumbers(100) {
+		encodedValue := EncodeUint64(value)
+
+		decodedValue, err := DecodeUint64(encodedValue)
+		assert.NoError(t, err)
+
+		assert.Equal(t, value, decodedValue)
+	}
```
Kourin1996: Why do you need to test with random numbers...? Min, max, and mid possible numbers are enough
+
Encode and Decode in a test is a different way to test. You should focus on testing new code in new test

Kourin1996 (changes requested):

helper/hex/hex_test.go
zivkovicmilos: We never hardcoded number values before in tests, nor should we - these numbers truly need to be random as we're not testing random number generation but encoding and decoding.

The functionality that was added as part of this PR (header encoding / decoding) is covered in another test

Kourin1996: I don't know why we need to test with random value...

There are 2 things I want to say

1. What is the purpose of the test? Do we need to test whether go can calculate `1+1` correctly? Or course, we don't need this. Likewise, it's not important to test that `strings.TrimPrefix` and `strconv.ParseUint` work correctly. The important thing is to check the DecodeUint64 can parse all of the values in the range we expect and return the correct value. (Otherwise, this function returns error or not) In such a case, do we need to test with random value? In such a case, we should do boundary value test. (In addition, we should give the value with wrong format to check the function returns error) You said "We never hardcoded number values" but the basis of unit test is to give sample inputs and check whether the outputs equal to expected values.

2. Why does new test encode and decode? The new code is to decode only. We should add the test to check the function first. But this test does both encode and decode. We should thing about unit test, e2e test, and fuzzy test separately. The test that checks many things is hard to read and difficult to fix.
 
And I'm not talking about the test for header encoding / decoding. But this test encodes uint64 values and decodes it.

Kourin1996: @zivkovicmilos

zivkovicmilos: Got it, no problem in keeping it minimal, you're right that we double verify stuff here.

Fixed:
[fe561e7](https://github.com/0xPolygon/polygon-edge/pull/642/commits/fe561e7fee2ca2975e121541ea2d65ce5a1ddee3)

Kourin1996 (approved):
LGTM

helper/hex/hex_test.go
Kourin1996: Thank you for fixing!

PR #650, opened 2022-07-22
IBFT 2.0

dbrajovic:
# Description

This PR provides the integration logic with the [go-ibft](https://github.com/0xPolygon/go-ibft) package, and introduces IBFT 2.0 into Polygon Edge.

# Changes include

# Breaking changes

The IBFT 2.0 upgrade introduces liveness fixes, which are by default breaking changes, as it alters the way messages are constructed.

This breaking change is breaking in a sense that an **entire** cluster running on **pre** IBFT 2.0 needs to be restarted to run on **post** IBFT 2.0 (all nodes in the cluster need to be running the latest version). The blockchain history remains intact.

This means that nodes running IBFT 2.0 version of Edge will not work with nodes not running IBFT 2.0 version of Edge (older versions) - all nodes in the cluster need to run IBFT 2.0 of Edge in order to work.

# Checklist

## Testing

### Manual tests

* Loadbot testing
* Manual voting / unvoting
* Time period tests

# Additional comments

Fixes EDGE-614


blockchain/blockchain.go
```diff
@@ -859,15 +862,15 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
...
-	)
+func (b *Blockchain) WriteBlock(block *types.Block, source string) error {
+	b.writeLock.Lock()
+	defer b.writeLock.Unlock()
+
+	if block.Number() <= b.Header().Number {
+		b.logger.Info("block already inserted", "block", block.Number(), "source", source)
+
+		return nil
+	}
```
Kourin1996, line 873: It's totally dangerous code. It's possible to ignore past blocks because of IBFT. You can handle this process on consensus layer and blockchain shouldn't do that.

+ If you want to support only IBFT, why doesn't this method return error? or Why does logger print on info level instead of  warning. In IBFT, this shouldn't reach hear usually.
+ Does it prevent restore command from overwriting blocks?

Or current blockchain didn't support overwriting?

consensus/ibft/extra.go
```diff
@@ -77,7 +77,7 @@ func getIbftExtra(h *types.Header) (*IstanbulExtra, error) {
 // IstanbulExtra defines the structure of the extra field for Istanbul
 type IstanbulExtra struct {
 	Validators    []types.Address
-	Seal          []byte
+	ProposerSeal  []byte
```
Kourin1996, line 80: Why renaming? It comes from document?

blockchain/blockchain.go
```diff
@@ -859,15 +862,15 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
...
-func (b *Blockchain) WriteBlock(block *types.Block) error {
-	// Log the information
-	b.logger.Info(
-		"write block",
-		"num",
-		block.Number(),
-		"parent",
-		block.ParentHash(),
-	)
+func (b *Blockchain) WriteBlock(block *types.Block, source string) error {
```
Kourin1996, line 865: I totally don't know why `source` is necessary after I read all changes in the PR.

Kourin1996 (changes requested):

consensus/dev/dev.go
```diff
@@ -197,7 +197,7 @@ func (d *Dev) writeNewBlock(parent *types.Header) error {
 	}
 
 	// Write the block to the blockchain
-	if err := d.blockchain.WriteBlock(block); err != nil {
+	if err := d.blockchain.WriteBlock(block, ""); err != nil {
```
Kourin1996: why empty?

archive/restore.go
```diff
@@ -78,7 +78,7 @@ func importBlocks(chain blockchainInterface, blockStream *blockStream, progressi
 			return err
 		}
 
-		if err := chain.WriteBlock(nextBlock); err != nil {
+		if err := chain.WriteBlock(nextBlock, "restore"); err != nil {
```
Kourin1996: how about defining constant values for reason?

consensus/ibft/consensus_backend.go
```diff
@@ -0,0 +1,319 @@
...
+)
+
+func (i *backendIBFT) BuildProposal(blockNumber uint64) []byte {
+	var (
+		latestHeader      = i.blockchain.Header()
+		latestBlockNumber = latestHeader.Number
+	)
+
+	if latestBlockNumber+1 != blockNumber {
+		return nil
```
Kourin1996, line 25: I thought "Why not return error?". I think core ibft should handle error from this function even if core ibft doesn't do anything except for logging.

consensus/ibft/consensus_backend.go
```diff
@@ -0,0 +1,319 @@
...
+	if err != nil {
+		i.logger.Error("cannot build block", "num", blockNumber, "err", err)
+
+		return nil
+	}
+
+	return block.MarshalRLP()
+}
+
+func (i *backendIBFT) InsertBlock(proposal []byte, committedSeals [][]byte) {
```
Kourin1996, line 45: This function is also. We need to think about where we call log.

syncer/syncer.go
```diff
@@ -293,7 +296,7 @@ func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.B
 				return lastReceivedNumber, false, fmt.Errorf("unable to verify block, %w", err)
 			}
 
-			if err := s.blockchain.WriteBlock(block); err != nil {
+			if err := s.blockchain.WriteBlock(block, "syncer"); err != nil {
```
Kourin1996: Why not define constant?

Kourin1996 (changes requested):

blockchain/blockchain.go
zivkovicmilos: We've discussed this internally, we need this logic here because the syncer and consensus modules concurrently write blocks now, and we need this safeguard in place.

Restore is not affected

consensus/ibft/extra.go
zivkovicmilos: Yes, @dbrajovic and I decided it's much clearer in code (and for us all) that we have a clear distinction between _proposer seals_ and _commit seals_

blockchain/blockchain.go
zivkovicmilos: We've discussed this internally, resolving this comment as you've left other comments below regarding `source`

syncer/syncer.go
zivkovicmilos: Defined:

[24b99b1](https://github.com/0xPolygon/polygon-edge/pull/650/commits/24b99b1298eb81b13e66aa46bfb57faa6d386d60)

consensus/dev/dev.go
zivkovicmilos: Defined:

[addcf45](https://github.com/0xPolygon/polygon-edge/pull/650/commits/addcf45a5be18f354cd6f0d3198ce9dd80d471be)

archive/restore.go
zivkovicmilos: Defined:

[e089b5c](https://github.com/0xPolygon/polygon-edge/pull/650/commits/e089b5c50d08a56ba2e091e317fc8e4ed964bbfe)

consensus/ibft/consensus_backend.go
zivkovicmilos: @dbrajovic and I had lots of discussions on this topic.

The spec does not define a situation where you can insert a block, and error out, or create a block and error out.
It assumes that these things are handled by the `Backend` implementation itself - either the implementation panics and nothing is proposed, or it sends out a nil proposal - both situations are completely valid.

Why are they valid?

Because if the proposer doesn't propose anything, or the proposal is invalid (`nil`, for example), the timer that's ticking on all nodes will eventually expire and the nodes will go into a round change state, where they'll start the new round and someone else will get a chance to propose a valid block. The timer is what's preserving your liveness, and you can always, _always_, fall back on it.

The burden of the consensus package is to handle consensus, not corrupted DBs or the nodes inability to create a block (the `go-ibft` package has no notion or concept of a block, you can run it for an arbitrary byte array)

consensus/ibft/consensus_backend.go
zivkovicmilos: Please see:

https://github.com/0xPolygon/polygon-edge/pull/650#discussion_r933090429

consensus/ibft/consensus_backend.go
zivkovicmilos: On that note, I've added a log to make it clear something funky happened:

[e9595a8](https://github.com/0xPolygon/polygon-edge/pull/650/commits/e9595a8de32840c705c316cc4bb93331816c7947)

blockchain/blockchain.go
Kourin1996: As discussed, we have decided to do that in the future task

Kourin1996 (approved):
Looks good. It has cleaned up IBFT completely

0xAleksaOpacic (approved):
LGTM, Let's merge this :D

PR #672, opened 2022-08-05
Add support for better client version details

### LINKED ISSUE #657: Inconsistent response from JSON-RPC GET method
### Description

While testing the JSON-RPC endpoints I decided to run

```
curl -X GET -H 'content-type: application/json' http://YOUR_NODE_URL
```

I was expecting a JSON response in a middleware I built but what I got was a plain text (not parseable to a JSON object), I went directly to repo's code and found this > https://github.com/0xPolygon/polygon-edge/blob/develop/jsonrpc/jsonrpc.go#L248-L252

This is not a big Issue but I'd suggest standardizing the response, and document it here > https://docs.polygon.technology/docs/edge/get-started/json-rpc-commands 

### Your environment

* OS and version: Kubernetes GCP/EKS
* version of the Polygon Edge: v0.4.1
* branch that causes this issue: `develop`

### Steps to reproduce

* Create a network with 1 Validator Node.
* From a terminal run 
```
curl -X GET -H 'content-type: application/json' http://YOUR_NODE_URL
```
You will see the response is a String that is not parseable to a JSON Object. 

### Expected behaviour

If the `Content-Type` is set to `application/json` > https://github.com/0xPolygon/polygon-edge/blob/develop/jsonrpc/jsonrpc.go#L237 I would expect a JSON Object.

### Actual behaviour

The response is just a plain text.

### Logs

NA

### Proposed solution

I'd suggest standardizing the response, and document it here > https://docs.polygon.technology/docs/edge/get-started/json-rpc-commands

zivkovicmilos:
# Description

This PR resolves a pending issue that doesn't return proper chain information on the client version.

It removes the base `GET` endpoint to the JSON-RPC layer, as it's not by the standard, nor should it be supported.
Users should use the `web3_clientVersion` endpoint instead, that is outlined in the specification.

# Changes include

# Checklist

## Testing

# Additional comments

Fix EDGE-709


jsonrpc/jsonrpc.go
```diff
@@ -261,12 +271,6 @@ func (j *JSONRPC) handle(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if req.Method == "GET" {
```
ZeljkoBenovic, line 267: Can we leave at least some GET endpoint as they are usually needed for default health-checks?   
Maybe we can have something like `/status` or `/health` that will return some data that will notify that the chain is running well.   

At the very least this health check endpoint should return 200 OK response.

ZeljkoBenovic (approved):
Overall looks good.   
We should just consider if we're totally removing GET endpoint

jsonrpc/jsonrpc.go
zivkovicmilos: By the ethereum JSON-RPC spec, there are no `GET` endpoints.

We can use the `web3_clientVersion` endpoint as a health check.

Do you think it's better to add custom ones?

cc @Kourin1996

zivkovicmilos: I also don't see an easy way of returning a JSON response from this point in the dispatcher, because the entire JSON-RPC package is as flexible as an old stick

inglkruiz: Allow me to give you some feedback from what a user of the platform I developed told me. 
He is using the GET method as a health-check. BUT! In my opinion the health-check endpoint should not be in the RPC port, it should be served from the operations port (from where the Prometheus' metrics are exposed), call it `/healthz` as it is kind of the standard as far as I can tell. Returning text or a JSON object is up to the implementers as long as it returns status 200 (OK). 

For me as a Node operator I could add it to my Grafana Dashboard.

zivkovicmilos: I love the idea honestly.

@ZeljkoBenovic ? What do you think? How should we expose a health ping endpoint?

ZeljkoBenovic: We should definitely keep GET `/health` , `/ping` or `/whatever` endpoint on `json-prc` API, as it can be used to quickly determine if the node is running or not.  

Especially useful in automated deployments, where you want to quickly check if the `json-rpc` API is running or not, which you can do with quick & dirty `curl` test ( instead of going with the whole POST `eth` endpoint shenanigans ) . 
And it really doesn't matter what it responds with as long as the status is 200 OK .

inglkruiz: @ZeljkoBenovic sorry to disagree. I think there should not be any endpoint using a pathname (`/health`) on the JSON-RPC port (doing it feels like implementing a REST endpoint). If you want to verify if the JSON-RPC port is working, you could use `web3_clientVersion` as @zivkovicmilos has said, but I would not use it as a constant health-check. 

Again, a health-check IMO should be part of the operations port, and its logic should check the status of the entire node, i.e. gRPC port, RPC port, more things if any.

zivkovicmilos: @inglkruiz 

We'll discuss internally as a team of adding a similar feature in the future 👍

Kourin1996 (approved):

0xAleksaOpacic (approved):
LGTM, thank you for exelent tests coverage 🚀

PR #682, opened 2022-08-11
[Fix] Add disable for batch request limit

ZeljkoBenovic:
# Description

This PR adds the ability to disable `json-rpc` batch request limits, introduced in #638.

Some tools, like Blockscout are using batch requests in order to efficiently gather data from the chain.    
With current default limits, Blockscout in production, is not working as it is hitting this limits.   
Also, this protection is not needed for nodes that are not interacting with general public, for example a dedicated non-validator node that will serve only Blockscout.

To disable this limits just set the values to `0`:     
`--json-rpc-block-range-limit 0 ` and     
`--json-rpc-batch-request-limit 0`

Also, this PR fixes the issue that these values could not be fed trough config file. Right now, they can be set only with flags.     

# Changes include

# Checklist

## Testing

### Manual tests

Set the flags `--json-rpc-block-range-limit 0 `  , `--json-rpc-batch-request-limit 0` and you should not see `Batch request length too long` error

### DIFF
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -194,14 +194,14 @@ func setFlags(cmd *cobra.Command) {
 	)
 
 	cmd.Flags().Uint64Var(
-		&params.jsonRPCBatchLengthLimit,
+		&params.rawConfig.JSONRPCBatchRequestLimit,
 		jsonRPCBatchRequestLimitFlag,
 		defaultConfig.JSONRPCBatchRequestLimit,
 		"max length to be considered when handling json-rpc batch requests, value of 0 disables it",
 	)
 
 	cmd.Flags().Uint64Var(
-		&params.jsonRPCBlockRangeLimit,
+		&params.rawConfig.JSONRPCBlockRangeLimit,
 		jsonRPCBlockRangeLimitFlag,
 		defaultConfig.JSONRPCBlockRangeLimit,
 		"max block range to be considered when executing json-rpc requests "+
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -197,15 +197,15 @@ func setFlags(cmd *cobra.Command) {
 		&params.jsonRPCBatchLengthLimit,
 		jsonRPCBatchRequestLimitFlag,
 		defaultConfig.JSONRPCBatchRequestLimit,
-		"the max length to be considered when handling json-rpc batch requests",
+		"max length to be considered when handling json-rpc batch requests, value of 0 disables it",
 	)
 
-	//nolint:lll
 	cmd.Flags().Uint64Var(
 		&params.jsonRPCBlockRangeLimit,
 		jsonRPCBlockRangeLimitFlag,
 		defaultConfig.JSONRPCBlockRangeLimit,
-		"the max block range to be considered when executing json-rpc requests that consider fromBlock/toBlock values (e.g. eth_getLogs)",
+		"max block range to be considered when executing json-rpc requests "+
+			"that consider fromBlock/toBlock values (e.g. eth_getLogs), value of 0 disables it",
 	)
 
 	cmd.Flags().StringVar(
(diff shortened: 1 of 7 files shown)


0xAleksaOpacic (changes requested):
There is an issue with this PR, i will take some time today to debug it:

The scenario that does not work:
1. change hardcoded DefaultJSONRPCBlockRangeLimit to 50 (just so you don't need to wait for 1k blocks)
2. start 4 node network wait til 50+ blocks
3. trigger getLogs jsonrpc with (from: 0, to: latest), i am expecting ErrBlockRangeTooHigh but call passes

This issue was introduced with adding initJSONRPCBulkLimits.

```
if p.rawConfig.JSONRPCBatchRequestLimit != config.DefaultJSONRPCBatchRequestLimit {
		p.jsonRPCBatchLengthLimit = p.rawConfig.JSONRPCBatchRequestLimit
	}

if p.rawConfig.JSONRPCBlockRangeLimit != config.DefaultJSONRPCBlockRangeLimit {
		p.jsonRPCBlockRangeLimit = p.rawConfig.JSONRPCBlockRangeLimit
	}
```

this method is called after we setFlags. p.rawConfig.JSONRPCBatchRequestLimit will always be 0 because the value is defined in p.JSONRPCBatchRequestLimit. Because 0 is != config.DefaultJSONRPCBlockRangeLimit we are going to override p.jsonRPCBlockRangeLimit with 0. And with the functionality, you've added 0 means disabled.

command/server/server.go
```diff
@@ -197,15 +197,15 @@ func setFlags(cmd *cobra.Command) {
 		&params.jsonRPCBatchLengthLimit,
 		jsonRPCBatchRequestLimitFlag,
 		defaultConfig.JSONRPCBatchRequestLimit,
```
Kourin1996, line 199: Just curious. Is it ok that default value is not zero?

0xAleksaOpacic: @zivkovicmilos I see that we approved that on this PR https://github.com/0xPolygon/polygon-edge/pull/638. But I agree, we should consider having default 0, and then if someone wants can adjust it

zivkovicmilos: I don't think having `0` as the default is very smart. That PR introduced the default limit so we can have better node protection, and so people can't just make the node busy with a single request. If you want to disable this functionality as a node operator you can - I just don't think out-of-the-box configurations should enable something as dangerous as this.

For example, we have peer limits by default, but you can disable those with a special value.

zivkovicmilos (approved):
Looks ok to me 💯 

Please see @Aleksao998's comment on removing that method, and we're good to go.

command/server/server.go
0xAleksaOpacic: Sounds reasonable, agree lets leave it like it is

0xAleksaOpacic:
> Looks ok to me 💯
> 
> Please see @Aleksao998's comment on removing that method, and we're good to go.

Quick question, why we have to have jsonRPCBatchLengthLimit/jsonRPCBlockRangeLimit in 2 places.

```
type serverParams struct {

           rawConfig{
                JSONRPCBatchRequestLimit
                JSONRPCBlockRangeLimit
           }
           jsonRPCBatchLengthLimit
           jsonRPCBlockRangeLimit

}
```

ZeljkoBenovic:
You were right @Aleksao998 ,

I've removed the method that doesn't do anything and I've removed redundant variables from the PR that introduced this functionality.

I've also added a small test, to verify that when set to `0`, there are no request limits.

0xAleksaOpacic (approved):
LGTM, thank you for fixing all comments. Double checked everything it works great now!

command/server/server.go
0xAleksaOpacic: @Kourin1996 please resolve so we can merge this

PR #687, opened 2022-08-16
[TxPool] Introduce limit on enqueued transactions

dbrajovic:
# Description

This PR introduces a configurable limit on the amount of transactions any account in the TxPool can hold at any given time. 

`--max-enqueued [DEFAULT= 128]`

As the `enqueued` part of an account just represents future transactions (higher nonce than expected), there are no particular benefits to keeping a large number of them in this queue as much as there are vulnerability risks that all result in an eventually bricked pool full of enqueued transactions. 

The limit also enforces a rate limiting on transaction acquisition for any account - in effect, an incoming transaction would be rejected if there is no space for it in its respective `enqueued` queue. What this means is that aggressive spams are greatly mitigated, delaying the point of pool overflow. 

# Changes include

# Checklist

## Testing

# Documentation update

# Additional comments

Fixes EDGE-732


txpool/txpool_test.go
```diff
@@ -1209,7 +1280,8 @@ func (e *eoa) signTx(tx *types.Transaction, signer crypto.TxSigner) *types.Trans
 
 var signerEIP155 = crypto.NewEIP155Signer(100)
 
-func TestAddTxns(t *testing.T) {
+func TestAddTxs(t *testing.T) {
+	t.SkipNow()
```
zivkovicmilos: Leftover skip?

zivkovicmilos (approved):
Looks great 💯 

Tested this out with [Pandora's box](https://github.com/Trapesys/pandoras-box), works like a charm 👏

0xAleksaOpacic (approved):
We need to add doc for this command MaxAccountEnqueued.

0xAleksaOpacic (approved):
LGTM

txpool/txpool_test.go
dbrajovic: Actually, no. The test is obsolete because of the limit. I think it's safe to drop it entirely

txpool/account.go
```diff
@@ -151,6 +157,9 @@ type account struct {
 	enqueued, promoted *accountQueue
 	nextNonce          uint64
 	demotions          uint
+
+	//	maximum number of enqueued transactions
+	maxEnqueued uint64
```
Kourin1996, line 162: Why each account has `maxEnqueued`? `maxEnqueued` is settable for each account? You can pass the pointer and read it by atomic.Load. Or give function object to `account`

Kourin1996 (commented):
Just a small comment

txpool/account.go
dbrajovic: Nice catch. 

I thought about how to circumvent this - the proper way to extend the `accountQueue` struct to have a limited queue as well. That way I could reject a tx based on the the length of the queue.

If I'm not mistaken, you're suggesting a global variable that can be read atomically. In that case, tests could not be run in parallel, as each test would unpredictably mutate this field, causing tests to randomly fail.

I realize this increases the memory limit, but I intend to rework `account` struct for TxPool v3 (so it's just one field).

txpool/txpool_test.go
dbrajovic: Removed in [latest commit](https://github.com/0xPolygon/polygon-edge/pull/687/commits/6725f590a2faf39a463bffe424abef3489160296)

Kourin1996 (approved):

PR #689, opened 2022-08-18
[TxPool] Prune accounts with nonce holes during high memory usage

dbrajovic:
# Description

This PR introduces an asynchronous handler that is invoked at high gauge levels. Due to a possibility of some accounts having nonce holes ("stuck" enqueued transactions) taking up slots, the handler prunes each account satisfying this criterium. After pruning the handler goes into a short cooldown to avoid successive calls. 

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EDGE-732


zivkovicmilos (approved):
Looks great 💯

txpool/txpool_test.go
```diff
@@ -2016,6 +2140,7 @@ func TestRecovery(t *testing.T) {
 }
 
 func TestGetTxs(t *testing.T) {
+	t.SkipNow()
```
0xAleksaOpacic: why?

0xAleksaOpacic (approved):
LGTM

txpool/slot_gauge.go
```diff
@@ -27,6 +31,12 @@ func (g *slotGauge) decrease(slots uint64) {
 	atomic.AddUint64(&g.height, ^(slots - 1))
 }
 
+// highPressure checks if the gauge level
+// is higher than the highPressureMark (0.8 * max)
+func (g *slotGauge) highPressure() bool {
+	return g.read() > uint64(highPressureMark*float64(g.max))
```
Kourin1996: There may be an error depends on `highPressureMark` and `g.max`. For example, the error of right term is 1 where highPressureMark=0.29 and max=100. I think it's ok as far as highPressureMark is 0.8. I recommend you to calculate by integer division as long as highPressureMark can be represented in percent without the value after decimal point.

txpool/slot_gauge.go
```diff
@@ -27,6 +31,12 @@ func (g *slotGauge) decrease(slots uint64) {
 	atomic.AddUint64(&g.height, ^(slots - 1))
 }
 
+// highPressure checks if the gauge level
+// is higher than the highPressureMark (0.8 * max)
```
Kourin1996: Why do you write actual value (0.8) instead of highPressureMark?

Kourin1996 (commented):
Just small comments

txpool/slot_gauge.go
dbrajovic: Can you write the suggestion as a code snippet ?

txpool/slot_gauge.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/689/commits/8529040d699d5df794073eb115f9f41d58514291)

txpool/txpool_test.go
dbrajovic: [Reversed](https://github.com/0xPolygon/polygon-edge/pull/689/commits/6538db785bd659d094b25f65d6bbcfe4adf07aa9)

txpool/slot_gauge.go
Kourin1996:
```go
const (
     highPressureMark = 80 // 80%
)

func (g *slotGauge) highPressure() bool {
        return g.read() > (highPressureMark*g.max) / 100
}
```

txpool/txpool_test.go
zivkovicmilos: @Aleksao998 ping

txpool/slot_gauge.go
dbrajovic: Great catch, I was not aware of [this behaviour](https://goplay.tools/snippet/RP60no2ZcyE) in Go.

(the `highPressureMark` is `const` anyhow, but maybe we'll make it configurable in the future)

[Refactored](https://github.com/0xPolygon/polygon-edge/pull/689/commits/e3eb589d38b2d4ff417f0a18db6cc50ff1e0bf48)

Kourin1996 (approved):

PR #705, opened 2022-08-30
Secrets print functionality for ecdsa/bls validator public key address and node id

epikichi:
# Description

Currently, `polygon-edge` only outputs Node ID, BLS, and consensus public address upon initialization/generation. The functionality to output the same data with provided secrets manager configuration would be a great benefit to help with further adoption considerations.

This PR adds in a new sub command `polygon-edge secrets output` that will do just that. This functionality does depend on the executor to have sufficient permissions to the configured secrets manager, which may be a security consideration and dependency. 

# Changes include

# Behavioral changes
Not necessarily breaking, but a behavioral changes. 
- The local secrets manager now checks for existing secrets at each secret initialization. rather than the local secrets manager setup.

# Checklist

- [X] I have updated the official documentation (https://github.com/maticnetwork/matic-docs/pull/1091)

## Testing

### Manual tests

Tested with the following secret managers:
* `local`
* `aws-ssm`
* `gcp-ssm`
* `hashicorp-vault` 

# Additional Information

Here's some examples of the feature using a pre-generated keys rooted in `/opt/polygon-edge/data` . 
```sh
# outputs without additional flags
$ polygon-edge secrets output  --data-dir /opt/polygon-edge/data

[SECRETS OUTPUT]
Public key (address) = 0x1e506242DB0b09Ae74b22516aA47FF266Ce65778
BLS Public key       = 0xb2fea46d1c7317c73d3c06d5cf4623dc8df3b9a74b86a4d9481edc2746df02deaf6b31c8766816b640437040624ef04a
Node ID              = 16Uiu2HAm588VnQtWvNSSHHGTKGBPHskuTJ9gezmkRokSVHdTRfYL

# outputs validator address only
$ polygon-edge secrets output  --data-dir /opt/polygon-edge/data --validator
0x1e506242DB0b09Ae74b22516aA47FF266Ce65778

# outputs BLS address only
$ polygon-edge secrets output  --data-dir /opt/polygon-edge/data --bls
0xb2fea46d1c7317c73d3c06d5cf4623dc8df3b9a74b86a4d9481edc2746df02deaf6b31c8766816b640437040624ef04a

#  outputs node id only
$ polygon-edge secrets output  --data-dir /opt/polygon-edge/data  --node-id
16Uiu2HAm588VnQtWvNSSHHGTKGBPHskuTJ9gezmkRokSVHdTRfYL

# outputs node id only in json
$ polygon-edge secrets output  --data-dir /opt/polygon-edge/data  --node-id --json
{"node_id":"16Uiu2HAm588VnQtWvNSSHHGTKGBPHskuTJ9gezmkRokSVHdTRfYL"}
```


epikichi:

command/secrets/output/result.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if r.outputBLS {
+		buffer.WriteString(r.BLSPubkey)
+
+		return buffer.String()
+	}
+
+	vals := make([]string, 0, 3)
+
+	vals = append(
```
0xAleksaOpacic, line 49: Nitpick: 
Because we don't have a guard here to check whether address and nodeId exists, if data-dir does not exists the output would look like this
```
[SECRETS OUTPUT]
Public key (address) = 0x0000000000000000000000000000000000000000
Node ID              = <none>
```

For NodeID it's simple we can add a same guard as you added for blsPubKey

```
if r.NodeID != "" {
		vals = append(vals, fmt.Sprintf("Node ID|%s", r.NodeID))
}
```

For address, it's a bit trickier because once you use types.Address.String(), if the address does not exist it will convert it into 0x0 address (types.ZeroAddress). We can add a guard 

```
if r.Address != types.ZeroAddress.String() {
		vals = append(
			vals,
			fmt.Sprintf("Public key (address)|%s", r.Address),
		)
	}
```
because we are sure that no node address can be ZeroAddress. 

Another option is to print an error if the dir does not exist but i think it's more complicated and doesn't add much to ux

command/secrets/output/params.go
```diff
@@ -0,0 +1,219 @@
...
+		return err
+	}
+
+	op.nodeID = nodeID
+
+	return nil
+}
+
+func (op *outputParams) getResult() command.CommandResult {
+	if op.outputNodeID {
```
0xAleksaOpacic, line 185: Is there any specific reason you need to fetch these commands individually? I assume this can be useful for fetching specific keys for kubernetes or aws :)

0xAleksaOpacic (approved):
LGTM, congrats on your first Polygon Edge contribution 😃

command/secrets/output/params.go
epikichi: Yeah, helps to not have to  `awk` it out later on.

```sh
polygon-edge genesis --ibft-validator $(polygon-edge secrets output --data-dir . --validator) # and so on...
```

command/secrets/output/result.go
epikichi: Discussed the behavior to only print out the line if the respective secrets exist. May help with identifying missing keys in the future implicitly. 

```sh
$ ./polygon-edge secrets output --data-dir /not/here

[SECRETS OUTPUT]

```

command/secrets/output/params.go
```diff
@@ -0,0 +1,248 @@
...
+	case secrets.GCPSSM:
+		GCPSSM, err := helper.SetupGCPSSM(op.secretsConfig)
+		if err != nil {
+			return err
+		}
+
+		secretsManager = GCPSSM
+	default:
+		return errUnsupportedType
+	}
```
Kourin1996: Could you create a helper function to return secret manager by config because both `command/secrets/init` and `command/secrets/output` package use

command/secrets/output/params.go
```diff
@@ -0,0 +1,248 @@
...
+
+	errs := make([]string, 0, 2)
+	if _, err := os.Stat(validatorPathPrefix); os.IsNotExist(err) {
+		errs = append(errs, fmt.Sprintf("no validator keys found in the data directory provided: %s", dataDirAbs))
+	}
+
+	if _, err := os.Stat(networkPathPrefix); os.IsNotExist(err) {
+		errs = append(errs, fmt.Sprintf("no network key found in the data directory provided: %s", dataDirAbs))
+	}
```
Kourin1996: There is `DirectoryExists` function in `common/helper` package

command/secrets/output/result.go
```diff
@@ -0,0 +1,70 @@
...
+	outputValidator bool `json:"-"`
+}
+
+func (r *SecretsOutputResult) GetOutput() string {
+	var buffer bytes.Buffer
+
+	if r.outputNodeID {
+		buffer.WriteString(r.NodeID)
+
+		return buffer.String()
```
Kourin1996: Why not just return the value? Buffer is used to concat string.

command/secrets/output/params.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return &SecretsOutputResult{
+		Address:   op.validatorAddress,
+		BLSPubkey: op.blsPubkey,
+		NodeID:    op.nodeID,
+
+		outputValidator: op.outputValidator,
+		outputNodeID:    op.outputNodeID,
+	}
+}
```
Kourin1996: There are many conditional branches in this function and `GetOutput`. You might want to define 2 result structs: one is the struct that has only one field and another is the struct that has all public keys and ID.

e2e/framework/testserver.go
```diff
@@ -187,7 +187,7 @@ func (t *TestServer) SecretsInit() (*InitIBFTResult, error) {
 
 	commandSlice := strings.Split(fmt.Sprintf("secrets %s", secretsInitCmd.Use), " ")
 	args = append(args, commandSlice...)
-	args = append(args, "--data-dir", t.Config.IBFTDir)
+	args = append(args, "--data-dir", filepath.Join(t.Config.IBFTDir, "tmp"))
```
Kourin1996, line 190: Is there any special reason?

Kourin1996 (changes requested):

command/secrets/output/params.go
epikichi: Yup! Good idea.

command/secrets/output/params.go
epikichi: I compacted the conditionals a bit to fix my laziness. Thanks for the suggestion. Let me know if you have any better direction 🙏

Ref: https://github.com/0xPolygon/polygon-edge/pull/705/commits/70f9884642c307c03187adb82b0e7c24cb85051f

command/secrets/output/result.go
epikichi: Resolved in https://github.com/0xPolygon/polygon-edge/pull/705/commits/70f9884642c307c03187adb82b0e7c24cb85051f

command/secrets/output/params.go
epikichi: Resolved in https://github.com/0xPolygon/polygon-edge/pull/705/commits/70f9884642c307c03187adb82b0e7c24cb85051f

e2e/framework/testserver.go
epikichi: A few lines later, https://github.com/0xPolygon/polygon-edge/blob/70f9884642c307c03187adb82b0e7c24cb85051f/e2e/framework/testserver.go#L213-L246 we are re-initializing it in the same directory runs into this check. https://github.com/0xPolygon/polygon-edge/pull/705/files#diff-af4842623f99504a537de510a8411702cde6ff5671554f5076c9337135dbcef9R135-R141. 

Test Result: https://github.com/0xPolygon/polygon-edge/runs/8143748563?check_suite_focus=true

Basically, `testserver.go` generated keys twice; once from the Cmd execute and another from the function call. 
Correct me if I mistook the behavior, or if there's a better suggestion. Thanks!

command/secrets/output/params.go
epikichi: I gave it a shot with https://github.com/0xPolygon/polygon-edge/pull/705/commits/98cf7ffd20f578b36bd3127d3c095a83c09a4fca 

I think more can be done for a larger effort later down the road.

.gitignore
```diff
@@ -14,6 +14,7 @@ polygon-edge-chain*
 dist/
 main.exe
 main
+/polygon-edge
```
Kourin1996: Why this is included?

.gitignore
```diff
@@ -23,4 +24,4 @@ main
 artifacts
 
 # Log files
-*.log
\ No newline at end of file
+*.log
```
Kourin1996: This is also

command/secrets/output/params.go
```diff
@@ -0,0 +1,208 @@
...
+
+		return outputResults
+	}
+
+	outputResults.NodeID = op.nodeID
+	outputResults.BLSPubkey = op.blsPubkey
+	outputResults.Address = op.validatorAddress
+
+	return outputResults
+}
```
Kourin1996: For JSON, `outputXXX` flags are not used because of omitempty setting which means command returns empty json if one of the flags is given and the value is empty

command/secrets/output/params.go
```diff
@@ -0,0 +1,208 @@
...
+		return err
+	}
+
+	if err := op.initValidatorAddress(); err != nil {
+		return err
+	}
+
+	if err := op.initBLSPublicKey(); err != nil {
+		return err
+	}
```
Kourin1996: Why loading all values is necessary if `outputXXX` flag is given?

command/secrets/output/params.go
epikichi: https://github.com/0xPolygon/polygon-edge/pull/705/commits/dafc86c9cb4a67a6554eb20082578493d24a785d condensed the results to having an object per output, as suggested.

.gitignore
epikichi: https://github.com/0xPolygon/polygon-edge/pull/705/commits/094c1b2216abd470d6c32d25eb2f2dbbae36da7a

.gitignore
epikichi: We do want this `newline` in the end of the file to follow POSIX standards on what a [definition of a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) is.

We can add this as a separate PR if preferred.

command/secrets/output/params.go
epikichi: https://github.com/0xPolygon/polygon-edge/pull/705/commits/b614ea1bd7e2f75ef0449f48d7a3a17482b7642c
https://github.com/0xPolygon/polygon-edge/pull/705/commits/18c49958292126f73db027738f714c44f9f25a33

Checks for flag or the condition to print all before loading.

Kourin1996 (approved):

ZeljkoBenovic (approved):
Great job 🏅 
This will definitely be a great help with all automated deployments and troubleshooting.

zivkovicmilos (approved):
Looks great 💯 

Sorry for the late review on this 🙏 

Welcome to the contributor club 🙂

PR #707, opened 2022-08-31
[Feature] Datadog profiler

ZeljkoBenovic:
This PR adds Datadog profiler into the codebase, so that it can be easily enabled on demand.
Profiler enables users to see profiles and tracing stats on their Datadog portal, but it requires some configuration of Datadog agent on the node. 

APM ( tracing ) should be enabled in Datadog agent configuration yaml file:
```
apm_config:

  enabled: true
  env: polygon-edge # this sets the environment name that will be shown on DD portal
```

To enable the profiler, a user must simply set `DD_PROFILING_ENABLED="True"` environment variable.
To set the IP and port that the DataDog agent, use `DD_AGENT_HOST` and `DD_TRACE_AGENT_PORT` environment variables.    
By default, it will try communicate with the agent on `localhost:8126`.

These env vars are now consistent with the DD doc https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/go/?tab=containers#configure-the-datadog-agent-for-apm

Additional parameters can be set by using env vars stated in the official Datadog doc https://docs.datadoghq.com/profiler/enabling/go/

# Changes include

# Checklist

## Testing

### Manual tests

Manual tests would require that the tester has access to Datadog portal.
Just set the `DD_ENABLE` env var, configure Datadog agent on the same node, and check Datadog portal if you can see your profile.

# Documentation update

Public doc will soon with the explanation on how to set this up will follow

# Additional comments

Fixes EDGE-764


server/server_metrics.go
```diff
@@ -29,3 +33,45 @@ func metricProvider(nameSpace string, chainID string, metricsRequired bool) *ser
...
+	}
+
+	// start the tracer
+	tracer.Start()
+	s.logger.Info("DataDog profiler started")
+
+	return nil
+}
+
+func (s *Server) CloseDataDogProfiler() {
```
zivkovicmilos: This shouldn't be an exported method, no need for that

zivkovicmilos (approved):
Looks great 💯

server/server_metrics.go
ZeljkoBenovic: unexported in be46b7199f54ae1cfd825dc31d9500004f810875

0xAleksaOpacic (approved):
Code looks good to me, WHy are we having issues with CLAssistant first time seeing this

0xAleksaOpacic:
Can you add a comment: I have read the CLA Document and I hereby sign the CLA

server/server_metrics.go
```diff
@@ -29,3 +33,45 @@ func metricProvider(nameSpace string, chainID string, metricsRequired bool) *ser
...
+func (s *Server) closeDataDogProfiler() {
+	defer func() {
+		s.logger.Debug("closing DataDog profiler")
+		profiler.Stop()
+	}()
+
+	defer func() {
+		s.logger.Debug("closing DataDog tracer")
+		tracer.Stop()
+	}()
```
Kourin1996: Is there any reason to use defer here?

server/server_metrics.go
```diff
@@ -29,3 +33,45 @@ func metricProvider(nameSpace string, chainID string, metricsRequired bool) *ser
 		txpool:    txpool.NilMetrics(),
 	}
 }
+
+// enableDataDogProfiler enables DataDog profiler. Enable it by setting DD_ENABLE env var.
+// Additional parameters can be set with env vars (DD_) - https://docs.datadoghq.com/profiler/enabling/go/
+func (s *Server) enableDataDogProfiler() error {
+	if os.Getenv("DD_ENABLE") == "" {
```
Kourin1996: Why not check the value?

Kourin1996 (approved):

server/server_metrics.go
epikichi: In addition, we can use the variable `DD_PROFILING_ENABLED` for consistency and anticipation for a similar option in the Datadog Java Profiler. https://docs.datadoghq.com/profiler/enabling/java/?tab=environmentvariables

server/server_metrics.go
```diff
@@ -29,3 +33,45 @@ func metricProvider(nameSpace string, chainID string, metricsRequired bool) *ser
...
+		s.logger.Debug("DataDog profiler disabled, set DD_ENABLE env var to enable it.")
+
+		return nil
+	}
+
+	if err := profiler.Start(
+		// enable all profiles
```
epikichi, line 61: - We can look to add `profiler.WithAgentAddr()`  for containerized solutions, as the default agent address in the profiler is implicitly `localhost:8126`.

Reference: [1.30.0](https://github.com/DataDog/dd-trace-go/releases/tag/v1.30.0)

This behavior does not hold true for the tracer, as it takes up the `DD_AGENT_HOST` env var. The release has further explanation on why they introduced the behavior.

epikichi (changes requested):

server/server_metrics.go
ZeljkoBenovic: You're right. Defers here makes no sense. Removed e0a262a286b1bb9df18d4811b4c28f8f359f0bd0

server/server_metrics.go
ZeljkoBenovic: Added the option to set IP and port, using env vars. `DD_PROF_IP` and `DD_PROF_PORT` in  e0a262a286b1bb9df18d4811b4c28f8f359f0bd0

ZeljkoBenovic: Changed env var to enable DD agent to `DD_PROFILING_ENABLED` and to define the agent host `DD_AGENT_HOST` and port `DD_TRACE_AGENT_PORT`

vcastellm (commented):
The approach looks good in general, but I don't feel confortable with adding features using only env vars, I mean, ideally for me, this can be configurable in CLI params, config file and env vars.

epikichi:
> The approach looks good in general, but I don't feel confortable with adding features using only env vars, I mean, ideally for me, this can be configurable in CLI params, config file and env vars.

I agree as well. Let's include the CLI and config equivalent to this PR. @ZeljkoBenovic 

Ideally, we can offload the implementation to https://github.com/spf13/viper @zivkovicmilos :eyes:

ZeljkoBenovic:
@vcastellm The primary idea was not to clutter the UX with additional flags as DataDog is not a feature that is so commonly used as opposed to Prometheus for example.    
Secondly, we wanted to be aligned with their docs on how to setup things, for example https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/go/?tab=containers#configure-the-datadog-agent-for-apm where they are using env vars.     
Additionally, this is not a standalone feature, it still requires DataDog Agent installed and configured on the host ( or as a remote instance ) 

If you guys think that we need flags/config file parameters for this, it is very easy to add them.

server/server_metrics.go
epikichi: Looks good now. Thanks!

epikichi (approved):

vcastellm (approved):
LGTM

PR #710, opened 2022-08-31
[FIX] Properly handle block-gas-target flag

ZeljkoBenovic:
# Description

This PR fixes `--block-gas-target` flag as it is currently broken - it can be set but gas limit wont budge.

# Changes include

# Checklist

## Testing

### Manual tests

Deploy a cluster let it run for a while, set your desired `--block-gas-target` and check if `gasLimit` value is changing towards set value. Eventually, it should fixate on your defined gas target value.

Fixes EDGE-800

### DIFF
--- a/command/server/init.go
+++ b/command/server/init.go
@@ -123,6 +123,11 @@ func (p *serverParams) initGenesisConfig() error {
 		return parseErr
 	}
 
+	// if block-gas-target flag is set override genesis.json value
+	if p.blockGasTarget != 0 {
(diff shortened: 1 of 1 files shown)


zivkovicmilos (approved):
I wrote an entire rant about how this doesn't make any sense (the whole server runtime config and genesis config dance), but decided it's better to just leave it as it is, and rework it soon in the future.

Looks good 💯

0xAleksaOpacic (approved):

Kourin1996 (approved):

PR #717, opened 2022-09-06
[Rootnet Staking] SAMUEL

zivkovicmilos:
# Description

This PR introduces the SAMUEL module for Rootnet Staking.

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EDGE-816


blockchain/storage/keyvalue.go
```diff
@@ -284,6 +287,21 @@ func (s *KeyValueStorage) ReadTxLookup(hash types.Hash) (types.Hash, bool) {
...
+		return "", false
+	}
+
+	return string(data), true
+}
+
+// WriteLastProcessedEvent writes the last processed event info
+func (s *KeyValueStorage) WriteLastProcessedEvent(data string, contractAddr string) error {
+	return s.set(LAST_PROCESSED_EVENT, []byte(contractAddr), []byte(data))
+}
```
Kourin1996, line 303: I don't see any meaningful reason why this domain appears under blockchain/storage package. I thought you didn't want to use KeyValueStorage without modification. Why don't you use own storage?

blockchain/storage/keyvalue.go
```diff
@@ -39,6 +39,9 @@ var (
 
 	// TX_LOOKUP_PREFIX is the prefix for transaction lookups
 	TX_LOOKUP_PREFIX = []byte("l")
+
+	// LAST_PROCESSED_EVENT is the prefix for the last processed event lookups
+	LAST_PROCESSED_EVENT = []byte("last_processed_event")
```
Kourin1996, line 44: What does `event` mean?

blockchain/storage/testing.go
```diff
@@ -469,31 +469,36 @@ type writeReceiptsDelegate func(types.Hash, []*types.Receipt) error
...
+	writeBodyFn               writeBodyDelegate
+	readBodyFn                readBodyDelegate
+	writeSnapshotFn           writeSnapshotDelegate
+	readSnapshotFn            readSnapshotDelegate
+	writeReceiptsFn           writeReceiptsDelegate
+	readReceiptsFn            readReceiptsDelegate
+	writeTxLookupFn           writeTxLookupDelegate
+	readTxLookupFn            readTxLookupDelegate
+	writeLastProcessedEventFn writeLastProcessedEvent
+	readLastProcessedEventFn  readLastProcessedEvent
```
Kourin1996, line 499: Now these two mock methods seem to be called from the tests in `rootchain/sam` package. I think it doesn't make sense to define them in MockStorage. You can define another MockStorage inside `rootchain/sam`.

rootchain/samuel/mock_test.go
```diff
@@ -0,0 +1,172 @@
...
+	stopFn      stopDelegate
+	subscribeFn subscribeDelegate
+}
+
+func (m mockEventTracker) Start(startBlock uint64) error {
+	if m.startFn != nil {
+		return m.startFn(startBlock)
+	}
+
+	return nil
```
Kourin1996, line 23: Is there any intended reason to check nil here? I think panic can tell us the existence of unexpected method call.

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+	Start(uint64) error
+
+	// Stop stops the event tracker
+	Stop() error
+
+	// Subscribe creates a rootchain event subscription
+	Subscribe() <-chan rootchain.Event
+}
+
+// samp defines the SAMP interface for SAMUEL
```
Kourin1996: To be honest, I don't see what samp means. Please write what it stands for in comment when you use abbreviation.

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+// eventData holds information on event data mapping
+type eventData struct {
+	payloadType  rootchain.PayloadType
+	eventABI     *abi.Event
+	methodABI    *abi.Method
+	localAddress types.Address
+}
+
+// SAMUEL is the module that coordinates activities with the SAMP and Event Tracker
+type SAMUEL struct {
```
Kourin1996, line 86: Please explain what SAMUEL means so that everyone that has not read the Bible can understand it.

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+		samp:         samp,
+		signer:       signer,
+		storage:      storage,
+		transport:    transport,
+	}
+}
+
+// initEventLookupMap generates the SAMUEL event data lookup map from the
+// passed in rootchain configuration
+func initEventData(
```
Kourin1996: Comment seems to explain different function

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+		eventABI:     abi.MustNewEvent(configEvent.EventABI),
+		methodABI:    abi.MustNewABI(configEvent.MethodABI).GetMethod(configEvent.MethodName),
+		localAddress: types.StringToAddress(configEvent.LocalAddress),
+	}
+}
+
+// Start starts the SAMUEL module
+func (s *SAMUEL) Start() error {
+	// Start the event loop for the tracker
+	s.startEventLoop()
```
Kourin1996: If the below initialization processes fail, goroutine will be left, I think. Is it impossible to move `startEventLoop` to the bottom of the method?

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+
+	data, exists := s.storage.ReadLastProcessedEvent(s.eventData.localAddress.String())
+	if exists && data != "" {
+		// index:blockNumber
+		values := strings.Split(data, ":")
+		if len(values) < 2 {
+			return 0, fmt.Errorf("invalid last processed event in DB: %v", values)
+		}
+
+		blockNumber, err := strconv.ParseUint(values[1], 10, 64)
```
Kourin1996: When is the first value in the tuple used?

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+		if err != nil {
+			return 0, fmt.Errorf("unable to parse last processed block number in DB: %w", err)
+		}
```
Kourin1996: How about defining functions to parse and generate the date to be stored in storage?

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+		marshalledEvent, err := localSAM.Event.Marshal()
+		if err != nil {
+			s.logger.Error(
+				fmt.Sprintf("unable to marshal event, %v", err),
+			)
+
+			return
+		}
+
+		hash := crypto.Keccak256(marshalledEvent)
```
Kourin1996: Why not define Hash method in rootchain.Event?

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+
+		return
+	}
+
+	switch s.eventData.payloadType {
+	case rootchain.ValidatorSetPayloadType:
+		// The method needs to contain
+		// (validatorSet[], index, blockNumber)
+		index, _ := castParams["index"].(uint64)
+		blockNumber, _ := castParams["blockNumber"].(uint64)
```
Kourin1996, line 350: Why not check the second returns?

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+		return &types.Transaction{
+			Nonce:    0,
+			From:     types.ZeroAddress,
+			To:       &s.eventData.localAddress,
+			GasPrice: big.NewInt(0),
+			Gas:      framework.DefaultGasLimit,
+			Value:    big.NewInt(0),
+			V:        big.NewInt(1), // it is necessary to encode in rlp,
+			Input:    append(s.eventData.methodABI.ID(), encodedArgs...),
+		}
```
Kourin1996: You might want to define a function to create the transaction from payload because this method is too long

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	if len(candidateBuckets) == 0 {
+		// No candidate SAMs
+		return nil
+	}
+
+	// Out of all the candidate buckets, pick the one with the
+	// lowest quorum threshold
```
Kourin1996, line 515: Need more detailed reason here

rootchain/payload/validator_set_payload.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func NewValidatorSetPayload(setInfo []ValidatorSetInfo) *ValidatorSetPayload {
+	return &ValidatorSetPayload{
+		setInfo: setInfo,
+	}
+}
+
+func (v *ValidatorSetPayload) Get() (rootchain.PayloadType, []byte) {
+	// TODO @dbrajovic, do we even need to handle this error?
```
dbrajovic: I don't think we need to check for this error, as there will only be this particular type (`ValidatorSetPayload`)

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+type storage interface {
+	// ReadLastProcessedEvent reads the last processed event data
+	ReadLastProcessedEvent(string) (string, bool)
+
+	// WriteLastProcessedEvent writes the last processed event data
+	WriteLastProcessedEvent(data string, contractAddr string) error
+}
+
+// eventData holds information on event data mapping
+type eventData struct {
```
dbrajovic: Can you define helper methods on this type? It is used in a bunch of places and I think having methods on it would improve readability

rootchain/samuel/samuel.go
dbrajovic: I would make this a method of `eventData` (`initFromConfig`)

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+
+	return nil
+}
+
+// getStartBlockNumber determines the starting block for the Event Tracker
+func (s *SAMUEL) getStartBlockNumber() (uint64, error) {
+	startBlock := rootchain.LatestRootchainBlockNumber
+
+	data, exists := s.storage.ReadLastProcessedEvent(s.eventData.localAddress.String())
+	if exists && data != "" {
```
dbrajovic: Nitpick: invert the if condition to avoid a big if block: `if exists == "" || data == "" -> return`

rootchain/samuel/samuel.go
```diff
@@ -0,0 +1,580 @@
...
+			s.logger.Error(
+				fmt.Sprintf("unable to add event with hash %s to SAMP, %v", sam.Hash, err),
+			)
+		}
+	})
+}
+
+// getEventPayload retrieves a concrete payload implementation
+// based on the passed in byte array and payload type
+func getEventPayload(
```
dbrajovic: This looks like it should reside in the `payload` package

blockchain/storage/keyvalue.go
zivkovicmilos: I don't see the reason to define and carry around a new DB instance, since we already have one present in Edge.
This is detailed in the RFC - we want to go with a single DB instance instead of managing a separate DB for a single value

blockchain/storage/keyvalue.go
zivkovicmilos: It's terminology from the rootnet staking efforts - it signifies the last processed event data for the specific smart contract

blockchain/storage/testing.go
zivkovicmilos: There is actually a separate mock storage implementation for the SAMUEL storage:
https://github.com/0xPolygon/polygon-edge/blob/95c8a400ba84901a0f867498e369b620de161f50/rootchain/samuel/mock_test.go#L150-L172

What you linked is required because `MockStorage` needs to implement the base `Storage` interface methods to be used as a `Storage` interface

rootchain/samuel/mock_test.go
zivkovicmilos: This is how we do mock methods for 100% of mocks in the codebase. If it's not defined, I don't need it to panic

rootchain/samuel/samuel.go
zivkovicmilos: Explained 👍 

[1bb2688](https://github.com/0xPolygon/polygon-edge/pull/717/commits/1bb26880acfe111c2e85d889d152c8973e3b3367)

rootchain/samuel/samuel.go
zivkovicmilos: We will have detailed explanations on the SAMUEL module after we complete the effort.
Right now, it doesn't make sense to go into detail on what it does in the comments

rootchain/samuel/samuel.go
zivkovicmilos: Leftover 🤦‍♂️ 

Fixed:

[77d8a8e](https://github.com/0xPolygon/polygon-edge/pull/717/commits/77d8a8e008a83e517d20627420177ba4368e9dba)

rootchain/samuel/samuel.go
zivkovicmilos: This is a good catch 👍 

Moved it, it doesn't make sense to listen for events if the event tracker start fails:

[2f29004](https://github.com/0xPolygon/polygon-edge/pull/717/commits/2f2900400f619dd610e4bc76ad9aae28a0fa6f87)

blockchain/storage/keyvalue.go
zivkovicmilos: @Kourin1996 
Completely disregard this, we will use a separate DB instance that will be added in a separate PR.

It will be a Key Value store, so we'll leave this as is

rootchain/samuel/samuel.go
zivkovicmilos: This was something that was unclear before.

Added it after consulting with @dbrajovic:

[2670f51](https://github.com/0xPolygon/polygon-edge/pull/717/commits/2670f51877cb301294c2924c233751a72f77c90d)

rootchain/samuel/samuel.go
zivkovicmilos: The snippet you linked is already part of the helper method.

Saving to the DB is a one-liner

rootchain/samuel/samuel.go
zivkovicmilos: What do you mean?

It's always going to be a keccak hash, regardless of the event type

rootchain/samuel/samuel.go
zivkovicmilos: We can assume the cast is correct - if the input params are not valid, they will not get decoded correctly

rootchain/samuel/samuel.go
zivkovicmilos: It actually doesn't matter which quorum bucket we choose - @dbrajovic and I just decided to go with the lowest possible quorum size to have the lowest transaction size

rootchain/samuel/samuel.go
zivkovicmilos: Added 👍 

[2670f51](https://github.com/0xPolygon/polygon-edge/pull/717/commits/2670f51877cb301294c2924c233751a72f77c90d)

rootchain/samuel/samuel.go
zivkovicmilos: Simplified 👍 

[cad229b](https://github.com/0xPolygon/polygon-edge/pull/717/commits/cad229b9d52bf7687eadc44f9466201eaff2075f)

rootchain/samuel/samuel.go
zivkovicmilos: Agreed 👍 

Moved out:

[432a1d5](https://github.com/0xPolygon/polygon-edge/pull/717/commits/432a1d5f5beaf7e92690477ac240b0317dc5312f)

rootchain/samuel/samuel.go
zivkovicmilos: I'm trying to figure out how this can be reworked to be cleaner, but I cannot find a smart way to do it - because you have these constraints to pass out many arguments. Apart from defining intermediary structs for packaging them up, I don't see a way to have something smart.

Do you have any suggestions?

dbrajovic (approved):

blockchain/storage/keyvalue.go
Kourin1996: I meant blockchain module doesn't know the term used in RootnetStaking, then supplement is needed.

rootchain/samuel/mock_test.go
Kourin1996: In the first place, a test causing panic is missing something. If the mock has default behaviour, hard to notice unintentionally method call.

rootchain/samuel/samuel.go
Kourin1996: Keeping this comment unresolved to remember what we talk here

blockchain/storage/testing.go
Kourin1996: Firstly, the added methods and the other functions are called from different modules completely. Why not consider about splitting interface?

rootchain/samuel/samuel.go
Kourin1996: To be honest, this part is a bit long to be called from this method. I thought defining method can simplify the flow.

rootchain/samuel/samuel.go
Kourin1996: I meant we should not write long code in switch-case

blockchain/storage/keyvalue.go
zivkovicmilos: The keyvalue store is in the blockchain module (the entire storage module for that matter), that's why we need to have this key prefix in there

rootchain/samuel/mock_test.go
zivkovicmilos: Resolving this as it's irrelevant

rootchain/samuel/samuel.go
zivkovicmilos: Resolving because documentation on SAMUEL will be added to the RFC later on

rootchain/payload/validator_set_payload.go
zivkovicmilos: Removed 👍 

[1b6d795](https://github.com/0xPolygon/polygon-edge/pull/717/commits/1b6d7953e2fa8f5ebb3a2580b89471a52c5f5471)

rootchain/samuel/samuel.go
zivkovicmilos: The switch case code is what I was talking about in the first place - you need to transfer the extracted arguments into it regardless.

Resolving this because we might tackle it in a future PR

rootchain/samuel/samuel.go
zivkovicmilos: Ah, I understand now what you meant. We always do this coupled operation -> marshal + hash, I've gone over the usages and that's the only reason we need the marshal exported, to hash it.

Moved it to a separate helper, it's much cleaner now:

[9217e48](https://github.com/0xPolygon/polygon-edge/pull/717/commits/9217e48af7bd637e4b88db341a436824859a99df)

blockchain/storage/testing.go
zivkovicmilos: The initialization is not present in this PR, but in #713, the DB interface is going to be of type `Storage` (key-value store), so we need to keep it

PR #718, opened 2022-09-07
[Rootnet Staking] SAM Pool

dbrajovic:
# Description

This PR introduces the `sampool` package that implements a container for Signed Arbitrary Messages (SAM). 

The container acts as a queue of messages order by their `Index`. At each level (index), messages are aggregated into buckets (in case of multiple message hashes with the same index) with each bucket containing uniquely signed messages. 

The message with the lowest `Index` is retrieved from the pool along with its signatures. 

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EDGE-775


rootchain/sampool/bucket.go
```diff
@@ -0,0 +1,63 @@
...
+
+func newUniqueSAMs() samSet {
+	return samSet{
+		messages:   make([]rootchain.SAM, 0),
+		signatures: make(map[string]bool),
+	}
+}
+
+func (s *samSet) add(msg rootchain.SAM) {
+	strSignature := string(msg.Signature)
```
Kourin1996, line 24: How about converting bytes to hex for easier debug

rootchain/sampool/bucket.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/0xPolygon/polygon-edge/rootchain"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type samSet struct {
+	messages   []rootchain.SAM
+	signatures map[string]bool
+}
+
+func newUniqueSAMs() samSet {
```
Kourin1996: What does `unique` mean? Do you need it in the function name?

rootchain/sampool/bucket.go
```diff
@@ -0,0 +1,63 @@
...
+	return samSet{
+		messages:   make([]rootchain.SAM, 0),
+		signatures: make(map[string]bool),
+	}
+}
+
+func (s *samSet) add(msg rootchain.SAM) {
```
Kourin1996, line 23: I want a comment describing that it's not safe against concurrent access.

rootchain/sampool/bucket.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if s.signatures[strSignature] {
+		return
+	}
+
+	s.messages = append(s.messages, msg)
+	s.signatures[strSignature] = true
+}
+
+func (s *samSet) get() []rootchain.SAM {
```
Kourin1996: I think `Messages` sounds better because I don't know what `get` returns from the name `get` and `samSet`.

rootchain/sampool/bucket.go
```diff
@@ -0,0 +1,63 @@
+package sampool
+
+import (
+	"github.com/0xPolygon/polygon-edge/rootchain"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type samSet struct {
+	messages   []rootchain.SAM
+	signatures map[string]bool
```
Kourin1996: I think it's just used to hold flags telling which messages are stored. You might want to rename it to some proper name like `isAdded`?

rootchain/sampool/pool.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil
+	}
+
+	messages := bucket.getQuorumMessages(p.verifier.Quorum)
+	if len(messages) == 0 {
+		return nil
+	}
+
+	result := make([]rootchain.SAM, len(messages))
+	copy(result, messages)
```
Kourin1996, line 99: Can you explain why shallow copy is necessary here?

rootchain/sampool/pool.go
```diff
@@ -0,0 +1,146 @@
...
+
+	//	verify message signature
+	if err := p.verifier.VerifySignature(msg); err != nil {
+		return fmt.Errorf("%w: %v", ErrInvalidSignature, err)
+	}
+
+	//	reject old message
+	if index := msg.Index; index <= p.lastProcessedIndex {
+		return fmt.Errorf("%w: message number %d", ErrStaleMessage, index)
+	}
```
Kourin1996, line 135: Why is this check the last one?

rootchain/sampool/pool.go
```diff
@@ -0,0 +1,146 @@
...
+
+	p.lastProcessedIndex = index
+}
+
+// TODO: Peek or Pop might be redundant
+func (p *SAMPool) Peek() rootchain.VerifiedSAM {
+	p.mux.Lock()
+	defer p.mux.Unlock()
+
+	expectedIndex := p.lastProcessedIndex + 1
```
Kourin1996, line 84: Is this `Index` sort of `nonce` to be set in original event?

rootchain/sampool/pool_test.go
```diff
@@ -0,0 +1,356 @@
...
+			}
+
+			pool := New(verifier)
+
+			assert.ErrorIs(t,
+				pool.AddMessage(rootchain.SAM{}),
+				ErrInvalidHash,
+			)
+		},
+	)
```
Kourin1996: I think you need to make sure the pool is empty in case of error

rootchain/sampool/pool_test.go
```diff
@@ -0,0 +1,356 @@
...
+				Hash: types.Hash{111},
+				Event: rootchain.Event{
+					Index: 10,
+				},
+			}
+
+			assert.NoError(t, pool.AddMessage(msg))
+
+			_, ok := pool.messages[msg.Index]
+			assert.True(t, ok)
```
Kourin1996: You can use `assert.Contains`

rootchain/sampool/bucket.go
dbrajovic: Do we have a helper utility for that?  (encoding/decoding)

rootchain/sampool/pool.go
dbrajovic: Exactly - the `lastProcessedIndex` helps deciding the next event to process from the SAM Pool. (Because the implementation is a `map` and not an ordered queue)

rootchain/sampool/bucket.go
dbrajovic: Leftover code. 

[Renamed](https://github.com/0xPolygon/polygon-edge/pull/718/commits/fe6278544c8f62a6f5938e8b3f7fb66f53ea94c8)

rootchain/sampool/bucket.go
dbrajovic: That's why I always look at the return type.

[Renamed](https://github.com/0xPolygon/polygon-edge/pull/718/commits/0e4b695ce070ddd52a5a2f7840d08c09cfd02c78)

rootchain/sampool/bucket.go
dbrajovic: [Renamed](https://github.com/0xPolygon/polygon-edge/pull/718/commits/6af174b0e8143c30e9b4f92e54fccbfe61030dcf)

rootchain/sampool/pool.go
dbrajovic: To protect against race conditions.

In an unlikely event that a call to `Peek` is made while there is a call to `AddMessage`, the original [slice reference](https://github.com/0xPolygon/polygon-edge/blob/7e2a9c397288c1edcb2c769c2a5d38ecd4430330/rootchain/sampool/bucket.go#L12) would be exposed for both read and write. The `copy` call is there just to ensure no 2 goroutines will be mutating this slice at the same time.

rootchain/sampool/pool.go
dbrajovic: It is now the only check - the verification logic will be present in the `SAMUEL` package. 

[Removed verification](https://github.com/0xPolygon/polygon-edge/pull/718/commits/57d614be5b069d8558a10662e3564149e513a438)

rootchain/sampool/bucket.go
Kourin1996: hex package has

rootchain/sampool/pool.go
Kourin1996: Why mux can't prepend this?

dbrajovic: If I return the underlying slice, then the caller can do a read on it while an event being added in the background is about to write to it

rootchain/sampool/bucket.go
dbrajovic: I think it's pretty apparent from the code

rootchain/sampool/pool_test.go
dbrajovic: Deprecated due to responsibility being delegated to `samuel`

rootchain/sampool/bucket.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/718/commits/4de7093b9051caaff411b7743d30395affb64ad1)

rootchain/sampool/pool_test.go
dbrajovic: [Fixed](https://github.com/0xPolygon/polygon-edge/pull/718/commits/fd6d964419635c317abbf5cf19a42827d3a89974)

Issue #684, opened 2022-08-12
From address is a zero address


semiPlanet:
### Description

I run a blockchain with 4 servers (4 bootnodes and 4 validators all) when I send ether to another address the from address is always zero address. I do not know what is the cause of this issue.
`{"jsonrpc":"2.0","id":1,"result":{"nonce":"0x7","gasPrice":"0x2540be400","gas":"0x5208","to":"0xaE65963832A2c4615CcC09e944aA3186B225Aa95","value":"0xde0b6b3a7640000","input":"0x","v":"0x648","r":"0xde9193e4a6de8f37f53a2c6513e3bc040ca46f7e9934f7a6e5c4924a8d5a2735","s":"0x1c1df31ab92926e501e47fb0c721e18d8bcabd61c894822fa83068fd8722c03","hash":"0x33572c89dcdac1607b79b8c43c301c3e76a3164f99e4b62e165ddf4b61faefe2","from":"0x0000000000000000000000000000000000000000","blockHash":"0x57b0890205bed1fa38c60ec641b4d0533325ff5542347ee9f48d0dca5c7b5bd4","blockNumber":"0x26a2","transactionIndex":"0x0"}}`

### Your environment

* OS and version: Ubuntu 20.04.4 LTS (GNU/Linux 5.4.0-122-generic x86_64)
* version of the Polygon Edge: Latest
* branch that causes this issue: Develop
* Go version: go1.17.11 linux/amd64

### Steps to reproduce

* Start a blochain with 4 bootnodes and 4 validator addresses <br />
* In every transactions from address is zero address <br />
* `polygon-edge server --price-limit "5000000000" --data-dir /home/ubuntu/node-dir --chain /home/ubuntu/node-dir/genesis.json  --libp2p 0.0.0.0:4546 --jsonrpc :4545 --nat 92.205.111.193 --seal`

### Expected behaviour

From address should be actual sender address

### Actual behaviour
The sender adddress is zero address
Link to block number 18106: https://nxchainscan.com/tx/0xfcf866b8fbcf88e28ae4c6c83a151a033dfd356ca354feac2ea9954e97365b0a

### Logs (Block Number: 18106)
Aug 11 20:44:18 193.111.205.92.host.secureserver.net run-node.sh[10556]: 2022-08-11T20:44:18.427Z [INFO]  polygon.blockchain: new block: number=18103 txs=0 hash=0xd6e0886dd27847c962cecf4550a2c0bc28dce0b561516282be>
[...]

PR #719, opened 2022-09-07
Fix zero from field in Tx

Kourin1996:
Fix EDGE-787

# Description

This PR fixes the issue that `from` field in the transaction data JSON-RPC API returns is zero address in the latest code. The issue is caused by that block data is unmarshalled before writing block in latest IBFT and this encoding doesn't add `from` field in the marshaled data.

This PR adds 2 fixes:
(1) Recover `from` field in transaction in blockchain module before writing the block into storage.
(2) Recover `from` field if the transaction in the block doesn't have `from` field when fetching block from storage and update the data in storage as well.

Relevant issue:
https://github.com/0xPolygon/polygon-edge/issues/684

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -580,7 +580,7 @@ func (b *Blockchain) readBody(hash types.Hash) (*types.Body, bool) {
 	}
 
 	// To return from field in the transactions of the past blocks
-	if updated := b.recoverAllFromFieldsInTransactions(bb.Transactions); updated {
+	if updated := b.recoverFromFieldsInTransactions(bb.Transactions); updated {
 		if err := b.db.WriteBody(hash, bb); err != nil {
 			b.logger.Warn("failed to write body into storage", "hash", hash, "err", err)
 		}
@@ -1010,7 +1010,7 @@ func (b *Blockchain) ReadTxLookup(hash types.Hash) (types.Hash, bool) {
 	return v, ok
 }
 
-// recoverFromFields recovers 'from' fields in the transactions of the given block
+// recoverFromFieldsInBlock recovers 'from' fields in the transactions of the given block
 // return error if the invalid signature found
 func (b *Blockchain) recoverFromFieldsInBlock(block *types.Block) error {
 	for _, tx := range block.Transactions {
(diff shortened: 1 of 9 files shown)


zivkovicmilos (approved):
Looks great 💯 

I would just like to note that this issue was present even before we upgraded our consensus layer - the previous IBFT implementation (incorrectly) had a side effect of instantiating the `From` field of a transaction when _verifying a block_ (even though it was not set), something that no normal consensus implementation should do. The new IBFT implementation doesn't do any such shenanigans, so this error propagated out (the transaction `From` not being marshaled in RLP) to other parts of the system.

Just make sure to resolve the linting errors.

Thank you for fixing this 🙏

dbrajovic (approved):

blockchain/blockchain.go
```diff
@@ -990,6 +1010,49 @@ func (b *Blockchain) ReadTxLookup(hash types.Hash) (types.Hash, bool) {
...
+
+		tx.From = sender
+	}
+
+	return nil
+}
+
+// recoverFromFieldsInTransactions recovers 'from' fields in the transactions
+// log as warning if failing to recover one address
+func (b *Blockchain) recoverAllFromFieldsInTransactions(transactions []*types.Transaction) bool {
```
0xAleksaOpacic: typo it should be called recoverFromFieldsInTransactions

blockchain/blockchain.go
```diff
@@ -990,6 +1010,49 @@ func (b *Blockchain) ReadTxLookup(hash types.Hash) (types.Hash, bool) {
 	return v, ok
 }
 
+// recoverFromFields recovers 'from' fields in the transactions of the given block
```
0xAleksaOpacic: typo should be recoverFromFieldsInTransactions not recoverFromFields

0xAleksaOpacic (approved):
LGTM

blockchain/blockchain.go
Kourin1996: Fixed 457977290e3bb91be48ba4e9bf2119baecffb787

blockchain/blockchain.go
Kourin1996: Fixed 457977290e3bb91be48ba4e9bf2119baecffb787

zivkovicmilos (approved):
Looks good again 💯

Issue #731, opened 2022-09-13
Promoted transactions in TxPool not demoting on recoverable error


lazartravica:
Was able to repro what we believe to be a similar scenario to the one we initially observed using the steps below:

1. assuming account balance is 5 ETH - submit a couple simple transfer txs we can currently afford with current nonce + 10 (for example each tx costing total of 3 ETH each), these txs will then appear within `queued` in txpool_inspect response (for example with nonces 450-451)
2. submit 10 txs we can currently afford using the current nonce (for example nonces 440-449 with each tx costing only 0.1 ETH each)
3. once 449 has been queued, promoted and mined, the queued txs with nonces 450 and 451 will be promoted to pending however at this point, the account balance cannot cover the tx costs:
```
{
    "jsonrpc": "2.0",
    "id": null,
    "result": {
        "pending": {
            "0xFd9E636956415cC447Fb621223d2fc779819E533": {
                "450": "3000000000000000000 wei + 21000 gas x 10000000000 wei",
                "451": "3000000000000000000 wei + 21000 gas x 10000000000 wei"
            }
        },
        "queued": {},
        "currentCapacity": 2,
        "maxCapacity": 4096
    }
}
```
4. these stuck pending txs do not go away on their own even once the account balance is topped up
---

Not sure but I would expect the queued txs to consider the account balance at the latest block during the promotion process and therefore not get promoted.

_Originally posted by @dankostiuk in https://github.com/0xPolygon/polygon-edge/issues/711#issuecomment-1244680548_

dbrajovic:
Possibly related:

https://github.com/0xPolygon/polygon-edge/issues/226

dbrajovic:
@dankostiuk 

This looks a lot like what you previously reported back in #226. The part which raises concern is:

> these stuck pending txs do not go away on their own 

even more so

> even once the account balance is topped up

The `Demote -> Drop` logic was implemented to solve this specific problem. Continuing from the previous issue, how come there is no `Demote/Drop` taking place for pending `450` and `451`? The only way I see that _not_ happening is that the txs eventually ended up in a block (due to balances being topped), but you claim this is not the case.

If you have the logs (or can reproduce them again), would you mind sharing?

dankostiuk:
Hey @dbrajovic ,

I've reproduced the issue with an account with 101 ETH and current nonce (getTransactionCount) 566. See below for the steps I took:

[...]

PR #742, opened 2022-09-20
Cleanup inactive transactions in non-validator

Kourin1996:
Fix EDGE-843

# Description

This PR adds a function to remove the transactions that have not been adopted for a certain number of transactions.
And also this PR disable `--seal` flag but keep CLI flag because this PR is intended to be merged in the next PATCH release

Relevant: https://github.com/0xPolygon/polygon-edge/issues/731

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

1. Send the multiple transactions with consecutive nonces to **non-validator node** so that validators process the first part of transactions and fail to include the second part of the transactions (e.g. Send the bigger amount than the value account holds in a multiple transactions)
2. Make sure there are pending transactions in TxPool of validator and non-validator
3. Make sure the first part of the transactions are processed
4. Wait for a while
5. Make sure there are no transactions in TxPool of validator and non-validator

# Documentation update

# Additional comments

### DIFF
--- a/consensus/dev/dev.go
+++ b/consensus/dev/dev.go
@@ -62,7 +62,7 @@ func Factory(
 // Initialize initializes the consensus
 func (d *Dev) Initialize() error {
 	d.txpool.SetSealing(true)
-	
+
 	return nil
 }
 
--- a/consensus/dev/dev.go
+++ b/consensus/dev/dev.go
@@ -61,6 +61,8 @@ func Factory(
 
 // Initialize initializes the consensus
 func (d *Dev) Initialize() error {
+	d.txpool.SetSealing(true)
+	
 	return nil
 }
 
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -41,6 +41,7 @@ type txPoolInterface interface {
 	Drop(tx *types.Transaction)
 	Demote(tx *types.Transaction)
 	ResetWithHeaders(headers ...*types.Header)
+	SetSealing(bool)
 }
 
 // backendIBFT represents the IBFT consensus mechanism object
@@ -392,6 +393,8 @@ func (i *backendIBFT) startConsensus() {
 		i.updateActiveValidatorSet(latest)
 
 		isValidator = i.isActiveValidator()
+
+		i.txpool.SetSealing(isValidator)
 
 		if isValidator {
 			sequenceCh = i.consensus.runSequence(pending)
--- a/consensus/ibft/ibft.go
+++ b/consensus/ibft/ibft.go
@@ -78,8 +78,6 @@ type backendIBFT struct {
 
 	blockTime time.Duration // Minimum block generation time in seconds
 
-	sealing bool // Flag indicating if the node is a sealer
-
 	closeCh chan struct{} // Channel for closing
 }
 
@@ -122,7 +120,6 @@ func Factory(params *consensus.Params) (consensus.Consensus, error) {
 		network:            params.Network,
 		epochSize:          epochSize,
 		quorumSizeBlockNum: quorumSizeBlockNum,
-		sealing:            params.Seal,
 		metrics:            params.Metrics,
 		secretsManager:     params.SecretsManager,
 		blockTime:          time.Duration(params.BlockTime) * time.Second,
@@ -465,11 +462,6 @@ func (i *backendIBFT) updateMetrics(block *types.Block) {
 var (
 	errBlockVerificationFailed = errors.New("block verification fail")
 )
-
-// isSealing checks if the current node is sealing blocks
-func (i *backendIBFT) isSealing() bool {
-	return i.sealing
-}
 
 // verifyHeaderImpl implements the actual header verification logic
 func (i *backendIBFT) verifyHeaderImpl(snap *Snapshot, parent, header *types.Header) error {
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -305,3 +305,23 @@ func (a *account) resetSkips()  {
 func (a *account) incrementSkips()  {
 	a.skips++
 }
+
+// getLowestTx returns the transaction with lowest nonce, which might be popped next
+// this method don't pop a transaction from both queues
+func (a *account) getLowestTx() *types.Transaction {
+	a.promoted.lock(true)
+	defer a.promoted.unlock()
+
+	if firstPromoted := a.promoted.peek(); firstPromoted != nil {
+		return firstPromoted
+	}
+
+	a.enqueued.lock(true)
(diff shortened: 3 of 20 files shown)


txpool/txpool.go
```diff
@@ -321,6 +323,16 @@ func (p *TxPool) SetSigner(s signer) {
 	p.signer = s
 }
 
+// SetSealing sets the sealing flag
+func (p *TxPool) SetSealing(sealing bool) {
+	p.sealing.Store(sealing)
+}
+
+// Sealing returns the current set sealing flag
+func (p *TxPool) Sealing() bool {
```
dbrajovic: You can un-export this method

txpool/account.go
```diff
@@ -157,6 +157,7 @@ type account struct {
 	enqueued, promoted *accountQueue
 	nextNonce          uint64
 	demotions          uint
+	unadopted          uint64
```
dbrajovic: Naming nitpick: unadopted -> skips

txpool/txpool.go
```diff
@@ -881,6 +898,49 @@ func (p *TxPool) resetAccounts(stateNonces map[types.Address]uint64) {
...
+	p.accounts.Range(
+		func(key, value interface{}) bool {
+			address, _ := key.(types.Address)
+			account, _ := value.(*account)
+
+			if _, ok := latestActiveAccounts[address]; ok {
+				account.unadopted = 0
+			} else {
+				account.unadopted++
+			}
```
dbrajovic: Nitpick: make this a method of `account`

txpool/txpool.go
```diff
@@ -881,6 +898,49 @@ func (p *TxPool) resetAccounts(stateNonces map[types.Address]uint64) {
...
+
+			account.enqueued.lock(true)
+			defer account.enqueued.unlock()
+
+			firstTx = account.enqueued.pop()
+			if firstTx != nil {
+				p.index.remove(firstTx)
+				p.gauge.decrease(slotsRequired(firstTx))
+				account.unadopted = 0
+			}
```
dbrajovic, line 937: This is the critical part the txpool needs to do correctly: since we're trying to mimic what validator nodes do when performing the `Demote->Drop` this piece of code should be replaced with a call to `Drop()` 

(Drop does all of this logic already and updates the metrics as well)

The core idea is symmetrical - `demotions` field is for validators `skips` (unadopted) for non-validators, but the behavior needs to be identical.

dbrajovic: This is the critical part the txpool needs to do correctly: since we're trying to mimic what validator nodes do when performing the `Demote->Drop`. This piece of code should be replaced with a call to `Drop()` 

(Drop does all of this logic already and updates the metrics as well)

The core idea is symmetrical - `demotions` field is for validators, `skips` (unadopted) for non-validators, but the behavior needs to be identical.

dbrajovic (changes requested):
Already looks good, just need to resolve the `updateUnadoptedCounts` detail 👍

consensus/dev/dev.go
```diff
@@ -61,6 +61,8 @@ func Factory(
 
 // Initialize initializes the consensus
 func (d *Dev) Initialize() error {
+	d.txpool.SetSealing(true)
```
zivkovicmilos, line 64: Why do you need this anymore?

consensus/ibft/ibft.go
```diff
@@ -41,6 +41,7 @@ type txPoolInterface interface {
 	Drop(tx *types.Transaction)
 	Demote(tx *types.Transaction)
 	ResetWithHeaders(headers ...*types.Header)
+	SetSealing(bool)
```
zivkovicmilos, line 44: You don't need this anymore

txpool/txpool.go
```diff
@@ -321,6 +323,16 @@ func (p *TxPool) SetSigner(s signer) {
 	p.signer = s
 }
 
+// SetSealing sets the sealing flag
+func (p *TxPool) SetSealing(sealing bool) {
+	p.sealing.Store(sealing)
+}
```
zivkovicmilos, line 329: Why is this required?

zivkovicmilos (approved):

txpool/txpool.go
Kourin1996: Fixed 52735acfc16c2ce6849fcc18334ad8c4c7180a0e

txpool/account.go
Kourin1996: Renamed e0372e5b909ab6f543b904505f9746abfe21a078

txpool/txpool.go
Kourin1996: Added 1f9344761d8c0676269fbf20d774528236c7f6c0

txpool/txpool.go
Kourin1996: Fixed　9c01589800042e14d3b408732fc4249457a0b292

txpool/account.go
```diff
@@ -293,3 +295,33 @@ func (a *account) promote() []*types.Transaction {
...
+		return firstPromoted
+	}
+
+	a.enqueued.lock(true)
+	defer a.enqueued.unlock()
+
+	if firstEnqueued := a.enqueued.peek(); firstEnqueued != nil {
+		return firstEnqueued
+	}
+
```
dbrajovic, line 325: This can potentially brick an account later on when `Drop` sets the nonce to that of the enqueued transaction. 

Here is my proposal on `updateAccountSkipsCounts` and `getLowestTx`:

```
	p.accounts.Range(
		func(key, value interface{}) bool {
			address, _ := key.(types.Address)
			account, _ := value.(*account)

			if _, ok := latestActiveAccounts[address]; ok {
				account.resetSkips()

				return true
			}

			firstPromoted := account.getLowestTx()
			if firstPromoted == nil {
				// no need to increment anything,
				// account has no txs
				return true
			}

			account.incrementSkips()

			if account.skips < maxAccountSkips {
				return true
			}

			// account has been skipped too many times
			p.Drop(firstPromoted)
			account.resetSkips()

			return true
		},
	)
```

```
func (a *account) getLowestTx() *types.Transaction {
	a.promoted.lock(false) // only read, no write
	defer a.promoted.unlock()

	return a.promoted.peek()
}
```

With these two, only accounts which are **skipped and contain promoted transactions** are processed. Skipped account with no transactions do not fit the criteria, so no need to increment `skips`.

Kourin1996: Fixed 5386a85520ccf47e44e84fbdb6f4ff9e3f4cd407

dbrajovic (approved):

0xAleksaOpacic (approved):

PR #747, opened 2022-09-21
Added docker setup for local env

begmaroman:
# Description

Added docker and docker-compose setup to run the service locally.

```bash
$ make run-local
...
$ make stop-local
```

# Changes include

# Checklist

## Testing


zivkovicmilos:
Hey @begmaroman,

Thank you for your first contribution to the Edge codebase 🙏 
I've added our DevOps wizards @ZeljkoBenovic and @epikichi to also give feedback.

Please follow the instructions by the CLA bot, so the contributor list is maintained correctly:
https://github.com/0xPolygon/polygon-edge/pull/747#issuecomment-1253563420

docker/local/polygon-edge.sh
```diff
@@ -0,0 +1,27 @@
...
+
+POLYGON_EDGE_BIN=./polygon-edge
+
+case "$1" in
+
+   "init")
+      echo "Generating secrets..."
+      node1id=$("$POLYGON_EDGE_BIN" secrets init --data-dir /data1 | grep Node | awk -F ' ' '{print $4}')
+      node2id=$("$POLYGON_EDGE_BIN" secrets init --data-dir /data2 | grep Node | awk -F ' ' '{print $4}')
+      node3id=$("$POLYGON_EDGE_BIN" secrets init --data-dir /data3 | grep Node | awk -F ' ' '{print $4}')
```
vcastellm: not necessary to store the id here and in the next line, we're not going to use it

docker/local/polygon-edge.sh
```diff
@@ -0,0 +1,27 @@
...
+      node4id=$("$POLYGON_EDGE_BIN" secrets init --data-dir /data4 | grep Node | awk -F ' ' '{print $4}')
+      echo "Secrets have been successfully generated"
+
+      echo "Generating genesis file..."
+      "$POLYGON_EDGE_BIN" genesis --dir /genesis/genesis.json --consensus ibft --ibft-validators-prefix-path test-chain- --bootnode /ip4/127.0.0.1/tcp/10001/p2p/"$node1id" --bootnode /ip4/127.0.0.1/tcp/20001/p2p/"$node2id"
```
vcastellm: Shouldn't this `--ibft-validators-prefix-path test-chain-` be `--ibft-validators-prefix-path data`?

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,77 @@
...
+      context: ../../
+      dockerfile: docker/local/Dockerfile
+    command: ["init"]
+    volumes:
+      - node-1:/data1
+      - node-2:/data2
+      - node-3:/data3
+      - node-4:/data4
+      - genesis:/genesis
+    network_mode: host
```
epikichi: Would it be possible to create a docker network for running this.  Removes the need to expose `libp2p` port on the host machine.  :thinking:

docker/local/polygon-edge.sh
begmaroman: agree, fixed

docker/local/polygon-edge.sh
begmaroman: does not really matter, changed to data

begmaroman:

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,77 @@
+version: '3.5'
+
+services:
+  ## INITIALIZE GENESIS AND SECRETS
+  init:
```
epikichi, line 5: The genesis file outputted seems to be missing or mis-interpolating the appropriate validator keys.

docker/local/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.18-alpine AS builder
+
+WORKDIR /polygon-edge
+
+ADD go.mod go.sum ./
+RUN go mod download
+
+COPY . .
```
epikichi, line 8: Can we also add in a `.dockerignore` to prevent potential binary inclusions during this build process? Helps reduce the room for human errors down the line. :pray:

epikichi (changes requested):
Thanks for the submission! This feature will help provide another reliable quick-start method for testing/sandboxing. :+1: Provided a few change requests out there to help better improve the experience.

docker/local/docker-compose.yml
ZeljkoBenovic: Agreed. Containers should run in an internal network, and only `json-rpc` and maybe `grpc` should be exposed on the host.

vcastellm: That will cause issues in Macos, most of us use Macos, and this in `genesis.json` won't work `/ip4/127.0.0.1/tcp/10001/p2p/` with port mapping, any suggested solution?

ZeljkoBenovic: You could probably use dockers internal dns. Something like `/dns4/node-1/tcp/1478/p2p/`

begmaroman: It would be much more difficult because the bootstrap node address must contain IP which is dynamic for docker networks. You cannot specify just an alias of the container.

docker/local/Dockerfile
begmaroman: Done

docker/local/docker-compose.yml
begmaroman: How did you verify it?

docker/local/docker-compose.yml
begmaroman: `/dns4/node-1/tcp/1478/p2p/` it won't pass an address validation because `node-1` is not an IP.

docker/local/Dockerfile
epikichi: To clarify, I was referring to preventing possible pre-built binaries for any users testing locally as a result of `go build` or `make build`, since those will be overwritten/unused.

docker/local/docker-compose.yml
epikichi: The `dns4` multiaddr header instructs lib2p2 to resolve the hostname rather than using a static IP, which works out great for dynamic network interface scenarios like this. (Ref: https://wiki.polygon.technology/docs/edge/get-started/set-up-ibft-locally/#step-2-prepare-the-multiaddr-connection-string-for-the-bootnode) 

It's not a functional requirement for local, but overall a cleaner adoption.

docker/local/Dockerfile
begmaroman: if you are talking about `polygon-edge` binary, it will be overwritten in the next step.

docker/local/docker-compose.yml
begmaroman: Good point, you are right, it works now. Updated docker-compose.

epikichi: Thanks. To finish up this change, we should be able to use the default respective ports `grpc`, `lib2p2`, and `json-rpc` within the container commands, and port map  `grpc` and `json-rpc` to use on the host. That way, users can send JSON-RPC commands to the nodes. :pray:

CC @vcastellm 

Example:
```yaml   
    command: ["server", "--data-dir", "/data", "--chain", "/genesis/genesis.json", "--grpc-address", "0.0.0.0:9632", "--libp2p", "0.0.0.0:1478", "--jsonrpc", "0.0.0.0:8545", "--seal"]
    ports:
      - '10000:9632'
      - '10002:8545'
```

begmaroman: Done

epikichi (approved):

docker/local/Dockerfile
epikichi: Yes, it will, but its unnecessary extra entropy to prebloat a cache layer. This was was more of a nitpick tbh. :+1:

docker/local/polygon-edge.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+
+POLYGON_EDGE_BIN=./polygon-edge
+
+case "$1" in
+
+   "init")
+      echo "Generating secrets..."
+      node1id=$("$POLYGON_EDGE_BIN" secrets init --data-dir data-1 | grep Node | awk -F ' ' '{print $4}')
```
ZeljkoBenovic: This works fine, but you could use `--json` flag, which will return `json` output so that you wouldn't need to do all this awk shenanigans. 
Not a big issue, but if something changes in the output format, you'd need to rewrite this command.

docker/local/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.18-alpine AS builder
```
ZeljkoBenovic: Not a big deal, but the official supported go version for Polygon Edge is `1.17`.

ZeljkoBenovic (approved):
Great job 🥇   
Left a few comments.

The only thing I would improve here to remove the need to build the binary for every container, as it takes a lot of time to actually start everything.    
It would be much better, if we could build the `polygon-edge` binary just one time, and then use that binary for all containers.

NWM, it's running great, but running it on WSL has some issues and it takes forever to build ...

docker/local/polygon-edge.sh
begmaroman: Good point, done

docker/local/Dockerfile
begmaroman: Done

vcastellm: Topic for another PR but we need to consider upgrading to 1.18

vcastellm (approved):
LGTM

lazartravica (approved):

PR #753, opened 2022-09-22
EVM-25 Create state abstraction boundaries for write-ops in the state…

igorcrevar:
# Description

`State Transition Function` should not call `Commit` from the `Snapshot` interface and should only return the entries in the state that have been updated. Then, as part of another workflow, these entries would be updated in the `merkle trie`.
[Jira Task Link](https://polygon.atlassian.net/browse/EVM-25)

# Changes include

-  New feature (non-breaking change that adds functionality)

# Checklist

-  I have assigned this PR to myself
-  I have added at least 1 reviewer
-  I have added the relevant labels

## Testing

-  I have tested this code with the official test suite

### DIFF
--- a/state/testing.go
+++ b/state/testing.go
@@ -126,7 +126,7 @@ func testDeleteCommonStateRoot(t *testing.T, buildPreState buildPreState) {
 	txn2.SetState(addr1, hash0, hash0)
 	txn2.SetState(addr1, hash1, hash0)
 
-	objs = txn.Commit(false)
+	objs = txn2.Commit(false)
 	snap3, _ := snap2.Commit(objs)
 
 	txn3 := newTxn(state, snap3)
--- a/state/testing.go
+++ b/state/testing.go
@@ -119,13 +119,15 @@ func testDeleteCommonStateRoot(t *testing.T, buildPreState buildPreState) {
 	txn.SetState(addr2, hash1, hash1)
 	txn.SetState(addr2, hash2, hash1)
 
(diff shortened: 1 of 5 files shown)


igorcrevar:

dbrajovic (approved):

state/testing.go
```diff
@@ -332,8 +332,14 @@ func testChangeAccountBalanceToZero(t *testing.T, buildPreState buildPreState) {
...
+	snap, _ = buildNewSnapshot(txn, snap, true)
 
 	txn = newTxn(state, snap)
 	assert.False(t, txn.Exist(addr1))
 }
+
+func buildNewSnapshot(txn *Txn, snapshot Snapshot, deleteEmptyObjects bool) (Snapshot, []byte) {
+	objs := txn.Commit(deleteEmptyObjects)
+
+	return txn.snapshot.Commit(objs)
```
ferranbt: I mentioned today on grooming that you should avoid relying on `txn.snapshot` object to perform write operations because that can be simplified too in a followup PR. Besides, you are already passing a `snapshot` reference to the function that you are not using.

Is this something you can do?
```suggestion
	return snapshot.Commit(objs)
```
Also, I would not put this in a separate function, a one-liner with both things would work for me too.

igorcrevar: Yes, that was just copy paste error. I have removed function also

Kourin1996 (approved):

ferranbt (approved):

PR #769, opened 2022-09-28
Automating e2e tests

vcastellm:
# Description

This is the first of a series of PRs to bring e2e testing to Edge, in this PR we enable automation of running e2e tests with GH actions.

# Changes include

# Checklist

## Testing

# Documentation update

Updated `e2e/README.md`

# Additional comments

One e2e test is failing, now detected because automation, but fixing the test itself is out of the scope of this PR.


vcastellm:
Additional comment: It looks like there's a flaky test in e2e
image

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,23 @@
...
+        with:
+          go-version: 1.17
+      - name: Run tests
+        run: make test-e2e
+      # - name: Archive test logs
+      #   if: always()
+      #   uses: actions/upload-artifact@v3
+      #   with:
+      #     name: e2e-logs
+      #     path: e2e-logs-*/
```
Stefan-Ethernal, line 29: Just checking, we are going to uncomment this once [Add logging into e2e framework](https://polygon.atlassian.net/jira/software/c/projects/EVM/boards/5?modal=detail&selectedIssue=EVM-51) gets implemented?

e2e/README.md
```diff
@@ -22,15 +13,8 @@ Now that the polygon-edge binary is in the `go/bin` folder, the e2e test server
...
-
-While running, the e2e server stores data needed for running in the /tmp/polygon-edge* folders. <br />
-To clean up these folders, simply run:
-
-````bash
-cd /tmp/
-rm -rf polygon-edge-*
-````
+Note the process id: `ps aux | grep polygon-edge`
+Kill the process `kill -9 [pid]`
```
Stefan-Ethernal: Could we implement this in the e2e framework itself or through some command, instead of doing it manually?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,23 @@
+name: E2E tests
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      E2E_TESTS: true
+      E2E_LOGS: true
+      CI_VERBOSE: true
```
Stefan-Ethernal, line 15: Currently we aren't utilizing anywhere in the edge codebase these environment variables. I don't mind leaving them if it doesn't break anything, but just saying. :slightly_smiling_face:

Stefan-Ethernal (commented):
Generally looks good, but posting comments with observations.

ferranbt (approved):
LGTM!

.github/workflows/e2e.yaml
vcastellm: Yup see https://github.com/0xPolygon/polygon-edge/pull/771

.github/workflows/e2e.yaml
vcastellm: Good question, also thought the same but considering that we will be moving all e2e from polybft here too, I opted to leave them here in advance, WDYT?

Makefile
```diff
@@ -51,3 +51,7 @@ run-local:
 stop-local:
 	docker-compose -f ./docker/local/docker-compose.yml stop
 
+.PHONY: test-e2e
+test-e2e:
+	go build -o polygon-edge main.go
+	env EDGE_BINARY=${PWD}/polygon-edge go test -v ./e2e/... -timeout=30m
```
Kourin1996: 1. Shall we move below `test`?
2. This is a suggestion. Can we prevent `test` command from running e2e test by the following command?

```
go test `go list ./... | grep -v e2e`
```

Kourin1996 (commented):
Thank you for PR! I've left only suggestions. Please check and change if you feel need. Thank you

.github/workflows/e2e.yaml
Stefan-Ethernal: Ok great. In that case, I'd consider maybe even removing this part in order to avoid unnecessary merge conflict (although it is pretty straight forward resolving it). WDYT? I'm ok with either approach.

.github/workflows/e2e.yaml
Stefan-Ethernal: Yes sure. I would also consider setting log level to debug (this is something I found useful multiple times on v3, although we are running there on `info` log level, which on some occasions wasn't enough).

Double check, but think this is the correct way to set it:
`E2E_LOG_LEVEL: debug`

Makefile
ferranbt: Though not enabled on this PR. What we did in `v3` was to enable `e2e` only when the `E2E_TESTS` env was set to true.

Stefan-Ethernal:
Also just another general feedback on e2e framework refactoring, which we can decide later on... 

On v3, we have introduced abstractions for cluster and bridge and I'd say that looks better than implementing all those functions, which don't belong to the `TestServer` itself in the `helper.go` ([test-cluster.go](https://github.com/maticnetwork/v3/blob/main/e2e/framework/test-cluster.go), [test-bridge.go](https://github.com/maticnetwork/v3/blob/main/e2e/framework/test-bridge.go) etc.)

zivkovicmilos (approved):
Looks good 💯

.github/workflows/e2e.yaml
vcastellm: Let's go with it as the next PR is branched from this branch so, no conflict

.github/workflows/e2e.yaml
vcastellm: Done in the next PR #771

Makefile
vcastellm: 1. I like it, moved
2. @ferranbt I found that env var switch a bit tricky, I like @Kourin1996's proposal, applied

vcastellm:
@Stefan-Ethernal on that observation, the corresponding abstraction to `test-cluster` here would be https://github.com/0xPolygon/polygon-edge/blob/develop/e2e/framework/ibft.go#L19 and I can move the resolve function up there and pass it as config to the test server, as I did in the #771, I'd say, let's go with this now, and I would propose this refactor in that PR to see how it looks.

e2e/README.md
vcastellm: Adapted docs, WDYT?

Stefan-Ethernal (approved):

Makefile
```diff
@@ -41,7 +41,12 @@ generate-bsd-licenses:
 test:
 	go build -o artifacts/polygon-edge .
 	$(eval export PATH=$(shell pwd)/artifacts:$(PATH))
-	go test -timeout 28m ./...
+	go test -timeout 28m `go list ./... | grep -v e2e`
+
+.PHONY: test-e2e
+test-e2e:
+	go build -o polygon-edge main.go
```
ZeljkoBenovic: Are we testing for race conditions in E2E as well?

ZeljkoBenovic (approved):
Great job. 🥇 
Left a minor comment.

epikichi (approved):

Makefile
vcastellm: Does it makes sense to test for race conditions when testing a binary?

Kourin1996 (approved):

Makefile
ZeljkoBenovic: In our current E2E tests, the binary is built with a `-race` flag and the tests should fail on race condition: https://github.com/0xPolygon/polygon-edge/blob/49b9e68a71252a06a99f05b7f572ca4c5d22e894/.github/workflows/main.yml#L39-L51   

But, maybe, we can have a dedicated test that checks for race conditions? 
What do you @Kourin1996 @zivkovicmilos  guys think ?

vcastellm: We can build with `-race` in the separated e2e tests too.

I've a question: why are you adding both CGO flag and GOOS in the action build?

vcastellm: This should address your concern @ZeljkoBenovic https://github.com/0xPolygon/polygon-edge/pull/769/commits/2107135c9ea2b1e3fec30529ca5f53d4f2cfa947

ZeljkoBenovic: Cool. All good then. 🚀 

The CGO and GOOS is just some legacy thing, which we never bothered to remove 😄

PR #770, opened 2022-09-28
EVM-50: Add --num feature to "secrets init" command

begmaroman:
# Description

In order to generate multiple secrets within one command, there should be `--num` feature flag that accepts an integer value of the number of secrets that should be generated. `--data-dir` flag value is used as a prefix of the directory. 

# Changes include

# Checklist

## Testing

### Manual tests

Tested manually using the local docker setup.

### DIFF
--- a/command/secrets/init/secrets_init.go
+++ b/command/secrets/init/secrets_init.go
@@ -8,7 +8,6 @@ import (
 )
 
 const (
-
 	// maxInitNum is the maximum value for "num" flag
 	maxInitNum = 1000
 )
@@ -116,6 +115,8 @@ func setFlags(cmd *cobra.Command, params *initParams, num *int) {
 	)
 }
 
+// newParamsList creates a list of initParams with num elements.
+// This function basically copies the given initParams but updating dataDir by applying an index.
 func newParamsList(params initParams, num int) []initParams {
 	if num == 1 {
 		return []initParams{params}
--- a/command/secrets/init/secrets_init.go
+++ b/command/secrets/init/secrets_init.go
@@ -3,8 +3,9 @@ package init
 import (
 	"fmt"
 
-	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/spf13/cobra"
+
+	"github.com/0xPolygon/polygon-edge/command"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


epikichi (approved):
LGTM. :+1:

Kourin1996 (approved):

command/secrets/init/secrets_init.go
```diff
@@ -1,25 +1,71 @@
...
 import (
+	"fmt"
+
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/spf13/cobra"
 )
 
+const (
+	// maxInitNum is the maximum value for "num" flag
+	maxInitNum = 1000
```
0xAleksaOpacic: Personally, I would make maxInitNum a much lower value, like 30. But it's optional, so if you don't think we should change this just resolve comment :D

command/secrets/init/secrets_init.go
```diff
@@ -1,25 +1,71 @@
...
+		number      int
+	)
+
 	secretsInitCmd := &cobra.Command{
 		Use: "init",
 		Short: "Initializes private keys for the Polygon Edge (Validator + Networking) " +
 			"to the specified Secrets Manager",
-		PreRunE: runPreRun,
-		Run:     runCommand,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
0xAleksaOpacic: Why not extract this into the separate method:
This is the standard we've been using so far and i think its useful
ibftSnapshotCmd := &cobra.Command{
		Use:     "propose",
		Short:   "Proposes a new candidate to be added or removed from the validator set",
		PreRunE: runPreRun,
		Run:     runCommand,
	}

0xAleksaOpacic (approved):

command/secrets/init/secrets_init.go
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/770/commits/df3173e8ff57e8e2ef4bc4f66b8b4f077087a325

command/secrets/init/secrets_init.go
begmaroman: Fixed: https://github.com/0xPolygon/polygon-edge/pull/770/commits/df3173e8ff57e8e2ef4bc4f66b8b4f077087a325

PR #773, opened 2022-09-30
EVM-48: Add RootChain helper

begmaroman:
# Description

Implement commands from v3: https://github.com/maticnetwork/v3/tree/main/internal/rootchain/cli

# Changes include

# Checklist

## Testing


command/output.go
```diff
@@ -7,34 +7,34 @@ import (
 // OutputFormatter is the standardized interface all output formatters
 // should use
 type OutputFormatter interface {
-	// getErrorOutput returns the CLI command error
-	getErrorOutput() string
-
-	// getCommandOutput returns the CLI command output
-	getCommandOutput() string
-
```
begmaroman, line 15: It does not make sense to have these funcs as part of the interface because they are used within the package only and as helper functions

contracts/smart_contracts/cache/solidity-files-cache.json
```diff
@@ -0,0 +1,408 @@
+{
```
Stefan-Ethernal: Should this be added to gitignore (if it is being produced as a result of contracts compilation)?

contracts/smart_contracts/scripts/sample-script.js
```diff
@@ -0,0 +1,32 @@
+// We require the Hardhat Runtime Environment explicitly here. This is optional
+// but useful for running the script in a standalone fashion through `node <script>`.
+//
+// When running the script with `npx hardhat run <script>` you'll find the Hardhat
+// Runtime Environment's members available in the global scope.
+const hre = require("hardhat");
```
Stefan-Ethernal: Think that we don't need this file, since this is only a small demo on how to use hardhat and ethers js. WDYT?

contracts/smart_contracts/cache/solidity-files-cache.json
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/773/commits/7b2cfb5a2a25d5e6e081b14a50ad87715e17571d

contracts/smart_contracts/scripts/sample-script.js
begmaroman: Agree: https://github.com/0xPolygon/polygon-edge/pull/773/commits/7b2cfb5a2a25d5e6e081b14a50ad87715e17571d

command/rootchain/emit/params.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	contractFlag = "contract"
+	walletsFlag  = "wallets"
+	amountsFlag  = "amounts"
+)
+
+var (
+	errWalletsMissing       = errors.New("wallet flag value is not provided")
+	errAmountsMissing       = errors.New("amount flag value is not provided")
+	errInconsistentAccounts = errors.New("wallets and amounts must be provided in same numbers")
```
zivkovicmilos: What do you think about changing this error message to:
`wallets and amounts must be provided in pairs`?

command/rootchain/emit/rootchain_emit.go
```diff
@@ -0,0 +1,143 @@
...
+	var prms []interface{}
+	prms = append(prms, parameters...)
+
+	wrapperInput, err := abi.MustNewType(paramsType).Encode(prms)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode parsed parameters: %w", err)
+	}
+
+	artifact := smartcontracts.MustReadArtifact("rootchain", "RootchainBridge")
+	method := artifact.Abi.Methods["emitEvent"]
```
zivkovicmilos: These artifact names should be extracted out to constants

vcastellm (commented):
As we've been discussing in Slack, I think this implementation needs first to be described, and discussed, before introducing it here.

My main problem is the `rootchain` command, IMO it should be managed outside the application, as it's part of the infra/environment needed to run the bridge, makes sense to me to separate that, it will be also less code to maintain.

command/rootchain/helper/txn.go
```diff
@@ -0,0 +1,176 @@
...
+		return false
+	}
+
+	return true
+}
+
+func GetPendingNonce(addr types.Address) (uint64, error) {
+	ipAddr := ReadRootchainIP()
+
+	provider, err := jsonrpc.NewClient(ipAddr)
```
Stefan-Ethernal: Not sure how expensive is constantly instantiating new `*jsonrpc.Client`, but I would consider maybe instantiating it only once and cache it somewhere and providing it here as a parameter. WDYT?

Stefan-Ethernal (commented):
Leaving some small suggestion, otherwise looks good.

command/rootchain/emit/rootchain_emit.go
Stefan-Ethernal: To be honest, not sure is it worth fixing it, since it is used just on a single place. Do you agree?

command/rootchain/emit/params.go
Stefan-Ethernal: Fixed [969c204](https://github.com/0xPolygon/polygon-edge/pull/773/commits/969c204c90702980f068399b6d9f3d5d3940fdfb)

Stefan-Ethernal (approved):

command/rootchain/helper/txn.go
begmaroman: done https://github.com/0xPolygon/polygon-edge/pull/773/commits/f4688414332cff3b5f9d54fb1516bcc445f60f16

vcastellm:
I'm advocating for removing the docker infrastructure in this PR, in other terms, not include the `rootchain server` command. This should be infra/environment provided, using `docker-compose` because:

1. It's not the responsibility of the client to maintain the lifecycle of an external service
2. It's more code to maintain while there are specific tools for doing that
3. it's not compatible with all dev platforms, we've had problems in the past with MacOS

I can create an RFC for this if you prefer to discuss that better.

ferranbt (approved):

vcastellm (approved):

command/common_output.go
```diff
@@ -1,14 +1,21 @@
...
+import "sync"
+
 type commonOutputFormatter struct {
+	sync.Mutex
 	errorOutput   error
 	commandOutput CommandResult
 }
 
 func (c *commonOutputFormatter) SetError(err error) {
+	c.Lock()
```
b00ris: I don't see any getError locking? Did you decide to make lock in place?

command/json_output.go
```diff
@@ -20,11 +30,21 @@ func (jo *JSONOutput) WriteOutput() {
...
+}
+
+// WriteOutput implements OutputFormatter plus io.Writer interfaces
+func (jo *jsonOutput) Write(p []byte) (n int, err error) {
+	return os.Stdout.Write(p)
 }
 
-func (jo *JSONOutput) getErrorOutput() string {
+func (jo *jsonOutput) getErrorOutput() string {
+	if jo.errorOutput == nil {
```
b00ris, line 41: Lock?

command/rootchain/helper/txn.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return receipt, nil
+}
+
+func ExistsCode(addr types.Address) bool {
+	provider := getJRPCClient()
+
+	code, err := provider.Eth().GetCode(ethgo.HexToAddress(addr.String()), ethgo.Latest)
+	if err != nil {
+		panic(err)
```
b00ris: Can we return error there?

command/common_output.go
begmaroman: Basically, this is not needed anymore. Removed: https://github.com/0xPolygon/polygon-edge/pull/773/commits/2a959a95d8e88584e7ec7a586b339d7412bc04d4.

command/json_output.go
begmaroman: No needed

command/rootchain/helper/txn.go
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/773/commits/2a959a95d8e88584e7ec7a586b339d7412bc04d4

b00ris (approved):

PR #775, opened 2022-10-03
[FIX] Bootnode re-discovery issue

### LINKED ISSUE #695: Unable to execute bootnode peer discovery
### Description

Sometimes my bootnodes reboot and the validators lose connection with the bootnodes

### Your environment

* 0.4.1_linux_amd64

### Steps to reproduce

Run a network with 2 bootnodes and 4 validators.
After stopping all bootnodes and starting the bootnodes again, the validators cannot connect to the bootnodes

### Expected behaviour

Expected would be that the validators would find the bootnodes again

### Actual behaviour

The validators issue an error that they cannot fetch the validators

### Logs

```
2022-08-24T20:11:11.251Z [ERROR] polygon.network.discovery: Unable to execute bootnode peer discovery, %w: EXTRA_VALUE_AT_END="unable to create new discovery client connection, peer disconnected before the discovery client was initialized"
```

### Proposed solution

I would like validators to be able to connect to bootnode without me running the command to add the peer.

When I add a peer to a validator, the network starts working again.

```
polygon-edge peers add --grpc-address 127.0.0.1:10000 --addr
```

ZeljkoBenovic:
# Description

Right now, in a running network, once the bootnodes lose all of its peers (server restart, network issues or similar) it does not get back any peers, without restarting all the nodes.    
There is a go routine that is supposed to ping random bootnode for the list of its peers, but the bootnodes stay disconnected from the rest of the network.

One can test this very easily, run a 4-node cluster with 1 bootnode. Keep all the nodes running but restart the bootnode. Once the bootnode goes back online, it will never join the network and it will always have 0 peers.

# Changes include

# Checklist

## Testing

### Manual tests

Run a 4-node cluster with 1 bootnode. Restart or power cycle the bootnode only.   
After the bootnode comes back online, after a while (no more than 60s), it should start peering with all the other nodes.

# Additional comments

Fixes EDGE-790


epikichi (approved):
Great catch. Confirmed locally that peers will dial back after bootnode(s) disruption occurs.

Kourin1996 (approved):
LGTM. Thank you for correcting minor things in addition to the fixes

0xAleksaOpacic (approved):
LGTM, tested manually everything and the issue is fixed with this

network/testing/testing.go
```diff
@@ -86,6 +86,8 @@ type getRandomPeerDelegate func() *peer.ID
 type fetchAndSetTemporaryDialDelegate func(peer.ID, bool) bool
 type removeTemporaryDialDelegate func(peer.ID)
 
+func (m *MockNetworkingServer) TemporaryDialPeer(peerAddrInfo *peer.AddrInfo) {}
```
zivkovicmilos: Nitpick:

Can you please introduce a delegate method like we did for other `MockNetworkingServer` methods, and populate the body of this callback, like we did for other methods?

It makes it easier in the future to mock, and is consistent with the way we write mocks

network/server.go
```diff
@@ -329,10 +330,17 @@ func (s *Server) checkPeerConnections() {
...
+				randPeer := s.GetRandomPeer()
+				if randPeer != nil && !s.IsConnected(*randPeer) {
+					s.addToDialQueue(s.GetPeerInfo(*randPeer), common.PriorityRandomDial)
+				}
 			} else {
+				// dial random unconnected bootnode
 				randomNode := s.GetRandomBootnode()
-				s.addToDialQueue(randomNode, common.PriorityRandomDial)
+				if randomNode != nil {
+					s.addToDialQueue(randomNode, common.PriorityRandomDial)
```
zivkovicmilos: You can inline this

zivkovicmilos:
```go
				if randomNode := s.GetRandomBootnode(); randomNode != nil {
					s.addToDialQueue(randomNode, common.PriorityRandomDial)
				}
```

zivkovicmilos (approved):
Looks good 💯 

Thank you for hunting down this bug 🙏 

I've left a few small comments that need to be addressed.

Please also resolve the conflicts in the `go.mod` file 🙏

network/server.go
ZeljkoBenovic: Good catch 🎖️  Fixed in: 1531d1aa1f19ed65d2448230e0b026c1488cac2b

network/testing/testing.go
ZeljkoBenovic: Fixed in 1531d1aa1f19ed65d2448230e0b026c1488cac2b

PR #778, opened 2022-10-06
EVM-69 Fix panic issue in opReturnDataCopy

Kourin1996:
# Description

This PR fixes the issue that may cause panic in opReturnDataCopy. `dataOffset` is expected to be uint64 value but this function doesn't check the negativity before using. This PR adds such a check.

In addition, this PR change the order of checks in `checkMemory` function because it always passes when the `size` argument is zero, even thought `offset` is negative

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/state/runtime/evm/state.go
+++ b/state/runtime/evm/state.go
@@ -272,7 +272,10 @@ func (c *state) Len() int {
 	return len(c.memory)
 }
 
-func (c *state) checkMemory(offset, size *big.Int) bool {
+// allocateMemory allocates memory to enable accessing in the range of [offset, offset+size]
+// throws error if the given offset and size are negative
+// consumes gas if memory needs to be expanded
+func (c *state) allocateMemory(offset, size *big.Int) bool {
 	if !offset.IsUint64() || !size.IsUint64() {
 		c.exit(errGasUintOverflow)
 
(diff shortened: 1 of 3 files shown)


ferranbt:
Should this check be implemented in the other instructions that have a `dataOffset`?

state/runtime/evm/state.go
```diff
@@ -273,16 +273,16 @@ func (c *state) Len() int {
 }
 
 func (c *state) checkMemory(offset, size *big.Int) bool {
```
zivkovicmilos: I know this is not part of this task, but can we please change the name of this method to be something more descriptive, or add a comment explaining what it does (and what it returns)? 🙏

zivkovicmilos (approved):
Looks great 💯 

Thank you for fixing this up, and for adding coverage as well 🙏

Kourin1996:
> Should this check be implemented in the other instructions that have a dataOffset?

Are you asking whether we need to do such a check in other instructions?
Basically `checkMemory` checks the negativities but this instruction didn't check all of values.

I did quick check and other instructions check the negativities.

Kourin1996:
@zivkovicmilos Renamed and added comment 87d7d41e9a629b12f714bb5f0f37c21de2108e3b

0xAleksaOpacic (approved):
LGTM! Thank you for fixing it and adding relevant tests :D

dbrajovic (approved):

PR #785, opened 2022-10-10
Integrate memdb to polybft

goran-ethernal:
# Description

Integrated `memdb` as another layer of persistance for the consensus protocol and bridge data.

`memdb` is used for performance reasons (since it is a in-memory database), building complex queries, and it's enabling of custom indexers and iterators.

# Changes include

# Breaking changes

# Checklist

## Testing


goran-ethernal:

consensus/polybft/state.go
```diff
@@ -51,123 +46,342 @@ const (
...
+		// if there is no upper bound provided, than we return everything
+		return math.MaxUint64
 	}
 
-	id, ok := raw["id"].(*big.Int)
+	upper, ok := upperBound.(uint64)
+
 	if !ok {
-		return nil, fmt.Errorf("failed to decode id field of log: %+v", log)
+		panic("state sync table upper bound not a uint64")
```
b00ris: Why do we need panic there?

goran-ethernal: I'll default it to math.MaxUint64, as in case if upper bound was not provided.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/b35d485640ae64d57ab444197381629b1c1916fe

consensus/polybft/state_test.go
```diff
@@ -0,0 +1,658 @@
...
+	t.Run("Delete state sync events lower or equal to index 4", func(t *testing.T) {
+		err := deleteFilteredFromMemDB(memdb, stateSyncTable, uint64(4))
+		require.NoError(t, err)
+		events, err := getFilteredFromMemDB[*StateSyncEvent](memdb, stateSyncTable, uint64(0), nil)
+		require.NoError(t, err)
+		require.Len(t, events, 5)
+	})
+}
+
+// TODO - uncomment once changes from the integration with v3-contracts get moved to edge
```
JekaMas: Do we need this commented out code?

JekaMas: And more a general question do we have all tests as positive cases in this PR?

consensus/polybft/data_types.go
```diff
@@ -0,0 +1,160 @@
...
+
+func decodeEvent(log *ethgo.Log) (*StateSyncEvent, error) {
+	raw, err := stateTransferEvent.ParseLog(log)
+	if err != nil {
+		return nil, err
+	}
+
+	id, ok := raw["id"].(*big.Int)
+	if !ok {
+		return nil, fmt.Errorf("failed to decode id field of log: %+v", log)
```
JekaMas, line 66: could we introduce a new error or const string for it?

consensus/polybft/consensus_runtime.go
```diff
@@ -400,15 +410,13 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) error {
...
-	if err != nil {
-		if errors.Is(err, ErrNotEnoughStateSyncs) {
-			c.logger.Info("[buildCommitment] Not enough state syncs to build a commitment",
-				"epoch", epoch, "from state sync index", fromIndex)
-			// this is a valid case, there is not enough state syncs
-			return nil, nil
-		}
-
+	stateSyncEvents, err := c.state.getStateSyncEventsForCommitment(fromIndex, fromIndex+stateSyncMainBundleSize-1)
+	if len(stateSyncEvents) == 0 || (err != nil && errors.Is(err, errNotEnoughStateSyncs)) {
```
JekaMas: I believe we don't log an error here. How do you think does it worth to distinguish this 2 cases with len(stateSyncEvents) == 0 and errors.Is() ?

consensus/polybft/consensus_runtime.go
```diff
@@ -277,26 +271,46 @@ func (c *consensusRuntime) FSM() (*fsm, error) {
...
-				if err != nil {
-					return nil, err
+			var bundleProofs []*BundleProof
+
+			var commitments []*CommitmentMessageSigned
+
+			processedBundles := 0
+			foundGap := false
+
+			for _, c := range nonExecutedCommitments {
```
JekaMas, line 281: do we have tests for it?

consensus/polybft/state.go
```diff
@@ -51,123 +46,342 @@ const (
...
-	parentBuckets = [][]byte{syncStateEventsBucket, commitmentsBucket, bundlesBucket,
-		epochsBucket, validatorSnapshotsBucket}
-	// ErrNotEnoughStateSyncs error message
-	ErrNotEnoughStateSyncs = errors.New("there is either a gap or not enough sync events")
-	// ErrCommitmentNotBuilt error message
-	ErrCommitmentNotBuilt = errors.New("there is no built commitment to register")
-)
+// getFilteredFromMemDB returns a filtered collection of desired memdb records
+func getFilteredFromMemDB[V MemDBRecord](memdb *memdb.MemDB, table string,
+	lowerBound, upperBound interface{}) ([]V, error) {
```
JekaMas: why do we want to use lowerBound and upperBound as empty interfaces?

JekaMas (commented):
just few comments to discuss. overall looks good to me

consensus/polybft/state_test.go
goran-ethernal: We can remove it from this PR, since it was added through integration with v3-contracts. We can always add it later, once we integrate with them on edge as well.

As for the second question. Yeah, we only have positive cases tests in the PR. I'll add some negative ones also, and leave a commit link for you to review.

consensus/polybft/consensus_runtime.go
goran-ethernal: Yes, this should be logged, you are right.

goran-ethernal: Yes, this error should definitely be returned.

Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/e0dc1bd33b0667bdac9a1142a52fb84f0fde749f

consensus/polybft/state_test.go
goran-ethernal: No. This test was added when we integrated with v3-contracts. I'll remove it for now, and we can add it once we integrate those changes to edge as well.

Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/a01ff43ddb674d1acd647a8355061503ae8b202f

consensus/polybft/state.go
goran-ethernal: It's a generic function, and lower and upper bounds are some fields on the struct we are querying from db. 

For example, we store this in memdb:

```
type StateSyncEvent struct {
   Id uint64
   ...
 }
```
 If we want to get state sync events from Id 100 - 200, the lower bound will be 100, upper bound will 200 (Id field on struct).

consensus/polybft/consensus_runtime.go
goran-ethernal: Yes, but those tests need to be integrated in edge through a separate PR, once we get the v3-contracts integration to edge as well.

consensus/polybft/state.go
JekaMas: `lowerBound, upperBound interface{}` - why not to have them as generics also? like numeric type?

consensus/polybft/consensus_runtime.go
JekaMas: got it, thank you!

consensus/polybft/state_test.go
JekaMas: I believe we need then a task to have negative tests also. How do you think?

consensus/polybft/state.go
goran-ethernal: Yep, you are right. I'll change that to be generics as well.

consensus/polybft/state_test.go
goran-ethernal: What do you think if I add some negative cases through this PR? It won't take me long to write them.

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/7b91a8734e719e42fac631fa4abd0bd629ae04ce

consensus/polybft/state_test.go
goran-ethernal: Added some new negative cases. There were some already implemented, but I added those which were missing:
https://github.com/0xPolygon/polygon-edge/pull/785/commits/6ef13e95425c38452637ec904bc1a0029ae5d334

JekaMas (approved):
Looks great to me.

consensus/polybft/consensus_runtime.go
```diff
@@ -400,15 +410,13 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) error {
...
-				"epoch", epoch, "from state sync index", fromIndex)
-			// this is a valid case, there is not enough state syncs
-			return nil, nil
-		}
-
+	stateSyncEvents, err := c.state.getStateSyncEventsForCommitment(fromIndex, fromIndex+stateSyncMainBundleSize-1)
+	if len(stateSyncEvents) == 0 {
+		c.logger.Info("[buildCommitment] Not enough state syncs to build a commitment",
+			"epoch", epoch, "from state sync index", fromIndex)
+		// this is a valid case, there is not enough state syncs
```
Stefan-Ethernal: I'd just switch these two lines, for the sake of consistency.
```suggestion
                // this is a valid case, there is not enough state syncs
		c.logger.Info("[buildCommitment] Not enough state syncs to build a commitment",
			"epoch", epoch, "from state sync index", fromIndex)
```

consensus/polybft/state.go
```diff
@@ -51,123 +46,350 @@ const (
...
+					Indexer: &memdb.UintFieldIndex{Field: "Epoch"},
+				},
+			},
+		},
+	},
+}
+
+func getUpperBoundUint64(upperBound interface{}) uint64 {
+	if upperBound == nil {
+		// if there is no upper bound provided, than we return everything
```
```suggestion
		// if there is no upper bound provided, then we return everything
```

consensus/polybft/state.go
```diff
@@ -51,123 +46,350 @@ const (
...
-	if !ok {
-		return nil, fmt.Errorf("failed to decode sender field of log: %+v", log)
+	return upper
+}
+
+var iterators = map[string]func(it memdb.ResultIterator, upperBound interface{}, first MemDBRecord) MemDBIterator{
+	stateSyncTable: func(it memdb.ResultIterator, upperBound interface{}, first MemDBRecord) MemDBIterator {
+		event, isOk := first.(*StateSyncEvent)
+		if !isOk {
+			panic("element not of type StateSyncEvent")
```
```suggestion
			panic("failed to create state sync events table iterator, because element is not of type StateSyncEvent")
```

consensus/polybft/state.go
```diff
@@ -51,123 +46,350 @@ const (
...
+
+		return res, nil
 	}
 
-	data, ok := raw["data"].([]byte)
-	if !ok {
-		return nil, fmt.Errorf("failed to decode data field of log: %+v", log)
+	obj, isOk := nextItem.(MemDBRecord)
+	if !isOk {
+		return nil, errors.New("item not a MemDBRecord")
```
Stefan-Ethernal: Maybe you could extract this error as well to package-level variable as you did with all the others?

Stefan-Ethernal (approved):
Left some minor suggestions, but otherwise looks good.

consensus/polybft/state_test.go
```diff
@@ -0,0 +1,614 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/umbracle/ethgo"
+)
+
+func newTestState(t *testing.T) *State {
+	t.Helper()
+
+	dir := fmt.Sprintf("/tmp/consensus-temp_%v", time.Now().Format(time.RFC3339))
+
```
ferranbt: Can we remove these extra lines over here?

consensus/polybft/state_test.go
```diff
@@ -0,0 +1,614 @@
...
+func TestState_insertAndGetValidatorSnapshot(t *testing.T) {
+	epoch := uint64(1)
+
+	state := newTestState(t)
+
+	keys, err := bls.CreateRandomBlsKeys(3)
+
+	require.NoError(t, err)
+
+	snapshot := AccountSet{
```
ferranbt: Didn't we have helper functions for this?

consensus/polybft/data_types.go
```diff
@@ -0,0 +1,160 @@
+package polybft
```
ferranbt: Can we move this back to `state.go`?

ferranbt (changes requested):

consensus/polybft/state_test.go
goran-ethernal: I think linter was unhappy about this for some reason, and that is why I had to separate them with new lines. I'll try to remove them, and see if it will report something.

consensus/polybft/data_types.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/56b13fd4757f4df9028d7c70d695afd781f389a6

consensus/polybft/state_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/e6ea05a82b32ac463830791a0ab9cf661f5fa87c

consensus/polybft/state_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/785/commits/e6ea05a82b32ac463830791a0ab9cf661f5fa87c

PR #789, opened 2022-10-11
Build block with Txs in PolyBFT

igorcrevar:
# Description

Include transactions from tx pool inside proposal.
Figure out the wiring for the building the block from IBFT

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

After generating polybft secrets with `--private switch` and starting 4 validators create js file with following node js code
and execute it

```
var Web3 = require("web3");

// place correct port here from your node
const web3 = new Web3("ws://localhost:30002/ws"); //example: ws://localhost:10002/ws
web3.eth.accounts
    .signTransaction(
        {
            to: "0x9d4042B5F03C89f76F1963d741d744014acFD20F",
            value: web3.utils.toWei("2"),
            gas: 21000,
            nonce: 0, // update each time with +1
        },
       // place here private key form any node which is outputted after generating secrets
        "29fe602f7723dd54ef908d06b9644ffe2a83a53d72ecc87a5a3643ba0092db9b"
    )
    .then((signedTxData) => {
        web3.eth
            .sendSignedTransaction(signedTxData.rawTransaction)
            .on('transactionHash', function(hash){
  		console.log("hash: ", hash);
	    })
	    .on('receipt', function(receipt){
	        console.log("receipt: ", receipt);
	    })
	    .on('confirmation', function(confirmationNumber, receipt){
	        console.log("confirmation: ", confirmationNumber, receipt);
	    })
	    .on('error', console.error);
    });
  ```

Script to start cluster of 4 polybft nodes:
```
#!/usr/bin/env bash
clear
set -e

# Reset test-dirs
rm -rf test-chain-*
rm -f genesis.json

go run ./main.go polybft-secrets --data-dir test-chain- --num 4 # --private true

go run ./main.go genesis --consensus polybft --block-gas-limit 10000000 --validator-set-size 4 --epoch-size 10 --premine-validators 1000000000000000000000 

go run ./main.go server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :10000 --libp2p :30301 --jsonrpc :10002 --seal --log-level debug &
go run ./main.go server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :20000 --libp2p :30302 --jsonrpc :20002 --seal --log-level debug &
go run ./main.go server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :30000 --libp2p :30303 --jsonrpc :30002 --seal --log-level debug &
go run ./main.go server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :40000 --libp2p :30304 --jsonrpc :40002 --seal --log-level debug &
wait
```

### CODE REVIEW

File: consensus/polybft/consensus_runtime.go [hunk 1]
```diff
@@ -39,6 +39,18 @@ type Transport interface {
...
+// txPoolInterface is an abstraction of transaction pool
+type TxPoolInterface interface {
+	Prepare()
+	Length() uint64
+	Peek() *types.Transaction
+	Pop(tx *types.Transaction)
+	Drop(tx *types.Transaction)
+	Demote(tx *types.Transaction)
+	ResetWithHeaders(headers ...*types.Header)
+	SetSealing(bool)
```
Stefan-Ethernal: Seems like we are not using these functions in the polybft, so maybe we should remove those from this interface?

consensus/polybft/consensus_runtime.go
```diff
@@ -39,6 +39,18 @@ type Transport interface {
 	Gossip(message interface{})
 }
 
+// txPoolInterface is an abstraction of transaction pool
+type TxPoolInterface interface {
```
Stefan-Ethernal: Seems like there is no need to export this interface (it is solely used by `consensus/polybft` package)

consensus/polybft/block_builder.go
```diff
@@ -234,29 +203,25 @@ write:
...
 
 		// continue processing
 		return fail, true
 	}
 
-	if err := b.params.Transition.Write(tx); err != nil {
+	if err := b.state.Write(tx); err != nil {
 		if _, ok := err.(*state.GasLimitReachedTransitionApplicationError); ok { //nolint:errorlint
 			// stop processing
 			return nothing, false
```
Stefan-Ethernal: This is a little bit strange. Shouldn't we create some `txExecutionResult` struct and return `nil` here instead. That way we wouldn't have `nothing` as tx status. Or if we are going to continue with this approach, then naming could be better as: `unknown` or something like that.

Also it is a little bit strange that we are returning multiple values (enum and bool) to determine behavior, that's one more argument why it seems like we should abstract this.

consensus/polybft/fsm.go
```diff
@@ -20,7 +20,7 @@ var _ pbft.Backend = &fsm{}
 type blockBuilder interface {
 	Reset()
 	// CommitTransaction(tx *types.Transaction) error
-	// Fill(ctx context.Context) error
+	Fill() (int, int, int, bool)
```
Stefan-Ethernal: Feel that these 4 return values should be abstracted to some struct. WDYT?
Also are you interested in all these values? As I can see it is used just for returning an error indication, but that doesn't seem like an essential data to me.

command/genesis/genesis.go
```diff
@@ -179,6 +179,15 @@ func setFlags(cmd *cobra.Command) {
 			[]string{},
 			"validators defined by user throughout a parameter (format: <address>:<blskey>)",
 		)
+		cmd.Flags().StringVar(
+			&params.premineValidators,
+			premineValidatorsFlag,
+			command.DefaultPremineBalance,
+			"the amount which will be premined for all the validators",
```
Stefan-Ethernal:
```suggestion
			"the amount which will be premined to all the validators",
```

consensus/polybft/fsm.go
```diff
@@ -143,37 +155,30 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
...
 		Data: stateBlock.Block.MarshalRLP(),
 		Hash: f.block.Block.Hash().Bytes(),
 	}
-	f.proposal = proposal
-	f.logger.Debug("[FSM Build Proposal]", "Proposal hash", hex.EncodeToHex(proposal.Hash))
 
-	return proposal, nil
+	f.logger.Debug("[FSM Build Proposal]",
+		"txs", len(stateBlock.Block.Transactions),
+		"time", time.Now(),
```
Stefan-Ethernal: Why do we care about time? We have time in the log anyways, right?

consensus/polybft/blockchain_wrapper.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/state"
 	"github.com/0xPolygon/polygon-edge/types"
+	hcf "github.com/hashicorp/go-hclog"
```
Stefan-Ethernal: You could use even default `hclog` alias, not much benefits naming it `hcf`, but that's minor thing.

consensus/polybft/block_builder.go
```diff
@@ -148,23 +133,9 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*StateBlock, error)
...
-	// calculate gas limit based on parent header
-	//gasLimit, err := b.blockchain.CalculateGasLimit(header.Number)
-	// if err != nil {
-	// 	return nil, err
-	// }
-	// header.GasLimit = gasLimit
-
-	// Nemanja - fill transactions
-	// txs := i.writeTransactions(gasLimit, header.Number, transition)
+	// Txs are filled outside - from FSM
```
Stefan-Ethernal: I'd consider removing this comment.

consensus/polybft/blockchain_wrapper.go
```diff
@@ -86,13 +91,17 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 	// TODO: Call validate block in polybft
```
Stefan-Ethernal, line 93: What about this TODO? Could it be removed?

0xSasaPrsic: This is a leftover from v3 code.

command/genesis/genesis.go
```diff
@@ -179,6 +179,15 @@ func setFlags(cmd *cobra.Command) {
 			[]string{},
 			"validators defined by user throughout a parameter (format: <address>:<blskey>)",
 		)
+		cmd.Flags().StringVar(
+			&params.premineValidators,
+			premineValidatorsFlag,
+			command.DefaultPremineBalance,
+			"the amount which will be premined to all the validators",
+		)
```
zivkovicmilos, line 187: Why do you need a special flag for this?

There is already a `premine` flag that you can utilize

consensus/polybft/block_builder.go
```diff
@@ -81,7 +53,8 @@ func NewBlockBuilder(params *BlockBuilderParams) *BlockBuilder {
 	}
 
 	if params.BlockTime == 0 {
-		params.BlockTime = time.Second * 2 // TODO: is this ok?
+		// default block time to 2 seconds, if it is not provided with the parameters
+		params.BlockTime = 2 * time.Second
```
zivkovicmilos: Why not define this duration in a constant?

zivkovicmilos (changes requested):
I don't believe we need to have the `premine-validators` flag, since the `premine` flag does the exact same functionality for arbitrary addresses.

Let's please discuss this before merging this out 🙏 

cc @lazartravica @vcastellm @Kourin1996

command/genesis/genesis.go
vcastellm: I wonder the same question, but the premine flag should be in the `format: <address>:<balance>` it's for specific addresses but not the validators addresses.

Do you have a suggestion on how to reuse the existing premine flag?

zivkovicmilos: I need to know the context of why the validators need to have a premined balance in the first place?
If it's orchestration, it can be handled outside of Edge

igorcrevar: Lets agree to remove this as soon as we finish all the child tasks if not needed. It makes our life easier while transferring code from v3 to the edge because we do not need to write any additional scripts

vcastellm (approved):
LGTM overall, but something weird, sending some transactions to it, it look like it's a but slow including transactions in blocks, do you know why?

`avg=6.9s`

igorcrevar:
> LGTM overall, but something weird, sending some transactions to it, it look like it's a but slow including transactions in blocks, do you know why?
> 
> `avg=6.9s`

didn't notice that, maybe the tx propagation or something? we can investigate that after everything is finished

Stefan-Ethernal (approved):

ferranbt (approved):

PR #794, opened 2022-10-12
Updating code scans to use Snyk for aligning with org.

epikichi:
# Description

Looking to update code scans use Snyk and integrate with Polygon's Snyk organization. Using default values for now as a baseline.

# Changes include

CI Update


vcastellm (approved):
LGTM

PR #796, opened 2022-10-13
POS v3 smart contracts integration

Stefan-Ethernal:
# Description

This PR integrates [v3 smart contracts ](https://github.com/maticnetwork/v3-contracts) for sidechain.
V3 smart contracts are introduced to Edge codebase as git submodule.

# Changes include

# Checklist

## Testing

#### Setup
initialize core-contracts git submodule
```go
git submodule init
git submodule update
```

compile core-contracts
```
make compile-core-contracts
```

#### Sidechain contracts integration workflow
1. Initialize test accounts
```go
go run . polybft-secrets --data-dir test-chain- --num 4
```

2. Initialize genesis configuration
```go
go run . genesis --block-gas-limit 10000000 --validator-set-size=4 --consensus polybft --epoch-size 10
```

3. Run validator nodes
```go
go run . server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :5001 --libp2p :30301 --jsonrpc :9545 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :5002 --libp2p :30302 --jsonrpc :10002 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :5003 --libp2p :30303 --jsonrpc :10003 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :5004 --libp2p :30304 --jsonrpc :10004 --seal --log-level DEBUG
```

4. Blocks are being produced without any errors.

#### Register new validator workflow
**Note:** Validators from above still need to be up and running before running following steps

1. Create new account
```go
go run . polybft-secrets --data-dir test-chain-5
```
2. Register validator
```go
go run . register-validator --data-dir test-chain-5 --registrator-data-dir test-chain-1 --jsonrpc <json_rpc_addr>
```

#### Rootchain workflow
1. start rootchain server
```go
go run . rootchain server
```
2. deploy and initialize rootchain contracts
```go
go run . rootchain init-contracts
```
3. run emit command
```go
go run . rootchain emit --wallets <existing addresses of accounts> --amounts 100
```


.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/ethereum/tests.git 
 [submodule "v3-contracts"]
 	path = v3-contracts
-	url = https://github.com/maticnetwork/v3-contracts.git
+	url = https://github.com/maticnetwork/v3-contracts
```
vcastellm: Use the new repo https://github.com/0xPolygon/core-contracts?

Makefile
```diff
@@ -64,5 +61,5 @@ stop-local:
 
 .PHONY: compile-v3-contracts
 compile-v3-contracts:
-	cd consensus/polybft/polybftcontracts && npm install && npm run compile
+	cd v3-contracts && npm install && npm run compile && cd ..
```
vcastellm: Rename to `core-contracts`?

.gitmodules
Stefan-Ethernal: Yep will do. That is the next step. 👍

consensus/polybft/polybft_config.go
```diff
@@ -46,9 +48,13 @@ func (p *PolyBFTConfig) IsBridgeEnabled() bool {
...
+	Balance *big.Int `json:"balance"`
 }
 
 // DebugConfig is a struct used for test configuration in init genesis
 type DebugConfig struct {
 	ValidatorSetSize uint64 `json:"validatorSetSize"`
 }
+
+// PolyBFTRegisterMessage is a challenge message which needs to be signed by each validator account
+var PolyBFTRegisterMessage = "Polybft validator"
```
vcastellm: const?

Makefile
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/796/commits/54880bbd64b294f330dcbfbb334a64c1b54a4245

.gitmodules
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/commit/54880bbd64b294f330dcbfbb334a64c1b54a4245

consensus/polybft/polybft_config.go
Stefan-Ethernal: Fixed [8783d52](https://github.com/0xPolygon/polygon-edge/pull/796/commits/8783d52f047ab0065341bfbc981391ce37e0a20a)

command/rootchain/initcontracts/rootchain_initcontracts.go
```diff
@@ -0,0 +1,321 @@
+package initcontracts
```
vcastellm: Why namespace this file name with `rootchain_`? same for server

types/rlp_unmarshal.go
```diff
@@ -361,11 +361,23 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
 
-	// Type
 	if len(elems) >= 10 {
+		// Type
 		if t.Type, err = ReadRlpTxType(elems[9]); err != nil {
 			return err
 		}
+
+		if t.IsStateTx() {
+			// From
```
vcastellm, line 373: Can you extend the comment with more info on why we're doing this address assignment?

crypto/crypto.go
```diff
@@ -250,6 +259,30 @@ func Keccak256(v ...[]byte) []byte {
...
+
+	return hash
+}
+
+// NewKeccakState creates a new KeccakState
+func NewKeccakState() KeccakState {
+	keccakState, isOk := sha3.NewLegacyKeccak256().(KeccakState)
+	if !isOk {
+		// this is only for linter, this will never happen
+		panic("newLegacyKeccak256 does not implement KeccakState interface")
```
vcastellm: Weird to introduce a panic here specially if it's not going to happen, can we just use linting exclusion directive?

command/rootchain/initcontracts/rootchain_initcontracts.go
Stefan-Ethernal: Just to make sure I've understood correctly:
you are suggesting naming package as `rootchain_initcontracts`, `rootchain_server`, `rootchain_emit` and so on?

state/runtime/precompiled/precompiled.go
```diff
@@ -177,3 +197,14 @@ func extendByteSlice(b []byte, needLen int) []byte {
 
 	return b[:needLen]
 }
+
+// abiBoolMustEncode encodes the given value using the given ABI type.
+// panics if there is an error occurred.
+func abiBoolMustEncode(v bool, t *abi.Type) []byte {
```
vcastellm: IDK if the second parameter makes sense, we can simply do `abi.MustNewType("bool")` as the declared var previously it's not used anywhere else.

QEDK:
Just a note that the contract repository has now moved into this org.

Makefile
```diff
@@ -41,28 +40,25 @@ generate-bsd-licenses:
...
 stop-local:
 	docker-compose -f ./docker/local/docker-compose.yml stop
 
-.PHONY: compile-v3-contracts
-compile-v3-contracts:
-	cd consensus/polybft/polybftcontracts && npm install && npm run compile
-
+.PHONY: compile-core-contracts
+compile-core-contracts:
+	cd core-contracts && npm install && npm run compile && cd ..
```
Kourin1996: I think the current directory is reverted after `make` command. Or you can wrap with parentheses like `(npm install && npm run compile)

command/e2e/register_validator.go
```diff
@@ -0,0 +1,462 @@
...
+		"the directory name where new validator key is stored",
+	)
+}
+
+func runPreRun(_ *cobra.Command, _ []string) error {
+	return params.validateFlags()
+}
+
+func runCommand(cmd *cobra.Command, _ []string) error {
+	secretsManager, err := secretsHelper.SetupLocalSecretsManager("./test-chain-1")
```
Kourin1996: I think this directory name should be changeable from CLI.

Stefan-Ethernal:
> Just a note that the contract repository has now moved into this org.

Yep, we've already accommodated it: [54880bb](https://github.com/0xPolygon/polygon-edge/pull/796/commits/54880bbd64b294f330dcbfbb334a64c1b54a4245). Tnx for reminding anyways 🙂

contracts/system_addresses.go
```diff
@@ -0,0 +1,25 @@
...
+	NativeTokenContract = types.StringToAddress("0x1010")
+	// SystemCaller is address of account, used for system calls to smart contracts
+	SystemCaller = types.StringToAddress("0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE")
+
+	// NativeTransferPrecompile is an address of native transfer precompile
+	NativeTransferPrecompile = types.StringToAddress("0x2020")
+	// BLSAggSigsVerificationPrecompile is an address of BLS aggregated signatures verificatin precompile
+	BLSAggSigsVerificationPrecompile = types.StringToAddress("0x2030")
+	// ConsolePrecompile is and address of Hardhat console precompile
+	ConsolePrecompile = types.StringToAddress("0x000000000000000000636F6e736F6c652e6c6f67")
```
Kourin1996, line 24: Just curious. Is it required?

state/runtime/precompiled/precompiled.go
goran-ethernal: Agreed. Don't need it.

command/genesis/genesis.go
```diff
@@ -185,6 +185,13 @@ func setFlags(cmd *cobra.Command) {
 			command.DefaultPremineBalance,
 			"the amount which will be premined to all the validators",
 		)
+		cmd.Flags().StringVar(
+			&params.smartContractsRootPath,
+			smartContractsRootPathFlag,
```
Kourin1996, line 191: Right now, is this only for `polybft` right?

contracts/system_addresses.go
Stefan-Ethernal: Not sure what are you referring to? Is it about this `ConsolePrecompile` we've added, or is it in general these addresses which we have put there as global variables?

Anyways I'll try to guess that you were referring to `ConsolePrecompile`:
it must be strictly deployed to the given address, because it intercepts Hardhat console logs (which turned out to be pretty handy when troubleshooting smart contracts execution). :slightly_smiling_face:

command/genesis/genesis.go
Stefan-Ethernal: Yes it is. Btw this question triggered me to think is it possible to "interlock" flags and expose them based on values of some other flags (e.g. if `--consensus` flag value is `polybft`, make polybft specific flags available, otherwise don't).

command/rootchain/initcontracts/rootchain_initcontracts.go
vcastellm: no, the other way around, naming the files `initcontracts` `server` `emit` 😄

state/runtime/precompiled/precompiled.go
Stefan-Ethernal: Fixed [97f3d78](https://github.com/0xPolygon/polygon-edge/pull/796/commits/97f3d78fdf9f276162c65df6b07b1d855b190e3d)

crypto/crypto.go
Stefan-Ethernal: Fixed [97f3d78](https://github.com/0xPolygon/polygon-edge/pull/796/commits/97f3d78fdf9f276162c65df6b07b1d855b190e3d)

types/rlp_unmarshal.go
Stefan-Ethernal: Fixed [97f3d78](https://github.com/0xPolygon/polygon-edge/pull/796/commits/97f3d78fdf9f276162c65df6b07b1d855b190e3d)

command/rootchain/initcontracts/rootchain_initcontracts.go
Stefan-Ethernal: Fixed [97f3d78](https://github.com/0xPolygon/polygon-edge/pull/796/commits/97f3d78fdf9f276162c65df6b07b1d855b190e3d)

Makefile
Stefan-Ethernal: Fixed [97f3d78](https://github.com/0xPolygon/polygon-edge/pull/796/commits/97f3d78fdf9f276162c65df6b07b1d855b190e3d)

command/e2e/register_validator.go
Stefan-Ethernal: Fixed [77f4f67](https://github.com/0xPolygon/polygon-edge/pull/796/commits/77f4f67a5fbcee764cf82c8540fc0a685375e643)

command/genesis/genesis.go
vcastellm: It's possible to make flags dependent on each other `MarkFlagsRequiredTogether("username", "password")`

igorcrevar: I think Stefan meant that it should not be possible to see/use/anything flags which belongs to some consensus if other consensus is set by `--consensus` flag

vcastellm: The thing is that you can not hide or show flags if using `--help` for example, this kind of behaviour is not very common.

consensus/polybft/consensus_runtime.go
```diff
@@ -231,6 +236,10 @@ func (c *consensusRuntime) populateFsmIfBridgeEnabled(
 	}
 
 	if isEndOfSprint {
+		if err != nil {
```
igorcrevar: This error check should be removed

command/rootchain/initcontracts/init_contracts.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/0xPolygon/polygon-edge/command/rootchain/helper"
+	"github.com/0xPolygon/polygon-edge/contracts"
+	"github.com/0xPolygon/polygon-edge/helper/hex"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+var (
+	params initContractsParams
+
+	initCheckpointManager, _ = abi.NewMethod("function initialize(" +
```
b00ris, line 28: Is there an error check?
If yes - could you move initialisation to init function and add error checking?

consensus/polybft/fsm.go
```diff
@@ -195,7 +196,14 @@ func (f *fsm) stateTransactions() []*types.Transaction {
 			}
 
 			txns = append(txns,
-				createStateTransactionWithData(f.config.SidechainBridgeAddr, inputData))
+				createStateTransactionWithData(f.config.StateReceiverAddr, inputData))
+
+			// since proposer does not execute Validate (when we see the commitment to register in state transactions)
```
igorcrevar: This assignment should be removed. Instance field is changed outside this function, this function should be used just to retrieve all the state transactions

consensus/polybft/block_builder.go
```diff
@@ -194,17 +195,8 @@ func (b *BlockBuilder) writeTxPoolTransaction(tx *types.Transaction) (bool, erro
 		return true, nil
 	}
 
-	if tx.ExceedsBlockGasLimit(b.params.GasLimit) {
```
b00ris, line 197: Don't we need this check?

consensus/polybft/consensus_runtime.go
b00ris: Why do we need this err check there?

consensus/polybft/block_builder.go
igorcrevar, line 197: This probably should be reverted to the old code. This way we lose `b.params.TxPool.Drop(tx)` if tx could not be written to the block

consensus/polybft/contracts_initializer.go
```diff
@@ -0,0 +1,79 @@
...
+	"math/big"
+
+	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	"github.com/0xPolygon/polygon-edge/contracts"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/umbracle/ethgo/abi"
+)
+
+var (
+	initCallStaking, _ = abi.NewMethod("function initialize(" +
```
b00ris, line 23: Could you add err check there?

consensus/polybft/block_builder.go
```diff
@@ -221,7 +213,6 @@ func (b *BlockBuilder) writeTxPoolTransaction(tx *types.Transaction) (bool, erro
 
 	// remove tx from the pool and add it to the list of all block transactions
 	b.params.TxPool.Pop(tx)
-	b.txns = append(b.txns, tx)
```
igorcrevar, line 224: I think we gonna need this because we lose some functionality if we simple call `err := b.WriteTx(tx)` (mentioned in previous comment)

consensus/polybft/contracts_initializer.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, err
+	}
+
+	return input, nil
+}
+
+func buildRegisterMessage(m string) [2]*big.Int {
+	data, err := bls.MarshalMessageToBigInt([]byte(m))
+	if err != nil {
+		panic("error building register message: " + err.Error())
```
b00ris: Can we get rid of panic there?

consensus/polybft/state_transaction.go
```diff
@@ -246,45 +263,23 @@ func (cm *CommitmentMessage) ContainsStateSync(stateSyncIndex uint64) bool {
 
 // BundlesCount calculates bundles count contained in given CommitmentMessge
 func (cm *CommitmentMessage) BundlesCount() uint64 {
-	return (cm.ToIndex - cm.FromIndex + cm.BundleSize) / cm.BundleSize
+	return uint64(math.Ceil((float64(cm.ToIndex) - float64(cm.FromIndex) + 1) / float64(cm.BundleSize)))
```
igorcrevar: I am not sure if original code is so hard to understand that we need to change it? It just simple integer divide arithmetics

types/rlp_marshal.go
```diff
@@ -194,5 +194,9 @@ func (t *Transaction) MarshalRLPWith(arena *fastrlp.Arena) *fastrlp.Value {
 		vv.Set(arena.NewBytes([]byte{byte(t.Type)}))
 	}
 
+	if t.IsStateTx() {
```
igorcrevar, line 197: Maybe we should serialize `from` only if it is not equal to `contracts.SystemCaller` and in that way save `20` bytes per state transaction

consensus/polybft/state_transaction.go
goran-ethernal: Yeah, this can be changed to the original.

consensus/polybft/fsm.go
```diff
@@ -108,6 +109,14 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
 		return nil, err
 	}
 
+	// set the timestamp
```
igorcrevar: Do we need to move `headerTime` assignment above `extraData` assignment? It's strange because `extraParent, err := GetIbftExtra(parent.ExtraData)` is kept at its place

consensus/polybft/block_builder.go
goran-ethernal: I think this is good, because, the same workflow is preserved. If an error happens in `WriteTxn` it will be handled in this function. As you see, the bottom else inside `if err != nil` block, drops the transaction.

consensus/polybft/block_builder.go
goran-ethernal: If `WriteTx` is successful (there is no error), it appends the transaction to the block builder. No need to have it in this function.

consensus/polybft/block_builder.go
goran-ethernal: It's inside `WriteTx`.

consensus/polybft/consensus_runtime.go
goran-ethernal: It's an error left by rebase and merge, will remove it.

Makefile
Kourin1996: So sorry. I made wrong suggestion. The correct command will be `(cd core-contracts && npm install && npm run compile)` to revert current directory after the command inside parentheses.

contracts/system_addresses.go
Kourin1996: Yes, I meant contract. I was not sure why you needed it. Thanks

consensus/polybft/contracts_initializer.go
igorcrevar: This can throw error only if we don't specify method signature correctly (so it can't in our case). Maybe we should add some linter flag or something?

command/genesis/polybft_params.go
```diff
@@ -127,6 +136,7 @@ func (p *genesisParams) getGenesisValidators(validators []GenesisTarget) (result
 			result = append(result, &polybft.Validator{
 				Address: types.StringToAddress(parts[0]),
 				BlsKey:  parts[1],
+				Balance: big.NewInt(1000), // hard-coded balance
```
igorcrevar: Can we specify these hardcoded values in some const or at least var (if we want bigint)?

command/rootchain/initcontracts/init_contracts.go
igorcrevar: This can throw error only if we don't specify method signature correctly (so it can't in our case). Maybe we should add some linter flag or something?

consensus/polybft/block_builder.go
igorcrevar: It's not that easy to see (I didn't also) but the code really works. Maybe we need some comment to explain why/how it is working?

Makefile
Stefan-Ethernal: It is working either way, we don't even need parentheses apparently, but works with them as well. :slightly_smiling_face: 
I'll remove them.

contracts/system_addresses.go
Stefan-Ethernal: It has proven multiple times that it is good to have console precompile when troubleshooting SC execution :smile: However, since it is a debugging tool, we can figure out in the future whether it makes sense to distribute it to production system (either remove it completely before release or deliver it with some custom build tags)...

command/genesis/genesis.go
Stefan-Ethernal: In that case, I'm suggesting leaving as it is currently if you all agree until we come up with some smarter approach?

consensus/polybft/contracts_initializer.go
Stefan-Ethernal: No need for adding linter ignore, because linter hasn't reported a warning. We deliberately ignored error handling here. As Igor explained, this would fail only in case bug in the code: either function signature is not correct or there is some bug in ethgo library and we should fail fast on such occasions as there is no point in handling error in this case.

consensus/polybft/contracts_initializer.go
Stefan-Ethernal: Fixed [85cacd4](https://github.com/0xPolygon/polygon-edge/pull/796/commits/85cacd45d458aaeada1a1c57ad7d24a131716bf0)

command/rootchain/initcontracts/init_contracts.go
Stefan-Ethernal: No need for adding linter ignore, because linter hasn't reported a warning. We deliberately ignored error handling here. As Igor explained, this would fail only in case bug in the code: either function signature is not correct or there is some bug in ethgo library and we should fail fast on such occasions as there is no point in handling error in this case.

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed [43050c7](https://github.com/0xPolygon/polygon-edge/pull/796/commits/43050c73e6a58bdc050ef8e7e0734e30b81072f8)

consensus/polybft/state_transaction.go
Stefan-Ethernal: Fixed [43050c7](https://github.com/0xPolygon/polygon-edge/pull/796/commits/43050c73e6a58bdc050ef8e7e0734e30b81072f8)

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed [43050c7](https://github.com/0xPolygon/polygon-edge/pull/796/commits/43050c73e6a58bdc050ef8e7e0734e30b81072f8)

consensus/polybft/block_builder.go
Stefan-Ethernal: Any suggestion how such comment should look like? It isn't meaningful to me adding comment where I'm explaining that given check is performed in the `WriteTx` function instead of this one :smile:

igorcrevar: I do not know :D Maybe we do not need it at all. @b00ris wdyt?

command/genesis/polybft_params.go
Stefan-Ethernal: Realized that was not correct way to set it and it is completely changed now. Since we have such information written in the genesis (and is dictated by genesis cmd premine flag), I've used that value to initialize validators balance).

@igorcrevar please take a look whether it makes sense to you: [dad6eba](https://github.com/0xPolygon/polygon-edge/pull/796/commits/dad6ebaa45aa22820d59d8f3f3efde388220bbc9)

consensus/polybft/fsm.go
Stefan-Ethernal: Fixed [dad6eba](https://github.com/0xPolygon/polygon-edge/pull/796/commits/dad6ebaa45aa22820d59d8f3f3efde388220bbc9). Sorry for making such unnecessary change :pray:  :slightly_smiling_face:

consensus/polybft/fsm.go
Stefan-Ethernal: Fixed [dad6eba](https://github.com/0xPolygon/polygon-edge/pull/796/commits/dad6ebaa45aa22820d59d8f3f3efde388220bbc9)

types/rlp_marshal.go
Stefan-Ethernal: In that case I would need to consider use case when `From` field is missing like it has `contracts.SystemCaller` value assigned, which is not entirely correct (unable to make difference between it and assigning `nil` value).

Btw I need `From` field serialized, because of this check:
https://github.com/0xPolygon/polygon-edge/pull/796/files#diff-f64dae91a36980ba02c538f69fb83b7c80a8b4345db96849127307d33ccfacd1R882

vcastellm (approved):
LGTM

Nemanja0x (approved):

consensus/polybft/merkle_tree.go
```diff
@@ -0,0 +1,130 @@
...
+// VerifyProof verifies a Merkle tree proof of membership for provided data using the default hash type (Keccak256)
+func VerifyProof(index uint64, leaf []byte, proof []types.Hash, root types.Hash) error {
+	return VerifyProofUsing(index, leaf, proof, root, crypto.NewKeccakState())
+}
+
+// VerifyProofUsing verifies a Merkle tree proof of membership for provided data using the provided hash type
+func VerifyProofUsing(index uint64, leaf []byte, proof []types.Hash, root types.Hash, hash hash.Hash) error {
+	proofHash := getProofHash(index, leaf, proof, hash)
+	if !bytes.Equal(root.Bytes(), proofHash) {
+		return fmt.Errorf("leaf with index %v, not a member of merkle tree. Merkle root hash: %v", index, root)
```
JekaMas, line 104: could we introduce a typed error to be able to use errors.Is(As)?

JekaMas (approved):

consensus/polybft/merkle_tree.go
goran-ethernal: We would need to build a custom error struct for this, because of the logging of leaf index and root hash. `errors.Is` can not be used on custom error structs :confused: . There is this solution though:

`targetError := &CustomError{}
require.True(t, errors.As(err, &targetError))`

Is it ok to go with this approach (custom error struct + this check).

igorcrevar (approved):

types/rlp_marshal.go
goran-ethernal: I made a Jira task for this, so we can come back to that optimization later: https://polygon.atlassian.net/browse/EVM-146

PR #809, opened 2022-10-17
EVM-57 PolyBft state transactions introduction

igorcrevar:
# Description

Up time state transaction is added on the end of epoch
Other bridge transactions will be added in the separate task

# Changes include

# Checklist

## Testing

### Manual tests

Starting a cluster with epoch size 10 creates a state transaction with uptime calculation on the end of the epoch.

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -8,13 +8,13 @@ import (
 	"sync/atomic"
 
 	"github.com/0xPolygon/pbft-consensus"
+	"github.com/0xPolygon/polygon-edge/blockchain"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/bitmap"
 	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/wallet"
 	"github.com/0xPolygon/polygon-edge/types"
 	hcf "github.com/hashicorp/go-hclog"
 	"github.com/umbracle/ethgo"
-	"github.com/umbracle/ethgo/abi"
 )
 
 const (
@@ -180,6 +180,82 @@ func (c *consensusRuntime) NotifyProposalInserted(b *StateBlock) {
 	}
 }
 
+func (c *consensusRuntime) populateFsmIfBridgeEnabled(
+	ff *fsm, epoch *epochMetadata, isEndOfEpoch, isEndOfSprint bool) error {
+	systemState, err := c.getSystemState(c.lastBuiltBlock)
+	if err != nil {
+		return err
+	}
+
+	nextStateSyncExecutionIdx, err := systemState.GetNextExecutionIndex()
+	if err != nil {
+		return err
+	}
+
+	ff.stateSyncExecutionIndex = nextStateSyncExecutionIdx
+
+	ff.postInsertHook = func() error {
+		if c.IsBridgeEnabled() {
+			if isEndOfEpoch && ff.commitmentToSaveOnRegister != nil {
+				if err := c.state.insertCommitmentMessage(ff.commitmentToSaveOnRegister); err != nil {
+					return err
+				}
+
+				if err := c.buildBundles(
+					epoch, ff.commitmentToSaveOnRegister.Message, nextStateSyncExecutionIdx); err != nil {
+					return err
+				}
+			}
+		}
+
+		c.NotifyProposalInserted(ff.block)
+
+		return nil
+	}
+
+	nextRegisteredCommitmentIndex, err := systemState.GetNextCommittedIndex()
+	if err != nil {
+		return err
+	}
+
+	if isEndOfEpoch {
+		commitment, err := c.getCommitmentToRegister(epoch, nextRegisteredCommitmentIndex)
+		if err != nil {
+			if errors.Is(err, ErrCommitmentNotBuilt) || errors.Is(err, errQuorumNotReached) {
+				c.logger.Info("[FSM] Retrieving commitment error",
+					"epoch", epoch.Number, "index", nextRegisteredCommitmentIndex, "err", err)
+			} else {
+				return err
+			}
+		}
+
+		ff.proposerCommitmentToRegister = commitment
+	}
+
+	if isEndOfSprint {
+		if err := c.state.cleanCommitments(nextStateSyncExecutionIdx); err != nil {
+			return err
+		}
+
+		nonExecutedCommitments, err := c.state.getNonExecutedCommitments(nextStateSyncExecutionIdx)
+		if err != nil {
+			return err
+		}
+
+		if len(nonExecutedCommitments) > 0 {
+			bundlesToExecute, err := c.state.getBundles(nextStateSyncExecutionIdx, maxBundlesPerSprint)
+			if err != nil {
+				return err
+			}
+
+			ff.commitmentsToVerifyBundles = nonExecutedCommitments
+			ff.bundleProofs = bundlesToExecute
+		}
+	}
+
+	return nil
+}
+
 // FSM creates a new instance of fsm
 func (c *consensusRuntime) FSM() (*fsm, error) {
 	// figure out the parent. At this point this peer has done its best to sync up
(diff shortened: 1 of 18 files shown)


state/executor.go
```diff
@@ -857,3 +814,114 @@ func TransactionGasCost(msg *types.Transaction, isHomestead, isIstanbul bool) (u
...
+// checkAndProcessLegacyTx - first check if this message satisfies all consensus rules before
+// applying the message. The rules include these clauses:
+// 0. resolve from if not specified
+// 1. the nonce of the message caller is correct
+// 2. caller has enough balance to cover transaction fee(gaslimit * gasprice)
+// 3. the amount of gas required is available in the block
+// 4. there is no overflow when calculating intrinsic gas
+// 5. the purchased gas is enough to cover intrinsic usage
+// 6. caller has enough balance to cover asset transfer for **topmost** call
+func checkAndProcessLegacyTx(msg *types.Transaction, t *Transition) (uint64, error) {
```
ferranbt: This naming is not correct, there are two types of transactions: legacy and typed transaction. A type of typed transaction is the state transaction which does not have any gas checks. But, that does not mean that gas checks are only done for legacy transactions.

state/executor.go
```diff
@@ -371,13 +360,35 @@ func (t *Transition) GetTxnHash() types.Hash {
...
 	s := t.state.Snapshot()
-	result, err := t.apply(msg)
 
+	switch msg.Type {
+	case types.LegacyTx:
+		checkAndProcessTx = checkAndProcessLegacyTx
+		finishApplyTx = finishApplyLegacyTx
+	case types.StateTx:
+		checkAndProcessTx = checkAndProcessStateTx
+		finishApplyTx = func(_ *runtime.ExecutionResult, _ *types.Transaction, _ *Transition) {}
```
ferranbt: Why don't we only do an if condition? This seems like an overoptimization. State transition-wise, there is little difference between the transaction executions. Only the state transactions have a special condition in which they do not have to pay any gas (but consume it)

ferranbt: Check the implementation of the State transaction logic in v3, it was only an if condition at the beginning of the transition function. Also, I would not do the check inside `Apply` if possible, since the point of this function is to revert from any prior-check errors.

state/executor.go
```diff
@@ -857,3 +814,114 @@ func TransactionGasCost(msg *types.Transaction, isHomestead, isIstanbul bool) (u
...
+
+	// 6. caller has enough balance to cover asset transfer for **topmost** call
+	if balance := t.state.GetBalance(msg.From); balance.Cmp(msg.Value) < 0 {
+		return 0, NewTransitionApplicationError(ErrNotEnoughFunds, true)
+	}
+
+	return gasLeft, nil
+}
+
+func finishApplyLegacyTx(
```
ferranbt: Why is this not valid for state transactions?

state/executor.go
```diff
@@ -857,3 +814,114 @@ func TransactionGasCost(msg *types.Transaction, isHomestead, isIstanbul bool) (u
...
+
+	// pay the coinbase
+	coinbaseFee := new(big.Int).Mul(new(big.Int).SetUint64(result.GasUsed), msg.GasPrice)
+	t.state.AddBalance(t.ctx.Coinbase, coinbaseFee)
+
+	// return gas to the pool
+	t.addGasPool(result.GasLeft)
+}
+
+func checkAndProcessStateTx(msg *types.Transaction, t *Transition) (uint64, error) {
```
ferranbt: State Transactions do have to consume gas from the gas pool, check `subGasPool`.

ferranbt (changes requested):
It needs a refactor on the transition part of the code. It does not cover some major points (pointed out in comments) and I think it is a bit over complicated. Check the code in `v3` for this feature since similar things were done https://github.com/maticnetwork/v3/blob/175c1a7eda5a1f512e0591a59d52040b7cdf00a9/core/state_transition.go#L287 in only a few lines.

To give some extra context, the only difference between `StateTransaction` and `Legacy` (as of now) is that `StateTransaction` does not consume any gas for the user. The rest should look similar.

state/executor.go
igorcrevar: If we execute for both types of transaction following code
```
        refund := t.state.GetRefund()
	result.UpdateGasUsed(msg.Gas, refund)

	// refund the sender
	remaining := new(big.Int).Mul(new(big.Int).SetUint64(result.GasLeft), msg.GasPrice)
	t.state.AddBalance(msg.From, remaining)

	// pay the coinbase
	coinbaseFee := new(big.Int).Mul(new(big.Int).SetUint64(result.GasUsed), msg.GasPrice)
	t.state.AddBalance(t.ctx.Coinbase, coinbaseFee)
```

 in logs I see errors like this:

```
2022-10-17T16:15:53.786+0200 [WARN]  polygon.server.syncer.syncer: failed to complete bulk sync with peer, try to next one: peer ID=error 16Uiu2HAkwmizH9TBo69reVFZzkuC4BytRKRoqQUQcnia5szWPhj8="unable to verify block, invalid gas limit, block gas used exceeds gas limit, limit = 10000000, used=9223372036855775855"
```

Why is that?

state/executor.go
```diff
@@ -456,75 +445,61 @@ func NewGasLimitReachedTransitionApplicationError(err error) *GasLimitReachedTra
...
 	if err != nil {
 		return nil, NewTransitionApplicationError(err, false)
 	}
 
-	// 5. the purchased gas is enough to cover intrinsic usage
-	gasLeft := msg.Gas - intrinsicGasCost
-	// Because we are working with unsigned integers for gas, the `>` operator is used instead of the more intuitive `<`
-	if gasLeft > msg.Gas {
+	// the purchased gas is enough to cover intrinsic usage
+	availableGas := msg.Gas - intrinsicGasCost
```
ferranbt: Could we keep the same variable names?

ferranbt (approved):

state/executor.go
igorcrevar: fixed

state/executor.go
igorcrevar: fixed

state/executor.go
igorcrevar: fixed

state/executor.go
igorcrevar: fixed

state/executor.go
igorcrevar: fixed

state/executor.go
```diff
@@ -272,17 +273,6 @@ func (t *Transition) WriteFailedReceipt(txn *types.Transaction) error {
...
-	signer := crypto.NewSigner(t.config, uint64(t.r.config.ChainID))
-
-	var err error
-	if txn.From == emptyFrom {
-		// Decrypt the from address
-		txn.From, err = signer.Sender(txn)
-		if err != nil {
-			return NewTransitionApplicationError(err, false)
-		}
-	}
```
Stefan-Ethernal: Is it ok to move this part from here to `apply` (i.e. [checkAndProcessLegacyTx](https://github.com/0xPolygon/polygon-edge/pull/809/files#diff-f64dae91a36980ba02c538f69fb83b7c80a8b4345db96849127307d33ccfacd1R843-R852)? For this reason, `TestFetchBLSValidators` and `TestFetchECDSAValidators` started failing.

Two options:
1. leave it as it was before (check sender in `Write` function)
2. introduce this change and set `Transaction.From` field to valid address

igorcrevar: fixed

consensus/polybft/consensus_runtime.go
```diff
@@ -180,6 +180,82 @@ func (c *consensusRuntime) NotifyProposalInserted(b *StateBlock) {
...
+
+	nextStateSyncExecutionIdx, err := systemState.GetNextExecutionIndex()
+	if err != nil {
+		return err
+	}
+
+	ff.stateSyncExecutionIndex = nextStateSyncExecutionIdx
+
+	ff.postInsertHook = func() error {
+		if c.IsBridgeEnabled() {
```
Stefan-Ethernal: Seems like you don't need this check here (since entire function is called only if bridge is enabled)?

types/receipt.go
```diff
@@ -32,6 +32,16 @@ type Receipt struct {
 	GasUsed         uint64
 	ContractAddress *Address
 	TxHash          Hash
+
+	TransactionType TxType
```
Stefan-Ethernal, line 36: Do you need this field? As I can see it is only used for marshaling (`IsLegacyTx` is used there). On the other hand, `IsStateTx` isn't used anywhere.

igorcrevar: We gonna keep field for now but I have removed `IsLegacyTx` func

types/transaction.go
```diff
@@ -29,16 +73,22 @@ func (t *Transaction) IsContractCreation() bool {
...
 func (t *Transaction) ComputeHash() *Transaction {
-	ar := marshalArenaPool.Get()
 	hash := keccak.DefaultKeccakPool.Get()
 
-	v := t.MarshalRLPWith(ar)
-	hash.WriteRlp(t.Hash[:0], v)
-
-	marshalArenaPool.Put(ar)
-	keccak.DefaultKeccakPool.Put(hash)
+	if _, err := hash.Write(t.MarshalRLP()); err == nil {
```
goran-ethernal: What happens if an error occurs on `Write`? Hash will not be computed? The older implementation of `ComputeHash` function also ignored the error on `Write` though.

Stefan-Ethernal (approved):

types/transaction.go
igorcrevar: I have reverted this function to original one. I have just copy pasted it from original state tx branch but the change does not make any difference. We should probably tackle this problem in the future.

goran-ethernal (approved):
LGTM

consensus/polybft/consensus_runtime.go
igorcrevar: fixed

PR #814, opened 2022-10-19
Integrate goleak to tests

begmaroman:
# Description

Integrate [goleak](https://github.com/uber-go/goleak) to tests

# Changes include

# Checklist

## Testing


archive/archive_test.go
```diff
@@ -0,0 +1,11 @@
+package archive
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
```
JekaMas: is it correct or we need m.Run() also? Could you check

begmaroman: It is difficult to identify the exact test if go leak happens. The solution from the official doc does not work so I decided to put it to each test in the repo.

Stefan-Ethernal: If we are going to keep this approach (invoking goroutine leaks detector in each unit test), it is worth considering how are we going to enforce invocation of it in each unit test?

On the other hand, it'd be great if we were able to detect leaks on package level, by introducing just a `TestMain` test for entire package.

PR #827, opened 2022-10-24
Add read abstraction to the State

ferranbt:
# Description

This PR is a follow-up for #753. It adds a new abstraction for the `State` such that instead of having access to a low-level interface with a method `Get` to query for account and storage, now, we get a much richer abstraction with two methods:
```
GetAccount() *Account
GetStorage(...) Hash
```
There are two benefits of this new data layer. First, the layers that want to access the state do not have to perform extra computation on top of the raw result of `Get` to parse the objects. Second, it gives more flexibility to the state storage to decide how it wants to store information and not force it to adapt to the `Get` interface (i.e Erigon).

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -460,6 +460,11 @@ func NewTestServers(t *testing.T, num int, conf func(*TestServerConfig)) []*Test
 		}
 	})
 
+	logsDir, err := initLogsDir(t)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	// It is safe to use a dummy MultiAddr here, since this init method
 	// is called for the Dev consensus mode, and IBFT servers are initialized with NewIBFTServersManager.
 	// This method needs to be standardized in the future
@@ -471,7 +476,11 @@ func NewTestServers(t *testing.T, num int, conf func(*TestServerConfig)) []*Test
 			t.Fatal(err)
 		}
 
-		srv := NewTestServer(t, dataDir, conf)
+		srv := NewTestServer(t, dataDir, func(c *TestServerConfig) {
+			c.SetLogsDir(logsDir)
+			c.SetSaveLogs(true)
+			conf(c)
+		})
 		srv.Config.SetBootnodes(bootnodes)
 
 		srvs = append(srvs, srv)
--- a/server/server.go
+++ b/server/server.go
@@ -303,10 +303,12 @@ func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
 	if err != nil {
 		return 0
 	}
+
 	account, err := snap.GetAccount(addr)
 	if err != nil {
 		return 0
 	}
+
 	return account.Nonce
 }
 
@@ -315,10 +317,12 @@ func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, e
 	if err != nil {
 		return nil, fmt.Errorf("unable to get snapshot for root, %w", err)
 	}
+
 	account, err := snap.GetAccount(addr)
 	if err != nil {
 		return big.NewInt(0), nil
 	}
+
 	return account.Balance, nil
 }
 
@@ -432,10 +436,12 @@ func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Acc
 	if err != nil {
 		return nil, err
 	}
+
 	account, err := snap.GetAccount(addr)
 	if err != nil {
 		return nil, err
 	}
+
 	return account, nil
 }
 
@@ -449,7 +455,9 @@ func (j *jsonRPCHub) GetStorage(root types.Hash, addr types.Address, slot types.
 	if err != nil {
 		return nil, err
 	}
+
 	res := snap.GetStorage(addr, root, slot)
+
 	return res.Bytes(), nil
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -431,7 +431,13 @@ func (j *jsonRPCHub) GetPeers() int {
 	return len(j.Server.Peers())
 }
 
+// GetForksInTime returns the active forks at the given block height
+func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
+	return j.Executor.GetForksInTime(blockNumber)
+}
+
 func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
+	fmt.Println("get account", root, addr, j.state)
 	snap, err := j.state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
@@ -442,15 +448,15 @@ func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Acc
 		return nil, err
 	}
 
+	if account == nil {
+		return nil, jsonrpc.ErrStateNotFound
+	}
+
 	return account, nil
 }
 
-// GetForksInTime returns the active forks at the given block height
-func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
-	return j.Executor.GetForksInTime(blockNumber)
-}
-
 func (j *jsonRPCHub) GetStorage(root types.Hash, addr types.Address, slot types.Hash) ([]byte, error) {
+	fmt.Println("get storage", root, addr, j.state)
 	snap, err := j.state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
@@ -463,7 +469,6 @@ func (j *jsonRPCHub) GetStorage(root types.Hash, addr types.Address, slot types.
 
 func (j *jsonRPCHub) GetCode(hash types.Hash) ([]byte, error) {
 	res, ok := j.state.GetCode(hash)
-
 	if !ok {
 		return nil, fmt.Errorf("unable to fetch code")
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -437,7 +437,6 @@ func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
 }
 
 func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
-	fmt.Println("get account", root, addr, j.state)
 	snap, err := j.state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
@@ -456,7 +455,6 @@ func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Acc
 }
 
 func (j *jsonRPCHub) GetStorage(root types.Hash, addr types.Address, slot types.Hash) ([]byte, error) {
-	fmt.Println("get storage", root, addr, j.state)
 	snap, err := j.state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
--- a/server/server.go
+++ b/server/server.go
@@ -320,7 +320,7 @@ func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, e
 
 	account, err := snap.GetAccount(addr)
 	if err != nil {
-		return big.NewInt(0), nil
+		return big.NewInt(0), err
 	}
 
 	return account.Balance, nil
--- a/server/server.go
+++ b/server/server.go
@@ -18,7 +18,6 @@ import (
 	"github.com/0xPolygon/polygon-edge/crypto"
 	"github.com/0xPolygon/polygon-edge/helper/common"
 	configHelper "github.com/0xPolygon/polygon-edge/helper/config"
-	"github.com/0xPolygon/polygon-edge/helper/keccak"
 	"github.com/0xPolygon/polygon-edge/helper/progress"
 	"github.com/0xPolygon/polygon-edge/jsonrpc"
 	"github.com/0xPolygon/polygon-edge/network"
@@ -299,22 +298,15 @@ type txpoolHub struct {
 }
 
 func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
+	// TODO: Use a function that returns only Account
 	snap, err := t.state.NewSnapshotAt(root)
 	if err != nil {
 		return 0
 	}
-
-	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
-	if !ok {
+	account, err := snap.GetAccount(addr)
+	if err != nil {
 		return 0
 	}
-
-	var account state.Account
-
-	if err := account.UnmarshalRlp(result); err != nil {
-		return 0
-	}
-
 	return account.Nonce
 }
 
@@ -323,17 +315,10 @@ func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, e
 	if err != nil {
 		return nil, fmt.Errorf("unable to get snapshot for root, %w", err)
 	}
-
-	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
-	if !ok {
+	account, err := snap.GetAccount(addr)
+	if err != nil {
 		return big.NewInt(0), nil
 	}
-
-	var account state.Account
-	if err = account.UnmarshalRlp(result); err != nil {
-		return nil, fmt.Errorf("unable to unmarshal account from snapshot, %w", err)
-	}
-
 	return account.Balance, nil
 }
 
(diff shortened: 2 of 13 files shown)


state/txn_test.go
```diff
@@ -1,128 +1,52 @@
...
-	}
-
-	account := &Account{
-		Nonce:   p.Nonce,
-		Balance: big.NewInt(int64(p.Balance)),
-		Root:    root,
-	}
-
-	return account, snap
+func newStateWithPreState(preState map[types.Address]*PreState) readSnapshot {
```
JekaMas, line 48: Can't we return just raw type instead of the interface and let user to decide how to use it?

state/txn_test.go
```diff
@@ -1,128 +1,52 @@
...
+		Nonce:   raw.Nonce,
 	}
-
-	return state, snapshot
+	return acct, nil
 }
 
-func newTestTxn(p map[types.Address]*PreState) *Txn {
-	return newTxn(newStateWithPreState(p))
+func (m *mockSnapshot) GetCode(hash types.Hash) ([]byte, bool) {
```
JekaMas, line 44: Do we need second output parameter instead of returning 'nil' or we might have 'nil' values as well?

state/txn_test.go
```diff
@@ -1,128 +1,52 @@
...
-		accountRlp := v.MarshalTo(nil)
-		/*
-			accountRlp, err := rlp.EncodeToBytes(account)
-			if err != nil {
-				panic(err)
-			}
-		*/
-		snapshot.data[hex.EncodeToHex(hashit(addr.Bytes()))] = accountRlp
+	acct := &Account{
+		Balance: new(big.Int).SetUint64(raw.Balance),
```
JekaMas, line 37: raw.Balance is uint64 type only for testing purposes?

server/server.go
```diff
@@ -304,22 +303,15 @@ type txpoolHub struct {
 }
 
 func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
+	// TODO: Use a function that returns only Account
 	snap, err := t.state.NewSnapshotAt(root)
 	if err != nil {
 		return 0
```
JekaMas, line 304: shall we log the error?

JekaMas: here and below

server/server.go
```diff
@@ -486,19 +451,12 @@ func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
...
-
-	if err != nil {
-		return nil, err
-	}
-
-	obj, err := j.getState(account.Root, slot.Bytes())
-
+	snap, err := j.state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
```
JekaMas, line 479: shall we add some context to the error? maybe root hash and address?

state/immutable-trie/trie.go
```diff
@@ -143,16 +143,11 @@ func (t *Trie) Commit(objs []*state.Object) (state.Snapshot, []byte) {
 			}
 
 			if len(obj.Storage) != 0 {
-				localSnapshot, err := t.state.NewSnapshotAt(obj.Root)
+				trie, err := t.state.newTrieAt(obj.Root)
 				if err != nil {
 					panic(err)
```
JekaMas, line 148: could we return the error with some context?

state/state.go
```diff
@@ -19,13 +19,9 @@ type State interface {
 }
 
 type Snapshot interface {
-	Get(k []byte) ([]byte, bool)
-	Commit(objs []*Object) (Snapshot, []byte)
-}
+	readSnapshot
```
JekaMas, line 22: why do we need a private interface instead of raw type?

state/txn.go
```diff
@@ -8,13 +8,18 @@ import (
...
 	"github.com/0xPolygon/polygon-edge/crypto"
-	"github.com/0xPolygon/polygon-edge/helper/keccak"
 	"github.com/0xPolygon/polygon-edge/state/runtime"
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
 var emptyStateHash = types.StringToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
 
+type readSnapshot interface {
+	GetStorage(addr types.Address, root types.Hash, key types.Hash) types.Hash
```
JekaMas, line 18: Could it be just 

```
Storage(...)
Account(...)
Code(...)
```
?

state/txn_test.go
ferranbt: Testing purposes.

server/server.go
ferranbt: These functions (both for txpoolHub and jsonrpcHub) happen because of user input (send a transaction to the pool, or make a Jsonrpc request). I think that if we log the error, someone could just spam the console of the client.

state/txn.go
ferranbt: Aren't those names less descriptive of what the functions do?

ferranbt:

state/state.go
ferranbt: The idea is to abstract the transition execution (i.e. `evm`) from the state model (`itrie`).

state/txn_test.go
ferranbt: Same point as with https://github.com/0xPolygon/polygon-edge/pull/827#discussion_r1003070989

state/txn_test.go
ferranbt: We cannot have nil values, this function has not changed from the already existing one. If we want to change it, I would do it on another PR.

server/server.go
ferranbt: `NewSnapshotAt` already returns in the error message the root that cannot be found.

state/immutable-trie/trie.go
ferranbt: Do you mean to remove the panic? or throw a panic with some context? The error already includes the root that cannot be found. I would like to remove the panic in another PR since it was already there. I have tried it but many parts of the code change.

server/server.go
```diff
@@ -328,17 +320,10 @@ func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, e
 	if err != nil {
 		return nil, fmt.Errorf("unable to get snapshot for root, %w", err)
 	}
-
-	result, ok := snap.Get(keccak.Keccak256(nil, addr.Bytes()))
-	if !ok {
+	account, err := snap.GetAccount(addr)
+	if err != nil {
 		return big.NewInt(0), nil
```
Stefan-Ethernal: Would you like to propagate error here perhaps, as `GetBalance` function signature implies returning an error?

server/server.go
```diff
@@ -304,22 +303,15 @@ type txpoolHub struct {
 }
 
 func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
+	// TODO: Use a function that returns only Account
```
Stefan-Ethernal, line 301: Isn't this comment already addressed on line 311? If it is, just remove it.

state/immutable-trie/state.go
```diff
@@ -27,6 +27,19 @@ func NewState(storage Storage) *State {
 }
 
 func (s *State) NewSnapshot() state.Snapshot {
+	t := s.newTrie()
+	return &Snapshot{state: s, trie: t}
```
Stefan-Ethernal: Minor thing:
We don't need trie reference (unless you've left it for the debugging purposes)? 🙂 

```suggestion
	return &Snapshot{state: s, trie: s.newTrie()}
```

state/immutable-trie/snapshot.go
```diff
@@ -0,0 +1,73 @@
...
+	return &account, nil
+}
+
+func (s *Snapshot) Commit(objs []*state.Object) (state.Snapshot, []byte) {
+	trie, root := s.trie.Commit(objs)
+
+	return &Snapshot{trie: trie, state: s.state}, root
+}
+
+func (s *Snapshot) GetCode(hash types.Hash) ([]byte, bool) {
```
Stefan-Ethernal: Nitpick: 
could you move this function above `Commit`, so that all the functions from `readSnapshot` are grouped together?

Stefan-Ethernal (commented):
Looking good apart from the failing e2e tests and some lint issues.

server/server.go
ferranbt: No, this is a reference to how `txpool` receives the data from the state. Instead of having two functions, one for the nonce and another one for the balance it should only receive one with the Account like `jsonrpc` does.

server/server.go
```diff
@@ -448,62 +437,44 @@ func (j *jsonRPCHub) GetPeers() int {
...
-func (j *jsonRPCHub) getState(root types.Hash, slot []byte) ([]byte, error) {
-	// the values in the trie are the hashed objects of the keys
-	key := keccak.Keccak256(nil, slot)
+// GetForksInTime returns the active forks at the given block height
+func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
+	return j.Executor.GetForksInTime(blockNumber)
+}
 
+func (j *jsonRPCHub) GetAccount(root types.Hash, addr types.Address) (*state.Account, error) {
+	fmt.Println("get account", root, addr, j.state)
```
Kourin1996: Nitpick. Leftover code?

server/server.go
```diff
@@ -448,62 +437,44 @@ func (j *jsonRPCHub) GetPeers() int {
...
-// GetForksInTime returns the active forks at the given block height
-func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
-	return j.Executor.GetForksInTime(blockNumber)
+	return account, nil
 }
 
 func (j *jsonRPCHub) GetStorage(root types.Hash, addr types.Address, slot types.Hash) ([]byte, error) {
-	account, err := j.GetAccount(root, addr)
-
+	fmt.Println("get storage", root, addr, j.state)
```
Kourin1996: It might be also leftover code

Kourin1996 (approved):

Stefan-Ethernal (approved):

state/immutable-trie/trie.go
```diff
@@ -118,7 +118,7 @@ var accountArenaPool fastrlp.ArenaPool
 
 var stateArenaPool fastrlp.ArenaPool // TODO, Remove once we do update in fastrlp
 
-func (t *Trie) Commit(objs []*state.Object) (state.Snapshot, []byte) {
+func (t *Trie) Commit(objs []*state.Object) (*Trie, []byte) {
```
igorcrevar, line 121: Just a question: do we need some interface for commit maybe? After these changes anyone who needs to change (commit) state will work with exact `Trie` implementation and not with an abstraction.

Otherwise, LGTM

igorcrevar (approved):
Approved, just left one question

e2e/framework/helper.go
```diff
@@ -460,6 +460,11 @@ func NewTestServers(t *testing.T, num int, conf func(*TestServerConfig)) []*Test
 		}
 	})
 
+	logsDir, err := initLogsDir(t)
```
vcastellm, line 463: 🤔 https://github.com/0xPolygon/polygon-edge/blob/develop/e2e/framework/ibft.go#L54

state/txn.go
vcastellm: Agree with @JekaMas, why use `Get` prefix?

e2e/framework/helper.go
ferranbt: There are two different e2e abstractions: one that deploys a chain of nodes with Ibft, and another one that deploys a dev chain. You added logs only for the first one (the link), this change adds logs for the second one.

state/immutable-trie/trie.go
ferranbt: I am thinking about removing the `*Trie` object in the `Commit` because we do not need it. So it might disappear in the future.

state/immutable-trie/snapshot.go
```diff
@@ -0,0 +1,78 @@
...
+		err  error
+		trie *Trie
+	)
+
+	if root == emptyStateHash {
+		trie = s.state.newTrie()
+	} else {
+		trie, err = s.state.newTrieAt(root)
+		if err != nil {
+			return types.Hash{}
```
b00ris, line 28: Is it correct?

state/immutable-trie/snapshot.go
```diff
@@ -0,0 +1,78 @@
...
+	val, ok := trie.Get(key)
+	if !ok {
+		return types.Hash{}
+	}
+
+	p := &fastrlp.Parser{}
+
+	v, err := p.Parse(val)
+	if err != nil {
+		return types.Hash{}
```
b00ris, line 43: The same. It'll be very hard to debug.

state/txn.go
b00ris, line 18: What is root there?

state/txn.go
```diff
@@ -106,32 +99,17 @@ func (txn *Txn) getStateObject(addr types.Address) (*StateObject, bool) {
 		return obj.Copy(), true
 	}
 
-	data, ok := txn.snapshot.Get(txn.hashit(addr.Bytes()))
-	if !ok {
+	account, err := txn.snapshot.GetAccount(addr)
+	if err != nil {
```
b00ris, line 103: Could you add a log there?

state/txn.go
ferranbt: `root` is the Storage Root for `addr`. In the future, The `readSnapshot` will know from the `addr` automatically which is the root.

state/txn.go
ferranbt: Similar to a previous comment by Evgeny. This function depends on arbitrary input from the smart contract, so if the account is not there (err != nil) is not an error. It is part of the workflow of the EVM. So if we were to add a log, there will be a smart contract combination that would print many logs on the screen.

state/immutable-trie/snapshot.go
ferranbt: Kind of, yes. The EVM does not have a way to signal that there was an error (and it has to stop) due to an external condition (like storage failure). If it fails, the storage entry returned will be wrong and the computation will be wrong so either the block created will not be valid (block creator) or it will not be able to validate a block.

state/txn.go
ferranbt: Do you feel that `Storage` alone represents what the method does?

server/server.go
JekaMas: On the other hand, not logging errors with user data makes debugging really hard. we can limit data to log + data logs rotation

server/server.go
JekaMas: do we have context in the error what root has failed?

state/immutable-trie/trie.go
JekaMas: as a context I mead to use something like `fmt.Errorf("%w on root %s", err, obj.Root)`

state/state.go
JekaMas: why do we need to abstract, do we have a few implementations? or just for testing?

state/txn.go
JekaMas: we have prefix as an object itself (or package if we have a function). In that case it's like `snap.Storage()`.

state/state.go
ferranbt: So this `Snapshot` object as it right now does two things: `readSnapshot` + `Commit`. It is like that because the `executor` object does the `EVM` processing (with `readSnapshot`) and computes the trie (with `Commit`).

However, I would like to move towards an architecture where those two things are separated from each other (in different packages) in order to support things like the Erigon flat data model (where commit does not happen on each block).

The `readSnapshot` is an iteration towards that model. It is also easier to test. I can leave it as it is right now since there is no need for it and introduce the change later as the use of the interface becomes more clear.

state/immutable-trie/trie.go
ferranbt: Fixed in 8c5f05599efa28b2a31c9a93db2878d7cc7ddd6a

server/server.go
ferranbt: Fixed in 8c5f05599efa28b2a31c9a93db2878d7cc7ddd6a

server/server.go
ferranbt: Let's address this in a follow-up PR. I think this is part of a bigger convo on what to do with errors in general from queries in the JsonRPC-API (I will create a Jira task).

My strategy would be to keep these functions on `server` as they are and have a special `Debug/Trace` on the Json-RPC Api that logs the errors (**only**) if Debug is enabled.

There was a situation before on the `txpool` where a Debug log was accounting for much of the memory usage on the pool because of string manipulation. So, I want to be really careful about these specific debugging scenarios.

Those that sound good to you?

JekaMas: works for me. maybe we can, later, have an observability epic or something to check all errors and their's contexts, metrics and tracing.

JekaMas (approved):

state/txn.go
ferranbt: I am going to resolve this comment, lets's go back to this later if we feel the names are not good.

b00ris (approved):
Good job!

PR #828, opened 2022-10-24
EVM-129 Implement bridge workflow

Stefan-Ethernal:
# Description

Add support for bridge state transactions.
e2e test that runs the bridge is passing (`TestE2E_Bridge_MainWorkflow` test; all the other `e2e` tests also)

# Changes include

# Checklist

## Testing


command/genesis/polybft_params.go
```diff
@@ -29,14 +31,18 @@ const (
...
 	blockTimeFlag        = "block-time"
 	validatorsFlag       = "polybft-validators"
+	bridgeFlag           = "bridge"
 
 	defaultEpochSize                  = uint64(10)
 	defaultSprintSize                 = uint64(5)
 	defaultValidatorSetSize           = 100
 	defaultBlockTime                  = 2 * time.Second
 	defaultPolyBftValidatorPrefixPath = "test-chain-"
+	defaultContractsRootFolder        = "core-contracts/artifacts/contracts"
```
Stefan-Ethernal: Maybe you could use this constant instead 
`contracts.ContractsRootFolder`?

igorcrevar: Yes, this is some leftover which is not deleted

command/rootchain/helper/txn.go
```diff
@@ -138,19 +150,25 @@ func FundAccount(account types.Address) (types.Hash, error) {
 	return types.BytesToHash(receipt.TransactionHash.Bytes()), nil
 }
 
-func getJSONRPCClient() *jsonrpc.Client {
-	jrpcClientOnce.Do(func() {
-		ipAddr := ReadRootchainIP()
+func getJSONRPCClient() (*jsonrpc.Client, error) {
+	jrpcClientLock.Lock()
```
b00ris, line 154: Is it a lock for jsonRPCClient(is it global variable?)?

b00ris (approved):

JekaMas (approved):

command/rootchain/helper/txn.go
igorcrevar: Yes there was two global variables before
```
	jrpcClientOnce sync.Once
	jsonRPCClient  *jsonrpc.
```
I wanted to get rid of panics in all the functions and because of that I think it is better to initialize `jsonRPCClient` if it is `nil` each time the function is called and not just first time. Lock is used because I suppose intention was to get `jsonRPCClient` from different routines

ferranbt (approved):

PR #836, opened 2022-10-25
Debug tracing

Kourin1996:
# Description

This PR adds the following JSON-RPC endpoints:

- `debug_traceBlockByNumber` => Trace all transactions in the finalized block specified by block number and return the results
- `debug_traceBlockByHash` => Trace all transactions in the finalized block specified by block hash and return the results
- `debug_traceBlock` => Trace all transactions in the block that is RLP encoded hex and return the results
- `debug_traceTransaction` => Trace a finalized transaction specified tx hash and return the result
- `debug_traceCall` => Trace a call (that is not applied as a transaction)

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

1. Setup and Start cluster of Polygon Edge in the same manner
2. Deploy contract and call its method by transaction
3. Call JSON-RPC API

# Documentation update

# Additional comments


mrwillis:
Overjoyed about this

jsonrpc/helper.go
```diff
@@ -0,0 +1,196 @@
...
+import (
+	"errors"
+	"fmt"
+	"math/big"
+
+	"github.com/0xPolygon/polygon-edge/state"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type endpointHelperTxPoolStore interface {
```
ferranbt: Is this mandatory for the debug endpoint? I would like to keep the PRs more isolated and targeted around a specific feature and this change seems an optimization not related to debugging.

state/executor.go
```diff
@@ -348,10 +346,13 @@ func (t *Transition) GetTxnHash() types.Hash {
 }
 
 // Apply applies a new transaction
-func (t *Transition) Apply(msg *types.Transaction) (*runtime.ExecutionResult, error) {
+func (t *Transition) Apply(
+	msg *types.Transaction,
+	tracer runtime.Tracer,
```
ferranbt: I am not comfortable changing the `Apply` function directly. I think `Tracer` should be an optional object for the `Transition` and the specific `Tracer` instance should figure out if it only wants to track a specific transaction.

state/executor.go
```diff
@@ -589,7 +603,48 @@ func (t *Transition) applyCall(
 		}
 	}
 
-	result := t.run(c, host)
+	var result *runtime.ExecutionResult
+
+	if t.ctx.Tracer != nil {
+		// XXX: Depth begins from 0 in geth
+		if c.Depth == 1 {
```
ferranbt: Not sure if I follow this. What is the difference between `CallStart` and `InnerCallStart`? Can we merge both? The `Tracer` implementation for `InnerCallStart` is void. Does this mean that only calls at depth 1 would be tracked?

state/runtime/evm/state.go
```diff
@@ -210,30 +213,84 @@ func (c *state) resetReturnData() {
...
-	codeSize := len(c.code)
 	for !c.stop {
-		if c.ip >= codeSize {
+		op, ok = c.CurrentOpCode()
+
+		if tracer != nil {
+			logged, ipCopy, gasCopy = false, c.ip, c.gas
+
+			// copy data before execution
+			tracer.CaptureMemory(c.memory)
```
ferranbt: Is there any reason for having 3 calls instead of a single one?

state/runtime/evm/state.go
```diff
@@ -210,30 +213,84 @@ func (c *state) resetReturnData() {
...
+
+	tracer := c.host.GetTracer()
+
+	defer func() {
+		if vmerr == nil || tracer == nil {
+			return
+		}
+
+		if !logged {
+			tracer.ExecuteState(
```
ferranbt: Could we do something like this to avoid the many function calls?
```
type State interface {
   GetReturnData() []byte
}

type state struct {
  returnData []byte
}

func (s *state) GetReturnData() []byte {
   return s.returnData
}
```

```
type Tracer interface {
   ExecuteState(s State)
}
```

state/runtime/evm/state.go
```diff
@@ -210,30 +213,84 @@ func (c *state) resetReturnData() {
...
+				// XXX: c.currentConsumedGas is not the total of gas assumption in a instruction
+				// when the execution terminated in the middle
+				c.currentConsumedGas,
+				c.returnData,
+				c.msg.Depth,
+				c.err,
+				c.host,
+			)
+		} else {
+			tracer.ExecuteFault(
```
ferranbt: Can we merge `ExecuteFault` and `ExecuteState`?

ferranbt (changes requested):

jsonrpc/dispatcher.go
```diff
@@ -92,12 +98,19 @@ func (d *Dispatcher) registerEndpoints(store JSONRPCStore) {
 		d.params.chainID,
 		d.params.chainName,
 	}
-	d.endpoints.TxPool = &TxPool{store}
+	d.endpoints.TxPool = &TxPool{
+		store,
+	}
+	d.endpoints.Debug = &Debug{
+		helper,
+		store,
```
zivkovicmilos: Just curious about this:

You initialize the `helper` with the `store`, but you instantiate the `Debug` endpoint with both. Could this be avoided?

jsonrpc/helper.go
Kourin1996: You're right, it's not necessary but I found some methods would be used in eth endpoint and debug endpoint that's why I created this. Totall agree with that. I will stop using this for now. And please give me a comment in the future push.

state/executor.go
Kourin1996: OK. I will create a method to attach tracer object to context and remove 2nd argument from Apply method

state/runtime/evm/state.go
Kourin1996: Actually no special reason and will merge these 3 methods

jsonrpc/helper.go
Kourin1996: 8db21ac958c26bbd5a5a6ea4efcb16c75b560ccc

state/executor.go
Kourin1996: 2a5da666eb66487793d7379d3f3683b591df486b

state/executor.go
Kourin1996: Merged 9654a544c54118078600a0b83bddc5ab3c1a24d5

state/runtime/evm/state.go
Kourin1996: Merged d730d2c56ed1aed10fe730b60d255a87293f458c

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,359 @@
...
+		if !ok {
+			return nil, fmt.Errorf("error fetching block number %d header", uint64(number))
+		}
+
+		return header, nil
+	}
+}
+
+// getHeaderFromBlockNumberOrHash returns a header using the provided number or hash
+func (d *Debug) getHeaderFromBlockNumberOrHash(bnh *BlockNumberOrHash) (*types.Header, error) {
```
ferranbt: Many of the helper methods have been copied from `eth` endpoint. What are your thoughts on integrating that?

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,359 @@
...
+	debugTxPoolStore
+	debugStateStore
+}
+
+// Debug is the debug jsonrpc endpoint
+type Debug struct {
+	store debugStore
+}
+
+type TraceConfig struct {
```
ferranbt, line 69: Add a test to make sure the encoding/decoding is as expected.

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,359 @@
...
+	tracer.Config
```
ferranbt: We are moving towards not cross-depending modules, especially for the `jsonrpc` layer (see RFC-145)

server/server.go
```diff
@@ -531,6 +530,121 @@ func (j *jsonRPCHub) ApplyTxn(
...
+
+	if _, err := transition.Apply(targetTx); err != nil {
+		return nil, err
+	}
+
+	return tracer.GetResult(), nil
+}
+
+func (j *jsonRPCHub) TraceCall(
+	tx *types.Transaction,
```
ferranbt, line 618: This is not correct. You cannot use core `Transaction` and `Header` types for encoding/decoding the jsonrpc call itself. The same applies for the other functions.

server/server.go
```diff
@@ -531,6 +530,121 @@ func (j *jsonRPCHub) ApplyTxn(
...
+	parentHeader *types.Header,
```
ferranbt, line 619: The param for `debug_traceCall` is a `BlockNumberOrHash` object, not the header itself.

ferranbt (changes requested):

state/runtime/evm/state.go
Kourin1996: I'm not sure whether this works because there are other several values to be passed to tracer. I added helper methods. Please check it. e803f497157cbcbfa133f82e10fd41d20ff62201

state/runtime/evm/state.go
Kourin1996: I removed `ExecutionFault` for now because struct tracer doesn't need it

jsonrpc/debug_endpoint.go
Kourin1996: Firstly I had reverted creating common struct after this discussion https://github.com/0xPolygon/polygon-edge/pull/836#discussion_r1009168456

server/server.go
ferranbt: My mistake here, I thought this was the endpoint specification.

server/server.go
ferranbt: My mistake here, I thought this was the endpoint specification.

jsonrpc/debug_endpoint.go
Kourin1996: I added helper functions instead of super struct b9b3d8e6e242ad4041cc53e5460ed2fffc9aeb42

jsonrpc/debug_endpoint.go
Kourin1996: Removed direct dependency 302c47a6ca4515c9960df67ea400f56d32919307

jsonrpc/debug_endpoint.go
Kourin1996: Added c19bb53a5cf11c2783cc804bdaa4686869c6dfba. Other tests will be added in other commits

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,209 @@
...
+	config *TraceConfig,
+) (interface{}, error) {
+	num, err := GetNumericBlockNumber(blockNumber, d.store)
+	if err != nil {
+		return nil, err
+	}
+
+	block, ok := d.store.GetBlockByNumber(num, true)
+	if !ok {
+		return nil, errors.New("block not found")
```
JekaMas: could we always include user's context to errors? like block number or hash?

JekaMas: as a separate task, how do you think, do we need error rate per API or something like that in our metrics?

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,209 @@
...
+	)
+
+	// The filter is empty, use the latest block by default
+	if filter.BlockNumber == nil && filter.BlockHash == nil {
+		filter.BlockNumber, _ = createBlockNumberPointer("latest")
+	}
+
+	header, err = d.getHeaderFromBlockNumberOrHash(&filter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get header from block hash or block number")
```
JekaMas: do we need fmt.Errorf here? maybe errors.New or fmt.Errorf + come additional data?

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,209 @@
...
+	tracer := newTracer(config)
+
+	return d.store.TraceCall(tx, header, tracer)
+}
+
+func (d *Debug) traceBlock(
+	block *types.Block,
+	config *TraceConfig,
+) (interface{}, error) {
+	if block.Number() == 0 {
```
JekaMas, line 180: is it possible to have block==nil here somehow?

JekaMas: if with some circumstances it is possible I believe we need to check this condition as well.

jsonrpc/helper.go
```diff
@@ -0,0 +1,191 @@
...
+
+	case EarliestBlockNumber:
+		return 0, nil
+
+	case PendingBlockNumber:
+		return 0, fmt.Errorf("fetching the pending header is not supported")
+
+	default:
+		if number < 0 {
+			return 0, fmt.Errorf("invalid argument 0: block number larger than int64")
```
JekaMas: could we add number to the error message?

JekaMas (approved):
looks good to me, could we add a few tests on API calls and tracing format? or we can do it in a separate task with fuzzing vs Geth.

jsonrpc/debug_endpoint.go
Kourin1996: Added 39fdd3601694f705f6bd563b89848aee94273fe1

Kourin1996: > as a separate task, how do you think, do we need error rate per API or something like that in our metrics?

I'll keep that in mind

jsonrpc/debug_endpoint.go
Kourin1996: Fixed 1c181d3f8ebdb9db2d0742371ef85c814307eb52

jsonrpc/debug_endpoint.go
Kourin1996: Not sure nil check is necessary here because this is helper method that is called only from other method in same file

jsonrpc/helper.go
Kourin1996: You can see the error has at the latest commit
https://github.com/0xPolygon/polygon-edge/blob/1c181d3f8ebdb9db2d0742371ef85c814307eb52/jsonrpc/helper.go#L70

jsonrpc/dispatcher.go
Kourin1996: This code has been removed at latest commit

jsonrpc/debug_endpoint.go
JekaMas: just checking that we won't get panic on user's call

state/runtime/tracer/structtracer/tracer.go
```diff
@@ -0,0 +1,374 @@
...
+
+func NewStructTracer(config Config) *StructTracer {
+	return &StructTracer{
+		Config:  config,
+		storage: make(map[types.Address]map[types.Hash]types.Hash),
+	}
+}
+
+func (t *StructTracer) Cancel(err error) {
+	t.reason = err
```
JekaMas, line 76: Why we don't need it trace safe?

Kourin1996: What do you mean? Did you mean you want me to use `atomic.Value` for this field?

jsonrpc/helper.go
```diff
@@ -0,0 +1,227 @@
...
+	"fmt"
+	"math/big"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+var (
+	ErrHeaderNotFound           = errors.New("header not found")
+	ErrPendingBlockNumber       = errors.New("fetching the pending header is not supported")
+	ErrNegativeBlockNumber      = errors.New("invalid argument 0: block number larger than int64")
```
Stefan-Ethernal: This error message seems misleading.

I would expect something like this:
"invalid argument 0: block number must not be negative"

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,231 @@
...
+
+	tracer := structtracer.NewStructTracer(structtracer.Config{
+		EnableMemory:     config.EnableMemory,
+		EnableStack:      !config.DisableStack,
+		EnableStorage:    !config.DisableStorage,
+		EnableReturnData: config.EnableReturnData,
+	})
+
+	// cancellation of context is done by caller
+	timeoutCtx, _ := context.WithTimeout(context.Background(), timeout)
```
Stefan-Ethernal: This line is giving me the following warning:

`[{
	"resource": "/home/stefan/go/src/polygon-edge/jsonrpc/debug_endpoint.go",
	"owner": "_generated_diagnostic_collection_name_#0",
	"severity": 4,
	"message": "the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak",
	"source": "lostcancel",
	"startLineNumber": 220,
	"startColumn": 14,
	"endLineNumber": 220,
	"endColumn": 15
}]`

It is complaining because we have discarded `cancelFn` callback return value. I'm aware of the above comment suggesting that cancellation of context is handled by the caller, but also from what I can see, we don't use `ctx` input parameter, which is also strange...

So I'm not sure, but maybe we should consider once again about invoking `defer cancelFn()` or if we are 100% sure, then just suppress the warning. WDYT?

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return d.store.TraceBlock(block, tracer)
+}
+
+// newTracer creates new tracer by config
+func newTracer(
+	ctx context.Context,
```
Stefan-Ethernal: Unused parameter. Double check do you need it?

jsonrpc/debug_endpoint.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/0xPolygon/polygon-edge/state/runtime/tracer/structtracer"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+var (
+	defaultTraceTimeout = 5 * time.Second
+
+	// ErrExecutionTimeout indicates the execution was terminated due to timeout
+	ErrExecutionTimeout  = errors.New("execution timeout")
+	ErrTraceGenesisBlock = errors.New("genesis is not traceable")
```
Stefan-Ethernal, line 21: Add a comment here as well?

jsonrpc/helper.go
```diff
@@ -0,0 +1,227 @@
...
+	ErrFailedFetchGenesis       = errors.New("error fetching genesis block header")
+	ErrNoDataInContractCreation = errors.New("contract creation without data provided")
+)
+
+type latestHeaderGetter interface {
+	Header() *types.Header
+}
+
+// GetNumericBlockNumber returns block number based on current state or specified number
+func GetNumericBlockNumber(number BlockNumber, store latestHeaderGetter) (uint64, error) {
```
Stefan-Ethernal, line 25: I'm aware that it is a bit out of scope for this PR, but would be nice if we would substitute `resolveNum` function in
`func (f *FilterManager) getLogsFromBlocks(query *LogQuery) ([]*Log, error)` with `GetNumericBlockNumber` and delete `resolveNum`. 
If not through this PR, then through some of the following ones.

jsonrpc/helper.go
```diff
@@ -0,0 +1,227 @@
...
+	switch number {
+	case LatestBlockNumber:
+		return store.Header().Number, nil
```
Stefan-Ethernal: Should we first test `Header` against nil and in that case return an error (just in case). WDYT?

jsonrpc/helper.go
```diff
@@ -0,0 +1,227 @@
...
+	default:
+		// Convert the block number from hex to uint64
+		header, ok := store.GetHeaderByNumber(uint64(number))
+		if !ok {
+			return nil, fmt.Errorf("error fetching block number %d header", uint64(number))
+		}
+
+		return header, nil
+	}
+}
```
Stefan-Ethernal, line 81: Small suggestion in order to keep code compact, although double check if my proposal is correct.

The only downside I see currently is that for the `EarliestBlockNumber` we would access to the `CurrentHeader` twice :slightly_frowning_face: 

```suggestion
func GetBlockHeader(number BlockNumber, store headerGetter) (*types.Header, error) {
	num, err := GetNumericBlockNumber(number, store)
	if err != nil {
		return nil, fmt.Errorf("error resolving block number %d: %w", number, err)
	}

	header, ok := store.GetHeaderByNumber(num)
	if !ok {
		return nil, fmt.Errorf("error fetching block number %d header", uint64(number))
	}

	return header, nil
}
```

state/runtime/tracer/types.go
```diff
@@ -0,0 +1,70 @@
...
+	"math/big"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+// RuntimeHost is the interface defining the methods for accessing state by tracer
+type RuntimeHost interface {
+	// GetRefund returns refunded value
+	GetRefund() uint64
+	// GetStorage access the storage slot at the given address and slot ha
```
lazartravica:
```suggestion
	// GetStorage access the storage slot at the given address and slot hash
```

jsonrpc/debug_endpoint.go
lazartravica: Passed in context is not used, but I see it is correctly created and defer cancelled in all usages - but without interacting with it.

I don't think we need it, and the only cancellation context we need is the timeout done below inside the newTracer (`timeoutCtx`). Do you agree?

state/runtime/tracer/structtracer/tracer.go
lazartravica: I don't think we need thread safety here since it's Cancel is only called from the timeoutCtx when creating the tracer, but let's make it thread safe as we might throw more `reason` errors in the future

lazartravica (changes requested):
@Kourin1996 Left a couple of comments, otherwise great job on the PR.

Please ping @JekaMas and @ferranbt to see if you can resolve the comments you fixed yourself 🙏

state/runtime/tracer/structtracer/tracer.go
```diff
@@ -0,0 +1,374 @@
...
+	opCode int,
+	contractAddress types.Address,
+	sp int,
+	host tracer.RuntimeHost,
+) {
+	if !t.Config.EnableStorage || (opCode != evm.SLOAD && opCode != evm.SSTORE) {
+		return
+	}
+
+	_, inited := t.storage[contractAddress]
```
Stefan-Ethernal: Rename to `initialized` perhaps?

state/runtime/tracer/structtracer/tracer.go
Kourin1996: Fixed 9601ce9adfbd3a7d1022a643f797f0cc82f925b3

state/runtime/tracer/types.go
Kourin1996: Fixed 1096028217694699f803388c321d6a061dd00de7

jsonrpc/debug_endpoint.go
Kourin1996: Honestly this function doesn't need to take context because no one needs to stop. I will remove it

Kourin1996: Removed f3fcdf99bf024ceb8f9be26c5dcfb60a8c0750ad

jsonrpc/helper.go
Kourin1996: Good catch. Thank you. Fixed it 6d43e8b1a65d494fcf8daa1de029636163bb72d4

jsonrpc/debug_endpoint.go
Kourin1996: You can't call cancel func inside this function because the process doesn't begin inside. The callee must call it but I added context instead of this but it wasn't used. I make this func return cancel func

Kourin1996: 3bf99bb2067186017e8f35cbc4fa85ad061a5579

jsonrpc/debug_endpoint.go
Kourin1996: Removed f3fcdf99bf024ceb8f9be26c5dcfb60a8c0750ad

jsonrpc/debug_endpoint.go
Kourin1996: Added 74e6f0fa823787496c32cd8803d68c0dddeae2f7

jsonrpc/helper.go
Kourin1996: it's nice suggestion but more DB accesses happen sometimes and need to fix mock

Kourin1996: for example, you can get latest header from DB directly but this new function fetch data twice (getting header number and header by number)

jsonrpc/helper.go
Kourin1996: Added a nil check b8ebca340a2cd9b30cc19ed9bc95ddfa3dd658f1

jsonrpc/helper.go
Kourin1996: Good catch. Fixed fa39661f82637a40e35f0ff30b5242da2f4596bb

state/runtime/tracer/structtracer/tracer.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/0xPolygon/polygon-edge/helper/hex"
+	"github.com/0xPolygon/polygon-edge/state/runtime"
+	"github.com/0xPolygon/polygon-edge/state/runtime/evm"
+	"github.com/0xPolygon/polygon-edge/state/runtime/tracer"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type Config struct {
+	EnableMemory     bool // enable memory capture
+	EnableStack      bool // disable stack capture
```
Stefan-Ethernal: Is this comment by purpose like that, since it is following a spec? I find it misleading though, because `EnableStack` and `EnableStorage` are set as negated values of `DisableStack` and `DisableStorage` set from the end-user.

```suggestion
	EnableStack      bool // enable stack capture
```

state/runtime/tracer/structtracer/tracer.go
```diff
@@ -0,0 +1,382 @@
...
+	EnableStorage    bool // disable storage capture
```
Stefan-Ethernal:
```suggestion
	EnableStorage    bool // enable storage capture
```

Stefan-Ethernal (commented):
Also run `go mod tidy`, because VS Code complains that go.sum is out of sync with go.mod. :slightly_smiling_face:

jsonrpc/helper.go
Stefan-Ethernal: Yep, aware of it. Ok we are going to leave it as it is currently. :slightly_smiling_face:

state/runtime/evm/state.go
```diff
@@ -210,37 +215,51 @@ func (c *state) resetReturnData() {
...
+		op OpCode
+		ok bool
+	)
 
-	codeSize := len(c.code)
 	for !c.stop {
-		if c.ip >= codeSize {
-			c.halt()
+		op, ok = c.CurrentOpCode()
+		ipCopy, gasCopy := c.ip, c.gas
```
Stefan-Ethernal: Aren't `int` and `uint64` passed by values? Trying to figure out why do you need copies here? I was thinking we are safe using `c.ip` and `c.gas` straight away.

state/runtime/tracer/structtracer/tracer.go
Kourin1996: I forgot changing when I created a config object in jsonrpc layer. The interface is following the spec but this Config object doesn't have to follow it.

state/runtime/tracer/structtracer/tracer.go
Kourin1996: 569552c745e4054b86bd3966bb2a1bdac1f048f3

state/runtime/tracer/structtracer/tracer.go
Kourin1996: 569552c745e4054b86bd3966bb2a1bdac1f048f3

state/runtime/tracer/structtracer/tracer.go
Kourin1996: 569552c745e4054b86bd3966bb2a1bdac1f048f3

Stefan-Ethernal (approved):

ferranbt (approved):

state/runtime/evm/state.go
Kourin1996: Removed only `ipCopy` because gas consuming happens during this loop 16e4bfc22d7cad9c04bb4746e34f8c2d77377a76

Stefan-Ethernal: > gas consuming happens during this loop

Ok. Just to make sure I have put it right: you don't want to report current state of the gas, but the original one?

lazartravica (approved):

state/runtime/evm/state.go
Kourin1996: Yes, gas must be the value before the evaluation

PR #838, opened 2022-10-25
[Checkpointing] `ExitEvent` storage and proof generation

goran-ethernal:
# Description

Added functions to store and query exit events in `state.go`.
Added event root hash calculation which will be used as a parameter in block extra (one of the parameters for checkpoint), that gets signed by validators.
Added exit proof generation for users that want to exit.
Added relevant UTs for positive and negative cases.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -780,8 +780,8 @@ func (c *consensusRuntime) getExitEventRootHash(epoch uint64) (types.Hash, error
 	return tree.Hash(), nil
 }
 
-// generateExitProof generates proof of exit
-func (c *consensusRuntime) generateExitProof(exitID, epoch, checkpointBlock uint64) ([]types.Hash, error) {
+// GenerateExitProof generates proof of exit
+func (c *consensusRuntime) GenerateExitProof(exitID, epoch, checkpointBlock uint64) ([]types.Hash, error) {
 	exitEvent, err := c.state.getExitEvent(exitID, epoch, checkpointBlock)
 	if err != nil {
 		return nil, err
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -150,7 +150,7 @@ func (c *consensusRuntime) AddLog(eventLog *ethgo.Log) {
 		"index", eventLog.LogIndex,
 	)
 
-	event, err := decodeEvent(eventLog)
+	event, err := decodeStateSyncEvent(eventLog)
 	if err != nil {
 		c.logger.Error("failed to decode state sync event", "hash", eventLog.TransactionHash, "err", err)
 
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -276,26 +276,21 @@ func (c *consensusRuntime) FSM() (*fsm, error) {
 		return nil, err
 	}
 
-	eventRoot, err := c.getExitEventRootHash(epoch.Number)
-	if err != nil {
-		return nil, err
-	}
-
 	pendingBlockNumber := c.getPendingBlockNumber()
 	isEndOfSprint := c.isEndOfSprint(pendingBlockNumber)
 	isEndOfEpoch := c.isEndOfEpoch(pendingBlockNumber)
 
 	ff := &fsm{
-		config:         c.config.PolyBFTConfig,
-		parent:         parent,
-		backend:        c.config.blockchain,
-		polybftBackend: c.config.polybftBackend,
-		blockBuilder:   blockBuilder,
-		validators:     newValidatorSet(types.BytesToAddress(parent.Miner), epoch.Validators),
-		isEndOfEpoch:   isEndOfEpoch,
-		isEndOfSprint:  isEndOfSprint,
-		eventRoot:      eventRoot,
-		logger:         c.logger.Named("fsm"),
+		config:            c.config.PolyBFTConfig,
+		parent:            parent,
+		backend:           c.config.blockchain,
+		polybftBackend:    c.config.polybftBackend,
+		blockBuilder:      blockBuilder,
+		validators:        newValidatorSet(types.BytesToAddress(parent.Miner), epoch.Validators),
+		isEndOfEpoch:      isEndOfEpoch,
+		isEndOfSprint:     isEndOfSprint,
+		generateEventRoot: c.getExitEventRootHash,
+		logger:            c.logger.Named("fsm"),
 	}
 
 	if c.IsBridgeEnabled() {
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -770,16 +770,16 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header) (*CommitE
 func (c *consensusRuntime) getExitEventRootHash(epoch uint64) (types.Hash, error) {
 	exitEvents, err := c.state.getExitEventsByEpoch(epoch)
 	if err != nil {
-		return types.Hash{}, err
+		return types.ZeroHash, err
 	}
 
 	if len(exitEvents) == 0 {
-		return types.Hash{}, nil
+		return types.ZeroHash, nil
 	}
 
 	tree, err := createExitTree(exitEvents)
 	if err != nil {
-		return types.Hash{}, err
+		return types.ZeroHash, err
 	}
 
 	return tree.Hash(), nil
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -276,6 +276,11 @@ func (c *consensusRuntime) FSM() (*fsm, error) {
 		return nil, err
 	}
 
+	eventRoot, err := c.getExitEventRootHash(epoch.Number)
+	if err != nil {
+		return nil, err
+	}
+
 	pendingBlockNumber := c.getPendingBlockNumber()
 	isEndOfSprint := c.isEndOfSprint(pendingBlockNumber)
 	isEndOfEpoch := c.isEndOfEpoch(pendingBlockNumber)
@@ -289,6 +294,7 @@ func (c *consensusRuntime) FSM() (*fsm, error) {
 		validators:     newValidatorSet(types.BytesToAddress(parent.Miner), epoch.Validators),
 		isEndOfEpoch:   isEndOfEpoch,
 		isEndOfSprint:  isEndOfSprint,
+		eventRoot:      eventRoot,
 		logger:         c.logger.Named("fsm"),
 	}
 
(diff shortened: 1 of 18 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -276,6 +276,11 @@ func (c *consensusRuntime) FSM() (*fsm, error) {
 		return nil, err
 	}
 
+	eventRoot, err := c.getExitEventRootHash(epoch.Number)
```
b00ris: Is it before block exsecution or after?

goran-ethernal: Before

consensus/polybft/state.go
```diff
@@ -256,6 +318,79 @@ func (s *State) list() ([]*StateSyncEvent, error) {
...
+func (s *State) insertExitEvent(event *ExitEvent) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		raw, err := json.Marshal(event)
+		if err != nil {
+			return err
+		}
+
+		bucket := tx.Bucket(exitEventsBucket)
+
+		return bucket.Put(bytes.Join([][]byte{itob(event.Epoch), itob(event.ID), itob(event.BlockNumber)}, nil), raw)
```
b00ris: to be able to find events by block number i think it worth to have event.BlockNumber first

goran-ethernal: It's much more better to have epoch number first. Because, on every block, we collect all the accumulated exit events for given epoch, and that is a function called much more frequently than generate proof in my opinion. Plus, we must collect all the events for generating proof that happened in the given epoch until provided block in which checkpoint was submitted.

consensus/polybft/state.go
```diff
@@ -256,6 +318,79 @@ func (s *State) list() ([]*StateSyncEvent, error) {
...
+		c := tx.Bucket(exitEventsBucket).Cursor()
+		prefix := itob(epoch)
+
+		for k, v := c.Seek(prefix); bytes.HasPrefix(k, prefix); k, v = c.Next() {
+			var event *ExitEvent
+			if err := json.Unmarshal(v, &event); err != nil {
+				return err
+			}
+
+			if filter(event) {
```
b00ris, line 404: if we want to make a proof up to specific block(N) - we'll have to read all epoch events. After block N we can stop reading new exit events.

goran-ethernal: That is what we are doing. Take a look at the function `getExitEventsForProof`.

consensus/polybft/state.go
b00ris: epoch, block number, event Number.

b00ris: I meant switch event number and block number

b00ris: But i realized that it's not important here

consensus/polybft/state.go
goran-ethernal: Yeah, I see your point now, so I was taking a look at this. It is not easy to do. Unfortunately, boltdb only supports seek by prefix :disappointed:, and since we need all the events that happened in that epoch, until given block, there is no optimization good enough. Even if I put block number at the first place in key, we would still need to get events that have block equal or lower than given block plus their epochs have to match. To me, what we have now is the most optimal we can get from boltdb in this case.

consensus/polybft/consensus_runtime.go
```diff
@@ -760,6 +766,50 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header) (*CommitE
 	return commitEpoch, nil
 }
 
+// getExitEventRootHash returns the exit event root hash from gathered exit events in given epoch
+func (c *consensusRuntime) getExitEventRootHash(epoch uint64) (types.Hash, error) {
+	exitEvents, err := c.state.getExitEventsByEpoch(epoch)
+	if err != nil {
+		return types.Hash{}, err
```
Stefan-Ethernal: Nitpick:
I would use `types.ZeroHash` everywhere where you are returning an "empty" hash.

consensus/polybft/state.go
```diff
@@ -113,11 +131,53 @@ func decodeEvent(log *ethgo.Log) (*StateSyncEvent, error) {
...
+	// ID is the decoded 'index' field from the event
+	ID uint64 `abi:"id"`
+	// Sender is the decoded 'sender' field from the event
+	Sender ethgo.Address `abi:"sender"`
+	// Receiver is the decoded 'receiver' field from the event
+	Receiver ethgo.Address `abi:"receiver"`
+	// Data is the decoded 'data' field from the event
+	Data []byte `abi:"data"`
+	// Epoch is the epoch number in which exit event was added
+	Epoch uint64 `abi:"-"`
```
Stefan-Ethernal: What do you think about renaming it to `EpochNumber` (I've named it like that in #840, so that we are consistent across the codebase)?

consensus/polybft/state.go
```diff
@@ -256,6 +335,79 @@ func (s *State) list() ([]*StateSyncEvent, error) {
...
+func (s *State) getExitEvent(exitEventID, epoch, checkpointBlockNumber uint64) (*ExitEvent, error) {
+	var exitEvent *ExitEvent
+
+	err := s.db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(exitEventsBucket)
+
+		key := bytes.Join([][]byte{itob(epoch), itob(exitEventID), itob(checkpointBlockNumber)}, nil)
+		v := bucket.Get(key)
+		if v == nil {
+			return &ExitEventNotFoundError{exitEventID, checkpointBlockNumber, epoch}
```
Stefan-Ethernal: You've switched `checkpointBlockNumber` and `epoch` fields initialization here.

I believe it is better to use field names when setting it, in order not to make permutations.

```suggestion
			return &ExitEventNotFoundError{
				exitID:          exitEventID,
				checkpointBlock: checkpointBlockNumber,
				epoch:           epoch,
			}
```

consensus/polybft/state.go
```diff
@@ -55,6 +61,17 @@ const (
 	stateSyncBundleSize = 5
 )
 
+type ExitEventNotFoundError struct {
```
Stefan-Ethernal: You can un-export this error struct, as it is used exclusively in the `polybft` package. Agree?

consensus/polybft/state_test.go
```diff
@@ -352,6 +352,101 @@ func TestState_insertAndGetBundles(t *testing.T) {
...
+				event := &ExitEvent{index, ethgo.HexToAddress("0x101"), ethgo.HexToAddress("0x102"), []byte{11, 22}, i, block}
+				assert.NoError(t, state.insertExitEvent(event))
+
+				index++
+			}
+			block++
+		}
+	}
+
+	fmt.Printf("Inserted %v exit events in: %v\n", numOfEpochs*numOfBlocksPerEpoch*numOfEventsPerBlock,
```
Stefan-Ethernal: Do you need this print?

consensus/polybft/state.go
```diff
@@ -256,6 +335,79 @@ func (s *State) list() ([]*StateSyncEvent, error) {
...
+// getExitEventsByEpoch returns all exit events that happened in the given epoch
+func (s *State) getExitEventsByEpoch(epoch uint64) ([]*ExitEvent, error) {
+	return s.getExitEvents(epoch, func(exitEvent *ExitEvent) bool {
+		return exitEvent.Epoch == epoch
+	})
+}
+
+// getExitEventsForProof returns all exit events that happened in and prior to the given checkpoint block number
+// with respect to the epoch in which block is added
+func (s *State) getExitEventsForProof(epoch, checkpointBlock uint64) ([]*ExitEvent, error) {
```
Stefan-Ethernal, line 384: Do you find `getExitEventsUntilBlockNumber` more appropriate perhaps?

Stefan-Ethernal (changes requested):
Generally looks good, but found some minor points to be addressed.

consensus/polybft/state.go
goran-ethernal: I would leave it as is. **Until** is not quite right, and it isn't used anywhere but to build proof.

Stefan-Ethernal: Function name should be based on what function **does** and not where it is invoked. But as you wish :stuck_out_tongue:

consensus/polybft/state_test.go
goran-ethernal: No, this is a leftover I used to measure different key compositions.

Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a298d0ecad0757efc27415f6149f45b1f1e4dc7d

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a298d0ecad0757efc27415f6149f45b1f1e4dc7d

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a298d0ecad0757efc27415f6149f45b1f1e4dc7d

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a298d0ecad0757efc27415f6149f45b1f1e4dc7d

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a298d0ecad0757efc27415f6149f45b1f1e4dc7d

Stefan-Ethernal (approved):

Kourin1996 (approved):

consensus/polybft/consensus_runtime.go
b00ris: Could you make event root calculation after block execution?

b00ris: Is it possible to integrate event root calculation to BuildProposal?

Stefan-Ethernal: We need to set `eventRoot` before `BuildProposal` is called, because that's where we are creating `Extra` and writing checkpoint data to it.
Why do you see that we need to calculate event root after block execution? I believe it is just a matter of question whether we are including exit events in the block N or block N+1?

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a6cbda397c6a46283dc10fab61f673a0e1de169c

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/838/commits/a6cbda397c6a46283dc10fab61f673a0e1de169c

ferranbt (approved):

PR #840, opened 2022-10-26
Checkpoint data signing

Stefan-Ethernal:
# Description

This PR introduces new fields to the `Header.Extra`, which are required for checkpointing mechanism and are part of consensus protocol.

Checkpoint data consists of: 
- `BlockRound` - round for which we are building block at given height, 
- `EpochNumber` - denoting current epoch,
- `CurrentValidatorsHash` - hash of the current validator set (for the current epoch), 
- `NextValidatorsHash` - hash of the next validator set (applicable for next epoch, namely epoch ending blocks which contain such information),
- `EventRoot` - merkle root hash of all exit events triggered by the appropriate smart contract.

Other than that, checkpoint is described with `BlockNumber`, `BlockHash` and `ChainID`, but we don't write those information into the blockchain, as it would be redundant.

Checkpoint data hash is in the following format:
```solidity
keccak256(abi.encode(uint256 chainId, uint256 blockNumber, bytes32 blockHash, uint256 blockRound, 
   uint256 epochNumber, bytes32 currentValidatorsHash, bytes32 nextValidatorsHash, bytes32 eventRoot))
```

Current validators and next validators hash is in the following format:
```solidity
keccak256(abi.encode(tuple(address _address, uint256[4] blsKey)[]))
```

Also format of the proposal hash being signed by the validators has changed. Prior to this PR, it was corresponding to `BlockHash`. However now it is the same as `checkpointHash`.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -132,7 +132,7 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*StateBlock, error)
 		Receipts: b.state.Receipts(),
 	})
 
-	_ = b.block.Header.ComputeHash()
+	b.block.Header.ComputeHash()
 
 	return &StateBlock{
 		Block:    b.block,
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -132,7 +132,7 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*StateBlock, error)
 		Receipts: b.state.Receipts(),
 	})
 
-	b.block.Header.ComputeHash()
+	_ = b.block.Header.ComputeHash()
 
 	return &StateBlock{
 		Block:    b.block,
--- a/consensus/polybft/extra.go
+++ b/consensus/polybft/extra.go
@@ -2,14 +2,15 @@ package polybft
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/hashicorp/go-hclog"
 
-	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
-
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/bitmap"
+	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	"github.com/0xPolygon/polygon-edge/crypto"
 	"github.com/0xPolygon/polygon-edge/types"
-	"github.com/umbracle/ethgo"
+	"github.com/umbracle/ethgo/abi"
 	"github.com/umbracle/fastrlp"
 )
 
@@ -21,10 +22,8 @@ const (
 	ExtraSeal = 65
 )
 
-var (
-	// PolyMixDigest represents a hash of "PolyBFT Mix" to identify whether the block is from PolyBFT consensus engine
-	PolyMixDigest = types.Hash(ethgo.HexToHash("adce6e5230abe012342a44e4e9b6d05997d6f015387ae0e59be924afc7ec70c1"))
-)
+// PolyBFTMixDigest represents a hash of "PolyBFT Mix" to identify whether the block is from PolyBFT consensus engine
+var PolyBFTMixDigest = types.StringToHash("adce6e5230abe012342a44e4e9b6d05997d6f015387ae0e59be924afc7ec70c1")
 
 // Extra defines the structure of the extra field for Istanbul
 type Extra struct {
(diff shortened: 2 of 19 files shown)


goran-ethernal (approved):
LGTM

consensus/polybft/extra.go
```diff
@@ -88,46 +95,46 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
 		return err
 	}
 
-	if num := len(elems); num != 4 {
-		return fmt.Errorf("not enough elements to decode extra, expected 4 but found %d", num)
+	if num := len(elems); num != 5 {
+		return fmt.Errorf("incorrect elements count to decode Extra, expected 5 but found %d", num)
 	}
```
Kourin1996, line 100: Just curious question. Checkpoint in Extra is required always? Please note it won't be able to encode the extra generated before.

Kourin1996 (approved):

consensus/polybft/extra.go
Stefan-Ethernal: > Checkpoint in Extra is required always?

Yes, so that we have all the checkpoint data on disposal for sending to the rootchain/Ethereum on arbitrary block.

> Please note it won't be able to encode the extra generated before.

:ok_hand: aware of it, but it shouldn't be the problem, since if I'm correct, PoS v3 chain is going to be run from the genesis block.

consensus/polybft/fsm.go
```diff
@@ -158,21 +171,46 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
 		headerTime = time.Now()
 	}
 
+	currentValidatorsHash, err := f.validators.Accounts().Hash()
+	if err != nil {
+		return nil, err
+	}
+
+	nextValidatorsHash, err := nextValidators.Hash()
```
b00ris, line 181: Is nextValidators empty for non-end-of-epoch blocks?

consensus/polybft/polybft_test.go
```diff
@@ -54,7 +54,7 @@ func TestPolybft_VerifyHeader(t *testing.T) {
 		Number:    0,
 		ExtraData: append(make([]byte, signer.IstanbulExtraVanity), genesisExtra.MarshalRLPTo(nil)...),
 	}
-	genesisHeader.ComputeHash()
+	_ = genesisHeader.ComputeHash()
```
b00ris: ?

consensus/polybft/block_builder.go
```diff
@@ -132,7 +132,7 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*StateBlock, error)
 		Receipts: b.state.Receipts(),
 	})
 
-	b.block.Header.ComputeHash()
+	_ = b.block.Header.ComputeHash()
```
b00ris: ?

consensus/polybft/fsm.go
Stefan-Ethernal: Yes it is:
https://github.com/0xPolygon/polygon-edge/pull/840/files#diff-e7ab7b5e6b57d0319949e67e49d31398cccb3e4625ff1d69a1403c1db6ba68c4R121.

Btw should we return `ZeroHash` or still calculate keccak256 of abi encoded empty tuple of validators when validator set is empty?

b00ris: I thought to make currentValidatorsHash== nextValidatorsHash for the blocks without validator set changes.  Do you see any benefits/drawbacks?

b00ris (approved):
LGTM

consensus/polybft/polybft_test.go
Stefan-Ethernal: I just wanted to stress out that I intentionally don't use return value from `ComputeHash` function. I thought it is a good coding practice in Go?
However if it bothers you I can revert it back, don't mind it either way. :slightly_smiling_face:

consensus/polybft/fsm.go
Stefan-Ethernal: Don't have problem with that, but hopefully smart contract is going to handle it properly...?

Stefan-Ethernal: Fixed [87de519](https://github.com/0xPolygon/polygon-edge/pull/840/commits/87de5193b197f89d24eb7f4644ca2a2a769ec43b)

consensus/polybft/block_builder.go
Stefan-Ethernal: Fixed [1dd5099](https://github.com/0xPolygon/polygon-edge/pull/840/commits/1dd5099063160f1f24890a46bcf728f6d9022c97)

ferranbt (approved):

b00ris (approved):

PR #844, opened 2022-10-27
[Checkpointing] JSON RPC point for generating exit proofs

goran-ethernal:
# Description

Added a new JSON RPC point for generating exit proofs.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -34,8 +34,8 @@ type Consensus interface {
 	// GetSyncProgression retrieves the current sync progression, if any
 	GetSyncProgression() *progress.Progression
 
-	// GenerateExitProof generates exit proof for given exit event id
-	GenerateExitProof(exitID, epoch, checkpointNumber uint64) ([]types.Hash, error)
+	// GetBridgeProvider returns an instance of BridgeDataProvider
+	GetBridgeProvider() BridgeDataProvider
 
 	// Initialize initializes the consensus (e.g. setup data)
 	Initialize() error
@@ -78,3 +78,9 @@ type Params struct {
 
 // Factory is the factory function to create a discovery consensus
 type Factory func(*Params) (Consensus, error)
+
+// BridgeDataProvider is an interface providing bridge related functions
+type BridgeDataProvider interface {
+	// GenerateExit proof generates proof of exit for given exit event
+	GenerateExitProof(exitID, epoch, checkpointBlock uint64) ([]types.Hash, error)
+}
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -34,6 +34,9 @@ type Consensus interface {
 	// GetSyncProgression retrieves the current sync progression, if any
 	GetSyncProgression() *progress.Progression
 
+	// GenerateExitProof generates exit proof for given exit event id
+	GenerateExitProof(exitID, epoch, checkpointNumber uint64) ([]types.Hash, error)
+
 	// Initialize initializes the consensus (e.g. setup data)
 	Initialize() error
 
--- a/jsonrpc/bridge_endpoint.go
+++ b/jsonrpc/bridge_endpoint.go
@@ -1,0 +1,20 @@
+package jsonrpc
+
+import (
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+// bridgeStore interface provides access to the methods needed by bridge endpoint
+type bridgeStore interface {
+	GenerateExitProof(exitID, epoch, checkpointBlock uint64) ([]types.Hash, error)
+}
+
+// Bridge is the bridge jsonrpc endpoint
+type Bridge struct {
+	store bridgeStore
+}
+
+// GenerateExitProof generates exit proof for given exit event
+func (b *Bridge) GenerateExitProof(exitID, epoch, checkpointBlock BlockNumber) (interface{}, error) {
+	return b.store.GenerateExitProof(uint64(exitID), uint64(epoch), uint64(checkpointBlock))
+}
--- a/jsonrpc/bridge_endpoint.go
+++ b/jsonrpc/bridge_endpoint.go
@@ -15,6 +15,6 @@ type Bridge struct {
 }
 
 // GenerateExitProof generates exit proof for given exit event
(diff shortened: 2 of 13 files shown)


jsonrpc/bridge_endpoint.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// bridgeStore interface provides access to the methods needed by bridge endpoint
+type bridgeStore interface {
+	GenerateExitProof(exitID, epoch, checkpointBlock uint64) ([]types.Hash, error)
+}
+
+// Bridge is the bridge jsonrpc endpoint
+type Bridge struct {
+	store bridgeStore
```
vcastellm, line 14: General question, I see that it's the pattern but why is it called `store`?

vcastellm (approved):
LGTM

jsonrpc/bridge_endpoint.go
goran-ethernal: No idea 😄 . I just went along with rest of the endpoints

Kourin1996 (approved):

jsonrpc/bridge_endpoint.go
ferranbt: Most likely it was me who called it `store`.

jsonrpc/bridge_endpoint.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// GenerateExitProof generates exit proof for given exit event
+func (b *Bridge) GenerateExitProof(exitID, epoch, checkpointBlock BlockNumber) (interface{}, error) {
```
Stefan-Ethernal: Are you going to create similar concept for `exitID` and `epoch` parameters, since I'm not sure is it entirelly correct to reuse `BlockNumber`? There is `argUint64`, so I'm thinking maybe it is worth trying using it for these two parameters...?

I'm thinking that it is more appropriate to use it for `checkpointBlock` parameter as well, since `BlockNumber` type supports `Pending`, `Latest` and `Earlier` as values, which doesn't mean anything in this context (since one should always provide exact block number and not those strings which you don't know to interpret in this context).

consensus/consensus.go
```diff
@@ -34,6 +34,9 @@ type Consensus interface {
 	// GetSyncProgression retrieves the current sync progression, if any
 	GetSyncProgression() *progress.Progression
 
+	// GenerateExitProof generates exit proof for given exit event id
+	GenerateExitProof(exitID, epoch, checkpointNumber uint64) ([]types.Hash, error)
```
Stefan-Ethernal: Just thinking out loud:
this function seems like it doesn't really belong here. How about creating new interface called e.g. `BridgeDataProvider`. It should contain only this function and it should be implemented by `consensus_runtime`. Next you would embed it to the `jsonRPCHub` struct and you would call it the same way you are currently doing? Not sure is it feasible and does it make sense.

jsonrpc/bridge_endpoint.go
Stefan-Ethernal: I would call it `bridgeDataProvider` or something like that. Also Goran put it GenerateExitProof to the Consensus inteface, but my gut feeling is that it doesn't belong there...

consensus/consensus.go
goran-ethernal: Can not be done easily. Only an instance of consensus interface (in our case `Polybft`) has a reference to the `consensusRuntime` object, which has that state where we keep all the consensus related data. Exposing runtime to some other struct or interface is possible, but is it ok to do so? That interface would also need to reference polybft, since consensus runtime is created in polybft when we are a validator.

jsonrpc/bridge_endpoint.go
Stefan-Ethernal: > Most likely it was me who called it `store`.

@ferranbt I know you like words such as `backend`, `state`, `snapshot`, but `store` is new one... haha 😄 Just kidding a little bit. 

Anyhow I'm aware that occasionally it is hard to come up with some meaningful name, but on the other hand it also can be hard to figure out which module is in charge for which part when we have repeating names here and there.

jsonrpc/bridge_endpoint.go
goran-ethernal: I'll change the function so that parameters use that `argUint64`.

consensus/consensus.go
Stefan-Ethernal: > That interface would also need to reference polybft, since consensus runtime is created in polybft when we are a validator.

Haven't entirely understood this part. We'll have a quick chat about it tomorrow.

Stefan-Ethernal: Anyhow, my point here is that this way, you are forcing other consensus protocols to implement `GenerateExitProof` function, although it is related only to bridge workflow, being introduced by polybft consensus protocol (and `consensusRuntime` to be more precise).

Stefan-Ethernal: After some more thinking, I think that I get it: at the point where you are creating `jsonRPCHub`, you aren't aware that there is `consensusRuntime`, since we don't have some proper abstraction for it.

Possibly some quick solution would be that `Consensus` interface expose some function called `GetBridgeDataProvider` which would return `consensusRuntime` instance in the `polybft` consensus protocol (otherwise `nil`). Otherwise the rest would remain as I've proposed earlier: `consensusRuntime` implements `BridgeDataProvider` interface. That approach would help us better isolate parts of the system (if we for example one day decide to expose more bridge-related functions to the JSON RPC or any other part of the system). WDYT?

jsonrpc/bridge_endpoint.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/844/commits/9b64cc29943e1df78393686cf9e9b189fc6b7ced

consensus/consensus.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/844/commits/9b64cc29943e1df78393686cf9e9b189fc6b7ced

Stefan-Ethernal (approved):
LGTM

b00ris (approved):

ferranbt (approved):

PR #849, opened 2022-10-31
EVM-151 Use (or not) edge hooks support

igorcrevar:
# Description

Currently polybft code uses some callbacks after certain operations are executed. For example, ‘NotifyProposalInserted’.

Edge has its own hooks for specific operations while using go-ibft.  Examine if (and how) we should  use hooks provided by edge.

Both sync-er and polybft should call the same method which should update tx pool and restart epoch whenever new block/proposal is inserted into chain.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -43,9 +43,11 @@ type txPoolInterface interface {
 	Prepare()
 	Length() uint64
 	Peek() *types.Transaction
-	Pop(tx *types.Transaction)
-	Drop(tx *types.Transaction)
-	Demote(tx *types.Transaction)
+	Pop(*types.Transaction)
+	Drop(*types.Transaction)
+	Demote(*types.Transaction)
+	SetSealing(bool)
+	ResetWithHeaders(...*types.Header)
 }
 
 // epochMetadata is the static info for epoch currently being processed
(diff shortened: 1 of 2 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -158,23 +160,26 @@ func (c *consensusRuntime) AddLog(eventLog *ethgo.Log) { //nolint
...
+	if c.isEndOfEpoch(block.Header.Number) {
 		// reset the epoch. Internally it updates the parent block header.
-		if err := c.restartEpoch(lastHeader); err != nil {
+		if err := c.restartEpoch(block.Header); err != nil {
 			c.logger.Error("failed to restart epoch after block inserted", "err", err)
 		}
 	} else {
-		// inside the epoch, update last built block header
-		c.lastBuiltBlock = lastHeader
+		c.lock.Lock()
```
goran-ethernal, line 178: Why do we need a lock around `lastBuiltBlock` if it's only changed in `restartEpoch` and on new block finalization? Those two functions should not be executed from two routines if I'm not mistaken.

consensus/polybft/consensus_runtime.go
```diff
@@ -253,8 +258,9 @@ func (c *consensusRuntime) populateFsmIfBridgeEnabled(
 func (c *consensusRuntime) FSM() (*fsm, error) {
 	// figure out the parent. At this point this peer has done its best to sync up
 	// to the head of their remote peers.
-	parent := c.lastBuiltBlock
-	epoch := c.getEpoch()
+	c.lock.RLock()
```
goran-ethernal: As per [comment](https://github.com/0xPolygon/polygon-edge/pull/849/files#r1012728673), I think you do not need a lock around `lastBuiltBlock`. You do need a lock on getting `epochMetadata`, but you have `getEpoch` function you can use here.

JekaMas: Could we introduce thread safe methods rather then explicit call on mutexes?

consensus/polybft/consensus_runtime.go
igorcrevar: OnBlockInserted is called from two go routines. Syncer and ibft

consensus/polybft/consensus_runtime.go
igorcrevar: Currently we have one get method for each field. 

if we call them from this place one after another than it is possible to retrieve (for example) epoch from previous state, and last block from the newest.

Technically these two fields should be in some other struct but that is a lot of work right now I think.

Is it ok to create another method which will return both of those fields together like `getEpochAndLastBuiltBlock() (uint64, *epochMetaData)`?

JekaMas: the last one works for me, as far as we don't change *epochMetaData.

JekaMas: or make a small struct and store it as atomic in future with atomic.Pointer[*T]

JekaMas (approved):

goran-ethernal (approved):

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed, @JekaMas please check https://github.com/0xPolygon/polygon-edge/pull/849/commits/55c2213f556ab9db005a6eb60203d4551feab5a2

Stefan-Ethernal (approved):

PR #851, opened 2022-10-31
Move metrics to global variable

ferranbt:
# Description

This PR replaces the per-module `Metrics` struct with a global `metrics` module. The advantage is that now it is possible to emit at any place a `metric` to the outside while before, that variable had to be associated with a `Metric` struct and a module for which the lifecycle might be a bit more convoluted. 

For example, let's say that the `consensus` module has a sub-module `block-builder` to build the blocks and we want to expose metrics there. With this PR, you only need one line to export metrics inside `block-builder`.

The PR also replaces the `go-kit` framework with `go-metrics` which is the internal library it uses to export metrics.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


Makefile
```diff
@@ -47,7 +47,7 @@ test-e2e:
     # because it will get picked up and run during e2e tests
     # and the e2e tests should error out if any kind of race is found
 	go build -race -o artifacts/polygon-edge .
-	env EDGE_BINARY=${PWD}/artifacts/polygon-edge go test -v -timeout=30m ./e2e/...
+	env EDGE_BINARY=${PWD}/artifacts/polygon-edge go test -v -timeout=30m ./e2e/... -run TestIbft_TransactionFeeRecipient
```
Stefan-Ethernal: This should be reverted I believe?

ferranbt: Yep, thank you

network/server.go
```diff
@@ -758,26 +754,22 @@ func (s *Server) SubscribeCh() (<-chan *peerEvent.PeerEvent, error) {
...
-		s.metrics.InboundConnectionsCount.Set(
-			float64(s.connectionCounts.GetInboundConnCount()),
-		)
+		metrics.SetGauge([]string{"inbound_connections_count"}, float32(s.connectionCounts.GetInboundConnCount()))
+
 	case network.DirOutbound:
-		s.metrics.OutboundConnectionsCount.Set(
-			float64(s.connectionCounts.GetOutboundConnCount()),
-		)
+		metrics.SetGauge([]string{"outbound_connections_count"}, float32(s.connectionCounts.GetInboundConnCount()))
```
vcastellm:
```suggestion
		metrics.SetGauge([]string{"outbound_connections_count"}, float32(s.connectionCounts.GetOutboundConnCount()))
```
copypasta error maybe?

vcastellm (approved):
LGTM aside from the suggestion.

network/server.go
ferranbt: Indeed. Good catch.

txpool/txpool.go
```diff
@@ -452,7 +448,7 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 	clearAccountQueue(dropped)
 
 	// update metrics
-	p.metrics.PendingTxs.Add(float64(-1 * len(dropped)))
+	metrics.SetGauge([]string{"pending_transactions"}, float32(-1*len(dropped)))
```
Kourin1996: Original code added a value with the previous one. `SetGauge` can do same?

Kourin1996 (approved):
LGTM. I just have small questions.

consensus/ibft/ibft.go
```diff
@@ -295,7 +294,7 @@ func (i *backendIBFT) startConsensus() {
 		}
 
 		// Update the No.of validator metric
-		i.metrics.Validators.Set(float64(i.currentValidators.Len()))
+		metrics.SetGauge([]string{"validators"}, float32(i.currentValidators.Len()))
```
JekaMas, line 297: Just curious, is it a thread-safe operation? Won't we get data races in tests?

ferranbt: No data races, it is thread-safe.

txpool/txpool.go
ferranbt: Good catch, I will address it.

consensus/ibft/ibft.go
JekaMas: I believe we don't have tests with race at the moment. Could we check?

ferranbt: I tested with the `txpool`, there are data races there but none related to the `metrics`. This library is used across many OSS projects (including Hashicorp tools) without explicit locks, so it is safely handled internally.

txpool/txpool.go
ferranbt: Fixed 01888aacd724fbf29459a959f97100a2cace724c

Stefan-Ethernal (approved):

JekaMas (approved):

PR #852, opened 2022-10-31
[Checkpoints] Parse receipt logs and insert new exit events

goran-ethernal:
# Description

Added parsing of receipt logs on `BuildProposal` which will identify any new exit events that happened and insert them to the `state.go` persistence storage.

Added appropriate UTs that test positive and negative cases.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -161,7 +161,7 @@ func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 }
 
 // Fill fills the block with transactions from the txpool
-func (b *BlockBuilder) Fill() error {
+func (b *BlockBuilder) Fill() []*types.Receipt {
 	blockTimer := time.NewTimer(b.params.BlockTime)
 
 	b.params.TxPool.Prepare()
@@ -169,7 +169,7 @@ write:
 	for {
 		select {
 		case <-blockTimer.C:
-			return nil
+			return b.state.Receipts()
 		default:
 			tx := b.params.TxPool.Peek()
 
@@ -188,7 +188,7 @@ write:
 	//	wait for the timer to expire
 	<-blockTimer.C
 
-	return nil
+	return b.state.Receipts()
 }
 
 func (b *BlockBuilder) writeTxPoolTransaction(tx *types.Transaction) (bool, error) {
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -161,7 +161,7 @@ func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 }
 
 // Fill fills the block with transactions from the txpool
-func (b *BlockBuilder) Fill() []*types.Receipt {
+func (b *BlockBuilder) Fill() {
 	blockTimer := time.NewTimer(b.params.BlockTime)
 
 	b.params.TxPool.Prepare()
@@ -169,7 +169,7 @@ write:
 	for {
 		select {
 		case <-blockTimer.C:
-			return b.state.Receipts()
+			return
 		default:
 			tx := b.params.TxPool.Peek()
 
@@ -187,7 +187,10 @@ write:
 
 	//	wait for the timer to expire
 	<-blockTimer.C
+}
 
+// Receipts returns the collection of transaction receipts for given block
+func (b *BlockBuilder) Receipts() []*types.Receipt {
 	return b.state.Receipts()
 }
 
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -1659,13 +1659,13 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 		tree, err := NewMerkleTree(encodedEvents)
 		require.NoError(t, err)
 
-		hash, err := runtime.getExitEventRootHash(1)
+		hash, err := runtime.BuildEventRoot(1)
 		require.NoError(t, err)
 		require.Equal(t, tree.Hash(), hash)
 	})
 
 	t.Run("Get exit event root hash - no events", func(t *testing.T) {
-		hash, err := runtime.getExitEventRootHash(2)
+		hash, err := runtime.BuildEventRoot(2)
 		require.NoError(t, err)
 		require.Equal(t, types.Hash{}, hash)
 	})
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -1659,13 +1659,13 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 		tree, err := NewMerkleTree(encodedEvents)
 		require.NoError(t, err)
 
-		hash, err := runtime.BuildEventRoot(1)
+		hash, err := runtime.BuildEventRoot(1, nil)
 		require.NoError(t, err)
 		require.Equal(t, tree.Hash(), hash)
 	})
 
 	t.Run("Get exit event root hash - no events", func(t *testing.T) {
-		hash, err := runtime.BuildEventRoot(2)
+		hash, err := runtime.BuildEventRoot(2, nil)
 		require.NoError(t, err)
 		require.Equal(t, types.Hash{}, hash)
 	})
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -1643,6 +1643,8 @@ func TestConsensusRuntime_FSM_EndOfEpoch_PostHook(t *testing.T) {
 }
 
 func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
+	t.Parallel()
+
 	const (
 		numOfBlocks         = 10
 		numOfEventsPerBlock = 2
@@ -1656,6 +1658,8 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 	encodedEvents := setupExitEventsForProofVerification(t, state, numOfBlocks, numOfEventsPerBlock)
 
 	t.Run("Get exit event root hash", func(t *testing.T) {
+		t.Parallel()
+
 		tree, err := NewMerkleTree(encodedEvents)
 		require.NoError(t, err)
 
@@ -1665,6 +1669,8 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 	})
 
 	t.Run("Get exit event root hash - no events", func(t *testing.T) {
+		t.Parallel()
+
 		hash, err := runtime.BuildEventRoot(2, nil)
 		require.NoError(t, err)
 		require.Equal(t, types.Hash{}, hash)
@@ -1672,6 +1678,8 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 }
 
 func TestConsensusRuntime_GenerateExitProof(t *testing.T) {
+	t.Parallel()
+
 	const (
 		numOfBlocks         = 10
 		numOfEventsPerBlock = 2
@@ -1694,11 +1702,14 @@ func TestConsensusRuntime_GenerateExitProof(t *testing.T) {
 	require.NotNil(t, proof)
 
 	t.Run("Generate and validate exit proof", func(t *testing.T) {
+		t.Parallel()
 		// verify generated proof on desired tree
 		require.NoError(t, VerifyProof(1, encodedEvents[1], proof, tree.Hash()))
 	})
 
 	t.Run("Generate and validate exit proof - invalid proof", func(t *testing.T) {
+		t.Parallel()
+
 		invalidProof := proof
 		invalidProof[0][0]++
 
@@ -1707,6 +1718,8 @@ func TestConsensusRuntime_GenerateExitProof(t *testing.T) {
 	})
 
 	t.Run("Generate exit proof - no event", func(t *testing.T) {
+		t.Parallel()
+
 		_, err := runtime.GenerateExitProof(21, 1, 1)
 		require.ErrorContains(t, err, "could not find any exit event that has an id")
 	})
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -538,11 +538,11 @@ func (f *fsm) Insert(p *pbft.SealedProposal) error {
 		Bitmap:              bitmap,
 	}
 
-	// Write extar data to header
+	// Write extra data to header
 	f.block.Block.Header.ExtraData = append(make([]byte, 32), extra.MarshalRLPTo(nil)...)
 
 	// commit exit events only when we finalize a block
-	events, err := getExitEventsFromReceipts(f.epochNumber, f.block.Block.Number(), f.blockBuilder.Receipts())
+	events, err := getExitEventsFromReceipts(f.epochNumber, f.block.Block.Number(), f.block.Receipts)
 	if err != nil {
 		return err
 	}
@@ -673,7 +673,7 @@ func getExitEventsFromReceipts(epoch, block uint64, receipts []*types.Receipt) (
 		}
 
 		for j := 0; j < len(receipts[i].Logs); j++ {
-			event, err := decodeExitEvent(convert(receipts[i].Logs[j]), epoch, block)
+			event, err := decodeExitEvent(convertLog(receipts[i].Logs[j]), epoch, block)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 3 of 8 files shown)


consensus/polybft/block_builder.go
```diff
@@ -161,15 +161,15 @@ func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 }
 
 // Fill fills the block with transactions from the txpool
-func (b *BlockBuilder) Fill() error {
+func (b *BlockBuilder) Fill() []*types.Receipt {
```
ferranbt: Could we have instead a function in `BlockBuilder` to return the receipts at any time?
```
func (b *BlockBuilder) Receipts() []*types.Receipt{
   return b.state.Receipts()
}
```

consensus/polybft/state.go
```diff
@@ -128,7 +129,12 @@ func decodeStateSyncEvent(log *ethgo.Log) (*StateSyncEvent, error) {
 }
 
 func decodeExitEvent(log *ethgo.Log, epoch, block uint64) (*ExitEvent, error) {
-	raw, err := exitEventABI.ParseLog(log)
+	if !exitEventABI.Match(log) {
+		// valid case, not an exit event
+		return nil, errNotAnExitEvent
```
ferranbt: Could we do `return nil, false, nil` and use the bool to signal if it's an exit event instead of playing with the error?

ferranbt (approved):
Other than minor comments LGTM!

consensus/polybft/state.go
goran-ethernal: Hmm, we could, but to me this looks much cleaner. Just my opinion, since you can easily use a predefined error to recognize certain cases. Whatever we agree upon, works for me.

consensus/polybft/fsm.go
```diff
@@ -161,8 +161,31 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
...
-		return nil, err
+	receipts := f.blockBuilder.Fill()
+	events := make([]*ExitEvent, 0)
+
+	for i := 0; i < len(receipts); i++ {
+		if len(receipts[i].Logs) == 0 {
+			continue
+		}
+
+		event, err := decodeExitEvent(convert(receipts[i].Logs[0]), f.epochNumber, parent.Number+1)
```
b00ris: Not sure about receipts[i].Logs[0]. I think we need to check all logs.
cc @QEDK

consensus/polybft/block_builder.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/f9e4392a8fbcad2f61c0eb784b1d48f4a324e38c

consensus/polybft/fsm.go
b00ris: One more thing. If log.Event is not exit contract event than we can skip other computations immideatly.

consensus/polybft/fsm.go
```diff
@@ -161,8 +161,31 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
...
+			}
+
+			return nil, err
+		}
+
+		events = append(events, event)
+	}
+
+	if len(events) > 0 {
+		if err := f.checkpointBackend.InsertExitEvents(events); err != nil {
```
b00ris: What happens if a proposer created an block(with exit events) and other network hasn't signed it and created a block on another round?

b00ris (changes requested):

consensus/polybft/fsm.go
goran-ethernal: We execute transactions on `BuildProposal`. If we are to move this on block insert, where we know that everything is signed, and block gets finalized, then, we must be aware, that those exit events, will not be included in that checkpoint. We need the event root hash before we sign a block.

goran-ethernal: Hmm, now that I think about this, this is not the right way to collect logs and exit events. Other validators will not have these events in their storage, since they are not proposers. We would need an event listener for `sidechain `events. Similar what we have for `rootchain` events (`State Syncs`).

Yeah, this needs more work.

consensus/polybft/state.go
vcastellm: If you want my two cents, I agree that if it's a valid event it should not return an err, I suggest `return nil, nil` should suffice, a `nil` `ExitEvent` indicates the event wasn't of that type but no error.

goran-ethernal: Thats totally fine by me. Will change it to return nil in this case.

consensus/polybft/fsm.go
```diff
@@ -161,8 +162,33 @@ func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
...
+		if len(receipts[i].Logs) == 0 {
+			continue
+		}
+
+		event, err := decodeExitEvent(convert(receipts[i].Logs[0]), f.epochNumber, parent.Number+1)
+		if err != nil {
+			if errors.Is(err, errNotAnExitEvent) {
+				// valid case, transaction is not an exit transaction
+				continue
+			}
```
Kourin1996: Nitpick. How about calling errors.Is(err, errNotAnExitEvent) before nil check? You can write shallower code.

consensus/polybft/mocks_test.go
```diff
@@ -298,6 +300,24 @@ func (t *transportMock) Gossip(message interface{}) {
 	_ = t.Called(message)
 }
 
+var _ checkpointBackend = &checkpointBackendMock{}
```
Kourin1996: Why do you need to init it? maybe it's another way to check statically.

```go
var _ checkpointBackend = (*checkpointBackendMock)(nil)
```

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/a82711ed9b25839a7f003c5f53018db390793148

consensus/polybft/fsm.go
goran-ethernal: Ok, regarding your comment, I added parsing on logs for proposer and non proposers on block finalization (`Insert` function). This will make sure we have a correct state of exit events on all validators.
Proposer will also parse logs to get exit events on `BuildProposal`, so he can include them in the exit event tree, but he will not insert them to storage until block is finalized.

Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/08930c53a30e19ba4fcd8c483c23475e1a07d92a

consensus/polybft/fsm.go
goran-ethernal: I removed that error completely. If the log does not contain an exit event (which is a valid case), that `decode` will return `nil, nil`, as per comment: https://github.com/0xPolygon/polygon-edge/pull/852#discussion_r1009656522

Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/08930c53a30e19ba4fcd8c483c23475e1a07d92a

consensus/polybft/fsm.go
goran-ethernal: OK, I added checking of all logs in receipt.

Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/08930c53a30e19ba4fcd8c483c23475e1a07d92a

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/08930c53a30e19ba4fcd8c483c23475e1a07d92a

consensus/polybft/state.go
```diff
@@ -181,6 +187,21 @@ func decodeEventData(eventDataMap map[string]interface{}, log *ethgo.Log,
 	return eventCreator(id, sender, receiver, data), nil
 }
 
+// convert converts types.Log to ethgo.Log
+func convert(log *types.Log) *ethgo.Log {
```
Stefan-Ethernal: How about calling it `convertLog` maybe? I find simply `convert` too generic name.

consensus/polybft/state.go
```diff
@@ -181,6 +187,21 @@ func decodeEventData(eventDataMap map[string]interface{}, log *ethgo.Log,
...
+	l := &ethgo.Log{
+		Address: ethgo.Address(log.Address),
+		Data:    log.Data,
+		Topics:  make([]ethgo.Hash, len(log.Topics)),
+	}
+
+	for i := 0; i < len(log.Topics); i++ {
+		l.Topics[i] = ethgo.Hash(log.Topics[i])
+	}
```
Stefan-Ethernal: Micro optimization :slightly_smiling_face: 

```suggestion
	for i, topic := range log.Topics {
		l.Topics[i] = ethgo.Hash(topic)
	}
```

consensus/polybft/fsm.go
```diff
@@ -535,15 +541,23 @@ func (f *fsm) Insert(p *pbft.SealedProposal) error {
...
+	if err != nil {
 		return err
 	}
 
-	if err := f.postInsertHook(); err != nil {
+	if len(events) > 0 {
+		if err := f.checkpointBackend.InsertExitEvents(events); err != nil {
+			return err
+		}
+	}
```
Stefan-Ethernal: Should we move this part after `f.backend.CommitBlock(f.block)` or even to the `postInsertHook` perhaps? 
Because at that point we are sure that block is written to the local storage (i.e. finalized).

consensus/polybft/fsm.go
```diff
@@ -535,15 +541,23 @@ func (f *fsm) Insert(p *pbft.SealedProposal) error {
 	// Write extar data to header
 	f.block.Block.Header.ExtraData = append(make([]byte, 32), extra.MarshalRLPTo(nil)...)
 
-	if err := f.backend.CommitBlock(f.block); err != nil {
+	// commit exit events only when we finalize a block
+	events, err := getExitEventsFromReceipts(f.epochNumber, f.block.Block.Number(), f.blockBuilder.Receipts())
```
Stefan-Ethernal: This is not correct: only proposer would have up to date `blockBuilder` right? Therefore you won't have receipts for non-proposer nodes...?

If I'm correct, you would need to execute transactions here in order to have receipts.

Stefan-Ethernal (changes requested):

consensus/polybft/fsm.go
goran-ethernal: Yes, but what happens if we do not insert these events for some reason? An error occurs. Should we move to another round and try again? To me that is a correct way to have a consistent state, but lets discuss it further on sync.

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/2905c39f46bf6a7af6d323b1e016924bb55a3573

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/2905c39f46bf6a7af6d323b1e016924bb55a3573

consensus/polybft/fsm.go
goran-ethernal: Yes, it's a copy/paste error. Good catch :100: . It should be f.block.Receipts there.

Fix: https://github.com/0xPolygon/polygon-edge/pull/852/commits/2905c39f46bf6a7af6d323b1e016924bb55a3573

consensus/polybft/fsm.go
Stefan-Ethernal: > Should we move to another round and try again?

Exactly. If `Insert` function fails for any reason, we are going to proceed to a new round where we would reset our internal state.
Although I don't see that we are re-injecting transactions back to the tx pool if `Insert` fails. That could be a problem, because in that case we would lose such transactions (with exit event logs).

consensus/polybft/consensus_runtime_test.go
```diff
@@ -1659,13 +1659,13 @@ func TestConsensusRuntime_getExitEventRootHash(t *testing.T) {
 		tree, err := NewMerkleTree(encodedEvents)
```
Stefan-Ethernal, line 1663: Add `t.Parallel()` (both in the scope `t.Run(...)` and `TestConsensusRuntime_getExitEventRootHash` itself).

Also `TestConsensusRuntime_GenerateExitProof` lacks parallel execution (cannot comment on it, because it was already added through prior PR).

consensus/polybft/fsm.go
```diff
@@ -648,3 +662,35 @@ func createStateTransactionWithData(target types.Address, inputData []byte) *typ
...
+func getExitEventsFromReceipts(epoch, block uint64, receipts []*types.Receipt) ([]*ExitEvent, error) {
+	events := make([]*ExitEvent, 0)
+
+	for i := 0; i < len(receipts); i++ {
+		if len(receipts[i].Logs) == 0 {
+			continue
+		}
+
+		for j := 0; j < len(receipts[i].Logs); j++ {
+			event, err := decodeExitEvent(convertLog(receipts[i].Logs[j]), epoch, block)
```
b00ris, line 676: The final thing there. We need to be sure that the event was from our exit contract. Do you have this check somewhere? I havent managed to find.

b00ris (approved):
Good job!

consensus/polybft/fsm.go
Stefan-Ethernal: This should do the trick: https://github.com/0xPolygon/polygon-edge/pull/852/files#diff-514a0d26a7bee06b50bd0d4d72613aa918eb7ae096be6ea973ea6e2680d4dd75R132. Hopefully there wouldn't be the same function signature on some other smart contracts? :slightly_smiling_face: 

@goran-ethernal However maybe we can try figuring out originating address, although chances are pretty low that we would have the same event signature emitted from multiple contracts.

consensus/polybft/state.go
```diff
@@ -335,20 +358,38 @@ func (s *State) list() ([]*StateSyncEvent, error) {
 	return events, nil
 }
 
-// insertStateSyncEvent inserts a new state sync event to state event bucket in db
-func (s *State) insertExitEvent(event *ExitEvent) error {
+// insertExitEvents inserts collection of exit events to exit event bucket in bolt db
```
```suggestion
// insertExitEvents inserts exit events slice to exit event bucket in bolt db
```

consensus/polybft/state_test.go
```diff
@@ -420,6 +421,53 @@ func TestState_Insert_And_Get_ExitEvents_ForProof_NoEvents(t *testing.T) {
 	assert.Nil(t, events)
 }
 
+func TestState_decodeExitEvent(t *testing.T) {
+	const (
```
Stefan-Ethernal, line 427:
```suggestion
        t.Parallel()
	const (
```

consensus/polybft/state_test.go
```diff
@@ -420,6 +421,53 @@ func TestState_Insert_And_Get_ExitEvents_ForProof_NoEvents(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, uint64(exitID), event.ID)
+	require.Equal(t, uint64(epoch), event.EpochNumber)
+	require.Equal(t, uint64(blockNumber), event.BlockNumber)
+
+	require.NoError(t, state.insertExitEvent(event))
+}
+
+func TestState_decodeExitEvent_NotAnExitEvent(t *testing.T) {
+	topics := make([]ethgo.Hash, 4)
```
Stefan-Ethernal, line 462:
```suggestion 
        t.Parallel()
	topics := make([]ethgo.Hash, 4)
```

Stefan-Ethernal (approved):
LGTM, major stuff is addressed.

consensus/polybft/fsm.go
goran-ethernal: Let's add that check of originating address once we implement deploying of checkpoint contracts. In my understanding, contracts will be deployed as part of genesis, and we will have a known (fixed) address, which we can use here to check if the address match. I'll write this down so we do not forget.

ferranbt: > event signature emitted from multiple contracts

This could be an attack vector on the exit contract if we do not hardcode the address. Anyone could create the same event and emit it, it is not per-se about the chances of generating the same event signature.

ferranbt (approved):

consensus/polybft/fsm.go
goran-ethernal: Created a separate issue for this address check, once we implement deployment of the exit contract:

https://polygon.atlassian.net/browse/EVM-162

consensus/polybft/fsm.go
goran-ethernal: Ok, so agreement here is to go as it is. The issue with not re-injecting transactions back to the tx pool if finalization of block fails will be addressed through separate issue, since it's not really an issue on checkpointing, but an existing one.

PR #854, opened 2022-10-31
Introduce block builder

ferranbt:
# Description

This PR introduces the `block builder` object to construct blocks with transactions from the pool. Any consensus protocol will use this object to abstract functionality.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/blockbuilder/blockbuilder.go
+++ b/blockbuilder/blockbuilder.go
@@ -66,6 +66,8 @@ func NewBlockBuilder(params *BlockBuilderParams) (*BlockBuilder, error) {
 
 	if params.Logger == nil {
 		params.Logger = hclog.NewNullLogger()
+	} else {
+		params.Logger = params.Logger.Named("block-builder")
 	}
 
 	builder := &BlockBuilder{
--- a/blockbuilder/blockbuilder.go
+++ b/blockbuilder/blockbuilder.go
@@ -1,0 +1,247 @@
+package blockbuilder
+
+import (
+	"github.com/0xPolygon/polygon-edge/consensus"
+	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
+	"github.com/0xPolygon/polygon-edge/helper/common"
+	"github.com/0xPolygon/polygon-edge/state"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/hashicorp/go-hclog"
+)
+
+// TODO: Add open tracing
+
+const (
+	BlockGasTargetDivisor uint64 = 1024 // The bound divisor of the gas limit, used in update calculations
+)
+
+// BlockBuilderParams are fields for the block that cannot be changed
+type BlockBuilderParams struct {
+	// Parent block
+	Parent *types.Header
+
+	// Executor. TODO: Abstract executor
+	Executor *state.Executor
+
+	// Coinbase that is signing the block
+	Coinbase types.Address
+
+	// Vanity extra for the block
+	VanityExtra []byte
+
+	// GasLimit is the gas limit for the block
+	BlockGasTarget uint64
+
+	// Logger
+	Logger hclog.Logger
+
+	// txPoolInterface implementation
+	TxPool txPoolInterface
+}
+
+type blockchainInterface interface {
+	// WriteBlock writes the block to the db but does not process it
+	WriteBlock(b *types.Block, source string) error
+}
+
+type txPoolInterface interface {
+	Prepare()
+	Peek() *types.Transaction
+	Pop(tx *types.Transaction)
+	Drop(tx *types.Transaction)
+	Demote(tx *types.Transaction)
+}
+
+func NewBlockBuilder(params *BlockBuilderParams) (*BlockBuilder, error) {
+	// vanity extra can only be 32 size max. it is better to trim that to return
+	// an error that we have to propagate. It should be up to higher level
+	// code to error if the vanity extra supplied by the user is too big.
+	if len(params.VanityExtra) > 32 {
+		params.VanityExtra = params.VanityExtra[:32]
+	}
+
+	if params.VanityExtra == nil {
+		params.VanityExtra = make([]byte, 0)
+	}
+
+	if params.Logger == nil {
+		params.Logger = hclog.NewNullLogger()
+	}
+
+	builder := &BlockBuilder{
+		params: params,
+	}
+	if err := builder.Reset(); err != nil {
+		return nil, err
+	}
+
+	return builder, nil
+}
+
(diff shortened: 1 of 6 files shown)


blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,249 @@
...
+
+	b.block.Header.ComputeHash()
+
+	return &StateBlock{
+		Block:    b.block,
+		Receipts: b.state.Receipts(),
+	}
+}
+
+// StateBlock is a block with the full state it modifies
```
igorcrevar, line 216: I think we can remove `StateBlock`, receipts can be retrieved from `types.Block`

consensus/ibft/consensus_backend.go
```diff
@@ -127,209 +125,46 @@ func (i *backendIBFT) Quorum(blockNumber uint64) uint64 {
...
-		if err := transition.WriteFailedReceipt(tx); err != nil {
-			i.logger.Error(
-				fmt.Sprintf(
-					"unable to write failed receipt for transaction %s",
-					tx.Hash,
-				),
-			)
+	stateBlock := bbuilder.Build(func(header *types.Header) {
+		if err := i.currentHooks.ModifyHeader(header, i.currentSigner.Address()); err != nil {
+			panic(err)
```
igorcrevar, line 151: Maybe we should change `Build` signature to
`func (b *BlockBuilder) Build(handler func(*types.Header) error) (*StateBlock, error) ` and get rid of panic

blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,249 @@
...
+		tx := b.params.TxPool.Peek()
+
+		// execute transactions one by one
+		finished, err := b.writeTxPoolTransaction(tx)
+		if err != nil {
+			b.params.Logger.Debug("Fill transaction error", "hash", tx.Hash, "err", err)
+		}
+
+		if finished {
+			break write
```
Kourin1996: I thought it could return here instead of break statement

blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (b *BlockBuilder) Reset() error {
+	// Generate the base block
+	header := &types.Header{
+		ParentHash: b.params.Parent.Hash,
+		Number:     b.params.Parent.Number + 1,
+		Miner:      types.ZeroAddress.Bytes(),
+		Nonce:      types.Nonce{},
+		MixHash:    signer.IstanbulDigest,
+		Difficulty: b.params.Parent.Number + 1,
```
igorcrevar, line 107: Is this copypaste error or difficulty should not be just
`Difficulty:   1,` ?

blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,249 @@
...
+	// transactions are the data included in the block
+	txns []*types.Transaction
+
+	// block is a reference to the already built block
+	block *types.Block
+
+	state *state.Transition
+}
+
+func (b *BlockBuilder) Reset() error {
```
Kourin1996, line 99: `Reset` seems not to be called after BlockBuilder is used. Right now `Init` is proper?

igorcrevar (changes requested):

blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+
+	b.state = transition
+
+	return nil
+}
+
+// Fill fills the block with transactions from the txpool
+func (b *BlockBuilder) Fill() error {
```
Kourin1996: Basically, `NewBlockBuilder()`, `Fill()`, and `BuildBlock()` seems to be called sequentially. How about defining helper function that takes argument and reference and returns constructed block?

Kourin1996 (commented):
Just comment & questions

blockbuilder/blockbuilder.go
igorcrevar: Also, I think `MixHash:    signer.IstanbulDigest,` should be passed by block builder params and not hard-coded

igorcrevar: And in polybft case we need to set `Miner`, so probably there should be
`Miner:        b.params.Coinbase[:],`

blockbuilder/blockbuilder.go
igorcrevar: in old v3, `Fill` was called inside `BuildBlock`, i think. but I think we should be able to call `NewBlockBuilder` and `BuildBlock` separately (at least for polybft because we are not creating and building from the same place)

blockbuilder/blockbuilder.go
ferranbt: Agree with the last two points. For difficulty, I think we will stick with Difficulty=Number since that is what Ethereum is doing also port-merge.

blockbuilder/blockbuilder.go
ferranbt: We use this functionality in the `v3-parity` branch to create a single instance of `block-builder` and reset it during rounds. With this, we manage to compute the block only once during sealing. I would keep it since we can streamline the review later on of that branch.

blockbuilder/blockbuilder.go
ferranbt: `NewBlockBuilder` does not need to be instantiated for each built block. It is done as part of this PR for simplicity. `v3-parity` branch has only one instance per sequence that gets reset.

As for `Fill` and `Build`. In `v3-parity` there is an extra step in between to include custom transactions (state transactions) before the block gets built. So, it does not make sense to unify both.

goran-ethernal: We are not returning any error in Fill, only nil, so I would remove this return value from the function signature.

blockbuilder/blockbuilder.go
Kourin1996: Noted. Will resolve

Kourin1996 (approved):

igorcrevar (approved):
approved

blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,246 @@
...
+		return err
+	}
+
+	b.txns = append(b.txns, tx)
+
+	return nil
+}
+
+func (b *BlockBuilder) writeTxPoolTransaction(tx *types.Transaction) (bool, error) {
+	if tx == nil {
```
b00ris, line 159: Could you move this check to the Fill method and add logging? 
If we have nil transactions there - it shouldnt be silent

b00ris (approved):

PR #861, opened 2022-11-01
Checkpoint submission

Stefan-Ethernal:
# Description

This PR implements checkpoint data submission to the rootchain contract.
Checkpoint data are submitted for the epoch-ending blocks and at each N blocks, where N represents some predefined value.

Checkpoint blocks are described on the image below:
image

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
-	"time"
 
 	"github.com/0xPolygon/polygon-edge/command/rootchain/helper"
 	"github.com/0xPolygon/polygon-edge/types"
@@ -25,7 +24,7 @@ var (
 		"bytes32 eventRoot, tuple(address _address, uint256[4] blsKey)[] nextValidators" + ")")
 
 	// frequency at which checkpoints are sent to the rootchain
-	checkpointTimeInterval = 30 * time.Minute
+	defaultCheckpointsOffset = uint64(900)
 )
 
 // checkpointManager encapsulates logic for checkpoint data submission
@@ -43,7 +42,7 @@ type checkpointManager struct {
 }
 
 // newCheckpointManager creates a new instance of checkpointManager
-func newCheckpointManager(sender types.Address, blockTime time.Duration, interactor rootchainInteractor,
+func newCheckpointManager(sender types.Address, checkpointOffset uint64, interactor rootchainInteractor,
 	blockchain blockchainBackend, backend polybftBackend) *checkpointManager {
 	r := interactor
 	if interactor == nil {
@@ -55,36 +54,34 @@ func newCheckpointManager(sender types.Address, blockTime time.Duration, interac
 		blockchain:        blockchain,
 		consensusBackend:  backend,
 		rootchain:         r,
-		checkpointsOffset: uint64(checkpointTimeInterval.Milliseconds() / blockTime.Milliseconds()),
+		checkpointsOffset: checkpointOffset,
 	}
 }
 
 // getCurrentCheckpointID queries CheckpointManager smart contract and retrieves current checkpoint id
-func (c checkpointManager) getCurrentCheckpointID(epochNumber uint64) (uint64, error) {
+func (c checkpointManager) getCurrentCheckpointID() (uint64, error) {
 	checkpointIDMethodEncoded, err := currentCheckpointIDMethod.Encode([]interface{}{})
 	if err != nil {
-		return 0, fmt.Errorf("failed to encode currentCheckpointId function parameters for epoch=%d: %w",
-			epochNumber, err)
+		return 0, fmt.Errorf("failed to encode currentCheckpointId function parameters: %w", err)
 	}
 
 	currentCheckpointID, err := c.rootchain.Call(c.sender, helper.CheckpointManagerAddress, checkpointIDMethodEncoded)
 	if err != nil {
-		return 0, fmt.Errorf("failed to invoke currentCheckpointId function on the rootchain for epoch=%d: %w",
-			epochNumber, err)
+		return 0, fmt.Errorf("failed to invoke currentCheckpointId function on the rootchain: %w", err)
 	}
 
 	checkpointID, err := strconv.ParseUint(currentCheckpointID, 0, 64)
 	if err != nil {
-		return 0, fmt.Errorf("failed to convert current checkpoint id '%s' to number for epoch=%d: %w",
-			currentCheckpointID, epochNumber, err)
+		return 0, fmt.Errorf("failed to convert current checkpoint id '%s' to number: %w",
+			currentCheckpointID, err)
 	}
 
 	return checkpointID, nil
 }
 
 // submitCheckpoint sends a transaction which with checkpoint data to the rootchain
-func (c checkpointManager) submitCheckpoint(latestHeader types.Header, epochNumber uint64) error {
-	checkpointID, err := c.getCurrentCheckpointID(epochNumber)
+func (c checkpointManager) submitCheckpoint(latestHeader types.Header) error {
+	checkpointID, err := c.getCurrentCheckpointID()
 	if err != nil {
 		return err
 	}
@@ -200,7 +197,12 @@ func (c *checkpointManager) setCheckpointsOffset(checkpointsOffset uint64) {
 // isCheckpointBlock returns true for epoch ending blocks and
 // blocks in the middle of the epoch which are offseted by predefined count of blocks
 func (c *checkpointManager) isCheckpointBlock(header types.Header) (bool, error) {
-	if header.Number%c.checkpointsOffset == 0 {
+	lastCheckpointBlockNumber, err := c.getCurrentCheckpointID()
+	if err != nil {
+		return false, err
+	}
+
+	if lastCheckpointBlockNumber+c.checkpointsOffset == header.Number {
 		return true, nil
 	}
 
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -141,7 +141,7 @@ func (c checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfEp
 			continue
 		}
 
-		err = c.submitCheckpointInternal(nonce, txn, *currentHeader, *currentExtra, true)
+		err = c.submitCheckpointInternal(nonce, txn, *parentHeader, *parentExtra, true)
 		if err != nil {
 			return err
 		}
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -2,6 +2,7 @@ package polybft
 
 import (
 	"fmt"
+	"math/big"
 	"strconv"
 
 	"github.com/0xPolygon/polygon-edge/command/rootchain/helper"
@@ -24,55 +25,148 @@ var (
 )
 
 type checkpointManager struct {
-	sender types.Address
+	sender           types.Address
+	blockchain       blockchainBackend
+	rootchain        rootchainInteractor
+	consensusBackend polybftBackend
+	epochSize        uint64
 }
 
 // getCurrentCheckpointID queries CheckpointManager smart contract and retrieves current checkpoint id
 func (c checkpointManager) getCurrentCheckpointID(epochNumber uint64) (uint64, error) {
 	checkpointIDMethodEncoded, err := currentCheckpointIDMethod.Encode([]interface{}{})
 	if err != nil {
-		return 0, fmt.Errorf("failed to encode currentCheckpointId function invocation for epoch=%d: %w",
+		return 0, fmt.Errorf("failed to encode currentCheckpointId function parameters for epoch=%d: %w",
 			epochNumber, err)
 	}
 
-	currentCheckpointID, err := helper.Call(ethgo.Address(c.sender),
-		ethgo.Address(helper.CheckpointManagerAddress), checkpointIDMethodEncoded)
+	currentCheckpointID, err := c.rootchain.Call(c.sender, helper.CheckpointManagerAddress, checkpointIDMethodEncoded)
 	if err != nil {
-		return 0, fmt.Errorf("failed to invoke currentCheckpointId on the rootchain for epoch=%d: %w",
+		return 0, fmt.Errorf("failed to invoke currentCheckpointId function on the rootchain for epoch=%d: %w",
 			epochNumber, err)
 	}
 
 	checkpointID, err := strconv.ParseUint(currentCheckpointID, 0, 64)
 	if err != nil {
-		return 0, fmt.Errorf("failed to convert current checkpoint id '%s' to number: %w", currentCheckpointID, err)
+		return 0, fmt.Errorf("failed to convert current checkpoint id '%s' to number for epoch=%d: %w",
+			currentCheckpointID, epochNumber, err)
 	}
 
 	return checkpointID, nil
 }
 
-// submitCheckpoint sends a transaction which contains checkpoint data to the rootchain
-func (c *checkpointManager) submitCheckpoint(header *types.Header) error {
-	extra, err := GetIbftExtra(header.ExtraData)
+// submitCheckpoint sends a transaction which with checkpoint data to the rootchain
+func (c checkpointManager) submitCheckpoint(latestHeader types.Header, epochNumber uint64) error {
+	checkpointID, err := c.getCurrentCheckpointID(epochNumber)
 	if err != nil {
 		return err
 	}
 
-	// TODO: REMOVE
-	fmt.Println(extra.Checkpoint)
+	pendingNonce, err := c.rootchain.GetPendingNonce(c.sender)
+	if err != nil {
+		return err
+	}
 
-	// TODO: Spawn a routine
-	// TODO: Encode input
 	checkpointManagerAddr := ethgo.Address(helper.CheckpointManagerAddress)
 	txn := &ethgo.Transaction{
 		To: &checkpointManagerAddr,
 	}
-	pendingNonce, err := helper.GetPendingNonce(c.sender)
 
+	// detect any pending (previously failed) checkpoints and send them
+	for blockNum := checkpointID + 1; blockNum < latestHeader.Number; blockNum++ {
+		header, exists := c.blockchain.GetHeaderByNumber(blockNum)
+		if !exists {
+			return fmt.Errorf("block %d was not found", blockNum)
+		}
+
+		extra, err := GetIbftExtra(header.ExtraData)
+		if err != nil {
+			return err
+		}
+
+		// send checkpoints only for epoch ending blocks (validators delta is non-empty)
+		if extra.Validators.IsEmpty() {
+			continue
+		}
+
+		err = c.submitCheckpointInternal(pendingNonce, txn, *header, *extra)
+		if err != nil {
+			return err
+		}
+		pendingNonce++
+	}
+
+	extra, err := GetIbftExtra(latestHeader.ExtraData)
 	if err != nil {
 		return err
 	}
 
-	helper.SendTxn(pendingNonce, txn)
+	return c.submitCheckpointInternal(pendingNonce, txn, latestHeader, *extra)
+}
 
+// submitCheckpointInternal encodes checkpoint data for the given block and
+// sends a transaction to the CheckpointManager rootchain contract
+func (c *checkpointManager) submitCheckpointInternal(nonce uint64, txn *ethgo.Transaction, header types.Header, extra Extra) error {
+	input, err := c.abiEncodeCheckpointBlock(header, extra)
+	if err != nil {
+		return fmt.Errorf("failed to encode checkpoint data to ABI for block %d: %w", header.Number, err)
+	}
+
+	txn.Input = input
+	receipt, err := c.rootchain.SendTransaction(nonce, txn)
+	if err != nil {
+		return err
+	}
+	if receipt.Status == uint64(types.ReceiptFailed) {
+		return fmt.Errorf("transaction execution failed for block %d", header.Number)
+	}
 	return nil
+}
+
+// abiEncodeCheckpointBlock encodes input data for submitCheckpoint function on CheckpointManager smart contract for a given header
+func (c *checkpointManager) abiEncodeCheckpointBlock(header types.Header, extra Extra) ([]byte, error) {
+	validators, err := c.consensusBackend.GetValidators(header.Number, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	validatorsAbi, err := validators.EncodeAbi()
+	if err != nil {
+		return nil, err
+	}
+
+	params := map[string]interface{}{
+		"chainID":             new(big.Int).SetUint64(c.blockchain.GetChainID()),
+		"aggregatedSignature": extra.Committed.AggregatedSignature,
+		"validatorsBitmap":    extra.Committed.Bitmap,
+		"epochNumber":         new(big.Int).SetUint64(extra.Checkpoint.EpochNumber),
+		"blockNumber":         new(big.Int).SetUint64(header.Number),
+		"blockHash":           header.Hash,
+		"blockRound":          new(big.Int).SetUint64(extra.Checkpoint.BlockRound),
+		"eventRoot":           extra.Checkpoint.EventRoot.Bytes(),
+		"nextValidators":      validatorsAbi,
+	}
+
+	return submitCheckpointMethod.Encode(params)
+}
+
+type rootchainInteractor interface {
+	Call(from types.Address, to types.Address, input []byte) (string, error)
+	SendTransaction(nonce uint64, transaction *ethgo.Transaction) (*ethgo.Receipt, error)
+	GetPendingNonce(address types.Address) (uint64, error)
+}
+
+type defaultRootchainInteractor struct {
+}
+
+func (d *defaultRootchainInteractor) Call(from types.Address, to types.Address, input []byte) (string, error) {
+	return helper.Call(ethgo.Address(from), ethgo.Address(to), input)
+}
+
+func (d *defaultRootchainInteractor) SendTransaction(nonce uint64, transaction *ethgo.Transaction) (*ethgo.Receipt, error) {
+	return helper.SendTxn(nonce, transaction)
+}
+
+func (d *defaultRootchainInteractor) GetPendingNonce(address types.Address) (uint64, error) {
+	return helper.GetPendingNonce(address)
 }
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -82,7 +82,7 @@ func (c checkpointManager) getCurrentCheckpointID() (uint64, error) {
 	return checkpointID, nil
 }
 
-// submitCheckpoint sends a transaction which with checkpoint data to the rootchain
+// submitCheckpoint sends a transaction with checkpoint data to the rootchain
 func (c checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfEpoch bool) error {
 	lastCheckpointBlockNumber, err := c.getCurrentCheckpointID()
 	if err != nil {
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -22,9 +22,6 @@ var (
 		"uint256 chainID, bytes aggregatedSignature, bytes validatorsBitmap, " +
 		"uint256 epochNumber, uint256 blockNumber, bytes32 blockHash, uint256 blockRound," +
 		"bytes32 eventRoot, tuple(address _address, uint256[4] blsKey)[] nextValidators" + ")")
-
-	burazBrat, _ = abi.NewMethod(
-		"function submitCheckpoint(tuple(address _address, uint256[4] blsKey)[] nextValidators)")
 )
 
 type checkpointManager struct {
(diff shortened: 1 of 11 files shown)


consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	return c.submitCheckpointInternal(nonce, txn, latestHeader, *extra)
+}
+
+// submitCheckpointInternal encodes checkpoint data for the given block and
+// sends a transaction to the CheckpointManager rootchain contract
+func (c *checkpointManager) submitCheckpointInternal(nonce uint64, txn *ethgo.Transaction,
+	header types.Header, extra Extra) error {
+	nextEpochValidators, err := c.consensusBackend.GetValidators(header.Number, nil)
```
goran-ethernal: If checkpoint is sent on a non-ending epoch block, `nextValidatorSet` would be the same. What did we agree, do we send an empty set or just send the same validator set? @b00ris

Stefan-Ethernal: We have agreed to send the same validator set at least when dealing with validator set hashes (https://github.com/0xPolygon/polygon-edge/pull/840#discussion_r1009087586). Therefore don't see why here would be any different, but I'm open for suggestions.

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,177 @@
...
+	currentCheckpointIDMethod, _ = abi.NewMethod("function currentCheckpointId() returns (uint256)")
+
+	// submitCheckpointMethod is an ABI method object representation for
+	// submit checkpoint function on CheckpointManager contract
+	submitCheckpointMethod, _ = abi.NewMethod("function submitCheckpoint(" +
+		"uint256 chainID, bytes aggregatedSignature, bytes validatorsBitmap, " +
+		"uint256 epochNumber, uint256 blockNumber, bytes32 blockHash, uint256 blockRound," +
+		"bytes32 eventRoot, tuple(address _address, uint256[4] blsKey)[] nextValidators" + ")")
+
+	burazBrat, _ = abi.NewMethod(
```
goran-ethernal: Remove this :joy:

consensus/polybft/consensus_runtime.go
```diff
@@ -115,22 +117,38 @@ type consensusRuntime struct {
 	// activeValidatorFlag indicates whether the given node is amongst currently active validator set
 	activeValidatorFlag uint32
 
+	// checkpointsOffset represents offset between checkpoint blocks (applicable only for non-epoch ending blocks)
+	checkpointsOffset uint64
```
goran-ethernal: I would move this `checkpointOffset` to `checkpointManager`, so that all checkpoint related configuration resides there.

consensus/polybft/consensus_runtime.go
```diff
@@ -20,9 +21,10 @@ import (
 )
 
 const (
-	eventsBufferSize   = 10
-	stateFileName      = "consensusState.db"
-	uptimeLookbackSize = 2 // number of blocks to calculate uptime from the previous epoch
+	eventsBufferSize       = 10
+	stateFileName          = "consensusState.db"
+	uptimeLookbackSize     = 2                // number of blocks to calculate uptime from the previous epoch
+	checkpointTimeInterval = 30 * time.Minute // frequency at which sending checkpoints are sent to the rootchain
```
goran-ethernal: I would move this `checkpointTimeInterval` to `checkpointManager`, so that all checkpoint related configuration resides there.

consensus/polybft/consensus_runtime.go
```diff
@@ -115,22 +117,38 @@ type consensusRuntime struct {
...
+		checkpointsOffset: uint64(checkpointTimeInterval.Milliseconds() / blockTimeInMillis),
 	}
 
 	if runtime.IsBridgeEnabled() {
 		err := runtime.startEventTracker()
 		if err != nil {
 			return nil, err
 		}
+
+		runtime.checkpointManager = &checkpointManager{
```
goran-ethernal: Can I suggest making a constructor function in `checkpointManager`, which will receive `rootchainInteractor` as a parameter. If its `nil`, we can use this `defaultRootchainInteractor` as an implementation. The constructor will also receive `blockTime` as a parameter, and if you move the default checkpoint submit time, you can calculate that `offset` in the constructor. That way you have that logic encapsulated in `checkpointManager` without the need that outside source knows all this.

consensus/polybft/consensus_runtime.go
```diff
@@ -646,9 +650,12 @@ func (c *consensusRuntime) deliverMessage(msg *TransportMessage) (bool, error) {
 	return true, nil
 }
 
-func (c *consensusRuntime) runCheckpoint(epoch *epochMetadata) error {
-	// TODO: Implement checkpoint
-	return nil
+// isCheckpointBlock returns indication whether given block is the checkpoint block.
+// Returns true for either epoch ending block or
+// at each N blocks, where N is calculated as division between predefined checkpoint interval and block time
+func (c *consensusRuntime) isCheckpointBlock(blockNumber uint64) bool {
```
goran-ethernal: I would move this as well to `checkpointManager`.

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Indeed!

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Fixed [6134a1a](https://github.com/0xPolygon/polygon-edge/pull/861/commits/6134a1a9e438cb2c9a4174f781cdb85b0a2bf02c)

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: The question is what to do with isEndOfEpoch function? It is not needed only for checkpointing as you know...
`func (c *consensusRuntime) isEndOfEpoch(blockNumber uint64) bool`

Are you suggesting splitting isCheckpointBlock into two logical conditions (`isEndOfEpoch || isCheckpointBlock`)?

Stefan-Ethernal: Fixed [cc68ae7](https://github.com/0xPolygon/polygon-edge/pull/861/commits/cc68ae76ef9282aa39319fade6386257e8bbb0ea)

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed [cc68ae7](https://github.com/0xPolygon/polygon-edge/pull/861/commits/cc68ae76ef9282aa39319fade6386257e8bbb0ea)

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed [cc68ae7](https://github.com/0xPolygon/polygon-edge/pull/861/commits/cc68ae76ef9282aa39319fade6386257e8bbb0ea)

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed [cc68ae7](https://github.com/0xPolygon/polygon-edge/pull/861/commits/cc68ae76ef9282aa39319fade6386257e8bbb0ea)

consensus/polybft/consensus_runtime.go
goran-ethernal: `isEndOfEpoch` just calls `isEndOfPeriod` function (which is not a method). You can call it from `checkpointManager`, just pass the epoch size to `checkpointManager.isCheckpointBlock` method.

consensus/polybft/checkpoint_manager.go
b00ris: We agreed to keep to sending  next validator set = current validator for non-end-of-epoch blocks.

b00ris: I think we should we ask to add a wrapper for submitCheckpoint without sending full validator set(send only nextValidatorSetHash).

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,209 @@
...
+			return fmt.Errorf("block %d was not found", blockNum)
+		}
+
+		extra, err := GetIbftExtra(header.ExtraData)
+		if err != nil {
+			return err
+		}
+
+		// send checkpoints only for epoch ending blocks (validators delta is non-empty)
+		if extra.Validators.IsEmpty() {
```
b00ris: Is it a marker of epoch ending?

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		return err
+	}
+
+	checkpointManagerAddr := ethgo.Address(helper.CheckpointManagerAddress)
+	txn := &ethgo.Transaction{
+		To: &checkpointManagerAddr,
+	}
+
+	// detect any pending (previously failed) checkpoints and send them
+	for blockNum := checkpointID + 1; blockNum < latestHeader.Number; blockNum++ {
```
b00ris: hmm. Do we need to handle that eth node of a validator got stuck? We can get dozens of the same checkpoint events.

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,209 @@
...
+	nextValidators AccountSet) ([]byte, error) {
+	params := map[string]interface{}{
+		"chainID":             new(big.Int).SetUint64(c.blockchain.GetChainID()),
+		"aggregatedSignature": extra.Committed.AggregatedSignature,
+		"validatorsBitmap":    extra.Committed.Bitmap,
+		"epochNumber":         new(big.Int).SetUint64(extra.Checkpoint.EpochNumber),
+		"blockNumber":         new(big.Int).SetUint64(header.Number),
+		"blockHash":           header.Hash,
+		"blockRound":          new(big.Int).SetUint64(extra.Checkpoint.BlockRound),
+		"eventRoot":           extra.Checkpoint.EventRoot.Bytes(),
```
b00ris, line 207: Is it bytes or bytes32 in solidity?

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		"nextValidators":      nextValidators.AsGenericMaps(),
+	}
+
+	return submitCheckpointMethod.Encode(params)
+}
+
+// isCheckpointIntervalElapsed returns indication whether given block is the checkpoint block.
+// Returns true at each N blocks, where N is calculated
+// as division between predefined checkpoint interval and block time
+func (c *checkpointManager) isCheckpointIntervalElapsed(blockNumber uint64) bool {
```
b00ris: i think it's not correct, because it can be changed(or terminated earlier).

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Yes: only epoch ending blocks have non-empty validator set delta.

At first I was thinking to determine epoch ending blocks by epoch size, but it turned out that it cannot be easily fitted here, because we don't know whether `checkpointID` corresponds to the epoch-ending or some block within the epoch. Therefore I find this check more robust (plus it is going to work properly with slashing as well).

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: `eventRoot` is `bytes32` solidity type (https://github.com/0xPolygon/polygon-edge/pull/861/files#diff-fce12e7457acbf9d31265d9e0f71a4340b87f09cb3dd52e2f61df69d0bbc142dR25), but apparently `ethgo` knows how to deal with it internally.

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: > I think we should we ask to add a wrapper for submitCheckpoint without sending full validator set(send only nextValidatorSetHash).

@b00ris could you elaborate a bit more? `submitCheckpoint` function expects full validator set. Are you talking about adding additional function on smart contract, through which we could provide `nextValidatorSetHash` (although SC should have enough information to calculate it on its own)...?

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: This for loop basically sends all the epoch ending blocks which were failed at some of the previous occasions. 

> We can get dozens of the same checkpoint events.

To me, `checkpointID` is the last checkpointed block. It is obtained from the smart contract (https://github.com/0xPolygon/polygon-edge/pull/861/files#diff-fce12e7457acbf9d31265d9e0f71a4340b87f09cb3dd52e2f61df69d0bbc142dR70). That should prevent sending same checkpoint data multiple times.

If you disagree, can you elaborate more what is the proper way to submit pending checkpoints?

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: We have agreed that this can be changed through on-chain governance, but for know `checkpointTimeInterval` is just a hardcoded const value (https://github.com/0xPolygon/polygon-edge/pull/861/files#diff-fce12e7457acbf9d31265d9e0f71a4340b87f09cb3dd52e2f61df69d0bbc142dR28).

Stefan-Ethernal: Fixed [732dd70](https://github.com/0xPolygon/polygon-edge/pull/861/commits/732dd70ad34721da848f89c8016ac39031ac3405). Added setter function for setting checkpoints offset.

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: This has changed per what we were talking this morning:
epoch ending blocks are determined by comparing epoch numbers in the checkpoint manager 
[732dd70](https://github.com/0xPolygon/polygon-edge/pull/861/commits/732dd70ad34721da848f89c8016ac39031ac3405)

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// detect any pending (previously failed) checkpoints and send them
+	for blockNum := checkpointID + 1; blockNum < latestHeader.Number; blockNum++ {
+		header, exists := c.blockchain.GetHeaderByNumber(blockNum)
+		if !exists {
+			return fmt.Errorf("block %d was not found", blockNum)
+		}
+
+		isEndOfEpoch, extra, err := c.isEndOfEpoch(*header)
```
goran-ethernal: You can optimize this further. Since you are already getting headers here, you can have a variable above for loop, called `parentHeader`, where you will initially get the header of `checkpointID`. Then in for loop, you will get the header of the next block, and compare the epoch numbers. Then, that header of the block in for loop will become `parentHeader`, and just use it to compare it to the next block.
As it is now, you are getting headers for the same block two times.

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,256 @@
...
+	return c.submitCheckpointInternal(nonce, txn, latestHeader, *extra)
+}
+
+// submitCheckpointInternal encodes checkpoint data for the given block and
+// sends a transaction to the CheckpointManager rootchain contract
+func (c *checkpointManager) submitCheckpointInternal(nonce uint64, txn *ethgo.Transaction,
+	header types.Header, extra Extra) error {
+	nextEpochValidators := AccountSet{}
+
+	isEndOfEpoch, _, err := c.isEndOfEpoch(header)
```
goran-ethernal: This check will be changed as well if you are to take in consideration this [comment](https://github.com/0xPolygon/polygon-edge/pull/861/files#r1012748750).

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Fixed [bc66984](https://github.com/0xPolygon/polygon-edge/pull/861/commits/bc66984233759009527557df6d0951b5fdc6239c)

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Fixed [bc66984](https://github.com/0xPolygon/polygon-edge/pull/861/commits/bc66984233759009527557df6d0951b5fdc6239c)

goran-ethernal (approved):
LGTM :100: :rocket: !

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,271 @@
...
+
+// setCheckpointsOffset sets new checkpointsOffset value
+func (c *checkpointManager) setCheckpointsOffset(checkpointsOffset uint64) {
+	c.checkpointsOffset = checkpointsOffset
+}
+
+// isCheckpointBlock returns true for epoch ending blocks and
+// blocks in the middle of the epoch which are offseted by predefined count of blocks
+func (c *checkpointManager) isCheckpointBlock(header types.Header) (bool, error) {
+	if header.Number%c.checkpointsOffset == 0 {
```
b00ris: It think its not quite right.
Eg:
start epoch block - 3
epoch offset - 5
epoch size - 20
we need to send checkpoints on 8th, 13th, 18th, 23th blocks.

ferranbt (approved):

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Good catch. Although it seemed to be working good enough when chain is starting from the genesis block... 😕 

In order to make it robust, I'd need to query smart contract to make sure what is the last checkpointed block and to add offset to it and compare it against `header.Number`. Do yo have any better solution?

Stefan-Ethernal: Fix proposal [50353e1](https://github.com/0xPolygon/polygon-edge/pull/861/commits/50353e1fe7b98b9b7c853ef1ac50814e0c192485)

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,237 @@
...
+		if !found {
+			return fmt.Errorf("block %d was not found", blockNumber)
+		}
+
+		currentExtra, err := GetIbftExtra(currentHeader.ExtraData)
+		if err != nil {
+			return err
+		}
+
+		if parentHeader == nil {
```
b00ris: Could you move it outside the cycle? We need it only once.

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,237 @@
...
+	checkpointManagerAddr := ethgo.Address(helper.CheckpointManagerAddress)
+	txn := &ethgo.Transaction{
+		To: &checkpointManagerAddr,
+	}
+
+	var parentHeader *types.Header
+
+	var parentExtra *Extra
+
+	initialBlockNumber := currentCheckpointID + 1
```
b00ris: mb+2? We are sending checkpoint for currentCheckpointID also.

b00ris (approved):

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Fixed [0eb1c68](https://github.com/0xPolygon/polygon-edge/pull/861/commits/0eb1c687b6f146a1b015090c240e1d832f389c61)

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: No, `currentCheckpointID` is in fact the last checkpointed block (the one which is already checkpointed on some previous occasion). 
I've changed comment and variable name so it is more self explanatory: [0eb1c68](https://github.com/0xPolygon/polygon-edge/pull/861/commits/0eb1c687b6f146a1b015090c240e1d832f389c61)

b00ris (approved):

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+	checkpointID, err := strconv.ParseUint(currentCheckpointID, 0, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert current checkpoint id '%s' to number: %w",
+			currentCheckpointID, err)
+	}
+
+	return checkpointID, nil
+}
+
+// submitCheckpoint sends a transaction which with checkpoint data to the rootchain
```
vcastellm:
```suggestion
// submitCheckpoint sends a transaction with checkpoint data to the rootchain
```

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+func (c checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfEpoch bool) error {
+	lastCheckpointBlockNumber, err := c.getCurrentCheckpointID()
```
vcastellm, line 87: Doesn't this have a chicken and egg problem? What's the checkid at the begining? If it returns an error again and again we'll never be able to submit checkpoint?

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+
+	checkpointManagerAddr := ethgo.Address(helper.CheckpointManagerAddress)
+	txn := &ethgo.Transaction{
+		To: &checkpointManagerAddr,
+	}
+
+	initialBlockNumber := lastCheckpointBlockNumber + 1
+
+	var parentExtra *Extra
+
```
vcastellm:
```suggestion
```

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+	if err != nil {
+		return err
+	}
+
+	return c.submitCheckpointInternal(nonce, txn, latestHeader, *extra, isEndOfEpoch)
+}
+
+// submitCheckpointInternal encodes checkpoint data for the given block and
+// sends a transaction to the CheckpointManager rootchain contract
+func (c *checkpointManager) submitCheckpointInternal(nonce uint64, txn *ethgo.Transaction,
```
vcastellm: A little bit confusing semantics, can this be `submitCheckpointTxn` or `submitCheckpointRootchain` or `sendToRootchain` `encodeAndSend`?

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+		currentEpochNumber := currentExtra.Checkpoint.EpochNumber
+		parentHeader = currentHeader
+		parentExtra = currentExtra
+
+		// send pending checkpoints only for epoch ending blocks
+		if parentEpochNumber == currentEpochNumber {
+			continue
+		}
+
+		err = c.submitCheckpointInternal(nonce, txn, *parentHeader, *parentExtra, true)
```
vcastellm: Can this incur in very elevated fees if there are a lot of them?

Stefan-Ethernal: Yes you are right and I've mentioned it once in the RFC-30 comment. It would be better if we would be able to submit batch of checkpoints through a single transaction. However current proposal for checkpointing doesn't  have such concept of submitting multiple checkpoints at once, but one by one exclusively.

Hopefully there shouldn't be too much of pending checkpoints waiting to be re-submitted.

(cc @b00ris)

consensus/polybft/fsm_test.go
```diff
@@ -285,7 +287,8 @@ func TestFSM_BuildProposal_WithoutUptimeTxGood(t *testing.T) {
 	checkpointBackendMock.On("BuildEventRoot", mock.Anything, mock.Anything).Return(types.ZeroHash, nil).Once()
 
 	fsm := &fsm{parent: parent, blockBuilder: mBlockBuilder, config: &PolyBFTConfig{}, backend: &blockchainMock{},
```
vcastellm, line 289: A bit out of the scope but, thinking on having a constructor for the FSM...

consensus/polybft/checkpoint_manager_test.go
```diff
@@ -0,0 +1,265 @@
...
+import (
+	"errors"
+	"math/big"
+	"strconv"
+	"testing"
+
+	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
+	"github.com/0xPolygon/polygon-edge/consensus/polybft/bitmap"
+	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	"github.com/0xPolygon/polygon-edge/types"
```
vcastellm, line 16: Nitpick but, is it worth it to separate local imports from external ones?

consensus/polybft/checkpoint_manager_test.go
```diff
@@ -0,0 +1,265 @@
...
+			rootchainMock.On("Call", mock.Anything, mock.Anything, mock.Anything).
+				Return(c.checkpointID, c.returnError).
+				Once()
+
+			checkpointMgr := &checkpointManager{rootchain: rootchainMock}
+			actualCheckpointID, err := checkpointMgr.getCurrentCheckpointID()
+			if c.errSubstring == "" {
+				expectedCheckpointID, err := strconv.ParseUint(c.checkpointID, 0, 64)
+				require.NoError(t, err)
+				require.NoError(t, err)
```
vcastellm: Duplicate line

consensus/polybft/checkpoint_manager.go
vcastellm: We should throttle this I think.

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: You are right, but don't see easy solution for it. One approach could be that validators listen to `CheckpointDoneEvent` and track submitted vs processed checkpoints and have it stored to the local db. Although not sure if that approach is bulletproof either.

Do you have any other idea, which is relatively simple to implement, but more robust?

vcastellm (changes requested):
Some pending changes

consensus/polybft/checkpoint_manager_test.go
Stefan-Ethernal: VS Code auto-formatter formats it like this. I could manually change this file, but I'm sure there are more files accross entire codebase where is the same situation as here.

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Unfortunatelly I've tried it, but linter complains. 

> declarations should never be cuddled (wsl).

consensus/polybft/fsm_test.go
Stefan-Ethernal: There are quite a few fields on the FSM to be initialized, however not all of them are required for all the tests. Therefore I don't clearly see benefit of introducing a constructor fn for it.

consensus/polybft/checkpoint_manager_test.go
Stefan-Ethernal: Fixed [9db0e5e](https://github.com/0xPolygon/polygon-edge/pull/861/commits/9db0e5e4367c209d373969accd0e7e2c1f33d1fc)

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Fixed [9db0e5e](https://github.com/0xPolygon/polygon-edge/pull/861/commits/9db0e5e4367c209d373969accd0e7e2c1f33d1fc)

consensus/polybft/checkpoint_manager.go
Stefan-Ethernal: Interesting point. 

However, I would say we need separate brainstorming session(s) and RFC for any further optimizations. At first, we need to see is it likely for this to happen and then to agree on exact strategy how to mitigate it.

consensus/polybft/checkpoint_manager_test.go
vcastellm: If you separate them manually, goimport respects that separation, we'll later add this https://github.com/0xPolygon/go-ibft/pull/40/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R86-R87

Stefan-Ethernal: Fixed [8d0b4fb](https://github.com/0xPolygon/polygon-edge/pull/861/commits/8d0b4fb01e34b1cacf29839a2432a4768cf3fab0)

consensus/polybft/checkpoint_manager.go
vcastellm: Creating an issue to track this https://polygon.atlassian.net/browse/EVM-186?atlOrigin=eyJpIjoiM2ZmNTVjZmRhMDRkNDkzNzkxMDYwYjVjN2VkMjE4ODQiLCJwIjoiaiJ9

consensus/polybft/checkpoint_manager.go
```diff
@@ -0,0 +1,241 @@
...
+	checkpointManagerAddr := ethgo.Address(helper.CheckpointManagerAddress)
+	txn := &ethgo.Transaction{
+		To: &checkpointManagerAddr,
+	}
+
+	initialBlockNumber := lastCheckpointBlockNumber + 1
+
+	var parentExtra *Extra
+
+	var parentHeader *types.Header
```
vcastellm:
```suggestion
	var(
		parentExtra *Extra
		parentHeader *types.Header
	)
```

Stefan-Ethernal: Fixed [12bdc98](https://github.com/0xPolygon/polygon-edge/pull/861/commits/12bdc983b2ae5339f54701adf635c5ea2c818a78)

consensus/polybft/checkpoint_manager.go
vcastellm: FTR After talking with Stefan: SC is initialized with checkid = 0, so we're good to go.

vcastellm (approved):
LGTM

PR #873, opened 2022-11-05
Uniform jsonrpc type return

ferranbt:
# Description

This PR use the return types in the `jsonrpc` package instead of an explicit conversion inside the endpoint.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -400,6 +400,11 @@ func (e *Eth) GetStorageAt(
 	// Get the storage for the passed in location
 	result, err := e.store.GetStorage(header.StateRoot, address, index)
 	if err != nil {
+		//nolint:govet
+		if errors.As(err, &ErrStateNotFound) {
+			return argBytesPtr(types.ZeroHash[:]), nil
+		}
+
 		return nil, err
 	}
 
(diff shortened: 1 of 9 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -433,12 +427,12 @@ func (e *Eth) GetStorageAt(
 
 // GasPrice returns the average gas price based on the last x blocks
 // taking into consideration operator defined price limit
-func (e *Eth) GasPrice() (string, error) {
+func (e *Eth) GasPrice() (interface{}, error) {
```
JekaMas, line 430: why empty interface is better? what user cases it solves?

ferranbt: That is how all the endpoints are defined. The `jsonrpc` library uses reflection to decide which methods export. The convention for all the methods is to return `(interface{}, error)`. This change just respects that convention.

Stefan-Ethernal (approved):
Overall looks good to me, although unit tests are failing, because it is needed to change assertions (due to changes in return types `string` vs `arg...` 
https://github.com/0xPolygon/polygon-edge/actions/runs/3401432822/jobs/5677755750

ferranbt:
> Overall looks good to me, although unit tests are failing, because it is needed to change assertions (due to changes in return types `string` vs `arg...`

Thank you! I did fail before due to the multi-lock error so I missed this error.

vcastellm (approved):
LGTM

JekaMas (approved):

PR #877, opened 2022-11-05
Use the Jsonrpc types as the abstraction layer in the store interfaces

ferranbt:
# Description

Use the types from the `jsonrpc` package on the abstraction interface.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -22,9 +22,6 @@ type ethTxPoolStore interface {
 
 	// AddTx adds a new transaction to the tx pool
 	AddTx(tx *types.Transaction) error
-
-	// GetPendingTx gets the pending transaction from the transaction pool, if it's present
-	GetPendingTx(txHash types.Hash) (*types.Transaction, bool)
 }
 
 type ethStateStore interface {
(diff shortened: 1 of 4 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -37,22 +34,16 @@ type ethStateStore interface {
 
 type ethBlockchainStore interface {
 	// Header returns the current header of the chain (genesis if empty)
-	Header() *types.Header
-
-	// GetHeaderByNumber returns the header by number
-	GetHeaderByNumber(block uint64) (*types.Header, bool)
+	Header() *Block
```
b00ris, line 36: Why does header function return block? =)

jsonrpc/eth_endpoint.go
```diff
@@ -821,13 +675,13 @@ func (e *Eth) Unsubscribe(id string) (bool, error) {
 	return e.filterManager.Uninstall(id), nil
 }
 
-func (e *Eth) getBlockHeader(number BlockNumber) (*types.Header, error) {
+func (e *Eth) getBlockHeader(number BlockNumber) (*Block, error) {
```
b00ris, line 635: the same

jsonrpc/eth_endpoint.go
ferranbt: Yes, the name should be changed. The idea is that for the `jsonrpc` module there is no difference between `Block` and `Header`. Thus, we should return only `jsonrpc.Block`. The name has to change too.

jsonrpc/eth_endpoint.go
ferranbt: Same as before.

jsonrpc/eth_endpoint.go
ferranbt: Now that I think about it, `Header` is not a good name either for what it does. It should be `LatestHeader` or something like that.

PR #865, opened 2022-11-02
FSM that supports go-ibft backend interface

0xSasaPrsic:
# Description

Please provide a detailed description of what was done in this PR

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [ ] I have assigned this PR to myself
- [ ] I have added at least 1 reviewer
- [ ] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [ ] I have tested this code with the official test suite
- [x] I have tested this code manually

# Additional comments

Unit tests were not completed since there are more pending changes.

PR #880, opened 2022-11-07
Add voting power to the validators

Stefan-Ethernal:
# Description

PR adds `VotingPower` field to the `ValidatorMetadata` (previously referred as `ValidatorAccount`). Currently `VotingPower` is equal to staked amount of tokens. However in the future we may opt to use scaled staked amount as voting power instead.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/validator_account.go
+++ b/consensus/polybft/validator_account.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/big"
 	"reflect"
 
 	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
@@ -16,8 +17,9 @@ import (
 
 // ValidatorAccount represents a validator from the validator set
 type ValidatorAccount struct {
-	Address types.Address
-	BlsKey  *bls.PublicKey
+	Address     types.Address
+	BlsKey      *bls.PublicKey
+	VotingPower uint64
 }
 
 // Equals compares ValidatorAccount equality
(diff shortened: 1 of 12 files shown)


consensus/polybft/validator_account.go
```diff
@@ -14,83 +15,100 @@ import (
 	"github.com/umbracle/fastrlp"
 )
 
-// ValidatorAccount represents a validator from the validator set
-type ValidatorAccount struct {
-	Address types.Address
-	BlsKey  *bls.PublicKey
+// ValidatorMetadata represents a validator metadata (its public identity)
+type ValidatorMetadata struct {
```
vcastellm, line 19: If we are refactoring struct name it's worth it to refactor filename too.

b00ris (approved):

goran-ethernal (approved):

ferranbt (approved):

igorcrevar (approved):
approved

consensus/polybft/validator_account.go
Stefan-Ethernal: Agreed, although this file consists of `ValidatorMetadata` and `AccountSet`, so not sure what is the proper name for it. :thinking: 

Btw, I think that we should reorganize code differently, because `polybft` package became pretty large. As part of such initiative, it is worth figuring out if whether we should split `AccountSet` and `ValidatorMetadata` to a separate files.
Although this is not a high priority at the moment I'd say.

Stefan-Ethernal: Fixed through a follow up PR [3e02cbf](https://github.com/0xPolygon/polygon-edge/pull/865/commits/3e02cbfb20a8c1d93f82be235b8a8a751a36bdbc)

PR #882, opened 2022-11-08
EVM-187 Handling bundles on both sync and consensus

igorcrevar:
# Description
Commitments and bundles should be handled when block is inserted in both cases: by syncer and by consensus.
Also consensus runtime tests are fixed (and a little bit extended) in this pr

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -200,6 +200,11 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	// after the block has been written we reset the txpool so that the old transactions are removed
 	c.config.txPool.ResetWithHeaders(block.Header)
 
+	// handle commitment and bundles creation
+	if err := c.createCommitmentAndBundles(block.Transactions); err != nil {
+		c.logger.Error("on block inserted error", "err", err)
+	}
+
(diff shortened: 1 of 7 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -191,6 +196,37 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	}
 }
 
+func (c *consensusRuntime) createCommitmentAndBundles(txs []*types.Transaction) error {
+	if commitment, err := getCommitmentMessageSignedTx(txs); err != nil {
```
```suggestion
	commitment, err := getCommitmentMessageSignedTx(txs); err != nil {
	   return err
	}
	if err := c.state.insertCommitmentMessage(commitment); err != nil {
	   return fmt.Errorf("insert commitment message error: %w", err)
	}
```

ferranbt (approved):

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/882/commits/c6f11711f7c003c40d0866bbd74a45c061b10b4a

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #894, opened 2022-11-09
Proposer selection

0xSasaPrsic:
# Description

Implementation of proposer selection algorithm from tendermint.

# Changes include

# Breaking changes

# Checklist

## Testing


consensus/polybft/validator_set.go
```diff
@@ -18,46 +86,430 @@ type ValidatorSet interface {
...
+	if overflow {
+		if b > 0 {
+			return math.MinInt64
+		}
+
+		return math.MaxInt64
+	}
+
+	return c
 }
```
Stefan-Ethernal: It'd be nice if we would extract this away from `validator_set` and put it to some `math.go` in the `helper/common` package?

0xSasaPrsic: This is not ready for review.

consensus/polybft/validator_set.go
```diff
@@ -18,46 +86,430 @@ type ValidatorSet interface {
...
+func NewValidatorSet(valz AccountSet) (*validatorSet, error) {
+	var validators = make([]*ValidatorAccount, len(valz))
+	for i, v := range valz {
+		validators[i] = NewValidator(v)
+	}
+
+	validatorSet := &validatorSet{
+		validators: validators,
+		// TODO voting power
+		// votingPowerMap: make(map[pbft.NodeID]uint64, len(validators)),
```
b00ris: Should we merge something first?

Nemanja0x: @b00ris that votingPoweMap is fixed here: https://github.com/0xPolygon/polygon-edge/pull/904

consensus/polybft/consensus_runtime.go
```diff
@@ -354,6 +354,40 @@ func (c *consensusRuntime) FSM() error {
 	isEndOfSprint := c.isEndOfSprint(pendingBlockNumber)
 	isEndOfEpoch := c.isEndOfEpoch(pendingBlockNumber)
 
+	valSet, err := NewValidatorSet(epoch.Validators)
+	if err != nil {
+		return fmt.Errorf("cannot create validator set for fsm: %w", err)
+	}
+
+	if epoch.Number > 0 { // it must be at least epoc
```
b00ris: this loop needs clarification

consensus/polybft/consensus_runtime.go
```diff
@@ -354,6 +354,40 @@ func (c *consensusRuntime) FSM() error {
...
+		iterationNumber := uint64(0)
+		currentHeader := parent
+		lastBlockOfPreviousEpoch := getEndEpochBlockNumber(epoch.Number-1, c.config.PolyBFTConfig.EpochSize)
+
+		for currentHeader.Number > lastBlockOfPreviousEpoch {
```
b00ris: do we need block  lastBlockOfPreviousEpoch - 1 or lastBlockOfPreviousEpoch?

consensus/polybft/safe_math.go
```diff
@@ -0,0 +1,51 @@
+package polybft
+
+import "math"
```
b00ris, line 3: Is it 'tendermint's part?

consensus/polybft/consensus_runtime.go
b00ris: Can we extract it to a separate function?

b00ris (approved):

consensus/polybft/consensus_runtime.go
0xSasaPrsic: yes it can, also I think this part needs optimization
cc @Nemanja0x

consensus/polybft/safe_math.go
0xSasaPrsic: yes it is from tendermint and it is needed because voting power is int64 which we will change in the future

consensus/polybft/consensus_runtime.go
0xSasaPrsic: lastBlockOfPreviousEpoch because we are interested in the blocks from current epoch.

consensus/polybft/validator_set_test.go
```diff
@@ -0,0 +1,528 @@
...
+
+	proposer, err := vs.CalcProposer(0)
+	require.NoError(t, err)
+	assert.Equal(t, proposer, addresses[1])
+
+	// validate no changes to validator set positions
+	for i, v := range vs.Accounts() {
+		assert.Equal(t, addresses[i], v.Address)
+	}
+}
```
Stefan-Ethernal: Suggesting this change, as it would make unit test more compact (similar goes to other tests as well, feel free to rely on `testValidators` as it makes life a lot easier :slightly_smiling_face:):

```suggestion
func TestValSetIndex(t *testing.T) {
	t.Parallel()

	validators := newTestValidatorsWithAliases([]string{"A", "B", "C", "D", "E"}, []uint64{10, 100, 1, 50, 30})
	metadata := validators.getPublicIdentities()
	vs := validators.toValidatorSetWithError(t)

	// validate no changes to validator set positions
	for i, v := range vs.Accounts() {
		assert.Equal(t, metadata[i].Address, v.Address)
	}

	proposer, err := vs.CalcProposer(0)
	require.NoError(t, err)
	assert.Equal(t, proposer, metadata[1].Address)

	// validate no changes to validator set positions
	for i, v := range vs.Accounts() {
		assert.Equal(t, metadata[i].Address, v.Address)
	}
}
```

consensus/polybft/consensus_runtime.go
goran-ethernal: I don't think we need this condition. First epoch will always be epoch 1 when we start the chain (as per smart contracts).

consensus/polybft/consensus_runtime.go
```diff
@@ -354,6 +354,40 @@ func (c *consensusRuntime) FSM() error {
...
+			blockExtra, err := GetIbftExtra(currentHeader.ExtraData)
+			if err != nil {
+				return fmt.Errorf("cannot get ibft extra: %w", err)
+			}
+
+			iterationNumber += blockExtra.Round + 1 // because round 0 is one of the iterations
```
goran-ethernal: I think we could optimize this, by keeping the `iterationNumber` in `epochMetadata` which we can update on `OnBlockInserted`, and only call this logic if node was dropped and restarted. We won't have a problem if `syncer` inserts the block first (since `OnBlockInserted` is called both from `syncer` and on `InsertBlock` from consensus), because if block was already inserted `OnBlockInserted` would not be called.

This is just a food for thought, we don't need to do it right away.

consensus/polybft/validator_set.go
```diff
@@ -18,46 +77,365 @@ type ValidatorSet interface {
...
+		return fmt.Errorf("cannot call IncrementProposerPriority with non-positive times")
+	}
+
+	// Cap the difference between priorities to be proportional to 2*totalPower by
+	// re-normalizing priorities, i.e., rescale all priorities by multiplying with:
+	//  2*totalVotingPower/(maxPriority - minPriority)
+	vp, err := v.TotalVotingPower()
+
+	if err != nil {
+		return fmt.Errorf("cannot calculate total voting power")
```
goran-ethernal:
```suggestion
		return fmt.Errorf("cannot calculate total voting power: %w", err)
```

consensus/polybft/validator_set.go
```diff
@@ -18,46 +77,365 @@ type ValidatorSet interface {
...
+	return v.totalVotingPower, nil
+}
+
+func (v *validatorSet) updateWithChangeSet() error {
+	err := v.updateTotalVotingPower()
+	if err != nil {
+		return fmt.Errorf("cannot update total voting power: %w", err)
+	}
+	// Scale and center.
+	totalVotingPower, err := v.TotalVotingPower()
```
goran-ethernal: Do we need this `TotalVotingPower` call here, since `updateTotalVotingPower` recalculates it? I think it is enough to call `v.totalVotingPower`.

go.sum
```diff
@@ -50,6 +50,8 @@ filippo.io/edwards25519 v1.0.0-rc.1/go.mod h1:N1IkdkCkiLB6tki+MYJoSx2JTY9NUlxZE7
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
 github.com/0xPolygon/go-ibft v0.0.0-20221109165724-ff0b77dd5f69 h1:7WV/2aoWQSEVdu4SglqKuwdKylhq1Cl/zXX/9GjreY0=
 github.com/0xPolygon/go-ibft v0.0.0-20221109165724-ff0b77dd5f69/go.mod h1:uJ9EHbl+8StbRhvKgfxDextK32FlVNOiBnRPlAgrQK8=
+github.com/0xPolygon/go-ibft v0.0.0-20221109135749-49b0ca1c3b8b h1:nkyfDDOxlLdrp1lTG1kmpn6v2wGekWkGB8xXD+Bh9ck=
+github.com/0xPolygon/go-ibft v0.0.0-20221109135749-49b0ca1c3b8b/go.mod h1:uJ9EHbl+8StbRhvKgfxDextK32FlVNOiBnRPlAgrQK8=
```
Stefan-Ethernal: Just run `go mod tidy`, as these two lines are not needed.

consensus/polybft/validator_set_test.go
0xSasaPrsic: not sure what I'm missing but this does not compile since some methods used do not exist.

consensus/polybft/consensus_runtime.go
0xSasaPrsic: yes, this needs refactoring

consensus/polybft/validator_set.go
0xSasaPrsic: good catch, it is not necessary since it will be calculate

consensus/polybft/validator_set_test.go
Stefan-Ethernal: Your local branch isn't synchronized to the remote. You are lacking this commit: 55b8e2059b5f5f1a43b55e078ce7b4de484556eb.

Stefan-Ethernal: Your local branch isn't synchronized with the remote. You are lacking this commit locally: 55b8e2059b5f5f1a43b55e078ce7b4de484556eb.

consensus/polybft/consensus_runtime.go
```diff
@@ -354,6 +354,40 @@ func (c *consensusRuntime) FSM() error {
...
+			var found bool
+			currentHeader, found = c.config.blockchain.GetHeaderByNumber(currentHeader.Number - 1)
+
+			if !found {
+				return fmt.Errorf("cannot get header by number: %d", currentHeader.Number)
+			}
+		}
+
+		if iterationNumber > 0 {
+			err = valSet.IncrementProposerPriority(iterationNumber)
```
igorcrevar: Can we extract whole logic for incrementing proposer priority to separate function?

consensus/polybft/validator_set_test.go
0xSasaPrsic: why tests were not updated with this commit https://github.com/0xPolygon/polygon-edge/commit/55b8e2059b5f5f1a43b55e078ce7b4de484556eb

consensus/polybft/extra.go
```diff
@@ -32,6 +30,7 @@ type Extra struct {
 	Parent     *Signature
 	Committed  *Signature
 	Checkpoint *CheckpointData
+	Round      uint64 // round in which the block was sealed
```
goran-ethernal: We already have `BlockRound` in the `CheckpointData`. You should use that one instead. Also, this newly introduced Round isn't set anywhere.

consensus/polybft/consensus_runtime.go
Nemanja0x: Here is the task for it:  https://polygon.atlassian.net/browse/EVM-205

consensus/polybft/consensus_runtime.go
Nemanja0x: Do not really understand, it is already extracted: IncrementProposerPriority

consensus/polybft/consensus_runtime.go
Nemanja0x: @goran-ethernal :  https://polygon.atlassian.net/browse/EVM-205

consensus/polybft/extra.go
0xSasaPrsic: not sure I understand, this is from rfc-21, also do you think this is not used anywhere https://github.com/0xPolygon/polygon-edge/blob/EVM-154_proposer_selection/consensus/polybft/extra.go#L83

goran-ethernal: Not used, but set. I only see it set in code when unmarshallng the `Extra`.

0xSasaPrsic: ok, it was already moved to checkpoint data in the extra field so redundant field is removed.

Nemanja0x (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -394,6 +411,31 @@ func (c *consensusRuntime) FSM() error {
 	return nil
 }
 
+// getNumberOfIteration returns number of iteration that are needed for the proposer calculation
+func getNumberOfIteration(parent *types.Header, epochNumber uint64, c *consensusRuntime) (uint64, error) {
```
Stefan-Ethernal, line 415: Change this to method of `consensusRuntime`, since it is required as a parameter?

consensus/polybft/extra_test.go
```diff
@@ -122,7 +120,7 @@ func TestExtra_UnmarshalRLPWith_NegativeCases(t *testing.T) {
 
 		extra := &Extra{}
 		ar := &fastrlp.Arena{}
-		require.ErrorContains(t, extra.UnmarshalRLPWith(ar.NewArray()), "incorrect elements count to decode Extra, expected 5 but found 0")
+		require.ErrorContains(t, extra.UnmarshalRLPWith(ar.NewArray()), "incorrect elements count to decode Extra, expected 6 but found 0")
```
Stefan-Ethernal: This assert isn't correct, test is failing.

We should require this error message instead:
"incorrect elements count to decode Extra, expected 5 but found 0"

consensus/polybft/consensus_runtime.go
Nemanja0x: Everything regrading this thing will be done through: https://polygon.atlassian.net/browse/EVM-205

consensus/polybft/extra_test.go
Nemanja0x: fixed here: https://github.com/0xPolygon/polygon-edge/pull/894/commits/a1c83a536745353298e74dd3c360efd8be8f08ca

Stefan-Ethernal (approved):

goran-ethernal (approved):

PR #897, opened 2022-11-10
Dynamic epoch size

goran-ethernal:
# Description

Current behavior of `polybft` heavily relied on the fixed epoch size (`currentBlockNumber % fixedEpochSize == 0`), either to calculate `uptime` or to check if its the end of epoch. This will not work with slashing, since a slashing transaction in block means the end of the current epoch, and that can happen on any block, which means that epoch can have an arbitrary size.

Epoch ending block is now detected by using two things:
1. If a slashing transaction occurred (a TODO is left in code once we agree how is this detected on client).
2. If a fixed epoch size is met. To do this, we needed to save the first block of a given epoch in the `epochMetadata` field, since a previous epoch could have ended in slashing, which means that we can't just check `currentBlockNumber % fixedEpochSize == 0`.

# Changes include

# Checklist

## Testing


consensus/polybft/validator_metadata.go
```diff
@@ -262,13 +262,13 @@ func (as AccountSet) ApplyDelta(validatorsDelta *ValidatorSetDelta) (AccountSet,
 	// Append added validators
 	for _, addedValidator := range validatorsDelta.Added {
 		if validators.ContainsAddress(addedValidator.Address) {
-			return nil, fmt.Errorf("validator %v is already present in the validators snapshot", addedValidator.Address.String())
+			continue
```
Stefan-Ethernal: I think that we should at least log this on warning level (if not returning an error). 

There should never be more than one same validators (and here we are using address as an identifier) added to the delta.

consensus/polybft/runtime_helpers.go
```diff
@@ -12,36 +19,54 @@ func getQuorumSize(validatorsCount int) int {
...
 
-// getEpochNumber returns epoch number for given blockNumber and epochSize.
-// Epoch number is derived as a result of division of block number and epoch size.
-// Since epoch number is 1-based (0 block represents special case zero epoch),
-// we are incrementing result by one for non epoch-ending blocks.
-func getEpochNumber(blockNumber, epochSize uint64) uint64 {
-	if isEndOfPeriod(blockNumber, epochSize) {
-		return blockNumber / epochSize
+// getFirstBlockOfEpoch returns the first block of epoch in which provided header resides
+func getFirstBlockOfEpoch(epochID uint64, header *types.Header, getBlock blockGetter) (uint64, error) {
```
Stefan-Ethernal: Maybe looks more natural to be a part of the `consensusRuntime` or `polybftBackend` (maybe latter that would be the best approach), talking because of this `getBlock` delegate you are sending?
Or if you don't want that, maybe send `blockchainBackend` instance instead?

consensus/polybft/runtime_helpers.go
```diff
@@ -12,36 +19,54 @@ func getQuorumSize(validatorsCount int) int {
 	return (validatorsCount + 1) / 2
 }
 
-// calculateFirstBlockOfPeriod calculates the first block of a period
-func calculateFirstBlockOfPeriod(currentBlockNumber, periodSize uint64) uint64 {
-	if currentBlockNumber <= periodSize {
-		return 1 // it's the first epoch
+// getBlockData returns block header and extra
+func getBlockData(blockNumber uint64,
```
Stefan-Ethernal: Similar goes as with `getFirstBlockOfEpoch`, it is strange that this is a helper function. To me it would be better to either have some other abstraction (maybe make it part of `polybftBackend`) or to move it to `consensusRuntime` (although I'm aware that `consensusRuntime` is huge).

consensus/polybft/validator_metadata.go
goran-ethernal: I returned the old behavior.

command/genesis/polybft_params.go
```diff
@@ -249,7 +249,7 @@ func generateExtraDataPolyBft(validators []GenesisTarget) []byte {
 		}
 	}
 
-	extra := polybft.Extra{Validators: delta}
+	extra := polybft.Extra{Validators: delta, Checkpoint: &polybft.CheckpointData{}}
```
Stefan-Ethernal: Just thinking out loud:
you don't need to set `NextValidatorsHash`, since genesis block is not going to be checkpointed. Although maybe we could set it (who knows whether we are going to need it eventually), not sure...

consensus/polybft/runtime_helpers.go
Stefan-Ethernal: Nitpick:
rename `epochID` to `epochNumber` (consistency)

consensus/polybft/runtime_helpers.go
```diff
@@ -12,36 +19,54 @@ func getQuorumSize(validatorsCount int) int {
...
+	blockExtra, err := GetIbftExtra(header.ExtraData)
+
+	if err != nil {
+		return 0, err
+	}
+
+	if epochID != blockExtra.Checkpoint.EpochNumber {
+		// its a regular epoch ending. No out of sync happened
+		return header.Number + 1, nil
+	}
```
Stefan-Ethernal: This seems a little bit odd, so just want to clear it to myself:
1. `header` is the last block of epoch N? therefore you are returning first block of next epoch basically?
2. Should this check be like this (more strict than it is currently)?
`if blockExtra.Checkpoint.EpochNumber == epochID+1`

consensus/polybft/runtime_helpers.go
Stefan-Ethernal: Also I'd probably rename `header` to `latestHeader`. WDYT?

consensus/polybft/validators_snapshot.go
```diff
@@ -38,58 +36,55 @@ func (v *validatorsSnapshotCache) GetSnapshot(blockNumber uint64, parents []*typ
...
 
 	var snapshot AccountSet
 
-	currentEpochNumber := epochNumber
+	lastBlockWithSnapshot := blockNumber
 
-	for ; ; currentEpochNumber-- {
-		snapshot = v.snapshots[currentEpochNumber]
+	for ; ; lastBlockWithSnapshot-- {
+		snapshot = v.snapshots[lastBlockWithSnapshot]
```
Stefan-Ethernal: If I'm correct, we could still persist snapshots per epoch, since validators are (still) changed per epoch (although we would need recognize epoch ending block by extracting `header.Extra` and comparing its `epochNumber`).

Of course we should weigh complexity vs benefits. This approach would give us better diversified epochs in the cache (and therefore better performances if we would like to query validators for different epochs). With this approach, we would commonly end up having snapshots which corresponds to the blocks of the same epoch.

consensus/polybft/consensus_runtime.go
```diff
@@ -870,9 +847,12 @@ func (c *consensusRuntime) isActiveValidator() bool {
...
 
-// isEndOfEpoch checks if an end of an epoch is reached with the current block
-func (c *consensusRuntime) isEndOfEpoch(blockNumber uint64) bool {
-	return isEndOfPeriod(blockNumber, c.config.PolyBFTConfig.EpochSize)
+// isFixedSizeOfEpochMet checks if epoch reached its end that was configured by its default size
+// this is only true if no slashing occurred in the given epoch
+func (c *consensusRuntime) isFixedSizeOfEpochMet(blockNumber uint64) bool {
+	epoch := c.getEpoch()
+
+	return (blockNumber - c.config.PolyBFTConfig.EpochSize + 1) == epoch.FirstBlockInEpoch
```
b00ris: It'd be great to do the check without subtraction. There will be uint64 underflow there for the first epoch. I don't see any bugs there, but i dont want to rely on it.

consensus/polybft/validators_snapshot.go
```diff
@@ -38,58 +36,55 @@ func (v *validatorsSnapshotCache) GetSnapshot(blockNumber uint64, parents []*typ
...
-		epochNumber--
-	}
-
-	v.logger.Trace("Retrieving snapshot started...", "Block", blockNumber, "Epoch", epochNumber)
+	v.logger.Trace("Retrieving snapshot started...", "Block", blockNumber)
 
 	var snapshot AccountSet
 
-	currentEpochNumber := epochNumber
+	lastBlockWithSnapshot := blockNumber
```
b00ris: We know here epoch number. Could we store validator set only for epochs and dont clean them?

b00ris: Is it worth to make https://polygon.atlassian.net/browse/EVM-155 there?

consensus/polybft/validators_snapshot.go
```diff
@@ -38,58 +36,55 @@ func (v *validatorsSnapshotCache) GetSnapshot(blockNumber uint64, parents []*typ
...
-		if currentEpochNumber == 0 {
+		if lastBlockWithSnapshot == 0 {
 			// prevent underflow of uint64
 			break
 		}
 	}
 
-	if snapshot != nil && currentEpochNumber == epochNumber {
+	if snapshot != nil && lastBlockWithSnapshot == blockNumber {
+		if err := v.cleanup(); err != nil {
```
b00ris: I think we can get rid of cleanup. 
marshalled validator size = 20b+ 32b*4+8b = 156b
validator set size = 15kb
Snapshot storage/year(for epoch=1d) = 15kb * 365 = 5.5mb

command/genesis/polybft_params.go
goran-ethernal: I am not setting the hash. I am just initializing the checkpoint object because of the epoch number, which is 0 in genesis block.

consensus/polybft/runtime_helpers.go
goran-ethernal: No. `header` is the last block we are synced to. So I am returning the first block of the current epoch.

consensus/polybft/validators_snapshot.go
goran-ethernal: I was analyzing this on Thursday. It complicates the logic quite a lot, but let me try. The previous logic relied heavily on epoch ending blocks.

consensus/polybft/consensus_runtime.go
goran-ethernal: No underflow: https://goplay.tools/snippet/Hn9vmkKfhZO

consensus/polybft/validators_snapshot.go
goran-ethernal: See https://github.com/0xPolygon/polygon-edge/pull/897#discussion_r1021137452. 

I think we could store them by hash. I've taken a look at the code, and we are aware of the block extra on each point where we call `GetValidators` function. Let's discuss on sync. Maybe this is a good idea to implement on another PR.

consensus/polybft/runtime_helpers.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/2b29277a2a91cb58037bdb5128a9eebe3c3e7069

consensus/polybft/runtime_helpers.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/2b29277a2a91cb58037bdb5128a9eebe3c3e7069

consensus/polybft/validators_snapshot.go
Stefan-Ethernal: > Snapshot storage/year(for epoch=1d) = 15kb * 365 = 5.5mb

If validator set changes more often than once/day, this can increase.

IMO we should keep cleanup logic, since it doesn't introduce too much complexity.

goran-ethernal: I agree. We can increase the number of snapshots we keep in storage though.

consensus/polybft/consensus_runtime.go
b00ris: (blockNumber - c.config.PolyBFTConfig.EpochSize + 1) - is an overflow for low block number. 
epoch.FirstBlockInEpoch + c.config.PolyBFTConfig.EpochSize - 1 == blockNumber will be without overflow.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/d4905bd03fe059a8cff181b91102d991b0819555

consensus/polybft/consensus_runtime.go
```diff
@@ -350,9 +345,12 @@ func (c *consensusRuntime) FSM() error {
 		return fmt.Errorf("cannot create block builder: %w", err)
 	}
 
+	// TODO - recognize slashing occurred
+	slash := false
+
 	pendingBlockNumber := parent.Number + 1
 	isEndOfSprint := c.isEndOfSprint(pendingBlockNumber)
```
Stefan-Ethernal: What shall we do with sprint blocks detection? Is it in the scope of this PR as well? To me it would be natural thing to consider with introduction of dynamic epochs.

consensus/polybft/consensus_runtime_test.go
```diff
@@ -195,31 +195,34 @@ func TestConsensusRuntime_getQuorumSize(t *testing.T) {
 func TestConsensusRuntime_isEndOfEpoch_NotReachedEnd(t *testing.T) {
```
Stefan-Ethernal: Rename test?

`TestConsensusRuntime_isFixedSizeOfEpochMet_NotReachedEnd`

consensus/polybft/consensus_runtime_test.go
```diff
@@ -231,17 +234,38 @@ func TestConsensusRuntime_isEndOfEpoch_NotReachedEnd(t *testing.T) {
 func TestConsensusRuntime_isEndOfEpoch_ReachedEnd(t *testing.T) {
```
Stefan-Ethernal: Rename test?
`TestConsensusRuntime_isFixedSizeOfEpochMet_ReachedEnd`

consensus/polybft/runtime_helpers.go
```diff
@@ -1,5 +1,11 @@
 package polybft
 
+import (
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type blockGetter func(uint64) (*types.Header, bool)
```
Stefan-Ethernal: You can remove this

consensus/polybft/consensus_runtime_test.go
```diff
@@ -921,154 +958,6 @@ func TestConsensusRuntime_FSM_EndOfSprint_HasBundlesToExecute(t *testing.T) {
...
-		t,
-		epochCases,
-		func(blockNum uint64) bool {
-			return runtime.isEndOfEpoch(blockNum + 1)
-		},
-		"end of epoch",
-		runtime,
-	)
-
-	var sprintCases = []struct {
```
Stefan-Ethernal, line 956: You've removed even is end of sprint test cases :crying_face:

consensus/polybft/consensus_runtime.go
goran-ethernal: If slashing occurs the end of epoch will also be the sprint block.

Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/dfc4c8c841fd4738b9946992c15d7b5b031e96a8

consensus/polybft/consensus_runtime_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/dfc4c8c841fd4738b9946992c15d7b5b031e96a8

consensus/polybft/consensus_runtime_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/dfc4c8c841fd4738b9946992c15d7b5b031e96a8

consensus/polybft/runtime_helpers.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/dfc4c8c841fd4738b9946992c15d7b5b031e96a8

consensus/polybft/consensus_runtime.go
goran-ethernal: And https://github.com/0xPolygon/polygon-edge/pull/897/commits/c05a67469b5b218909632b17e7b4cadca5dbea63 :smile:

consensus/polybft/consensus_runtime.go
```diff
@@ -205,7 +198,9 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 		c.logger.Error("on block inserted error", "err", err)
 	}
 
-	if c.isEndOfEpoch(block.Header.Number) {
+	// TODO - this condition will need to be changed to recognize that either slashing happened
+	// or epoch reached its fixed size
```
vcastellm, line 261: Why it that different?

consensus/polybft/consensus_runtime.go
```diff
@@ -1199,6 +1182,57 @@ func (c *consensusRuntime) BuildRoundChangeMessage(
 	return signedMsg
 }
 
+// getFirstBlockOfEpoch returns the first block of epoch in which provided header resides
+func (c *consensusRuntime) getFirstBlockOfEpoch(epochNumber uint64, latestHeader *types.Header) (uint64, error) {
```
vcastellm, line 1302: it can be any header not necessarily the latest, `latestHeader` => `header`?

consensus/polybft/consensus_runtime.go
goran-ethernal: `OnBlockInserted` is called on block insert through `consensus` and through `syncer` so we can not just check if fixed size of epoch is met. Maybe this was the epoch ending block because we had slashing. We would need to persist some flag that slashing occurred and check it here. It won't be a big problem to do that, but we need the slashing implemented both on client and on smart contract.

consensus/polybft/consensus_runtime.go
```diff
@@ -1199,6 +1182,57 @@ func (c *consensusRuntime) BuildRoundChangeMessage(
...
+	if latestHeader.Number == 0 {
+		// if we are starting the chain, we know that the first block is block 1
+		return 1, nil
+	}
+
+	blockHeader := latestHeader
+	blockExtra, err := GetIbftExtra(latestHeader.ExtraData)
+
```
vcastellm, line 1309:
```suggestion
```

consensus/polybft/consensus_runtime.go
goran-ethernal: For now this is only called on `restartEpoch` where we pass the latest header we are synced on.

consensus/polybft/consensus_runtime.go
goran-ethernal: Linter gets unhappy if I rewmove this :confused: : 

> only one cuddle assignment allowed before if statement (wsl)go-golangci-lint

consensus/polybft/consensus_runtime.go
```diff
@@ -731,51 +723,37 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
...
-			if err := calculateUptimeForBlock(blockHeader, validators); err != nil {
+			if err := calculateUptimeForBlock(blockExtra, validators); err != nil {
 				return nil, err
 			}
 
-			blockHeader, found = c.config.blockchain.GetHeaderByNumber(blockHeader.Number - 1)
-			if !found {
-				return nil, blockchain.ErrNoBlock
-			}
+			blockHeader, blockExtra, err = c.getBlockData(blockHeader.Number - 1)
```
vcastellm: This code is a bit confusing, we're overwriting `blockHeader` and `blockExtra` several times, would be great if we can simplify it.

goran-ethernal: I'll try. Not gonna leave any promises though :sweat_smile:

consensus/polybft/consensus_runtime.go
vcastellm: Then cuddle with the next line and separate from the previous, I find it more clean.

consensus/polybft/consensus_runtime.go
```diff
@@ -205,7 +198,9 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 		c.logger.Error("on block inserted error", "err", err)
 	}
 
-	if c.isEndOfEpoch(block.Header.Number) {
+	// TODO - this condition will need to be changed to recognize that either slashing happened
+	// or epoch reached its fixed size
+	if c.isFixedSizeOfEpochMet(block.Header.Number) {
```
b00ris: Should we restart an epoch on slashing also?

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/b28485cc0c255875b9603276f1a6cbb4bacaf70e

consensus/polybft/consensus_runtime.go
goran-ethernal: Tried it. It gets even more complicated. I'm reusing `blockHeader` and `blockExtra` because uptime is calculated like this:

**Block that is currently being built**: 20
**Parent block**: 19
**EpochSize**: 10

We start reading parent signatures from block 19 (which are the only certain way to know who signed a block), until we reach the first block in epoch, which is block 11 (This is the first for loop). This means we will not calculate uptime for the last block in epoch since we are currently building it (block is not sealed nor finalized).

To put it more graphical, uptime is calculated like this in our example (first number is block number, second number is for which block we are counting signatures when we are checking parent signatures):

======epoch3====================================================
21 (20)
================================epoch 2==========================
20 (19) 
This is where the first loop starts.
19 (18)
18 (17)
17 (16)
16 (15)
15 (14)
14 (13)
13 (12)
12 (11)
This is where the first for loop stops, and second loop begins.
11 (10)
================================epoch 1==========================
10 (9)
// This is where the second loop stops.
9 (8)

That is why we need to calculate uptime for the last two blocks of previous epoch as well. (That is the second for loop). To actually reduce the logic, I am reusing the `blockExtra` and `blockHeader` local variables, so when the first loop is done, I know from which block I need to continue (the last block of previous epoch), since their validator sets are not the same.

I can introduce new variables to make it more readable, but I think it is not necessary, and just makes the code a bit complicated and adds unnecessary lines.

What are your thoughts on this? I'll take any recommendation to improve it that's for sure. I'm just not seeing it right know :confused: .

consensus/polybft/consensus_runtime.go
```diff
@@ -705,20 +707,10 @@ func (c *consensusRuntime) deliverMessage(msg *TransportMessage) (bool, error) {
...
-		return nil, errors.New("epoch size not large enough to calculate uptime")
-	}
-
-	calculateUptimeForBlock := func(header *types.Header, validators AccountSet) error {
-		blockExtra, err := GetIbftExtra(header.ExtraData)
-		if err != nil {
-			return err
-		}
-
+	calculateUptimeForBlock := func(blockExtra *Extra, validators AccountSet) error {
```
vcastellm: I would move this closure to a function returning `uptimeCounter, err`, it will be more performant and simplify enclosing function.

consensus/polybft/consensus_runtime.go
```diff
@@ -731,51 +723,37 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
...
 	}
 
-	// since we need to calculate uptime for the last block of the previous epoch,
-	// we need to get the validators for the that epoch from the smart contract
-	// this is something that should probably be optimized
-	if lastBlockInPreviousEpoch > 0 { // do not calculate anything for genesis block
+	// calculate uptime for blocks from previous epoch that were not processed in previous uptime
+	// since we can not calculate uptime for the last block in epoch (because of parent signatures)
+	if blockHeader.Number > uptimeLookbackSize {
 		for i := 0; i < uptimeLookbackSize; i++ {
```
vcastellm, line 778: This loop is just 1 iteration, when do we want to change `lookbackSize`?

consensus/polybft/consensus_runtime.go
vcastellm: Ok, now I understand it better, it's still a big func so adding some other comments for improvements.

Why sorting here: https://github.com/0xPolygon/polygon-edge/pull/897/files#diff-e5b2e93010ff84c3777aef136e0834d710dd22861ca62ffbab9d5baa12a52b3cR766

consensus/polybft/consensus_runtime.go
```diff
@@ -731,51 +723,37 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
...
-			if !found {
-				return nil, blockchain.ErrNoBlock
-			}
+			blockHeader, blockExtra, err = c.getBlockData(blockHeader.Number - 1)
 		}
 	}
 
-	epochID := epoch.Number
-
 	uptime := Uptime{EpochID: epochID}
```
vcastellm: We can move this declaration down just before it's used

consensus/polybft/consensus_runtime.go
goran-ethernal: Its actually 2 iterations. It was an error from my side. In that example I left, wr would have first iteration for the first block in the current epoch, where we read parent signatures (who sealed block 10), and then we would read parent signatures on block 10 (which is the last block of previous epoch, and indicator who sealed block 9).

consensus/polybft/validators_snapshot.go
b00ris: if our epoch length will be 30m -  our data consumtion will be:
15kb * 365 * 48 = 264mb/year.
I'm not against cleaning up in general, but I'm confused that we want to remove our knowlidge about epoch's validator set. It's not taking much space.

Stefan-Ethernal: > 15kb * 365 * 48 = 264mb/year.

So is snapshot storing frequency and 48 here represents epoch ending blocks. Although we are not persisting just epoch ending blocks with this PR I'd say, but arbitrary blocks if I'm not wrong (@goran-ethernal right?).

This PR introduces persistence of snapshot per blocks (instead of per epoch) therefore it almost certainly is going to be larger, since we are retrieving validators snapshot per each block when verifying signatures. 
If we assume that block time is 2 s (which was by default on v3), it means that during a single day we should produce 43200. So my formula looks like this (please check if this makes sense):
15kb * 365 * 43200 = ~236k mb/year, which is quite a lot :slightly_smiling_face:.
Whenever we construct a snapshot we store it in the cache and db. Block within the epoch will have the redundant data (i.e. same validator sets).

Even if we preserve cleanup logic, we can always reconstruct snapshot based on last persisted snapshot and deltas.

goran-ethernal: I started working on that task for storing snapshots by validator set hash, so it won't be by block anymore:

https://polygon.atlassian.net/browse/EVM-155

consensus/polybft/validators_snapshot.go
goran-ethernal: Task EVM-155 that I am working on, will change this. Snapshota will be store by validator set hash, so lets keep it like this in this PR.

consensus/polybft/consensus_runtime.go
goran-ethernal: Yes, I left a TODO here.

consensus/polybft/consensus_runtime.go
goran-ethernal: Because `uptimeCounter` is a map. We can not ensure that all validators will have the same order of addresses in the `Uptime` object lists of addresses and times they sealed a block (which is important when we calculate a hash of uptime trasnaction), so we make a slice of addresses and sort it.

consensus/polybft/validators_snapshot.go
goran-ethernal: I would say let's decide about cleanup on that following PR. For now, let's keep it like this.

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/7e3625bb48d731068c5c64aebc56aafe8d42ffa1

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/897/commits/7e3625bb48d731068c5c64aebc56aafe8d42ffa1

b00ris (approved):

Stefan-Ethernal (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -205,7 +198,9 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 		c.logger.Error("on block inserted error", "err", err)
 	}
 
-	if c.isEndOfEpoch(block.Header.Number) {
+	// TODO - this condition will need to be changed to recognize that either slashing happened
```
JekaMas, line 260: could we rather have tasks in jira then TODOes and FIXMEs in the code?

goran-ethernal: I agree, but, we will need to do a planning about slashing, since this PR is done as a sort of Proof of Concept that happened to be a way to go.

@b00ris Maybe you already have some tasks in mind about slashing detection on client?

vcastellm (approved):
LGTM

igorcrevar (approved):

Nemanja0x (approved):

PR #901, opened 2022-11-10
Use jsonrpc block header in subscription

ferranbt:
# Description

This PR does a couple of things:
- It uses the local `jsonrpc.Block` instead of `types.Header` for the `jsonrpc` filter methods. Note that in the Go-ethereum codebase, it returns a `Header` (like `types.Header`) object with fewer fields than our `jsonrpc.Block` object. However, I do not think that is a problem since `types.Header` is a subset of `jsonrpc.Block`.
- It also removes all the JSON methods to encode `types.Header`.

I think overall, the tradeoffs are okay and it simplifies the code a lot.

There is one more thing to fix in the copy method of `jsonrpc.Block`.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


jsonrpc/types.go
```diff
@@ -104,6 +104,20 @@ type block struct {
 	Uncles          []types.Hash        `json:"uncles"`
 }
 
+func (b *block) Copy() *block {
```
JekaMas, line 107: Do we need copy transactions and uncles as well, via copy() and allocating new slices?

JekaMas: And bloom filter

ferranbt: Yep, it's in a TODO in the code. For sure the arrays, not sure about the bloom. I will test it anyway.

ferranbt: Ok, so this `Clone` function is there because `types.Header` also had one. But two things, in this case, I do not think `Clone` is totally necessary (I do not want to remove it though since I do not remember why was there). Second, the block is copied only on filter updates which do never have transactions. So, to me, we can keep it like this.

Otherwise, we would have to complicate it a little bit creating also a `Clone` function for `jsonrpc.Transaction`.

vcastellm (approved):
LGTM

jsonrpc/filter_manager.go
```diff
@@ -265,7 +265,10 @@ func NewFilterManager(logger hclog.Logger, store filterManagerStore, blockRangeL
 
 	// start blockstream with the current header
 	header := store.Header()
-	m.blockStream.push(header)
+
+	// TODO: Make Header return jsonrpc.block object directly
```
JekaMas, line 269: could we rather create tasks then putting todoes?

jsonrpc/types.go
JekaMas: works for me.

JekaMas (approved):

PR #902, opened 2022-11-10
Use a local account object and refactor the store account interface

ferranbt:
# Description

This PR replaces the `state.Account` object used in `jsonrpc` with a local `Account` object with only the required fields. It also refactors the `getCode` interface method as a side effect since the new `Account` object does not include the `codeHash`.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -27,11 +27,16 @@ type ethTxPoolStore interface {
 	GetPendingTx(txHash types.Hash) (*types.Transaction, bool)
 }
 
+type Account struct {
+	Balance *big.Int
+	Nonce   uint64
+}
+
 type ethStateStore interface {
-	GetAccount(root types.Hash, addr types.Address) (*state.Account, error)
+	GetAccount(root types.Hash, addr types.Address) (*Account, error)
 	GetStorage(root types.Hash, addr types.Address, slot types.Hash) ([]byte, error)
 	GetForksInTime(blockNumber uint64) chain.ForksInTime
-	GetCode(hash types.Hash) ([]byte, error)
+	GetCode(root types.Hash, addr types.Address) ([]byte, error)
 }
 
 type ethBlockchainStore interface {
@@ -729,7 +734,7 @@ func (e *Eth) GetCode(address types.Address, filter BlockNumberOrHash) (interfac
 	}
 
 	emptySlice := []byte{}
-	acc, err := e.store.GetAccount(header.StateRoot, address)
+	code, err := e.store.GetCode(header.StateRoot, address)
 
 	if errors.Is(err, ErrStateNotFound) {
 		// If the account doesn't exist / is not initialized yet,
@@ -737,12 +742,6 @@ func (e *Eth) GetCode(address types.Address, filter BlockNumberOrHash) (interfac
 		return "0x", nil
 	} else if err != nil {
 		return argBytesPtr(emptySlice), err
-	}
-
-	code, err := e.store.GetCode(types.BytesToHash(acc.CodeHash))
-	if err != nil {
-		// TODO This is just a workaround. Figure out why CodeHash is populated for regular accounts
-		return argBytesPtr(emptySlice), nil
 	}
 
 	return argBytesPtr(code), nil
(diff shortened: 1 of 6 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -27,11 +27,16 @@ type ethTxPoolStore interface {
 	GetPendingTx(txHash types.Hash) (*types.Transaction, bool)
 }
 
+type Account struct {
+	Balance *big.Int
```
b00ris, line 31: Why do we need narrowed Account structure?

jsonrpc/eth_endpoint.go
```diff
@@ -27,11 +27,16 @@ type ethTxPoolStore interface {
...
+	Nonce   uint64
+}
+
 type ethStateStore interface {
-	GetAccount(root types.Hash, addr types.Address) (*state.Account, error)
+	GetAccount(root types.Hash, addr types.Address) (*Account, error)
 	GetStorage(root types.Hash, addr types.Address, slot types.Hash) ([]byte, error)
 	GetForksInTime(blockNumber uint64) chain.ForksInTime
-	GetCode(hash types.Hash) ([]byte, error)
+	GetCode(root types.Hash, addr types.Address) ([]byte, error)
```
b00ris, line 39: Is it get contract code? Why do we need address for it?

jsonrpc/eth_state_test.go
```diff
@@ -823,12 +820,12 @@ func (m *mockSpecialStore) GetStorage(root types.Hash, addr types.Address, slot
 	return val, nil
 }
 
-func (m *mockSpecialStore) GetCode(hash types.Hash) ([]byte, error) {
-	if bytes.Equal(m.account.account.CodeHash, hash.Bytes()) {
-		return m.account.code, nil
+func (m *mockSpecialStore) GetCode(root types.Hash, addr types.Address) ([]byte, error) {
+	if m.account.address.String() != addr.String() {
```
b00ris: You can compare addresses without string conversion

jsonrpc/eth_endpoint.go
ferranbt: It is part of the conversation from yesterday on removing cross-dependencies between the packages. `jsonrpc` as it now relies on a structure targeted to work on the state transition (it even has some internal objects for merkle trie). Now, it only contains the required fields to make it work on `jsonrpc`.

jsonrpc/eth_endpoint.go
ferranbt: The `GetCode` endpoint (eth_getCode) returns the code for an address. Thus, first, it resolves the address, retrieves the code hash, and then resolves the code (since the code is internally stored as `codeHash:code`)

b00ris: I just wanted to mention that to return contract code you need code hast(it's unique) or account(to get code hash). I dont understand well why do you need both of them.

jsonrpc/eth_endpoint.go
b00ris: I may havent heard it. Could you give me a link to a RFC about it?

jsonrpc/eth_endpoint.go
ferranbt: Following [this](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getcode) spec, you can get the code at some point in time. That root is used to resolve the `Account` at some point in time.

vcastellm: Does it makes sense to reverse the param order as in the spec?

Kourin1996 (approved):

jsonrpc/eth_endpoint.go
ferranbt: [This](https://www.notion.so/polygontechnology/RFC-145-0b3ac0015b2047b0b655fc66c2f028fc#5aa35c955f414602a43d4f355b6db662) is the example of the implementation. It uses local objects instead of the ones from `types`. The same reasoning is applied here.

JekaMas (approved):

b00ris (approved):

PR #904, opened 2022-11-13
Polybft consensus voting power quorum size

Nemanja0x:
# Description

Using voting power to check if the consensus is reached for:
1. consensus engine
2. block verification 
3. state transactions

# Changes include

# Checklist

## Testing


consensus/polybft/validator_set.go
```diff
@@ -323,6 +340,27 @@ func (v *validatorSet) rescalePriorities(diffMax int64) error {
...
 
+// HasQuorum determines if there is quorum of enough signers reached,
+// based on its voting power and quorum size
+func (v validatorSet) HasQuorum(signers []types.Address) bool {
+	return v.calculateVotingPower(signers) >= v.quorumSize
+}
+
+// checks if submitted signers have reached prepare quorum
+func (v validatorSet) HasPrepareQuorum(signers []types.Address) bool {
+	return v.calculateVotingPower(signers) >= v.quorumSize-1
```
Stefan-Ethernal: I'd say we should substract local node's voting power here instead of 1 (we were substracting 1 because of the local node). 
With voting power established, current change would imply that local node has `votingPower=1` which may be wrong.

```go
return v.calculateVotingPower(signers) >= v.quorumSize-<localNodeVotingPower>
```

consensus/polybft/validator_set.go
```diff
@@ -425,6 +463,30 @@ func (v *validatorSet) Len() int {
...
+	for _, v := range v.votingPowerMap {
+		totalVotingPower += v
+	}
+
+	if totalVotingPower == 0 {
+		return ErrInvalidTotalVotingPower
+	}
+
+	maxFaultyVotingPower := (totalVotingPower - 1) / 3
+	v.quorumSize = 2*maxFaultyVotingPower + 1
```
Stefan-Ethernal: We should change this to 2/3 supermajority (Edge adopted that as a quorum size and we were discussing it back when pbft was consensus engine that we should eventually change it):

Correct me if I'm wrong, but think it should be calculated like this:
`math.Ceil(2/3*totalVotingPower)`

consensus/polybft/validator_set.go
```diff
@@ -425,6 +463,30 @@ func (v *validatorSet) Len() int {
...
+	totalVotingPower := uint64(0)
+	for _, v := range v.votingPowerMap {
+		totalVotingPower += v
+	}
+
+	if totalVotingPower == 0 {
+		return ErrInvalidTotalVotingPower
+	}
+
+	maxFaultyVotingPower := (totalVotingPower - 1) / 3
```
Stefan-Ethernal: This should be deleted if we are going to adopt 2/3 supermajority quorum size.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/904/commits/f3573424174f3f8cdcf6c2ad31ac9dbfebb0ed44

consensus/polybft/validator_set.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/904/commits/f3573424174f3f8cdcf6c2ad31ac9dbfebb0ed44

igorcrevar (approved):

consensus/polybft/validator_set.go
```diff
@@ -80,6 +87,12 @@ type ValidatorSet interface {
 
 	// IncrementProposerPriority increments priorities number of times
 	IncrementProposerPriority(times uint64) error
+
+	// checks if submitted signers have reached quorum
+	HasQuorum(signers []types.Address) bool
+
+	// checks if submitted signers have reached prepare quorum
```
Stefan-Ethernal: Should we change this like this?

`// checks if submitted signers have reached quorum without proposer voting power`

consensus/polybft/consensus_runtime.go
```diff
@@ -1085,34 +1091,36 @@ func (c *consensusRuntime) ID() []byte {
...
+	// if we are not using consensus engine for current block return false
+	if c.fsm.parent.Number+1 != blockNumber {
+		c.logger.Warn("HasQuorum checking against stale block",
+			"blockNumber", blockNumber, "parentBlock", c.fsm.parent.Number)
+
+		return false
+	}
 
+	// extract the addresses of all the senders of the messages
+	senders := make([]types.Address, len(messages))
```
Stefan-Ethernal: Should we call this `senders` or `signers`? I'd like to call it the same both here and in the `HasQuorum` and `HasQuorumWithoutProposer` functions.

Stefan-Ethernal (approved):
Generally looks good, with minor consideration.

The only thing which bothers me a little bit is about accessing to the `proposer` field (not sure if we should ensure thread safety there, because looking at the code, `backend.IsProposer` function seems like can be called from multiple routines on IBFT consensus engine side):
https://github.com/0xPolygon/polygon-edge/pull/904/files#diff-297043a61d1a54ac38dc24c4a73a7de2fac85bd5567b58b1cfb47d8bdff663bdR376

command/genesis/polybft_params.go
```diff
@@ -235,20 +271,27 @@ func (p *genesisParams) deployContracts() (map[types.Address]*chain.GenesisAccou
...
 	}
 
 	for i, validator := range validators {
 		delta.Added[i] = &polybft.ValidatorMetadata{
-			Address: types.Address(validator.Account.Ecdsa.Address()),
-			BlsKey:  validator.Account.Bls.PublicKey(),
+			Address: validator.Address,
+			BlsKey:  publicKeys[i],
+			// convert from wei to token amount (voting power == tokens amount)
+			VotingPower: validator.Balance.Div(validator.Balance, big.NewInt(polybft.WeiScalingFactor)).Uint64(),
```
b00ris: Could you extract voting power scaling to a separate function?

consensus/polybft/consensus_runtime.go
b00ris: Are messages unique?

consensus/polybft/consensus_runtime.go
```diff
@@ -1085,34 +1091,36 @@ func (c *consensusRuntime) ID() []byte {
...
+
+	for i, message := range messages {
+		senders[i] = types.BytesToAddress(message.From)
+	}
+
+	// check quorum
 	switch msgType {
-	case proto.MessageType_PREPREPARE:
-		return len(messages) >= 0
+	default: // proto.MessageType_PREPREPARE etc
```
b00ris: ?

consensus/polybft/validator_set.go
```diff
@@ -91,27 +104,80 @@ type validatorSet struct {
...
+	}
+
+	if totalVotingPower == 0 {
+		return errInvalidTotalVotingPower
+	}
+
+	// If there cannot be faulty nodes (less than 4 nodes in the network),
+	// then quorum size is determined as total voting power (namely all the nodes must send vote).
+	// Otherwise quorum size is calculated as 2/3 supermajority
+	if v.calcMaxFaultyNodes() == 0 {
```
b00ris: This confuses me. Why do we need it?

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: This is reverted through
https://github.com/0xPolygon/polygon-edge/pull/904/commits/5c5c4689626cc81ca01289e9892811de9eed72be#diff-e5b2e93010ff84c3777aef136e0834d710dd22861ca62ffbab9d5baa12a52b3cR1117-R1118

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: They should be, but I think we should rely on a map of senders here (instead of relying on slice).

Let me try change it that way.

consensus/polybft/consensus_runtime.go
Nemanja0x: @vcastellm will give explanation

consensus/polybft/consensus_runtime.go
Nemanja0x: Will have another task for that validation, at the moment we are relying on go-ibft to send unique and proper messages. https://polygon.atlassian.net/browse/EVM-206

consensus/polybft/validator_set.go
Stefan-Ethernal: I followed existing implementation on the Edge. It sets more strict quorum size for small networks, i.e. less than 4 nodes where we don't have BFT properties met. All the validators need to send a message in that case so that we have a quorum:
https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/consensus/ibft/validators.go#L38

However we can remove it if you don't find it necessary.

Nemanja0x: If there are 3 or less validators then according to 
```
func (v validatorSet) calcMaxFaultyNodes() uint64 {
	return uint64((v.Len() - 1) / 3)
}
```
number of faulty validators is 0. If that is the case, all of the validators must vote positively, so the qourum size is equal to total voting power.

Stefan-Ethernal: I've removed it, so the code is simpler and behavior consistent: [daae347](https://github.com/0xPolygon/polygon-edge/pull/904/commits/daae347f6e768e6e825c349d673706504c9610e1).

command/genesis/polybft_params.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/904/commits/8c7c342c70b9c63d3e47b79173689704f4361b7f

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fix [f5c5c74](https://github.com/0xPolygon/polygon-edge/pull/904/commits/f5c5c74162b118ae9948bb6a302b23b1b970188a)

goran-ethernal (approved):

b00ris (approved):

JekaMas (approved):

PR #908, opened 2022-11-15
Jsonrpc fix race conditions

ferranbt:
# Description

This PR fixes two race conditions in `jsonrpc` and one mutex lock. You can check the commits for each specific change.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/filter_manager.go
+++ b/jsonrpc/filter_manager.go
@@ -849,7 +849,6 @@ type headElem struct {
 
 func (h *headElem) getUpdates() ([]*block, *headElem) {
 	res := make([]*block, 0)
-
 	cur := h
 
 	for {
--- a/jsonrpc/filter_manager.go
+++ b/jsonrpc/filter_manager.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/0xPolygon/polygon-edge/blockchain"
@@ -255,7 +256,6 @@ func NewFilterManager(logger hclog.Logger, store filterManagerStore, blockRangeL
 		logger:          logger.Named("filter"),
 		timeout:         defaultTimeout,
 		store:           store,
-		blockStream:     &blockStream{},
 		blockRangeLimit: blockRangeLimit,
 		filters:         make(map[string]filter),
 		timeouts:        timeHeapImpl{},
@@ -268,7 +268,7 @@ func NewFilterManager(logger hclog.Logger, store filterManagerStore, blockRangeL
 
 	// TODO: Make Header return jsonrpc.block object directly
 	block := toBlock(&types.Block{Header: header}, false)
-	m.blockStream.push(block)
+	m.blockStream = newBlockStream(block)
 
 	// start the head watcher
 	m.subscription = store.SubscribeEvents()
@@ -809,14 +809,23 @@ func (t *timeHeapImpl) Pop() interface{} {
 // of the stream at any point
 type blockStream struct {
 	lock sync.Mutex
-	head *headElem
+	head atomic.Value
+}
+
+func newBlockStream(head *block) *blockStream {
+	b := &blockStream{}
+	b.head.Store(&headElem{header: head})
+
+	return b
 }
 
 func (b *blockStream) Head() *headElem {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	return b.head
+	head, _ := b.head.Load().(*headElem)
+
+	return head
 }
 
 func (b *blockStream) push(header *block) {
@@ -827,16 +836,15 @@ func (b *blockStream) push(header *block) {
 		header: header.Copy(),
 	}
 
-	if b.head != nil {
-		b.head.next = newHead
-	}
+	oldHead, _ := b.head.Load().(*headElem)
+	oldHead.next.Store(newHead)
 
-	b.head = newHead
+	b.head.Store(newHead)
 }
 
 type headElem struct {
 	header *block
-	next   *headElem
+	next   atomic.Value
 }
 
 func (h *headElem) getUpdates() ([]*block, *headElem) {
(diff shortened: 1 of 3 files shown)


jsonrpc/filter_manager_test.go
```diff
@@ -511,9 +511,7 @@ func newMockWsConnWithMsgCh() (*mockWsConn, <-chan []byte) {
 func TestHeadStream(t *testing.T) {
```
JekaMas: as far as we testing data races, could we have a test for it? something simple like calling each API in a separate goroutine with tickers like 1mcs?

ferranbt: Let me know what you think. I added a new `TestHeaderStream_Concurrent` test.

b00ris (approved):

jsonrpc/filter_manager.go
```diff
@@ -818,14 +809,23 @@ func (t *timeHeapImpl) Pop() interface{} {
 // of the stream at any point
 type blockStream struct {
 	lock sync.Mutex
```
Stefan-Ethernal, line 811: Do we still need lock here, since we have introduced `atomic`?

jsonrpc/filter_manager.go
```diff
@@ -836,26 +836,35 @@ func (b *blockStream) push(header *block) {
...
 type headElem struct {
 	header *block
-	next   *headElem
+	next   atomic.Value
 }
 
 func (h *headElem) getUpdates() ([]*block, *headElem) {
 	res := make([]*block, 0)
 
 	cur := h
```
Stefan-Ethernal:
```suggestion
	res := make([]*block, 0)
	cur := h
```

jsonrpc/filter_manager.go
ferranbt: Let me try, the test for concurrency is better now.

jsonrpc/filter_manager.go
ferranbt: Fixed 9658318d96398862a3ce41db26c35841fb3cbd43

jsonrpc/filter_manager.go
ferranbt: So, the lock is there to avoid concurrent `push`. In theory, there cannot be any because of how `push` is being used, but I wanted to be sure.

jsonrpc/filter_manager.go
```diff
@@ -836,26 +836,34 @@ func (b *blockStream) push(header *block) {
...
+			// no more messages
+			break
+		} else {
+			nextElem, _ := next.(*headElem)
+
+			if nextElem.header != nil {
+				res = append(res, nextElem.header)
+			}
+			cur = nextElem
+		}
```
vcastellm, line 863:
```suggestion
	for next := cur.next.Load(); next != nil; next = cur.next.Load() {
		nextElem, _ := next.(*headElem)

		if nextElem.header != nil {
			res = append(res, nextElem.header)
		}

		cur = nextElem
```

jsonrpc/filter_manager_test.go
```diff
@@ -532,6 +532,70 @@ func TestHeadStream(t *testing.T) {
...
+					if expect == uint64(nMessages) {
+						// Succeeded
+						errCh <- nil
+
+						return
+					}
+				}
+
+				item = next
+			}
```
vcastellm, line 589:
```suggestion
			for blocks, next := item.getUpdates(); ; blocks, next = next.getUpdates() {
				for _, block := range blocks {
					if num := uint64(block.Number); num != expect {
						errCh <- fmt.Errorf("subscriber %05d bad event want=%d, got=%d", i, num, expect)

						return
					}
					expect++

					if expect == uint64(nMessages) {
						// Succeeded
						errCh <- nil

						return
					}
				}
			}
```

vcastellm (approved):
LGTM aside from the suggestions

Kourin1996 (approved):

jsonrpc/filter_manager.go
Stefan-Ethernal: > So, the lock is there to avoid concurrent push.

What do you think about keeping lock used only in `push` and remove it from `Head`, since `Head` completely relies on atomic (just loading a value)?

jsonrpc/filter_manager.go
```diff
@@ -818,14 +809,23 @@ func (t *timeHeapImpl) Pop() interface{} {
...
+}
+
+func newBlockStream(head *block) *blockStream {
+	b := &blockStream{}
+	b.head.Store(&headElem{header: head})
+
+	return b
 }
 
 func (b *blockStream) Head() *headElem {
```
Stefan-Ethernal: Nitpick, unrelated to this PR:
`Head` can be unexported, as it is used only by `jsonrpc` package.

Stefan-Ethernal (approved):
LGTM

JekaMas (approved):

jsonrpc/filter_manager.go
ferranbt: You got it! I also removed the lock in Head.

jsonrpc/filter_manager.go
ferranbt: I will keep it as it is since I do not like that much those inlined for loops. They confuse me a lot.

jsonrpc/filter_manager_test.go
ferranbt: Same as before.

jsonrpc/filter_manager.go
vcastellm: It's the other way around for me 😂 , when I see an empty `for` loop I immediately think of an endless loop until checking the inside.

Being a bit picky 😛 checking with https://github.com/uudashr/gocognit:

Before `6 jsonrpc (*headElem).getUpdates jsonrpc/filter_manager.go:850:1`
After `3 jsonrpc (*headElem).getUpdates jsonrpc/filter_manager.go:850:1`

The nesting hurts it

PR #912, opened 2022-11-15
Introduce local development cluster script

vcastellm:
# Description

Introduce local development cluster script

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

Run the cluster

### DIFF
--- a/scripts/cluster
+++ b/scripts/cluster
@@ -11,28 +11,21 @@ go build -o polygon-edge .
 if [ "$1" == "ibft" ]; then
   echo "Running with ibft consensus"
   nodeid=$(./polygon-edge secrets init --data-dir test-chain- --num 4 | grep Node | head -n 1 | awk -F ' ' '{print $4}')
+  genesis_params="--consensus ibft --ibft-validators-prefix-path test-chain- --bootnode /ip4/127.0.0.1/tcp/10001/p2p/$nodeid"
 
-  ./polygon-edge genesis --consensus ibft --ibft-validators-prefix-path test-chain- \
-    --bootnode /ip4/127.0.0.1/tcp/10001/p2p/"$nodeid" \
-    --block-gas-limit 10000000 \
-    --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000
-
-  ./polygon-edge server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :10000 --libp2p :10001 --jsonrpc :10002 --seal --log-level DEBUG &
-  ./polygon-edge server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :20000 --libp2p :20001 --jsonrpc :20002 --seal &
-  ./polygon-edge server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :30000 --libp2p :30001 --jsonrpc :30002 --seal &
-  ./polygon-edge server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :40000 --libp2p :40001 --jsonrpc :40002 --seal &
+  nodeid=$(./polygon-edge secrets init --data-dir test-chain- --num 4 | grep Node | head -n 1 | awk -F ' ' '{print $4}')
 else
   echo "Running with polybft consensus"
+  genesis_params="--consensus polybft --validator-set-size=4"
   ./polygon-edge polybft-secrets --data-dir test-chain- --num 4
-  ./polygon-edge genesis --consensus polybft \
-    --block-gas-limit 10000000 \
-    --validator-set-size=4 \
-    --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000
-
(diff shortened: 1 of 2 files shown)


scripts/cluster
```diff
@@ -0,0 +1,31 @@
...
+# Reset test-dirs
+rm -rf test-chain-*
+rm -f genesis.json
+
+go build -o polygon-edge .
+
+nodeid=$(./polygon-edge secrets init --data-dir test-chain- --num 4 | grep Node | head -n 1 | awk -F ' ' '{print $4}')
+
+./polygon-edge genesis --consensus ibft --ibft-validators-prefix-path test-chain- \
+  --bootnode /ip4/127.0.0.1/tcp/10001/p2p/"$nodeid" \
```
ferranbt: Didn't we have a PR to use all nodes as bootnodes?

ferranbt (approved):

scripts/cluster
```diff
@@ -0,0 +1,31 @@
...
+  --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000 \
+  --block-gas-limit 10000000
+
+./polygon-edge server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :10000 --libp2p :10001 --jsonrpc :10002 --seal --log-level DEBUG &
+./polygon-edge server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :20000 --libp2p :20001 --jsonrpc :20002 --seal &
+./polygon-edge server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :30000 --libp2p :30001 --jsonrpc :30002 --seal &
+./polygon-edge server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :40000 --libp2p :40001 --jsonrpc :40002 --seal &
+
+# Uncomment to run with polybft
```
Stefan-Ethernal: Could we enhance script so that it contains flag which consensus we want to run, instead of (un)commenting code?

scripts/cluster
```diff
@@ -0,0 +1,31 @@
...
+# ./polygon-edge polybft-secrets --data-dir test-chain- --num 4
+# ./polygon-edge genesis --consensus polybft --block-gas-limit 10000000 --validator-set-size=4 --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000
```
Stefan-Ethernal:
```suggestion
# ./polygon-edge genesis --consensus polybft --block-gas-limit 10000000 --validator-set-size=4
```

scripts/cluster
```diff
@@ -0,0 +1,31 @@
...
+rm -rf test-chain-*
+rm -f genesis.json
+
+go build -o polygon-edge .
+
+nodeid=$(./polygon-edge secrets init --data-dir test-chain- --num 4 | grep Node | head -n 1 | awk -F ' ' '{print $4}')
+
+./polygon-edge genesis --consensus ibft --ibft-validators-prefix-path test-chain- \
+  --bootnode /ip4/127.0.0.1/tcp/10001/p2p/"$nodeid" \
+  --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000 \
```
Stefan-Ethernal: If I'm correct, this doesn't do anything, since we are not sure that given address 0x85da99c8a7c2c95964c8efd687e95e632fc533d6 exists. Am I right? I would omit premine flag and rely on default premine balance (1M tokens/10^24 weis). WDYT?

```suggestion
```

JekaMas (approved):

scripts/cluster
igorcrevar: Currently, there is switch on `v3-parity`: `--premine-validators 1000000000000000000000` which can be used to premine some balance for all the validators. I am not sure if we gonna keep that switch after merging to develop

scripts/cluster
igorcrevar: The functionality is only on `v3-parity` I think. If `--bootnode` is not specified, all the validators are set as bootnodes

scripts/cluster
Stefan-Ethernal: @igorcrevar No need to do that because all the validators are going to get 1M of tokens if premine flag is not provided. 🙂

vcastellm: It surely does, it's the same as in `v3` it pre-mines the account.

vcastellm: FTR: The problem is that we don't have a fixed address for external testing, load testing tools and that, you know? so instead of update the address for testing on each run we rely on a fixed address that can be hardcoded in external tools.

Stefan-Ethernal: We have discussed this offline and it turns out that address which we are going to premine is going to be used by the load testing tools, so we want to make sure that we have premined appropriate address.

scripts/cluster
vcastellm: Applied in https://github.com/0xPolygon/polygon-edge/pull/912/commits/ae15c6e795a0f5f8f7bb18d650102f6cc68a2747

scripts/cluster
vcastellm: Not sure about this

vcastellm: Checked and in `develop` the `bootnode` flag is required 🤔

Stefan-Ethernal (approved):
LGTM.

b00ris (approved):

PR #919, opened 2022-11-16
Revert removal of MarshalText function

ferranbt:
# Description

This PR fixes an issue with the encoding of blocks in jsonrpc. In a previous PR, the `Nonce` encoding for `json` was removed. Thus, it was not possible to render the block correctly. 

Tasks:
- [x] Add some unit tests to the encoding.
- [x] Update `ethgo` library (with https://github.com/umbracle/ethgo/pull/228) since it would have catch up this issue during e2e.
# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


epikichi:
Confirming that performance test tools affected now return expected results with the changes. :pray:

jsonrpc/types_test.go
```diff
@@ -137,3 +138,63 @@ func TestBlock_Copy(t *testing.T) {
...
+		b.Transactions = []transactionOrHash{
+			transactionHash{0x8},
+		}
+		testBlock("testsuite/block-with-txn-hashes.json")
+	})
+
+	t.Run("block with transactions", func(t *testing.T) {
+		// TODO: do the same tests for the transaction object and enable this test
+		t.Skip("TODO")
+	})
```
Stefan-Ethernal, line 190: Are we going to implement this test now, otherwise maybe delete it?

You wanted here to create transaction and make sure that nonce is correctly marshaled and unmarshaled (since `Transcation` also has `Nonce`)? In that case, this test doesn't belong here, as this tests block struct.

begmaroman (approved):

vcastellm (approved):

types/header.go
```diff
@@ -54,6 +54,11 @@ func (n Nonce) String() string {
 	return hex.EncodeToHex(n[:])
 }
 
+// MarshalText implements encoding.TextMarshaler
+func (n Nonce) MarshalText() ([]byte, error) {
+	return []byte(n.String()), nil
+}
+
```
Stefan-Ethernal, line 61: Nitpick:

could we remove `Nonce` type outside of Header implementation (below `Header` methods implementation and above `Block`)? It is strange having it interleaved.

jsonrpc/types_test.go
lazartravica, line 190: redundant? or you wanted a test case with a block with many transactions?

lazartravica (approved):
Left one minor comment that may end up in deleting that skippped test case altogether,  otherwise LGTM

Stefan-Ethernal (approved):
No major complaints. LGTM.

jsonrpc/types_test.go
ferranbt: Nope, this is a test for a block with transactions (with body), not only transaction hashes.

ferranbt: The block can also return the body of the transactions, so we need to make sure that the transactions there are correct too. See comment to Lazar question too.

types/header.go
ferranbt: Yes, I have similar thoughts. Do not want to touch that yet.

goran-ethernal (approved):

jsonrpc/types_test.go
Stefan-Ethernal: Saw response to Lazar's same question (we were posting comments at about same time :slightly_smiling_face:). 

Although not sure why we don't implement such test right now?

zivkovicmilos (approved):
[provider.getBlock](https://docs.ethers.io/v5/api/providers/provider/#Provider-getBlock) is now working 👍

jsonrpc/types_test.go
ferranbt: This is a bug fix for the block encoding. I will do a followup PR similar to this for transaction.

ferranbt: Now that I see it, I understand the confusion. The name of this test is not correct. It should be `block with transaction hashes` for the first one and this one should be `block with transaction bodies`.

I will do a followup PR to add this test and encoding for `transaction`.

PR #925, opened 2022-11-17
Normalize block timestamps

zivkovicmilos:
# Description

This PR normalizes block timestamps, so blocks are created in clocks of `blockTime`.

We currently have a block generation process that follows this flow:
- create an empty header
- set the header timestamp to be the current time
- add transactions for however long blockTime is
- finish building the block

This is actually a correct way to create blocks, it respects the exact moment your blocks are created and that’s reflected in the timestamp. But imagine you have for example a 1.1s delay between two build block calls (1s is for the block time 100ms are for all other protocols running).

This would mean that whenever you grab the current time it would look something like this (assuming we start at `T=0.458`):
```
B0 = 0.458
B1 = 1.558
B2 = 2.658
B3 = 3.758
B4 = 4.858
B5 = 5.958
B6 = 7.058
B7 = 8.158
```

The Ethereum block timestamp standard is in seconds, and not milliseconds, which is very limiting since you lose a lot of information (for example, that `B5` was created much closer to the next second than to the previous one. Given this, when you make these times into unix timestamps, you are rounded down to the last second.
So for this example, the timestamps would be:

```
B0 = 0
B1 = 1
B2 = 2
B3 = 3
B4 = 4
B5 = 5
B6 = 7
B7 = 8
```

Notice how the gap between blocks is always 1s, except for `B7`, the gap to `Block 5` is by this data 2s.
If you looked at the ms time, you’d see that it’s indeed 1.2s apart, but from the pure unix second data, you conclude that it’s 2s aprat.

This PR normalizes this value, where the block is not being generated for `blockTime`, but being generated _within_ `blockTime`. If the block cannot be generated within the expected block time since the parent block, it is generated to the nearest multiple.

# Changes include

# Checklist

## Testing

# Additional comments

Fixes EVM-178


consensus/ibft/consensus_backend.go
```diff
@@ -152,8 +153,22 @@ func (i *backendIBFT) buildBlock(parent *types.Header) (*types.Block, error) {
...
+		now                = time.Now()
+	)
+
+	if potentialTimestamp.Before(now) {
+		// The deadline for creating this next block
+		// has passed, round it to the nearest
+		// multiple of block time
+		// t........t+blockT...x (t+blockT.x; now).....t+blockT (potential)
+		potentialTimestamp = now.Round(i.blockTime)
+	}
```
Kourin1996: Can you extract to a new function and write a test?

zivkovicmilos: Extracted, and added unit tests 👍 

[da5c0e9](https://github.com/0xPolygon/polygon-edge/pull/925/commits/da5c0e9e7e0d6537338e0ac82806c875f6d53a69)

epikichi (approved):
Block time metrics report consistent values as expected after these changes. LGTM!

consensus/ibft/consensus_backend.go
```diff
@@ -225,6 +212,37 @@ func (i *backendIBFT) buildBlock(parent *types.Header) (*types.Block, error) {
...
+// nearest higher multiple
+func roundUpTime(t time.Time, roundOn time.Duration) time.Time {
+	roundedTime := t.Round(roundOn)
+
+	if t.Sub(roundedTime) >= 0 {
+		roundedTime = roundedTime.Add(roundOn)
+	}
+
+	return roundedTime
+}
```
Kourin1996: FYI, you can write like this

```go
func roundUpTime(t time.Time, roundOn time.Duration) time.Time {
	return t.Add(roundOn / 2).Round(roundOn)
}
```

Kourin1996 (approved):

consensus/ibft/consensus_backend.go
zivkovicmilos: Good catch 💯 

Simplified:
[78d0254](https://github.com/0xPolygon/polygon-edge/pull/925/commits/78d02540b764b22a8acb29ed631f2c6097e9b54a)

JekaMas:
Just to double check, will it work with things like leap seconds?

trimixlover:

PR #927, opened 2022-11-17
introducing workflow dispatch trigger for Devnet V3 workflow.

epikichi:
# Description

Introducing DevNet V3 workflow as a `workflow_dispatch` trigger. Should be able to use this workflow for testing polybft changes. Workflow need to be in the default `develop` branch to enable dispatching.


JekaMas (approved):

vcastellm (approved):
LGTM

PR #928, opened 2022-11-17
Epikichi/edge 904 investigate and add reproducable build

epikichi:
# Description

Adding build reproducible validation step in the build workflow. 

## Notes on Behavior
- Reproducibility errors  will not fail the job nor the step, but will show in the workflow.
- Builds with `-trimpath` and `-buildvcs=false` to prevent build context info from being injected into the artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,6 @@ name: Pull Request CI
 on:  # yamllint disable-line rule:truthy
   workflow_dispatch: {}
   pull_request:
-    paths:
-      - '**.go'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,8 +3,6 @@ name: Pull Request CI
 on:  # yamllint disable-line rule:truthy
   workflow_dispatch: {}
   pull_request:
-    paths:
```
epikichi, line 6: Removing since it the behavior providing much value.

JekaMas (approved):

vcastellm (approved):
LGTM

PR #939, opened 2022-11-20
Track updates for validators

Stefan-Ethernal:
# Description

This PR introduces another field to the `ValidatorSetDelta` which allows us to track updates in `ValidatorMetadata` instances.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/extra.go
+++ b/consensus/polybft/extra.go
@@ -144,7 +144,8 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
 // createValidatorSetDelta calculates ValidatorSetDelta based on the provided old and new validator sets
 func createValidatorSetDelta(log hclog.Logger, oldValidatorSet,
 	newValidatorSet AccountSet) (*ValidatorSetDelta, error) {
-	var addedValidators AccountSet
+	addedValidators := make(AccountSet, 0)
+	updatedValidators := make(AccountSet, 0)
 
 	oldValidatorSetMap := make(map[types.Address]*ValidatorMetadata)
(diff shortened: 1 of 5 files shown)


igorcrevar (approved):

b00ris (approved):

consensus/polybft/extra.go
```diff
@@ -158,15 +159,19 @@ func createValidatorSetDelta(log hclog.Logger, oldValidatorSet,
...
+		if validatorExists {
+			if !oldValidator.EqualAddressAndBlsKey(newValidator) {
 				return nil, fmt.Errorf("validator '%s' found in both old and new validator set, but its BLS keys differ",
 					newValidator.Address.String())
 			}
 
 			// If it is, then discard it from removed validators...
 			delete(removedValidators, newValidator.Address)
+
+			if !oldValidator.Equals(newValidator) {
```
goran-ethernal, line 172: You could just check if their voting power is not equal here. No need to check the address and bls key, since if they are not equal, the previous condition will fail.

goran-ethernal (approved):
Left a short comment, but in general looks good.

consensus/polybft/extra.go
Stefan-Ethernal: True story, but I'm not entirely sure if this condition should exist in the first place: https://github.com/0xPolygon/polygon-edge/blob/293b050f9557f71f29c34525071e0789f09a8468/consensus/polybft/extra.go#L164 Therefore if we opt to remove it one day, it is better and I'm more confident to test equality in order to determine updates.

PR #942, opened 2022-11-21
Add Dockerfile

scalalang2:
# Description

I put a Dockerfile to compile polygon-edge with docker.
This can improve the development productivity as we can create a sample [polygon-edge network locally](https://github.com/scalalang2/polygon-edge-docker-compose).

# Changes include

# Breaking changes

No Breaking changes

# Checklist

## Testing

### Manual tests

let's execute below command

```shell
docker build -t polygon-edge .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM golang:1.18.8-alpine as build_base
+
+WORKDIR /src
+
+COPY go.mod .
+COPY go.sum .
+
+RUN apk add --no-cache gcc librdkafka-dev libc-dev
+RUN cd /src && go mod download
+
+FROM build_base AS builder
+
+ADD . /src
+RUN cd /src && go build --tags musl -o polygon-edge main.go
+
+FROM alpine:3.14
+
+RUN set -x \
+    && apk add --update --no-cache \
+       ca-certificates \
+    && rm -rf /var/cache/apk/*
+
+COPY --from=builder /src/polygon-edge /usr/local/bin
+
+EXPOSE 8545 9632 1478
+ENTRYPOINT ["polygon-edge"]


scalalang2:

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY go.mod .
+COPY go.sum .
+
+RUN apk add --no-cache gcc librdkafka-dev libc-dev
+RUN cd /src && go mod download
+
+FROM build_base AS builder
+
+ADD . /src
+RUN cd /src && go build --tags musl -o polygon-edge main.go
```
epikichi, line 14: Can we include the necessary `-ldflags`  to inject versioning context into the image builds as well? We can include them as `--build-args` if needed. 
- `-X "github.com/0xPolygon/polygon-edge/versioning.Version=${GITHUB_REF_NAME}"` 
- `-X "github.com/0xPolygon/polygon-edge/versioning.Commit=${GITHUB_SHA}"`

[Reference for flags](https://github.com/0xPolygon/polygon-edge/blob/f80b615edb8a69edea2a363b85a7f42c4a191251/.github/workflows/build.yml#L20)

epikichi (changes requested):
Thanks for the contribution! This Dockerfile should help offer a familiar approach for those testing/developing in containerized environments. Added one change request to help provide additional context for developers while working with the image. 

@ZeljkoBenovic We have 3 separate Dockerfiles, including this one after acceptance. In the future, we can do something to [consolidate the one made for compose](https://github.com/0xPolygon/polygon-edge/blob/develop/docker/local/Dockerfile) with the proposed one in this PR, since they achieve similar goals.

Dockerfile
scalalang2: thanks for feedback : ) I'll apply those changes soon.

ZeljkoBenovic:
Don't we already have [Dockerfile](https://github.com/0xPolygon/polygon-edge/blob/develop/docker/local/Dockerfile) in docker folder? In the same folder there is `docker-compose` as well.
Why is this `Dockerfile` different than the one we already have?

scalalang2:
@ZeljkoBenovic 
At the time of writing this pull request, I couldn't find the `Dockerfile` since It's supposed to be on `root` directory.
There is no `Dockerfile` on root dir, so then, I thought that it doesn't exsit.

scalalang2:
I agree that so I'll close this pr.

PR #949, opened 2022-11-21
Docker setup: Wait for genesis file to exist before starting nodes

mediremi-antithesis:
# Description

The Docker image entrypoint `docker/local/polygon-edge.sh` had a race condition where if the `init` container took too long to generate the `genesis.json` genesis file, nodes would fail with `Error: open /genesis/genesis.json: no such file or directory`.

This fix ensures that nodes wait for this file to exist (by polling for its existence in a loop) prior to starting up.

# Changes include

# Checklist

## Testing

### Manual tests

Manually tested by running `docker-compose up`

### DIFF
--- a/docker/local/docker-compose.yml
+++ b/docker/local/docker-compose.yml
@@ -24,7 +24,8 @@ services:
       dockerfile: docker/local/Dockerfile
     command: ["server", "--data-dir", "/data", "--chain", "/genesis/genesis.json", "--grpc-address", "0.0.0.0:9632", "--libp2p", "0.0.0.0:1478", "--jsonrpc", "0.0.0.0:8545", "--seal"]
     depends_on:
-      - init
+      init:
+        condition: service_completed_successfully
     ports:
       - '10000:9632'
       - '10002:8545'
@@ -41,7 +42,8 @@ services:
       dockerfile: docker/local/Dockerfile
     command: ["server", "--data-dir", "/data", "--chain", "/genesis/genesis.json", "--grpc-address", "0.0.0.0:9632", "--libp2p", "0.0.0.0:1478", "--jsonrpc", "0.0.0.0:8545", "--seal"]
     depends_on:
-      - init
+      init:
+        condition: service_completed_successfully
     ports:
       - '20000:9632'
       - '20002:8545'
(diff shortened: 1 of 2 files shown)


mediremi-antithesis:

epikichi (commented):
Thanks for the identifying this behavior and adding a remediation! This change looks good functionally.

I want to see your thoughts on implementing  the `depends_on` [long syntax condition](https://docs.docker.com/compose/compose-file/#long-syntax-1) to handle this race scenario on the compose level for better `init` feedback. We should be able to set the dependency to wait for the `service_completed_successfully` condition  to avoid premature node starts and uncaught init exceptions.

This approach does bring up an inevitable downstream condition where if the `genesis.json` already exists in the defined volumes (`make run-local; make stop-local; make run-local` to simply reproduce), then the `init` service will exit with status code `1` and fail the condition.  Adding a catch for this scenario in the [polygon-edge.sh](https://github.com/0xPolygon/polygon-edge/blob/develop/docker/local/polygon-edge.sh) file should be sufficient for that case.

vcastellm (approved):
LGTM

mediremi-antithesis:
I didn't know about the `depends_on` long syntax but that definitely sounds like a cleaner way of implementing this.

Unfortunately I use `podman-compose` instead of `docker-compose`, which doesn't support the long syntax version of `depends_on`, so I'm unable to test it. So please feel free to make that change and push to this branch.

I have made the change you suggested regarding the `init` command, which will now only attempt to generate the genesis file if it does not already exist.

epikichi:
> I didn't know about the `depends_on` long syntax but that definitely sounds like a cleaner way of implementing this.
> 
> Unfortunately I use `podman-compose` instead of `docker-compose`, which doesn't support the long syntax version of `depends_on`, so I'm unable to test it. So please feel free to make that change and push to this branch.
> 
> I have made the change you suggested regarding the `init` command, which will now only attempt to generate the genesis file if it does not already exist.

Thanks for the amendments! Made a quick https://github.com/mediremi-antithesis/polygon-edge/pull/1 to add in the changes as suggested.

epikichi (approved):
Tested and worked with the mentioned above changes regarding docker compose conditions. LGTM!

ZeljkoBenovic (approved):
LGTM 🏅

PR #951, opened 2022-11-22
Add discard reason in eth_getTransactionReceipt

Kourin1996:
# Description

This PR lets `eth_getTransactionReceipt` return message indicating the reason of tx discard.

```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "eth_getTransactionReceipt",
  "params": [
    "0x...." 
  ]
}
```

If the reason is recorded in journal, JSON-RPC returns string reason as an error instead of nil.

```json
{
  "message": "Tx discarded by reaching recover limit"
}
```

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

1. Setup & Start Edge clusters in the usual manner
2. Send a transaction
3. Call `eth_txStatus` giving the tx hash and make sure the status is right

# Documentation update

# Additional comments


txpool/journal.go
```diff
@@ -0,0 +1,91 @@
...
+
+// memoryJournal is an implementation of Journal and it saves data in-memory
+type memoryJournal struct {
+	txHashMap sync.Map
+}
+
+// newMemoryJournal initializes memoryJournal
+func newMemoryJournal() *memoryJournal {
+	return &memoryJournal{
+		txHashMap: sync.Map{},
```
ferranbt: Does the map implement a cache? Otherwise, a malicious actor could attack the pool sending thousands of dropped transactions.

e2e/framework/config.go
```diff
@@ -46,6 +46,7 @@ type TestServerConfig struct {
 	ValidatorType           validators.ValidatorType // Validator Type
 	Bootnodes               []string                 // Bootnode Addresses
 	PriceLimit              *uint64                  // Minimum gas price limit to enforce for acceptance into the pool
+	MaxSlots                *uint64                  // Maximum available slots in TxPool
```
ferranbt, line 49: Could we move this to another PR or is required for this?

jsonrpc/eth_endpoint.go
```diff
@@ -769,6 +775,11 @@ func (e *Eth) Unsubscribe(id string) (bool, error) {
 	return e.filterManager.Uninstall(id), nil
 }
 
+// TxStatus returns a status of the Tx that has been added in TxPool
+func (e *Eth) TxStatus(txHash types.Hash) (*string, error) {
```
ferranbt: Could this be merged with `eth_getTransactionReceipt`?

txpool/journal.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	TxSuccessful TxStatus = iota // Tx has been mined in a block successfully
+	TxDropped                    // Tx was dropped for some reason
+)
+
+var (
+	// txStatusToString is a mapping from TxStatus to text status
+	txStatusToString = map[TxStatus]string{
+		TxSuccessful: "success",
```
ferranbt: I think the journal should be used only to signal when a transaction was dropped from the pool and why.

txpool/journal.go
```diff
@@ -0,0 +1,91 @@
...
+		TxDropped:    "dropped",
```
ferranbt: Dropped should have a bit more context on why was it dropped: gas limit, not enough space in pool, nonce gap too big...

ferranbt (changes requested):

txpool/journal.go
Kourin1996: it holds tx statuse like mined （processed in a block already）or dropped.
Yes, you're right. Someone can send many txs that can't be processed （unconsective nonces） and occupy node's memory.
I think tx has 4 statuses
unconfirmed: tx doesn't reach to node
pending: tx is in TxPool
dropped: tx has been removed from TxPool some reason (unrecoverable error, recoverable error but error happened many times, and cleaning up nonce holes)
successful: tx has been processed and packed in a block

You can get tx with pending or successful status by existing API. But you can't get dropped or unconfirmed tx. That's why I added the dropped tx's hash in journal now

Do you think we don't need record dropped tx? or need cleanup function to keep memory usage stable?

e2e/framework/config.go
Kourin1996: It's used in new e2e test to drop tx from tx pool due to nonce hole cleanups. It's easy and simple way. but let me try to think about another approach

jsonrpc/eth_endpoint.go
Kourin1996: So eth_getTransactionReceipt will return the following types of value?
・receipt body
・"success"
・"dropped"
・null

txpool/journal.go
Kourin1996: it can be after we merge with 'eth_getTransactionReceipt'

txpool/journal.go
Kourin1996: Do we need to return this reason as a part of JSONRPC response? 😄

e2e/framework/config.go
lazartravica: This is only for controlling the txpool max memory footprint in the e2e tests, it's not a MaxSlot implementation as that is already present in the txpool just hasn't been used in tests before @ferranbt @Kourin1996

txpool/journal.go
```diff
@@ -0,0 +1,91 @@
...
+// newMemoryJournal initializes memoryJournal
+func newMemoryJournal() *memoryJournal {
+	return &memoryJournal{
+		txHashMap: sync.Map{},
+	}
+}
+
+// logSuccessfulTx marked the Tx as mined
+func (j *memoryJournal) logSuccessfulTx(txHash types.Hash) {
+	j.txHashMap.Store(txHash, true)
```
lazartravica: We don't need to log successful txs.

lazartravica (changes requested):
I don't think that we need journaling of successful transactions.

The only purpose of this new journal is to keep track of the transactions that are acceptable to the TxPool, but were dropped from it (for a multitude of reasons).

To summarize, a couple of big changes that need to be done on the PR:
- Log only dropped transactions from TxPool (ones that are rejected from JSON-RPC we don't care about, and ones that are included in a block there is the getTransactionReceipt call)
- The journal should be bounded in memory, something like a circular list where the eleements would overwrite in chronological order would be perfect (ideally without allocating memory all at once, imo).
- Add descriptive reasons (there's a finite number of them to why it got dropped).

As for reusing the `eth_getTransactionReceipt` or the `eth_getTransactionByHash` endpoint with these informations, I suggest to think about that later, not to start breaking web3 clients on accident and just provide this endpoint.

Also, @Kourin1996 we need a linked Docs PR for this endpoint. There's no docs on this other than the ones we write :) 

@ferranbt @Kourin1996 please let's reach consensus first here in the discussion , before going into implementation 🙏

Kourin1996:
@lazartravica 
> The journal should be bounded in memory, something like a circular list where the eleements would overwrite in chronological order would be perfect (ideally without allocating memory all at once, imo)

I have no idea of the best data structure but journal can have ring buffer (circular list) to hold keys and lookup map from key to data (flag and reason,etc)

ferranbt:
Agree with everything that @lazartravica mentions, some of my reviews go in that direction too.

> As for reusing the eth_getTransactionReceipt or the eth_getTransactionByHash endpoint with these informations, I suggest to think about that later, not to start breaking web3 clients on accident and just provide this endpoint.

I think it is better to update `eth_getTransactionReceipt` since that one represents the end status of a transaction.

I am not worried about breaking the endpoint, this is just an error message (if as discussed we only journal the drop reasons). I think it is way more cumbersome to add support for new endpoints in web3 clients (I am not sure I would add it on `ethgo` only to give support for `edge` for example).

> I have no idea of the best data structure but journal can have ring buffer (circular list) to hold keys and lookup map from key to data (flag and reason,etc)

Use a simple cache like https://github.com/hashicorp/golang-lru. Lets not overcomplicate the design.

Kourin1996:
@ferranbt Thank you for reviewing the discussion!
cc: @lazartravica 

> Use a simple cache like https://github.com/hashicorp/golang-lru. Lets not overcomplicate the design

In such a case, I thought the data in journal wouldn't be discarded in FIFO. Is it ok for you?

txpool/journal.go
JekaMas: > Does the map implement a cache? Otherwise, a malicious actor could attack the pool sending thousands of dropped transactions.

Had exactly the same question in the RFC :)

It might make sense to have cache based also on accounts or have the bigger TTL the greater nonce is.

jsonrpc/eth_endpoint.go
JekaMas: I don't think that it is related to receipts. A transaction might not have a receipt at all.

txpool/journal.go
```diff
@@ -0,0 +1,91 @@
...
+	"sync"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+// TxStatus indicates the status of tx that is added to TxPool
+type TxStatus int8
+
+const (
+	TxSuccessful TxStatus = iota // Tx has been mined in a block successfully
```
JekaMas: Why we need this statuses as ints, not strings?

JekaMas: Is it for performance?

JekaMas: if so then it's better to start with initial value 1, not 0 and treat 0 as incorrect.

txpool/journal.go
JekaMas: "gas limit, not enough space in pool, nonce gap too big" we might need types errors for users with error codes as well. what do you think?

txpool/journal.go
JekaMas: Do we define 'slot' the same way as in Geth?

lazartravica: @JekaMas yep, https://github.com/0xPolygon/polygon-edge/blob/b97e753e8300e488e88a9046bf2aa2628bd269cc/txpool/txpool.go#L24-L25

lazartravica:
> @ferranbt Thank you for reviewing the discussion! cc: @lazartravica
> 
> > Use a simple cache like https://github.com/hashicorp/golang-lru. Lets not overcomplicate the design
> 
> In such a case, I thought the data in journal wouldn't be discarded in FIFO. Is it ok for you?

I agree completely with @ferranbt suggestion let's not introduce complexity and reuse the package we already have, the LRU cache itself will prove to be good.

I am also in-line with `eth_getTransactionReceipt`, though not aware of the complexity of making that change.

cc @Kourin1996

jsonrpc/eth_endpoint.go
```diff
@@ -295,7 +298,20 @@ func (e *Eth) GetTransactionByHash(hash types.Hash) (interface{}, error) {
...
+		if err != nil {
+			e.logger.Warn(
+				"failed to get the reason of tx discard: hash=%s, err=%s",
+				hash,
+				err.Error(),
+			)
+		}
+
+		if reason != nil {
+			return *reason, nil
```
ferranbt: This is not API-compatible with the JsonRPC spec, it might cause problems with the client implementation. I was under the impression that we would return it as an error of the JsonRPC call, which would be more compatible, and it kind of makes sense since this is an error on the transaction.

ferranbt (changes requested):

jsonrpc/eth_endpoint.go
Kourin1996: Fixed 05377c9d21c5dc0241a8c9637eafdfc17b0b0bc9

PR #957, opened 2022-11-22
Epikichi/edge 973 change v3 parity workflow to devnet

epikichi:
# Description

Adding workflow changes to this branch to allow ephemeral environment deployments. This change is necessary for us to enable workflow dispatches, as it requires the workflow to exist in the `develop` branch as well as the target branch.

Also minor change, re-naming the `devnet-v3-parity` environment to a more general `devnet-ephemeral`


vcastellm (approved):
GO

PR #963, opened 2022-11-25
Create the JSON RPC endpoint to generate state sync proofs

vcastellm:
# Description

Create the JSON RPC endpoint to generate state sync proofs

This goes after:
* #964 

# Changes include

# Checklist

## Testing

### Manual tests

Querying the RPC endpoint

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -856,6 +856,20 @@ func (c *consensusRuntime) GenerateExitProof(exitID, epoch, checkpointBlock uint
 	return tree.GenerateProofForLeaf(e, 0)
 }
 
+// GetStateSyncProof returns the proof of the bundle for the state sync
+func (c *consensusRuntime) GetStateSyncProof(stateSyncID uint64) ([]types.Hash, error) {
+	bundlesToExecute, err := c.state.getBundles(stateSyncID, 1)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get bundles: %w", err)
+	}
+
+	if len(bundlesToExecute) == 0 {
+		return nil, fmt.Errorf("cannot find StateSync with id %d", stateSyncID)
+	}
+
+	return bundlesToExecute[0].Proof, nil
+}
+
 // setIsActiveValidator updates the activeValidatorFlag field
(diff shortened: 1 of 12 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -856,6 +856,29 @@ func (c *consensusRuntime) GenerateExitProof(exitID, epoch, checkpointBlock uint
...
+	bundlesToExecute, err := c.state.getBundles(stateSyncID, 1)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get bundles: %w", err)
+	}
+
+	if len(bundlesToExecute) == 0 {
+		return nil, fmt.Errorf("cannot find bundle containing StateSync id %d", stateSyncID)
+	}
+
+	for _, bundle := range bundlesToExecute[0].StateSyncs {
```
goran-ethernal, line 855: Could we remove that `BundleProof` struct we had earlier, and when we build bundles, just directly build this new `StateSyncProof`, or will we do it in another PR?
It would even be easier to save these `StateSyncProof`s in `boltdb` and get them when `GetStateSyncProof` is called, because its key would be the `StateSync.ID`.

vcastellm: We are going to do that, in RPC we're only exposing StateSyncs, I can start right away with the bigger refactor after merging the complete workflow we have for the demo.

goran-ethernal (approved):
LGTM :100:

consensus/polybft/consensus_runtime.go
```diff
@@ -856,6 +856,29 @@ func (c *consensusRuntime) GenerateExitProof(exitID, epoch, checkpointBlock uint
...
+		if bundle.ID == stateSyncID {
```
goran-ethernal, line 856: Small thing, I think we don't need this condition, since we have only one state sync event per bundle now, so it will contain only the one we are looking for.

vcastellm: Right, this was leftover from previous tests, but let's leave it as is at this time just in case, we'll remove it together with bundles in next PRs.

lazartravica (approved):

Stefan-Ethernal (approved):

Stefan-Ethernal (approved):

PR #965, opened 2022-11-25
Fix data race to polybft unit test

0xSasaPrsic:
# Description

Enables polybft unit tests to run with race flag.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ test-e2e:
 test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
-	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true go test -race -v -timeout=30m ./e2e-polybft/...
+	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true go test -v -timeout=30m ./e2e-polybft/...
 
 .PHONY: run-local
 run-local:
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ test-e2e:
 test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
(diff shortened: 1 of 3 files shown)


consensus/polybft/event_tracker_test.go
```diff
@@ -13,17 +14,28 @@ import (
...
 	}
 
 	m.logs = append(m.logs, log)
 }
 
+func (m *mockEventSubscriber) getLogs() []*ethgo.Log {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	return m.logs
```
JekaMas: It won't work with slices, if we can overwrite the slice elements for example? Can we have like thread safe Len() method instead?

0xSasaPrsic: Updated with https://github.com/0xPolygon/polygon-edge/pull/965/commits/1ce9901abef40e7b193850fe78678855d677bef1

Makefile
```diff
@@ -54,7 +54,7 @@ test-e2e:
 test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
-	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true go test -v -timeout=30m ./e2e-polybft/...
+	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true go test -race -v -timeout=30m ./e2e-polybft/...
```
Stefan-Ethernal: This target relates to polybft e2e tests. It doesn't harm adding it here, though unit tests are executed using `test` target. Therefore maybe we should consider adding race flag there (although not sure do we want to open that Pandora box, since it executes all the unit tests both from Edge and Polybft ones)?

0xSasaPrsic: Indeed, but this can't run with flag because it reports some issues that are not polybft. 
However, polybft e2e tests are testing specific target and that is why I added it. Do you think we should skip race flag until boltdb issue is fixed?

b00ris (approved):

goran-ethernal (approved):

PR #975, opened 2022-11-28
Extracted GetAccountImpl into fn for use in txPoolHub and jsonRPCHub 

lazartravica:
Fixes EVM-241

# Description

Streamlined error handling on non-existent accounts which caused a runtime error on pending block param with `eth_getTransactionCount`.

# Changes include

# Checklist

## Testing

### Manual tests

Ran `eth_getTransactionCount` with pending block param.

# Additional comments

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -305,10 +305,6 @@ func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
 		return 0
 	}
 
-	if account == nil {
-		return 0
-	}
-
 	return account.Nonce
 }
 
@@ -317,10 +313,6 @@ func (t *txpoolHub) GetBalance(root types.Hash, addr types.Address) (*big.Int, e
 
 	if err != nil {
 		return big.NewInt(0), err
-	}
-
-	if account == nil {
-		return big.NewInt(0), nil
 	}
 
 	return account.Balance, nil
--- a/server/server.go
+++ b/server/server.go
@@ -298,6 +298,25 @@ type txpoolHub struct {
 	*blockchain.Blockchain
 }
 
+// getAccountImpl is used for fetching account state from both TxPool and JSON-RPC
+func getAccountImpl(state state.State, root types.Hash, addr types.Address) (*state.Account, error) {
+	snap, err := state.NewSnapshotAt(root)
+	if err != nil {
+		return nil, err
+	}
+
+	account, err := snap.GetAccount(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	if account == nil {
+		return nil, jsonrpc.ErrStateNotFound
+	}
+
+	return account, nil
+}
+
 func (t *txpoolHub) GetNonce(root types.Hash, addr types.Address) uint64 {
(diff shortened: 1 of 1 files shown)


b00ris (approved):

server/server.go
```diff
@@ -432,8 +423,8 @@ func (j *jsonRPCHub) GetPeers() int {
 	return len(j.Server.Peers())
 }
 
-func (j *jsonRPCHub) getAccountImpl(root types.Hash, addr types.Address) (*state.Account, error) {
-	snap, err := j.state.NewSnapshotAt(root)
+func getAccountImpl(state state.State, root types.Hash, addr types.Address) (*state.Account, error) {
+	snap, err := state.NewSnapshotAt(root)
 	if err != nil {
 		return nil, err
```
Stefan-Ethernal: Just a minor suggestion if it makes sense to you (inspiration found in `GetBalance` function which formatted error message in a similar manner).

```suggestion
		return nil, fmt.Errorf("unable to get snapshot for root '%s': %w", root, err)
```

server/server.go
```diff
@@ -470,7 +461,7 @@ func (j *jsonRPCHub) GetForksInTime(blockNumber uint64) chain.ForksInTime {
 }
 
 func (j *jsonRPCHub) GetStorage(stateRoot types.Hash, addr types.Address, slot types.Hash) ([]byte, error) {
-	account, err := j.getAccountImpl(stateRoot, addr)
+	account, err := getAccountImpl(j.state, stateRoot, addr)
```
Stefan-Ethernal, line 463: What do you think about returning `snapshot` instance from the `getAccountImpl`? Asking this, because in this use case, we are going to call `NewSnapshotAt` twice for the same root. 

Although it probably doesn't matter, as we rely on LRU cache when accessing to the trie, and when we call `NewSnapshotAt` second time, we will most probably have the trie in the cache.

Stefan-Ethernal (approved):
LGTM, neat :slightly_smiling_face:

server/server.go
```diff
@@ -432,8 +423,8 @@ func (j *jsonRPCHub) GetPeers() int {
 	return len(j.Server.Peers())
 }
 
-func (j *jsonRPCHub) getAccountImpl(root types.Hash, addr types.Address) (*state.Account, error) {
-	snap, err := j.state.NewSnapshotAt(root)
+func getAccountImpl(state state.State, root types.Hash, addr types.Address) (*state.Account, error) {
```
vcastellm: Nitpick: move this method to top of the file?

vcastellm (approved):
LGTM

ferranbt (approved):

server/server.go
lazartravica: The fact that it is cached from the previous call in getAccountImpl, and that only the GetStorage fetches it twice (which is not a frequently used endpoint) I think it's better to make a compromise to have a smaller function signature in this case.

Stefan-Ethernal: Yep agreed, just was thinking out loud and wasn't sure myself either :+1:

PR #978, opened 2022-11-29
Replace BLS library with github.com/kilic/bn254

igorcrevar:
# Description

Replace BLS library with github.com/kilic/bn254

# Changes include

# Checklist

## Testing


goran-ethernal (approved):
LGTM. Great job :100:

b00ris (approved):

consensus/polybft/signer/public.go
```diff
@@ -6,36 +6,46 @@ import (
...
 }
 
 // aggregate adds the given public keys
-func (p *PublicKey) aggregate(onemore *PublicKey) *PublicKey {
-	var g2 *bn256.G2
-	if p.p == nil {
-		g2 = new(bn256.G2).Set(&zeroG2)
-	} else {
-		g2 = new(bn256.G2).Set(p.p)
+func (p *PublicKey) aggregate(next *PublicKey) *PublicKey {
```
JekaMas, line 18: As a separate task could we remove this code duplication(publicKey, signature)?

JekaMas (approved):

Stefan-Ethernal (approved):

consensus/polybft/signer/public.go
Stefan-Ethernal: At first glance it seemed like it can be unified into a single implementation, but eventually it cannot. The reason is because signature aggregation works with G1 point, whereas public key aggregates G2 points.

PR #982, opened 2022-11-30
Fix data race in fsm

0xSasaPrsic:
# Description

Lock is added to prevent data race in hasQuorum method.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -394,6 +394,7 @@ func (c *consensusRuntime) FSM() error {
 	c.lock.Lock()
 	c.fsm = ff
 	c.lock.Unlock()
+
 	return nil
 }
 
@@ -999,6 +1000,7 @@ func (c *consensusRuntime) IsValidBlock(proposal []byte) bool {
 func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
 	c.lock.Lock()
 	defer c.lock.Unlock()
+
 	err := c.fsm.ValidateSender(msg)
 	if err != nil {
 		c.logger.Error("invalid sender", "error", err)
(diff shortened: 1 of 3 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -996,6 +998,9 @@ func (c *consensusRuntime) IsValidBlock(proposal []byte) bool {
 }
 
 func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
+	c.lock.Lock()
```
JekaMas: is it Lock or Rlock?

0xSasaPrsic: good catch, I think it can be RL on all checks.

JekaMas: agree, only `c.fsm = fsm` at the very beginning of the PR should be a write lock.

JekaMas (approved):

JekaMas (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -996,6 +998,9 @@ func (c *consensusRuntime) IsValidBlock(proposal []byte) bool {
 }
 
 func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
+	c.lock.RLock()
```
igorcrevar, line 1001: I suggest doing on the top:
```
c.lock.RLock()
fsm := c.fsm
c.lock.RUnlock()
```
Or even better make `func getFsm` which will call every other function that needs fsm -> on the top
`fsm := c.getFsm()`

JekaMas: getFSM() should do deep copying then. I believe it's a pointer type so just returning won't help much.

JekaMas: @igorcrevar

b00ris (approved):

JekaMas (approved):

PR #984, opened 2022-11-30
EVM-208 Avoid necessary creation of validator set + EVM-211 Extract proposer algorithm from validator set

igorcrevar:
# Description

While validating headers we need to check quorum. For that purposes we obtain `AccountSet` and create `ValidatorSet` each time, even though the `ValidatorSet` is the same for an epoch. Need to examine if this could be avoided.

`ValidatorSet` structure inside `validator_set.go` is becoming huge, maybe proposer algorithm does not have to be a part of valdator set structure (this is realy, realy nice to have and the task has been created as a possible reminder)

# Changes include

# Checklist

## Testing


consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,355 @@
...
+	var validators = make([]*validatorCalcMetadata, len(valz))
+	for i, v := range valz {
+		validators[i] = newValidatorCalcMetadata(v, 0)
+	}
+
+	proposerCalc := &proposerCalculator{
+		totalVotingPower: totalVotingPower,
+		validators:       validators,
+		lock:             &sync.RWMutex{},
+		round:            uint64(math.MaxUint64),
```
goran-ethernal: Just curious, why are we setting the round to `MaxUint64` initially? Shouldn't it just be 0?

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,355 @@
...
+	clone := pc.Copy()
+
+	// if round = 0 then we need one iteration
+	if err := clone.IncrementProposerPriority(round + 1); err != nil {
+		return types.ZeroAddress, fmt.Errorf("cannot increment proposer priority: %w", err)
+	}
+
+	var err error
+
+	proposer := clone.proposer // no need for lock here because clone is temporary copy
```
goran-ethernal, line 148: Do we need to clone entire proposer calculator object if we only need its `proposer` field?

igorcrevar: We are cloning because whole calculation algorithm is executed on copy proposer calculator object. After algorithm is done, new proposer is set on that clone object. 
We than pick from cloned object pointer to newly calculated proposer and set original one to that pointer.
The only  "problem" is that new proposer variable points to proposer on cloned object and not to some of the objects inside `pc.validators`.

consensus/polybft/proposer_calculator.go
igorcrevar: If its 0, than the first `CalcProposer` call will return "cached" value

goran-ethernal: Just leave a comment for this when setting it to `MaxUint64` :pray:

igorcrevar: Will work with zero too. Changed that with https://github.com/0xPolygon/polygon-edge/pull/984/commits/4050ffe6ebbc88e2b3fa15075acc343e14071771

goran-ethernal (approved):

Nemanja0x (approved):

consensus/polybft/fsm.go
```diff
@@ -46,6 +46,9 @@ type fsm struct {
 	// validators is the list of validators for this round
 	validators ValidatorSet
 
+	// proposerCalculator is the object which calculates new proposer
+	proposerCalculator ProposerCalculator
```
Stefan-Ethernal, line 50: Why this field is not part of `consensusRuntime` but in the FSM? Understand that it's lifespan is the same as single block lifespan, but again it is used by consensusRuntime, therefore I was expecting to see it there.

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+
+	return nil
+}
+
+// isNilOrEmpty returns true if validator set is nil or empty.
+func (pc proposerCalculator) isNilOrEmpty() bool {
+	return len(pc.validators) == 0
+}
```
Stefan-Ethernal: What do you think about inlining this function? It is a one-liner and only checks single thing. If it is going to remain, then I think it should be renamed to e.g. `isEmptyValidatorSet`.

Stefan-Ethernal (approved):
LGTM apart from minor considerations.

consensus/polybft/fsm.go
igorcrevar: Because the lifetime of this object (with currently implementing algorithm) is one sequence (which is exactly the lifetime of FSM)

consensus/polybft/proposer_calculator.go
igorcrevar: Removed https://github.com/0xPolygon/polygon-edge/pull/984/commits/a3e993bda514d24947454feaf47d95846ce64edd

consensus/polybft/consensus_runtime.go
```diff
@@ -1144,7 +1151,26 @@ func (c *consensusRuntime) HasQuorum(
 			return c.fsm.validators.HasQuorum(signers)
 		}
 
-		return c.fsm.validators.HasQuorumWithoutProposer(signers)
+		if len(messages) == 0 {
+			return false
+		}
+
+		propAddress, exist := c.fsm.proposerCalculator.GetLatestProposer(messages[0].View.Round)
```
ferranbt, line 1172: What is this? This `HasQuorum` function is getting too clumsy.

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,351 @@
...
+		"invalid voting power configuration provided: total voting power must be greater than 0")
+)
+
+// ProposerCalculator interface of the current validator set
+type ProposerCalculator interface {
+	// CalcProposer calculates next proposer based on the passed round
+	CalcProposer(round uint64) (types.Address, error)
+
+	// IncrementProposerPriority increments priorities number of times
+	IncrementProposerPriority(times uint64) error
```
ferranbt, line 43: Is this an interface for `proposerCalculator`?

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,351 @@
...
+)
+
+var (
+	// errInvalidTotalVotingPower is returned if the total voting power is zero
+	errInvalidTotalVotingPower = errors.New(
+		"invalid voting power configuration provided: total voting power must be greater than 0")
+)
+
+// ProposerCalculator interface of the current validator set
+type ProposerCalculator interface {
```
ferranbt, line 38: Is this an interface for proposerCalculator?

consensus/polybft/consensus_runtime.go
Nemanja0x: Oh yea. There was a discussion about that not so long ago. There is a situation when go-ibft calls backend  and asks for quorum, but that quorum must include proposer even though a message form the proposer hasn't been submitted to the backend. Therefore, we have to find the proposer by ourselves and include it in the quorum calculation.

consensus/polybft/proposer_calculator.go
igorcrevar: Yes, it is. In this pr main thing is separation between `validatorsSet` and `calculatorProposer`. But you should not look at proposer calculator itself here, because it is changed a lot in https://github.com/0xPolygon/polygon-edge/pull/1018. We did not know how exactly will it look like before we start implementing everything

consensus/polybft/consensus_runtime.go
igorcrevar: here, we do not need to calculate again proposer, we just need to retrieve previously calculated proposer for specific round

PR #989, opened 2022-12-01
Provide rootchain IP address from CLI

Stefan-Ethernal:
# Description

Provide JSON RPC endpoint IP address using CLI flags, instead of relying on reading locally running Geth docker instance.
Implemented `TxRelayer` abstraction which enables sending of transactions and call messages to the Ethereum-compatible blockchains.

# Changes include

# Breaking changes

Genesis configuration isn't backward compatible, due to a couple of renames in `bridge`configuration part (`bridgeAddr` -> `stateSenderAddr`)

# Checklist

## Testing


command/genesis/genesis.go
```diff
@@ -192,11 +192,11 @@ func setFlags(cmd *cobra.Command) {
...
 			"the smart contracts folder",
 		)
-		cmd.Flags().BoolVar(
-			&params.bridgeEnabled,
+		cmd.Flags().StringVar(
+			&params.bridgeJSONRPCAddr,
 			bridgeFlag,
-			defaultBridge,
-			"enabling bridge, default is not enabled",
+			"",
```
vcastellm, line 198: Can we setup here the default address?

command/rootchain/emit/emit.go
```diff
@@ -59,6 +59,13 @@ func setFlags(cmd *cobra.Command) {
 		nil,
 		"list of amounts to fund wallets",
 	)
+
+	cmd.Flags().StringVar(
+		&params.jsonRPCAddress,
+		jsonRPCFlag,
+		"",
```
ferranbt: If it is required for the command, default to `localhost:8545`.

vcastellm: Default to localhost?

command/rootchain/fund/fund.go
```diff
@@ -51,6 +51,13 @@ func setFlags(cmd *cobra.Command) {
 		"the flag indicating the number of accounts to be funded",
 	)
 
+	cmd.Flags().StringVar(
+		&params.jsonRPCAddress,
+		jsonRPCFlag,
+		"",
```
vcastellm: Same here

command/rootchain/emit/emit.go
Stefan-Ethernal: It is required, but it will fallback to resolving docker IP address. But I can default it to localhost if that makes more sense...?

command/rootchain/helper/rootchain_interactor.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	txnHash, err := d.provider.Eth().SendRawTransaction(data)
+	if err != nil {
+		return nil, err
+	}
+
+	receipt, err := d.waitForReceipt(txnHash)
```
JekaMas: maybe just `return d.waitForReceipt(txnHash)` ?

command/rootchain/helper/rootchain_interactor.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return receipt, nil
+}
+
+func (d *DefaultRootchainInteractor) GetPendingNonce(address types.Address) (uint64, error) {
+	nonce, err := d.provider.Eth().GetNonce(ethgo.Address(address), ethgo.Pending)
```
JekaMas: same here - `return d.provider.Eth().GetNonce(ethgo.Address(address), ethgo.Pending)`

JekaMas (approved):
lgtm with minor comments

command/rootchain/helper/rootchain_interactor.go
```diff
@@ -0,0 +1,156 @@
...
+func (d *DefaultRootchainInteractor) ExistsCode(contractAddr types.Address) (bool, error) {
+	code, err := d.provider.Eth().GetCode(ethgo.Address(contractAddr), ethgo.Latest)
+	if err != nil {
+		return false, err
+	}
+
+	return code != "0x", nil
+}
+
+func (d *DefaultRootchainInteractor) FundAccount(account types.Address) (types.Hash, error) {
```
b00ris, line 106: You dont use account there. You are funding accounts[0].

e2e-polybft/framework/test-cluster.go
```diff
@@ -250,7 +251,9 @@ func NewTestCluster(t *testing.T, validatorsCount int, opts ...ClusterOption) *T
 		}
 
 		if cluster.Config.HasBridge {
-			args = append(args, "--bridge")
+			rootchainIP, err := command.ResolveRootchainIP("")
```
b00ris: ?

b00ris (approved):

e2e-polybft/framework/test-cluster.go
b00ris: Is it expeced default value there?

command/genesis/genesis.go
Stefan-Ethernal: No I won't say so, because if `--bridge-json-rpc` not provided, it means that we are creating genesis in a "non-bridge-mode" (this flag replaced the previously used `--bridge`).

vcastellm: * `--bridge-json-rpc` present without value, bridge enabled with default address, value provided in other case.
* Flag not provided bridge disabled.

Can we do that?

command/rootchain/helper/rootchain_interactor.go
Stefan-Ethernal: Going to be changed very soon (refactoring it), so comment won't be applicable anymore (going to generalize a bit more `DefaultRootchainInteractor` by making it act like `TxRelayer`, which Ferran suggested https://github.com/0xPolygon/polygon-edge/commit/7484fb2cb0d0794a62ec8e8044f277796b33ecc7#diff-c83c8a1e714409dd49d38a1ac6f69af3a22eeee9947090430b37619db121ccf7) :slightly_smiling_face:

command/genesis/genesis.go
Stefan-Ethernal: Seems we cannot.

I've tried running genesis command like this:
`go run . genesis --block-gas-limit 10000000 --validator-set-size=4 --consensus polybft --epoch-size 5 --bridge-json-rpc`
and it returns a following error:

> flag needs an argument: --bridge-json-rpc

Tried this as a workaround (https://stackoverflow.com/a/70186443), but no luck as well :thinking: :
`cmd.Flags().Lookup(bridgeFlag).NoOptDefVal = "http://127.0.0.1:8545"`

e2e-polybft/framework/test-cluster.go
Stefan-Ethernal: If it is passed empty string for an IP address, that meant it should be figured out from the locally running docker instance.
Anyhow since it is only for running e2e tests, it is safe assuming that Geth is up and running. [3732f6e](https://github.com/0xPolygon/polygon-edge/pull/989/commits/3732f6e0b662b6ace6f9c297999c9ff6ffc5efd1)

command/rootchain/fund/fund.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/989/commits/3732f6e0b662b6ace6f9c297999c9ff6ffc5efd1

command/rootchain/emit/emit.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/989/commits/3732f6e0b662b6ace6f9c297999c9ff6ffc5efd1

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/989/commits/3732f6e0b662b6ace6f9c297999c9ff6ffc5efd1

A note there:
I've tried supplying `localhost:8545`, but seems `ethgo` cannot handle it since it expects specifying protocol, IP address and port notation instead.

command/rootchain/helper/rootchain_interactor.go
Stefan-Ethernal: `DefaultRootchainInteractor` is gone in favor of `TxRelayer`, but applied feedback :wink: https://github.com/0xPolygon/polygon-edge/pull/989/commits/3bd627d9aa808d76e850cb9575e7d1a38d4d5007#diff-c83c8a1e714409dd49d38a1ac6f69af3a22eeee9947090430b37619db121ccf7R93

b00ris (approved):

txrelayer/txrelayer.go
```diff
@@ -0,0 +1,189 @@
...
+		err     error
+	)
+
+	if !t.localAccount {
+		txnHash, err = t.sendSignedTransaction(txn, key)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		txnHash, err = t.sendNonSignedTransaction(txn)
```
ferranbt: Split this into two functions, it is confusing in the `SendTransaction` API call that the `key` field is not used.

Besides, this is only required for funding the main account, once we find a way to do that without the local account of the Geth node (which is possible), this functionality will be removed since it is quite legacy in Geth at this point.

command/rootchain/helper/txn.go
```diff
@@ -48,166 +44,33 @@ func GetRootchainAdminKey() ethgo.Key {
...
-}
-
-func ExistsCode(addr types.Address) (bool, error) {
-	provider, err := getJSONRPCClient()
-	if err != nil {
-		return false, err
-	}
-
-	code, err := provider.Eth().GetCode(ethgo.HexToAddress(addr.String()), ethgo.Latest)
+func ContractExists(client *jsonrpc.Client, contractAddr types.Address) (bool, error) {
```
ferranbt: Inline this on the call, This function is only used once. I do not think it has enough merit to be moved to an external function now.

command/rootchain/helper/txn.go
```diff
@@ -48,166 +44,33 @@ func GetRootchainAdminKey() ethgo.Key {
...
-func GetPendingNonce(addr types.Address) (uint64, error) {
-	provider, err := getJSONRPCClient()
-	if err != nil {
-		return 0, err
-	}
-
-	nonce, err := provider.Eth().GetNonce(ethgo.HexToAddress(addr.String()), ethgo.Pending)
+// FundAccount funds provided account.
+// Used only for testing purposes.
+func FundAccount(ipAddress string, address types.Address) (types.Hash, error) {
```
ferranbt: Same as with `ContractExists`, this function is only here to avoid 2 lines of code. I would inline this on the function call as well. Besides, at least in one instance, the relayer was already defined.

command/rootchain/emit/emit.go
```diff
@@ -76,7 +86,14 @@ func runCommand(cmd *cobra.Command, _ []string) {
...
 
-	pendingNonce, err := helper.GetPendingNonce(helper.GetRootchainAdminAddr())
+	txRelayer, err := txrelayer.NewTxRelayer(jsonRPCAddress)
+	if err != nil {
+		outputter.SetError(fmt.Errorf("could not create rootchain interactor: %w", err))
+
+		return
+	}
+
+	pendingNonce, err := txRelayer.GetNonce(ethgo.Address(helper.GetRootchainAdminAddr()))
```
ferranbt: The `relayer` handles the nonce already.

txrelayer/txrelayer.go
```diff
@@ -0,0 +1,189 @@
...
+	defaultGasPrice = 1879048192 // 0x70000000
+	defaultGasLimit = 5242880    // 0x500000
+)
+
+var (
+	errNoAccounts               = errors.New("no accounts registered")
+	errNonSignedTxnNotSupported = errors.New("sending non signed transactions supported only for local account mode")
+)
+
+type TxRelayer interface {
```
ferranbt, line 25: I would prefer if we move the interfaces to the place where we need them. In the end, an interface defines a level of abstraction, it should be the module that uses the relayer the one that describes which is the abstraction it requires.

This is, if the `consensusRuntime` needs the relayer, it should be the one defining the interface of that relayer, not relying on this.

ferranbt (changes requested):

command/genesis/genesis.go
vcastellm: I tried it, it works as expected 🤔

command/rootchain/emit/emit.go
Stefan-Ethernal: Yes it does, but it doesn't cover all the use cases.
`rootchain emit` command is submitting transactions in parallel, which means that nonce must be set from the outside in order to be valid (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/command/rootchain/emit/emit.go#L93).
Therefore I thought this is the proper way to handle it, expose `GetNonce` and use it to set nonce when creating transaction instances. Do you have any more elegant approach?

txrelayer/txrelayer.go
Stefan-Ethernal: It is being used on two places:
1. `polybft` package (`checkpointManager` uses it) and
2. `initcontracts` package (`rootchain init-contracts` uses it).

I don't like duplicating code (e.g. defining same interface twice), also it is good practice to hide implementation details from the caller (that's why I have introduced interface in the first place) therefore I would prefer to keep it in `txrelayer` package.

command/rootchain/helper/txn.go
Stefan-Ethernal: Fixed [fb9abbe](https://github.com/0xPolygon/polygon-edge/pull/989/commits/fb9abbe8661f3f160cba5dd828b5fc3b14e542bf)

command/rootchain/helper/txn.go
Stefan-Ethernal: Fixed [fb9abbe](https://github.com/0xPolygon/polygon-edge/pull/989/commits/fb9abbe8661f3f160cba5dd828b5fc3b14e542bf)

txrelayer/txrelayer.go
Stefan-Ethernal: Fixed [fb9abbe](https://github.com/0xPolygon/polygon-edge/pull/989/commits/fb9abbe8661f3f160cba5dd828b5fc3b14e542bf)

command/genesis/genesis.go
Stefan-Ethernal: Fixed [fb9abbe](https://github.com/0xPolygon/polygon-edge/pull/989/commits/fb9abbe8661f3f160cba5dd828b5fc3b14e542bf)

command/rootchain/emit/emit.go
ferranbt: Lock the `SendTransaction` call. That way the nonce will be correct. It is a naive approach and it does not work for all use cases but it works for this.

Stefan-Ethernal: Been there done that. If I remember correctly bridge e2e test started to fail, since rootchain emit command runs sequentially in that case. :disappointed: 
So there is no point we made emit command sending transactions in parallel (which was done in order to make bridge e2e test works faster).

So if you don't mind, let's keep it this way: we are expecting fully constructed transaction object provided by the caller.

ferranbt: Ok, so it is not that it does not work, but that it is not as fast for e2e which is another thing. Ok, I remember now, you have to lock only between getting the nonce and sending the transaction, unlock for receipt wait.

Once you send the transaction, the internal txpool gets updated and the next query nonce is sequential. This should fix both things.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/989/commits/f2a1c83e5f8631e495529caa64ce2e381553e249

ferranbt (approved):

vcastellm (approved):
LGTM

goran-ethernal (approved):

igorcrevar (approved):

PR #993, opened 2022-12-02
Only setup telemetry if `prometheusAddr` is set

tuommaki:
Prometheus metric collector registration fails if performed multiple times with the same configuration. This can happen if multiple server instances are started within the same process, which is sometimes useful in e2e test cases.

# Description

This change will only call `setupTelemetry()` if prometheus address has been configured.

Prometheus metric collector registrations fail if performed multiple times. This can happen if multiple server instances are started within the same process, which is sometimes useful in e2e test cases.

# Changes include

# Checklist

## Testing

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -140,11 +140,12 @@ func NewServer(config *Config) (*Server, error) {
 		return nil, fmt.Errorf("failed to create data directories: %w", err)
 	}
 
-	if err := m.setupTelemetry(); err != nil {
-		return nil, err
-	}
-
 	if config.Telemetry.PrometheusAddr != nil {
+		// Only setup telemetry if `PrometheusAddr` has been configured.
+		if err := m.setupTelemetry(); err != nil {
+			return nil, err
+		}
+
 		m.prometheusServer = m.startPrometheusServer(config.Telemetry.PrometheusAddr)
 	}
 


tuommaki:

server/server.go
```diff
@@ -140,11 +140,12 @@ func NewServer(config *Config) (*Server, error) {
 		return nil, fmt.Errorf("failed to create data directories: %w", err)
 	}
 
-	if err := m.setupTelemetry(); err != nil {
-		return nil, err
-	}
-
 	if config.Telemetry.PrometheusAddr != nil {
+		// Only setup telemetry if `PrometheusAddr` has been configured.
+		if err := m.setupTelemetry(); err != nil {
```
Stefan-Ethernal, line 145: Not sure is this change safe, because `setupTelemetry` initializes global telemetry object, which is latter used when reporting metrics:
- https://github.com/0xPolygon/polygon-edge/blob/develop/txpool/txpool.go#L277,
- https://github.com/0xPolygon/polygon-edge/blob/develop/network/server.go#L517,
- https://github.com/0xPolygon/polygon-edge/blob/develop/network/server.go#L772 etc. 
Since we don't check is `PrometheusAddr` provided when we report metrics from different modules, I believe it would fail whenever we try to report it (because as I've mentioned previously, we wouldn't have global metrics object initialized).

So to me, if we are going to use this approach, we would need to check when we are reporting metrics is `PrometheusAddr` provided in order not to cause `panic`.

Stefan-Ethernal (commented):
Hello and thanks for your PR. :slightly_smiling_face:  Posting a comment with my concern. Hopefully I'm correct there.

server/server.go
lazartravica: @Stefan-Ethernal It is safe. The https://github.com/armon/go-metrics package defaults to a black hole sink, I'd say it's safe because of that. Checked locally

lazartravica (approved):
LGTM

server/server.go
lazartravica: https://github.com/hashicorp/go-metrics/blob/v0.4.1/start.go#L46-L49

Stefan-Ethernal: Awesome 👍

Stefan-Ethernal (approved):

PR #997, opened 2022-12-05
Modify snapshot calculation for dynamic epoch

goran-ethernal:
# Description

Changed the validator snapshot storage back to storing it by epoch number, but added necessary changes to accommodate dynamic epoch size, mainly how we figure out which was the epoch ending block.

Added a helper function (`isEpochEndingBlock`) that checks if any arbitrary block is an epoch ending block, so that it can be used by different features (such as storing a validator snapshot, or  It is located in `runtime_helpers.go`. It checks two things to know if a block is an epoch ending block:
1. If the `validatorSetDelta` is not empty in the header `extra`, it means that epoch change happened on that block.
2. If there is a newer block than the block for which we are checking the epoch ending exists in blockchain, it checks if it has a different epoch number (in header `extra`). If it has, that block we are checking is the epoch ending block.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -290,29 +290,28 @@ func initMainDBBuckets(db *bolt.DB) error {
 	return err
 }
 
+const snapshotKeySeparator = "edge-snapshot-key-separator"
+
 // insertValidatorSnapshot inserts a validator snapshot for the given block to its bucket in db
-func (s *State) insertValidatorSnapshot(block uint64, validatorSnapshot AccountSet) error {
+func (s *State) insertValidatorSnapshot(validatorSnapshot *validatorSnapshot) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
-		raw, err := validatorSnapshot.Marshal()
+		raw, err := json.Marshal(validatorSnapshot)
 		if err != nil {
 			return err
 		}
 
-		bucket := tx.Bucket(validatorSnapshotsBucket)
-
-		return bucket.Put(itob(block), raw)
+		return tx.Bucket(validatorSnapshotsBucket).Put(itob(validatorSnapshot.Epoch), raw)
 	})
 }
 
 // getValidatorSnapshot queries the validator snapshot for given block from db
-func (s *State) getValidatorSnapshot(block uint64) (AccountSet, error) {
-	var validatorSnapshot AccountSet
+func (s *State) getValidatorSnapshot(epoch uint64) (*validatorSnapshot, error) {
+	var validatorSnapshot *validatorSnapshot
 
 	err := s.db.View(func(tx *bolt.Tx) error {
-		bucket := tx.Bucket(validatorSnapshotsBucket)
-		v := bucket.Get(itob(block))
+		v := tx.Bucket(validatorSnapshotsBucket).Get(itob(epoch))
 		if v != nil {
-			return validatorSnapshot.Unmarshal(v)
+			return json.Unmarshal(v, &validatorSnapshot)
 		}
 
 		return nil
@@ -321,7 +320,27 @@ func (s *State) getValidatorSnapshot(block uint64) (AccountSet, error) {
 	return validatorSnapshot, err
 }
 
-// list iterates through all events in events bucket in db, unmarshals them, and returns as array
+// getLastSnapshot returns the last snapshot saved in db
+// since they are stored by epoch number (uint64), they are sequentially stored,
+// so the latest epoch will be the last snapshot in db
+func (s *State) getLastSnapshot() (*validatorSnapshot, error) {
+	var snapshot *validatorSnapshot
+
+	err := s.db.View(func(tx *bolt.Tx) error {
+		c := tx.Bucket(validatorSnapshotsBucket).Cursor()
+		k, v := c.Last()
+		if k == nil {
+			// we have no snapshots in db
+			return nil
+		}
+
+		return json.Unmarshal(v, &snapshot)
+	})
+
+	return snapshot, err
+}
+
+// list iterates through all events in events bucket in db, un-marshals them, and returns as array
 func (s *State) list() ([]*types.StateSyncEvent, error) {
 	events := []*types.StateSyncEvent{}
 
@@ -705,7 +724,7 @@ func (s *State) cleanEpochsFromDB() error {
 
 // cleanValidatorSnapshotsFromDB cleans the validator snapshots bucket if a limit is reached,
 // but it leaves the latest (n) number of snapshots
-func (s *State) cleanValidatorSnapshotsFromDB(block uint64) error {
+func (s *State) cleanValidatorSnapshotsFromDB(epoch uint64) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket(validatorSnapshotsBucket)
 
@@ -713,14 +732,14 @@ func (s *State) cleanValidatorSnapshotsFromDB(block uint64) error {
 		keys := make([][]byte, 0)
 		values := make([][]byte, 0)
 		for i := 0; i < numberOfSnapshotsToLeaveInDB; i++ { // exclude the last inserted we already appended
-			key := itob(block)
+			key := itob(epoch)
 			value := bucket.Get(key)
 			if value == nil {
 				continue
 			}
 			keys = append(keys, key)
 			values = append(values, value)
-			block--
+			epoch--
 		}
 
 		// removing an entire bucket is much faster than removing all keys
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -290,8 +290,6 @@ func initMainDBBuckets(db *bolt.DB) error {
 	return err
 }
 
-const snapshotKeySeparator = "edge-snapshot-key-separator"
-
 // insertValidatorSnapshot inserts a validator snapshot for the given block to its bucket in db
 func (s *State) insertValidatorSnapshot(validatorSnapshot *validatorSnapshot) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
--- a/consensus/polybft/validators_snapshot.go
+++ b/consensus/polybft/validators_snapshot.go
@@ -1,16 +1,34 @@
 package polybft
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
+	"github.com/0xPolygon/polygon-edge/blockchain"
 	"github.com/0xPolygon/polygon-edge/types"
 	"github.com/hashicorp/go-hclog"
 )
 
+type validatorSnapshot struct {
+	Epoch            uint64     `json:"epoch"`
+	EpochEndingBlock uint64     `json:"epochEndingBlock"`
+	Snapshot         AccountSet `json:"snapshot"`
+}
+
+func (vs *validatorSnapshot) copy() *validatorSnapshot {
+	copiedAccountSet := vs.Snapshot.Copy()
+
+	return &validatorSnapshot{
+		Epoch:            vs.Epoch,
+		EpochEndingBlock: vs.EpochEndingBlock,
+		Snapshot:         copiedAccountSet,
+	}
+}
+
 // TODO: Should we switch validators snapshot to Edge implementation? (validators/store/snapshot/snapshot.go)
 type validatorsSnapshotCache struct {
-	snapshots  map[uint64]AccountSet
+	snapshots  map[uint64]*validatorSnapshot
 	state      *State
 	blockchain blockchainBackend
 	lock       sync.Mutex
@@ -22,7 +40,7 @@ func newValidatorsSnapshotCache(
 	logger hclog.Logger, state *State, blockchain blockchainBackend,
 ) *validatorsSnapshotCache {
 	return &validatorsSnapshotCache{
-		snapshots:  map[uint64]AccountSet{},
+		snapshots:  map[uint64]*validatorSnapshot{},
 		state:      state,
 		blockchain: blockchain,
 		logger:     logger.Named("validators_snapshot"),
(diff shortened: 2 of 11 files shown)


consensus/polybft/state.go
```diff
@@ -294,29 +294,28 @@ func initMainDBBuckets(db *bolt.DB) error {
 	return err
 }
 
+const snapshotKeySeparator = "edge-snapshot-key-separator"
```
Stefan-Ethernal: This isn't used as far as I can see?

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+
+		epoch = extra.Checkpoint.EpochNumber
+	}
+
+	return blockNumber - 1, nil
+}
+
+// isEpochEndingBlock checks if given block is an epoch ending block
+func (v *validatorsSnapshotCache) isEpochEndingBlock(blockNumber uint64, extra *Extra) (bool, error) {
+	if !extra.Validators.IsEmpty() {
```
Stefan-Ethernal: As we have discussed earlier: you cannot rely that delta is going to be non-empty for epoch ending blocks (e.g. PoA systems won't have changes in validator sets AFAIK). Only relevant data is `EpochNumber` from the header.

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+	_, nextBlockExtra, err := getBlockData(blockNumber+1, v.blockchain)
+	if err != nil {
+		if errors.Is(err, blockchain.ErrNoBlock) {
+			return true, nil
+		}
+
+		return false, nil
+	}
+
+	return extra.Checkpoint.EpochNumber != nextBlockExtra.Checkpoint.EpochNumber, nil
```
Stefan-Ethernal: Just thinking out loud, does enforcing that next block epoch must be strictly greater than the current one (maybe we could even insist that it is greater by one comparing to current block)?

```suggestion
       	return extra.Checkpoint.EpochNumber < nextBlockExtra.Checkpoint.EpochNumber, nil
```

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+func (v *validatorsSnapshotCache) isEpochEndingBlock(blockNumber uint64, extra *Extra) (bool, error) {
+	if !extra.Validators.IsEmpty() {
+		return true, nil
+	}
+
+	_, nextBlockExtra, err := getBlockData(blockNumber+1, v.blockchain)
+	if err != nil {
+		if errors.Is(err, blockchain.ErrNoBlock) {
+			return true, nil
+		}
```
Stefan-Ethernal: This doesn't seem quite correct: how can you be sure that if there isn't next block in the blockchain, given block is an epoch-ending block? :thinking:

consensus/polybft/validators_snapshot.go
Stefan-Ethernal: Please double check me if all the use cases are covered, but I would expect something like this for epoch-ending blocks check:

```suggestion
	_, nextExtra, err := getBlockData(blockNumber+1, v.blockchain)
	if err != nil {
		if errors.Is(err, blockchain.ErrNoBlock) {
			return false, nil
		}

		return false, err
	}

	return nextExtra.Checkpoint.EpochNumber > extra.Checkpoint.EpochNumber, nil
```

Stefan-Ethernal (commented):
First batch of feedback

consensus/polybft/validators_snapshot.go
goran-ethernal: We do not solely rely on it. It's just one of the things that indicate an epoch end. That is why on the bottom of the function you have epoch number check.

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+
+		return false, nil
```
Stefan-Ethernal: You aren't propagating error anywhere. Is that deliberate? If yes, we don't need to return error at all from this function.

consensus/polybft/validators_snapshot.go
goran-ethernal: It doesn't matter really. We are checking it on the `currentBlock + 1`, so both conditions are correct to use (`<` or `!=`).

consensus/polybft/validators_snapshot.go
Stefan-Ethernal: Ok I see, doesn't harm here.

consensus/polybft/validators_snapshot.go
goran-ethernal: We can't be sure. This can only happen if you are trying to get the validator snapshot for the latest block in chain. You can't be certain that it is either epoch ending block or not. That is why I assume it is, because snapshot calculation algorithm will either take the snapshot from the last epoch if this block does not have a validator delta in it, or it will take the validator delta from the given block and apply it to the last snapshot.

consensus/polybft/validators_snapshot.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/6239de4e5f05874320d4906782c6294f293fa677

consensus/polybft/validators_snapshot.go
goran-ethernal: Just needed to return an error there. Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/6239de4e5f05874320d4906782c6294f293fa677

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/6239de4e5f05874320d4906782c6294f293fa677

consensus/polybft/validators_snapshot.go
Stefan-Ethernal: I know, I was just thinking about setting more strict condition (either next block strictly greater than or even next block epoch number incremented by one comparing to the current one).

Bottomline it really doesn't matter here, but we should definitely add validation of epochs in the fsm and header verifications in polybft: next epoch should be strictly incremented by one comparing to the previous one. I'm going to create a task for it.

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+		if err != nil {
+			if errors.Is(err, blockchain.ErrNoBlock) {
+				return blockNumber - 1, nil
+			}
+
+			return 0, err
+		}
+
+		epoch = extra.Checkpoint.EpochNumber
+	}
```
Stefan-Ethernal, line 320: What do you think about altering this for loop so it iterates through blocks and checks is it epoch ending block using this function `isEpochEndingBlock` instead?
Asking because this part looks similar to the logic implemented in the `isEpochEndingBlock` function.

```go
_, extra, err = getBlockData(blockNumber, v.blockchain)
if err != nil {
	if errors.Is(err, blockchain.ErrNoBlock) {
		return blockNumber - 1, nil
	}

	return 0, err
}
```

consensus/polybft/validators_snapshot.go
```diff
@@ -36,127 +54,106 @@ func (v *validatorsSnapshotCache) GetSnapshot(blockNumber uint64, parents []*typ
...
+		return latestValidatorSnapshot.Snapshot, nil
 	}
 
-	if snapshot == nil {
-		// Haven't managed to retrieve snapshot for any block from the cache.
+	if latestValidatorSnapshot == nil {
+		// Haven't managed to retrieve snapshot for any epoch from the cache.
 		// Build snapshot from the scratch, by applying delta from the genesis block.
-		initialSnapshot, err := v.computeSnapshot(nil, 0, parents)
+		// log.Trace("Building validators snapshot from scratch...")
```
Stefan-Ethernal: Remove this log or change it a little bit (like "Building validators snapshot from genesis block...") and uncomment it?

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+// If it doesn't have snapshot cached for desired epoch, it will return the latest one it has
+func (v *validatorsSnapshotCache) getLastCachedSnapshot(currentEpoch uint64) (*validatorSnapshot, error) {
+	snapshot := v.snapshots[currentEpoch]
+	if snapshot != nil {
+		v.logger.Trace("Found snapshot in memory cache", "Epoch", currentEpoch)
+
+		return snapshot, nil
+	}
+
+	snapshot, err := v.state.getLastSnapshot()
```
Stefan-Ethernal, line 261: Why you couldn't use `state.getValidatorSnapshot` here? You have the key, which is `currentEpoch` and have direct access to the required snapshot. If you can rely on `getValidatorSnapshot`, then you could easily remove `getLastSnapshot` function. Am I missing out something?

consensus/polybft/validators_snapshot.go
```diff
@@ -36,127 +54,106 @@ func (v *validatorsSnapshotCache) GetSnapshot(blockNumber uint64, parents []*typ
...
+		intermediateSnapshot, err := v.computeSnapshot(latestValidatorSnapshot, nextEpochEndBlockNumber, parents)
 		if err != nil {
-			return nil, fmt.Errorf("failed to store validators snapshot for block %d: %w", lastBlockWithSnapshot, err)
+			return nil, fmt.Errorf("failed to compute snapshot for epoch %d: %w", latestValidatorSnapshot.Epoch+1, err)
 		}
+
+		latestValidatorSnapshot = intermediateSnapshot
+		if err = v.storeSnapshot(latestValidatorSnapshot); err != nil {
+			return nil, fmt.Errorf("failed to store validators snapshot for epoch %d: %w", latestValidatorSnapshot.Epoch, err)
+		}
```
Stefan-Ethernal: Couldn't you done this outside of the for loop? Not sure why you need to store all those intermediary snapshots persisted?

If you remember, previous logic was relying on `deltasCount` in order to determine is snapshot altered in the for loop (and only in that case we were saving it to the cache and DB).

goran-ethernal: If we already calculated snapshot for a certain epoch (even if it was an intermediate snapshot), why not store it? Maybe a use case will happen where we would need a snapshot for an epoch which was intermediate, and we would need to recalculate it.

Stefan-Ethernal: I'd say that the odds are pretty low that we are going to need some of the past snapshots (most of the time we need the latest one or the one before it). Also when saving the most recent one only, it is good enough, because it would represent a starting point if there should be built some more recent snapshot.

I agree that persistence of snapshot is "cheaper" than calculation of snapshots, but to me it is really a question whether we are ever need those intermediary snapshots.

consensus/polybft/validators_snapshot.go
goran-ethernal: We could reuse it here, but it will mean that we would retrieve some blocks 2 times.

For example, you are on block 10 (which is the epoch ending block of epoch 1). You are searching for the epoch ending block for epoch 2, so you start from block 11. You get the header and extra for block 11, and call the `isEpochEndingBlock`. `isEpochEndingBlock` will get the block 12 to check if epoch on block 11 and 12 are different. If they are not, you return to the main loop, and get the extra and header for block 12, and call `isEpochEndingBlock` with its data, which will then get the data for block 13, to check if epochs are different, etc. That means some blocks are retrieved twice.

Yes, I can return the extra and header of the next block retrieved in `isEpochEndingBlock`, but that will that function to complex, and the plan is to reuse it (make it a helper function).

consensus/polybft/validators_snapshot.go
goran-ethernal: I agree. That is something we can discuss further. To me it's better safe then sorry. Save it, let it stay in cache and db, it's not a huge amount of data we are saving. Plus, we have that cleaning mechanism, so it won't stay in cache and db for too long.

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +221,96 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+		return snapshot, nil
+	}
+
+	snapshot, err := v.state.getLastSnapshot()
+	if err != nil {
+		return nil, err
+	}
+
+	return snapshot, nil
+}
```
Stefan-Ethernal: WDYT about this change? Basically idea is to return the more recent snapshot (either it is fetched from the cache or db). Although it makes code slightly complex, I think it is worth considering it.

```suggestion
func (v *validatorsSnapshotCache) getLastCachedSnapshot(currentEpoch uint64) (*validatorSnapshot, error) {
	var (
		cachedSnapshot *validatorSnapshot
	)

	for epoch := currentEpoch; epoch >= 0; epoch-- {
		cachedSnapshot = v.snapshots[epoch]
		if cachedSnapshot != nil {
			v.logger.Trace("Found snapshot in memory cache", "Epoch", currentEpoch)

			break
		}
	}

	dbSnapshot, err := v.state.getLastSnapshot()
	if err != nil {
		return nil, err
	}

	if dbSnapshot != nil && cachedSnapshot != nil &&
		dbSnapshot.Epoch > cachedSnapshot.Epoch {
		return dbSnapshot, nil
	}

	return cachedSnapshot, nil
}
```

consensus/polybft/validators_snapshot.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/9045a89e0bbf81028a0dbe958b390f8b7075b11f

consensus/polybft/validators_snapshot.go
```diff
@@ -179,9 +176,15 @@ func (v *validatorsSnapshotCache) computeSnapshot(
 		return nil, fmt.Errorf("failed to decode extra from the block#%d: %w", header.Number, err)
 	}
 
-	snapshot := existingSnapshot
+	var snapshot AccountSet
+
+	var snapshotEpoch uint64
```
Stefan-Ethernal: Nitpick (spare single line :stuck_out_tongue: )

```suggestion
	var  (
	    snapshot AccountSet
            snapshotEpoch uint64
        )
```

consensus/polybft/validators_snapshot.go
goran-ethernal: Modified this to return false, and handle the error outside the `isEpochEndingBlock` function. Also, made that function a helper function, so that we can reuse it on other places, because it is quite handy and useful. Plus added UTs for it:
https://github.com/0xPolygon/polygon-edge/pull/997/commits/9045a89e0bbf81028a0dbe958b390f8b7075b11f

Stefan-Ethernal (approved):
Aside from the suggestions I've made, no blockers and LGTM :fire: :rocket:

consensus/polybft/validators_snapshot.go
goran-ethernal: It makes sense.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/a4709904d8b5cfc71dab457a4bf18bbb1338367b

consensus/polybft/validators_snapshot.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/a4709904d8b5cfc71dab457a4bf18bbb1338367b

b00ris (approved):
Good job!

consensus/polybft/validators_snapshot.go
```diff
@@ -214,29 +224,98 @@ func (v *validatorsSnapshotCache) storeSnapshot(block uint64, snapshot AccountSe
...
+	dbSnapshot, err := v.state.getLastSnapshot()
+	if err != nil {
+		return nil, err
+	}
+
+	if dbSnapshot != nil {
+		// if we do not have any snapshot in memory, or db snapshot is newer than the one in memory
+		// return the one from db
+		if cachedSnapshot == nil || dbSnapshot.Epoch > cachedSnapshot.Epoch {
+			cachedSnapshot = dbSnapshot
```
vcastellm, line 285: Don't we need to store the snapshot in the cache at this point?

goran-ethernal: Good point. Fix: https://github.com/0xPolygon/polygon-edge/pull/997/commits/43c3ecea6729a138f43da4378fab505569002594

igorcrevar (approved):

PR #1004, opened 2022-12-06
Implement abstraction of transactions by adding tx interface

begmaroman:
# Description

Geth introduced concept of envelope+payload for transactions, before EIP-1559 so that interface of transaction remains the same for the "outer" world and created common interface for both legacy and dynamic fee (EIP-1559) transactions:

- https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go (this is envelope struct I was referring to and it embeds either legacy tx, access list tx or dynamic fee/EIP-1559 tx)

These implementations are embedded as the payload of the transaction object itself:

- EIP-1559 transactions https://github.com/ethereum/go-ethereum/blob/master/core/types/tx_dynamic_fee.go
- legacy transactions(gas price only transactions, pre-EIP-1559 hardfork) https://github.com/ethereum/go-ethereum/blob/master/core/types/tx_legacy.go

Introduce `Transaction` interface and add two implementations: state tx and legacy tx 

# Changes include

# Checklist

## Testing


types/transaction.go
```diff
@@ -12,73 +13,102 @@ import (
...
+	time  time.Time // Time first seen locally (spam avoidance)
+
+	// Caches
+	hash Hash
 	size atomic.Value
+	from Address
+}
+
+// NewTx creates a new transaction.
+func NewTx(inner TxData) *Transaction {
```
JekaMas, line 83: when we start using go 1.19 it'd be nice to use generics here and in NewTxWithSender. Maybe for Transaction also and have something like

```
type Transaction[T TxData] struct {
...
    inner T
...
}

// maybe even the interface
type txData interface {
        txType() TxType // returns the type ID
    	input() []byte
	gas() uint64
	setGas(gas uint64)
	gasPrice() *big.Int
	gasTipCap() *big.Int
	gasFeeCap() *big.Int
	value() *big.Int
	nonce() uint64
	to() *Address

	rawSignatureValues() (v, r, s *big.Int)
	setSignatureValues(v, r, s *big.Int)
}

type TxData[T txData] interface {
	Copy() T   // creates a deep copy and initializes all fields
        txData
}
```

Please, make a task for it.

begmaroman: Very nice point! Done: EVM-301

JekaMas (approved):

types/transaction.go
```diff
@@ -150,9 +161,94 @@ func (t *Transaction) Size() uint64 {
...
+// SetSignatureValues sets the given signature values
+func (t *Transaction) SetSignatureValues(v, r, s *big.Int) *Transaction {
+	t.inner.setSignatureValues(v, r, s)
+	t.ComputeHash()
+
+	return t
+}
+
+// setDecoded sets the inner transaction and size after decoding.
+func (t *Transaction) setDecoded(inner TxData, size uint64) {
```
Stefan-Ethernal: What's the purpose of `size` param here? As I can see we are always providing 0 for it.

types/rlp_unmarshal.go
```diff
@@ -307,79 +307,156 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
+
+		// to
+		if vv, _ := v.Get(3).Bytes(); len(vv) == 20 {
+			// address
+			addr := BytesToAddress(vv)
+			lTx.To = &addr
+		} else {
+			// reset To
+			lTx.To = nil
+		}
```
Stefan-Ethernal, line 356: Nitpick: does it make sense to bring `to` field unmarshal above `value` and `input` fields, because it is on index 3, comparing `value` and `input` which are on indices 4 and 5 respective?

types/transaction.go
```diff
@@ -87,48 +117,29 @@ func (t *Transaction) ComputeHash() *Transaction {
...
-		tt.GasPrice.Set(t.GasPrice)
-	}
-
-	tt.Value = new(big.Int)
-	if t.Value != nil {
-		tt.Value.Set(t.Value)
+	return &Transaction{
+		inner: t.inner.Copy(),
+		time:  t.time,
+		hash:  t.hash,
```
Stefan-Ethernal, line 133: Create copy of `hash` field as well?

types/rlp_encoding_test.go
```diff
@@ -59,8 +59,10 @@ func TestRLPMarshall_And_Unmarshall_Transaction(t *testing.T) {
 
 	unmarshalledTxn.ComputeHash()
 
-	txn.Hash = unmarshalledTxn.Hash
-	if !reflect.DeepEqual(txn, unmarshalledTxn) {
+	// txn.Hash = unmarshalledTxn.Hash
```
Stefan-Ethernal: Think that you could have left this line in this particular case. However if `ComputeHash` is going to be kept instead, then just remove this commented line.

txpool/event_subscription_test.go
```diff
@@ -141,13 +143,14 @@ func TestEventSubscription_ProcessedEvents(t *testing.T) {
...
 			defer eventWaitFn()
-			if _, err := tests.RetryUntilTimeout(eventWaitCtx, func() (interface{}, bool) {
+
+			_, err := tests.RetryUntilTimeout(eventWaitCtx, func() (interface{}, bool) {
 				return nil, atomic.LoadInt64(&processed) < int64(testCase.expectedProcessed)
-			}); err != nil {
-				t.Fatalf("Unable to wait for events to be processed, %v", err)
-			}
+			})
+			require.NoError(t, err)
```
Stefan-Ethernal: Would you like to keep this error message as third parameter of `require.NoError`?

`Unable to wait for events to be processed`

types/transaction.go
```diff
@@ -12,73 +13,102 @@ import (
...
 
 func (t *Transaction) IsLegacyTx() bool {
-	return t.Type == LegacyTx
+	return t.inner.txType() == LegacyTxType
 }
 
 func (t *Transaction) IsStateTx() bool {
-	return t.Type == StateTx
+	return t.inner.txType() == StateTxType
 }
```
Stefan-Ethernal: Just thinking out loud and I'm not biased towards any approach:
maybe remove these functions and rely on `Type()` function instead whenever needed? Otherwise, soon we are about to add `DynamicFeeTxType` and possibly add `IsDynamicFeeTx` helper function...

command/e2e/register_validator.go
```diff
@@ -294,22 +294,22 @@ type txnSender struct {
 	account *wallet.Account
 }
 
-func (t *txnSender) sendTransaction(txn *types.Transaction) asyncTxn {
-	if txn.GasPrice == nil {
-		txn.GasPrice = big.NewInt(defaultGasPrice)
+func (t *txnSender) sendTransaction(txnData *types.LegacyTx) asyncTxn {
```
Stefan-Ethernal, line 325: Didn't want to have `*types.Transaction` here as a parameter?

Stefan-Ethernal (commented):
Generally looks good, although leaving a couple of comments to consider.

types/transaction.go
begmaroman: We copy all data here so the hash would remain the same

command/e2e/register_validator.go
begmaroman: Not really because it is used within this package only and adds some extra fields population within the function itself. It would not be possible if we pass the common transaction model.

txpool/event_subscription_test.go
begmaroman: Done

types/rlp_encoding_test.go
begmaroman: ok, done

types/rlp_unmarshal.go
begmaroman: Done

types/transaction.go
begmaroman: Makes sense, removed

types/transaction.go
begmaroman: > Otherwise, soon we are about to add DynamicFeeTxType and possibly add IsDynamicFeeTx helper function...

That's right. Why not?

Stefan-Ethernal: I see it as a redundant in a way, since we can always query `TxType`, but however you want.

types/transaction.go
Stefan-Ethernal: I was referring to mutation of `hash` field on the `Transaction` instance itself and in that regard copying that byte slice, but that's probably not a valid use case...

Stefan-Ethernal (approved):

types/transaction.go
begmaroman: Well, this is what it is right now. I would change it within one of the next PRs anyways.

types/transaction.go
begmaroman: It is an array so this statement is not applicable. It copies an original array.

Stefan-Ethernal: I was having this example on mind: https://go.dev/play/p/JP6kGT2XVwx. Am I missing out something?

begmaroman: Yes, there is a slice in your example, but the hash is an array, they behave differently.

Stefan-Ethernal: Ok I see now: https://go.dev/play/p/UlXTOwSOih0 🙂

types/rlp_unmarshal.go
```diff
@@ -307,79 +307,156 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
 	}
 
 	if len(elems) < 9 {
-		return fmt.Errorf("incorrect number of elements to decode transaction, expected 9 but found %d", len(elems))
+		return fmt.Errorf("incorrect number of elements to decode transaction, expected at leaset 9 but found %d", len(elems))
```
```suggestion
		return fmt.Errorf("incorrect number of elements to decode transaction, expected at least 9 but found %d", len(elems))
```

PR #1005, opened 2022-12-06
Decouple rootchain contracts deployment from development environment

Stefan-Ethernal:
# Description

This PR introduces new command called `polybft-manifest`, which creates a `manifest.json` file. Manifest file is input for `rootchain init-contracts` and `genesis` commands. It contains public validators information (namely: ECDSA address, public BLS key, P2P node id and balance info) as well as addresses of rootchain smart contracts.

Several rootchain commands are now expanded so that accept manifest file: `rootchain emit` and `rootchain init-contracts` as well as `genesis` command itself.

Commands workflow along with artefacts it makes is illustrated by the image below (dashed rectangles denotes commands intended for testing purposes):

CLI workflow (1)

# Changes include

# Breaking changes

Genesis configuration (genesis.json) layout has changed in a way that `bridge` configuration contains addresses of all rootchain contracts and also has address of admin account for the rootchain.

# Checklist

## Testing

### Manual tests

**Running polybft without bridge**
1. Run `scripts/cluster` script
2. Make sure blocks are being produced

**Running polybft with bridge**
Run these commands in this exact order:
```go
// 1. init accounts
go run . polybft-secrets --data-dir test-chain- --num 4

// 2. start rootchain server
go run . rootchain server

// 3. generate manifest file
go run . polybft-manifest

// 4. deploy and init rootchain contracts
go run . rootchain init-contracts

// 5. create genesis file
go run . genesis --block-gas-limit 10000000 --validator-set-size=4 --consensus polybft --epoch-size 5 --bridge-json-rpc

// 6. fund validators on rootchain
go run . rootchain fund --data-dir test-chain- --num 4

// 7. run sidechain cluster (which consists of 4 Edge clients)
go run . server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :5001 --libp2p :30301 --jsonrpc :9545 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :5002 --libp2p :30302 --jsonrpc :10002 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :5003 --libp2p :30303 --jsonrpc :10003 --seal --log-level DEBUG
go run . server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :5004 --libp2p :30304 --jsonrpc :10004 --seal --log-level DEBUG
```


consensus/polybft/polybft_config.go
```diff
@@ -3,60 +3,91 @@ package polybft
...
+	BlsKey  string
+	Balance *big.Int
+	NodeID  string
 }
 
 type validatorRaw struct {
 	Address types.Address `json:"address"`
 	BlsKey  string        `json:"blsKey"`
 	Balance *string       `json:"balance"`
+	NodeID  string        `json:"nodeId"`
```
Stefan-Ethernal, line 84: Important point:
the same `Validator` struct is being used for both `manifest` and `genesis` files generation, therefore both of these would have p2p `NodeID` persisted (although it is not needed by the `genesis` per se, it is required in manifest, because that info is being used to populate bootnodes if those aren't provided via CLI).

command/genesis/params.go
```diff
@@ -75,15 +75,12 @@ type genesisParams struct {
...
-	blockTime        time.Duration
-	validators       []string
-
-	polyBftValidatorPrefixPath string
-	premineValidators          string
-	smartContractsRootPath     string
-	bridgeJSONRPCAddr          string
+	manifestPath           string
+	smartContractsRootPath string
+	validatorSetSize       int
```
goran-ethernal, line 80: Maybe we should rename this to `activeValidatorSetSize` to reflect more its true purpose (it's flag should be renamed as well).

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,193 @@
...
+var (
+	params = &manifestInitParams{}
+)
+
+// GetCommand returns the rootchain emit command
+func GetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "polybft-manifest",
+		Short: "Initializes manifest file",
+		// PreRunE: runPreRun,
```
goran-ethernal: Remove this since we don't have a pre-run.

Stefan-Ethernal: Now we have it, good catch! :stuck_out_tongue: 
https://github.com/0xPolygon/polygon-edge/pull/1005/commits/8a965c3c489384f6e32ea9ace2338dd3b7ceb997

command/genesis/params.go
Stefan-Ethernal: I feel that this part is going to be changed with on-chain governance RFC, as we need to agree what is going to be parameterized and what is it semantic on SC side, so this is going to be revisited anyways. Therefore I'd prefer keep it this way for now.

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,211 @@
...
+	nodeIDLength       = 53
+	ecdsaAddressLength = 42
+	blsKeyLength       = 2
+)
+
+var (
+	params = &manifestInitParams{}
+)
+
+// GetCommand returns the rootchain emit command
```
vcastellm:
```suggestion
// GetCommand returns the polybft-manifest command
```

vcastellm: even just `the command`

Stefan-Ethernal: Good catch :eagle: :+1:

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,211 @@
...
+func GetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "polybft-manifest",
```
vcastellm: I would call it just `manifest` and specify in the docs, `--help`, etc, that it only applies to `polybft`, it will anyway in the future

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,211 @@
...
+		"./",
+		"root path containing polybft validators private keys",
+	)
+
+	cmd.Flags().StringVar(
+		&params.validatorsPrefixPath,
+		validatorsPrefixFlag,
+		defaultValidatorPrefixPath,
+		"folder prefix names for polybft validator private keys",
+	)
```
vcastellm: If this need to read the private keys, it should be compatible with secrets manager, should be capable of reading secrets from the `secretsManagerConfig.json`

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,211 @@
...
+func runPreRun(_ *cobra.Command, _ []string) error {
+	return params.validateFlags()
+}
+
+func setFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(
+		&params.manifestPath,
+		manifestPathFlag,
+		defaultManifestPath,
+		"the file path where manifest file is going to be stored",
```
vcastellm, line 60: file path or file path + name?

command/polybftmanifest/manifest_init.go
```diff
@@ -0,0 +1,211 @@
...
+		validatorsPrefixFlag,
+		defaultValidatorPrefixPath,
+		"folder prefix names for polybft validator private keys",
+	)
+
+	cmd.Flags().StringArrayVar(
+		&params.validators,
+		validatorsFlag,
+		[]string{},
+		"validators defined by user throughout a parameter (format: <node id>:<ECDSA address>:<public BLS key>)",
```
vcastellm:
```suggestion
		"validators defined by user (format: <node id>:<ECDSA address>:<public BLS key>)",
```

consensus/polybft/polybft.go
```diff
@@ -134,8 +126,17 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
...
+		}
+
+		rootchainAdmin := types.ZeroAddress
+		if polyBFTConfig.IsBridgeEnabled() {
+			rootchainAdmin = polyBFTConfig.Bridge.AdminAddress
+		}
+
 		input, err = nativeTokenInitializer.Encode(
-			[]interface{}{helper.GetRootchainAdminAddr(), nativeTokenName, nativeTokenSymbol})
+			[]interface{}{rootchainAdmin, nativeTokenName, nativeTokenSymbol})
```
vcastellm, line 139: Can be inited with `ZeroAddress`?

server/builtin.go
```diff
@@ -22,7 +22,7 @@ type ConsensusType string
 const (
 	DevConsensus     ConsensusType = "dev"
 	IBFTConsensus    ConsensusType = "ibft"
-	PolyBFTConsensus ConsensusType = "polybft"
+	PolyBFTConsensus ConsensusType = consensusPolyBFT.PolyBFTConsensusName
```
vcastellm: This doesn't seem consistent with the other constants in this group

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f432f4babd18c4081d8bd25c8ade9a73badfbc28

command/polybftmanifest/manifest_init.go
Stefan-Ethernal: Removed comment entirely, as don't find it necessary https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f432f4babd18c4081d8bd25c8ade9a73badfbc28

command/polybftmanifest/manifest_init.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f432f4babd18c4081d8bd25c8ade9a73badfbc28

command/polybftmanifest/manifest_init.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f432f4babd18c4081d8bd25c8ade9a73badfbc28

consensus/polybft/polybft.go
Stefan-Ethernal: My understanding, based on `MRC20.sol` and `StateReceiver.sol` implementations, is that the `predicate` parameter is used only for the state syncs workflow https://github.com/0xPolygon/core-contracts/blob/main/contracts/child/MRC20.sol#L59. As it is not mandatory running an Edge client in the bridge mode, returning an error here wouldn't be desired behavior I'd say.

command/polybftmanifest/manifest_init.go
Stefan-Ethernal: Maybe comment is imprecise as we don't need private keys per se, but are deducing ECDSA address, BLS public key and p2p node id. So this works exactly the same as polybft genesis population did: either provide path and prefixes in order to read validator information from the local storage, or provide such information using `--validators` flag. This is not reading single validator private key, but all the genesis validators information. 
Can you give an example of code snippet in IBFT genesis command which consults secrets manager in order to extract such information, so that I get more context?

consensus/polybft/polybft.go
Stefan-Ethernal: Although it is questionable should we provide another rootchain private key for bridging, instead of relying on `rootchainAdmin` :thinking: We can decide that in the future, as it is not breaking change...

goran-ethernal (approved):
LGTM. I think you covered and addressed everything :100:

consensus/polybft/contracts_initializer.go
```diff
@@ -40,11 +40,11 @@ var (
 )
 
 func getInitChildValidatorSetInput(polyBFTConfig PolyBFTConfig) ([]byte, error) {
-	validatorAddresses := make([]types.Address, len(polyBFTConfig.InitialValidatorSet))
-	validatorPubkeys := make([][4]*big.Int, len(polyBFTConfig.InitialValidatorSet))
-	validatorStakes := make([]*big.Int, len(polyBFTConfig.InitialValidatorSet))
+	validatorAddresses := make([]types.Address, len(polyBFTConfig.Validators))
```
ferranbt: Was it necessary to change the name of this field?

consensus/polybft/consensus_runtime.go
```diff
@@ -957,7 +959,10 @@ func (c *consensusRuntime) getSystemState(header *types.Header) (SystemState, er
 		return nil, err
 	}
 
-	return c.config.blockchain.GetSystemState(c.config.PolyBFTConfig, provider), nil
+	return c.config.blockchain.GetSystemState(
+		contracts.ValidatorSetContract,
```
ferranbt: Though this is correct since these contracts are hardcoded, changing this is out of the scope of the decoupling of the rootchain contracts.

consensus/polybft/polybft_config.go
```diff
@@ -3,60 +3,89 @@ package polybft
...
+}
+
 // BridgeConfig is the rootchain bridge configuration
 type BridgeConfig struct {
-	BridgeAddr      types.Address `json:"stateSenderAddr"`
-	CheckpointAddr  types.Address `json:"checkpointAddr"`
-	JSONRPCEndpoint string        `json:"jsonRPCEndpoint"`
+	StateSenderAddress       types.Address `json:"stateSenderAddress"`
+	CheckpointManagerAddress types.Address `json:"checkpointManagerAddress"`
+	BLSAddress               types.Address `json:"blsAddress"`
```
ferranbt: Does it need to have the BLS addresses here?

consensus/polybft/polybft_config.go
```diff
@@ -3,60 +3,89 @@ package polybft
...
+	return polyBFTConfig, nil
+}
+
 // BridgeConfig is the rootchain bridge configuration
 type BridgeConfig struct {
-	BridgeAddr      types.Address `json:"stateSenderAddr"`
-	CheckpointAddr  types.Address `json:"checkpointAddr"`
-	JSONRPCEndpoint string        `json:"jsonRPCEndpoint"`
+	StateSenderAddress       types.Address `json:"stateSenderAddress"`
+	CheckpointManagerAddress types.Address `json:"checkpointManagerAddress"`
```
ferranbt: I was unaware that changing these `json` tags and the fields was part of the scope of the decoupling of the rootchain contracts. Any change on the `json` config should be carefully studied and discussed.

consensus/polybft/polybft_config.go
```diff
@@ -3,60 +3,89 @@ package polybft
...
 
-	ValidatorSetSize int `json:"validatorSetSize"`
+	// Bridge is the rootchain bridge configuration
+	Bridge *BridgeConfig `json:"bridge"`
 
-	// Address of the system contracts, as of now (testing) this is populated automatically during genesis
-	ValidatorSetAddr  types.Address `json:"validatorSetAddr"`
-	StateReceiverAddr types.Address `json:"stateReceiverAddr"`
+	// ActiveValidatorSetSize denotes how many validators are active per each epoch
+	ActiveValidatorSetSize int `json:"activeValidatorSetSize"`
```
ferranbt: Is this the new parametrizable field for the validator set size during init? If so, this is a different feature.

ferranbt (changes requested):
I think this PR ended up covering more ground than it should have. It is doing several things at the same time:
- Adding the new `manifest.json` file and using it to initialize the genesis (intended goal).
- Rewrites `polybft` to use hardcoded addresses directly. (not against it).
- Changes the name of `json` config tags. (not against it but it would need a reason).

I think only with the scope of the PR, much of the changes on the `polybft` package would be avoided and it would be simpler to review.

consensus/polybft/contracts_initializer.go
Stefan-Ethernal: No it wasn't, but I felt it is a good opportunity since I was changing genesis configuration layout anyways (adding rootchain contracts addresses to it, rootchain account address etc.). Do you agree?

consensus/polybft/polybft_config.go
Stefan-Ethernal: It isn't right now, although I'm not sure whether we are going to need it in the future. I'm going to remove it for now (as well as `BN256G2` address as it is not needed as well).

consensus/polybft/contracts_initializer.go
ferranbt: I agree with the change but not as part of this PR.

consensus/polybft/polybft_config.go
Stefan-Ethernal: > I was unaware that changing these json tags and the fields was part of the scope of the decoupling of the rootchain contracts.

As I've already mentioned: https://github.com/0xPolygon/polygon-edge/pull/1005#discussion_r1049497065

> Any change on the json config should be carefully studied and discussed.

This is just renaming for the sake of consistency, I don't see it as a major issue which needs deep discussion to be honest.

consensus/polybft/polybft_config.go
Stefan-Ethernal: It is not, although in my opinion it should be parameterized in the future (not sure if it is going to be part of genesis or on-chain governance though). I'm going to remove it for now and we are going to add it when it is needed.

consensus/polybft/polybft_config.go
ferranbt: As with the other changes, to me it is not about if this change is correct or not (I think it is correct). But we should keep one PR, one feature and the feature of this PR is the manifest workflow, not the `genesis.json` parameter consistency.

consensus/polybft/polybft_config.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f35ba13ffe9fa37fd11178fe49e55f2f7c6f73d4

consensus/polybft/polybft_config.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f35ba13ffe9fa37fd11178fe49e55f2f7c6f73d4

consensus/polybft/polybft_config.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f35ba13ffe9fa37fd11178fe49e55f2f7c6f73d4

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f35ba13ffe9fa37fd11178fe49e55f2f7c6f73d4

consensus/polybft/contracts_initializer.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f35ba13ffe9fa37fd11178fe49e55f2f7c6f73d4

consensus/polybft/polybft_config.go
```diff
@@ -3,60 +3,89 @@ package polybft
...
 }
 
 // BridgeConfig is the rootchain bridge configuration
 type BridgeConfig struct {
-	BridgeAddr      types.Address `json:"stateSenderAddr"`
-	CheckpointAddr  types.Address `json:"checkpointAddr"`
-	JSONRPCEndpoint string        `json:"jsonRPCEndpoint"`
+	BridgeAddr            types.Address `json:"stateSenderAddr"`
+	CheckpointManagerAddr types.Address `json:"checkpointAddr"`
+	AdminAddress          types.Address `json:"adminAddress"`
```
ferranbt: What is this admin address?

consensus/polybft/consensus_runtime_test.go
```diff
@@ -1621,7 +1616,7 @@ func TestConsensusRuntime_FSM_EndOfEpoch_OnBlockInserted(t *testing.T) {
 		epoch:             metadata,
 		config:            config,
 		lastBuiltBlock:    lastBuiltBlock,
-		checkpointManager: newCheckpointManager(wallet.NewEcdsaSigner(signer), 5, nil, nil, nil, hclog.NewNullLogger()),
+		checkpointManager: newCheckpointManager(wallet.NewEcdsaSigner(signer), 5, types.ZeroAddress, nil, nil, nil, hclog.NewNullLogger()),
```
ferranbt, line 1623: Except for this change, are the other changes required for this PR?

consensus/polybft/fsm_test.go
```diff
@@ -647,7 +647,7 @@ func TestFSM_VerifyStateTransactions_EndOfEpochWrongValidatorsUptimeTx(t *testin
 	uptimeCounter, err := createTestUptimeCounter(t, nil, 5).EncodeAbi()
 	require.NoError(t, err)
 
-	commitEpochTx := createStateTransactionWithData(contracts.ValidatorSetContract, uptimeCounter)
+	commitEpochTx := createStateTransactionWithData(fsm.config.ValidatorSetAddr, uptimeCounter)
```
ferranbt: Do we need these changes?

ferranbt (changes requested):

command/polybftmanifest/manifest_init.go
vcastellm: Makes sense, then the flag help message is not totally correct.

consensus/polybft/polybft.go
vcastellm: This add another point of difficulty to enabling bridge post genesis

consensus/polybft/polybft_config.go
Stefan-Ethernal: That is a (rootchain) account address which is needed as a predicate for bridging workflow in `MRC20.sol`
(https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/MRC20.sol#L59). 
For now we are using singular account for:
1. deployment (and initialization) of rootchain contracts
2. sending emit transactions to the rootchain
3. registering that address as a predicate for the MRC20. Based on the code, it is related only for the state syncs/bridge workflow.

I'm not sure should we eventually split these accounts into two separate ones: 1st should cover only 1. point and other would be needed for 2. and 3. points from above...?

consensus/polybft/fsm_test.go
Stefan-Ethernal: https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f84bf34abe858c448eaa998b6e389c301ba58708

consensus/polybft/consensus_runtime_test.go
Stefan-Ethernal: https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f84bf34abe858c448eaa998b6e389c301ba58708

command/polybftmanifest/manifest_init.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1005/commits/f84bf34abe858c448eaa998b6e389c301ba58708

vcastellm (approved):
LGTM

ferranbt (approved):

PR #1013, opened 2022-12-08
Remove code related to bundles for state syncs

vcastellm:
# Description

Relate to RFC-139, we're removing bundles concept from the codebase. 

* Asked for some changes to the SCs in order to bring consistency in naming.

# Changes include

# Breaking changes

We're not using bundles anymore, so we'll need to use `dev` branch from SCs

# Checklist

## Testing


types/state_transaction.go
```diff
@@ -0,0 +1,135 @@
...
+	// ID is the decoded 'index' field from the event
+	ID uint64
+	// Sender is the decoded 'sender' field from the event
+	Sender ethgo.Address
+	// Receiver is the decoded 'receiver' field from the event
+	Receiver ethgo.Address
+	// Data is the decoded 'data' field from the event
+	Data []byte
+	// Skip is the decoded 'skip' field from the event
+	Skip bool
```
ferranbt: We can remove this field, it is part of the old spec.

consensus/polybft/state_test.go
```diff
@@ -300,17 +300,17 @@ func TestState_cleanCommitments(t *testing.T) {
...
 	}
 
-	bundles, err := state.getBundles(0, maxBundlesPerSprint)
+	bundles, err := state.getStateSyncProof(0)
 	assert.NoError(t, err)
 	assert.Nil(t, bundles)
 }
 
-func TestState_insertAndGetBundles(t *testing.T) {
+func TestState_insertAndGetStateSyncProof(t *testing.T) {
```
ferranbt: Could we namespace the tests with something like `TestState_StateSync_XXX`? It would be easier to know what section they cover.

consensus/polybft/commitment.go
```diff
@@ -224,58 +119,41 @@ func (cm *CommitmentMessage) Hash() (types.Hash, error) {
...
+	return cm.ToIndex - cm.FromIndex
 }
 
-// VerifyProof validates given bundle proof against merkle trie root hash contained in the CommitmentMessage
-func (cm CommitmentMessage) VerifyProof(bundle *BundleProof) error {
-	if len(bundle.StateSyncs) == 0 {
-		return errors.New("no state sync events")
+// VerifyStateSyncProof validates given state sync proof
+// against merkle trie root hash contained in the CommitmentMessage
+func (cm CommitmentMessage) VerifyStateSyncProof(stateSyncProof *types.StateSyncProof) error {
```
goran-ethernal: Only used in tests. What do you think, should we remove it, since now we don't have verification on client?
Maybe it's good to leave it in tests, like a way if we mess something up, should show us that something is wrong with state sync proofs?

consensus/polybft/commitment.go
```diff
@@ -303,21 +180,21 @@ func (cm *CommitmentMessageSigned) EncodeAbi() ([]byte, error) {
 	}
 
 	data := map[string]interface{}{
-		"bundle":    commitment,
-		"signature": cm.AggSignature.AggregatedSignature,
-		"bitmap":    blsVerificationPart,
+		"commitment": commitment,
```
goran-ethernal: Is there a PR on `core-contracts` for refactoring names on `StateReceiver`? This is still called `bundle` on their side.

consensus/polybft/commitment.go
vcastellm: Agree

types/state_transaction.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/25d02cb3f3a0cfb83b0bd6fb62afc2984a452960

consensus/polybft/state_test.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/7e8df2f49cb95d9fce492cb61862eed4c7c4b748

consensus/polybft/commitment.go
vcastellm: Asked the SCs team to change it

types/state_transaction.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var ExecuteStateSyncABIMethod, _ = abi.NewMethod("function execute(" +
 	"bytes32[] proof, " +
-	"tuple(uint256 id, address sender, address receiver, bytes data, bool skip) stateSync)")
+	"tuple(uint256 id, address sender, address receiver, bytes data) stateSync)")
```
Stefan-Ethernal: Note: Seems like test is still going to fail, until skip field is not removed on `StateReceiver` SC .

goran-ethernal: I think they still need to change the names of some fields in `StateReceiver`.

consensus/polybft/commitment.go
```diff
@@ -4,32 +4,30 @@ import (
...
+	commitABIMethod, _ = abi.NewMethod("function commit(" +
+		"tuple(uint256 startId, uint256 endId, bytes32 root) commitment," +
 		"bytes signature," +
 		"bytes bitmap)")
 
-	executeBundleABIMethod, _ = abi.NewMethod("function execute(" +
+	executeStateSyncABIMethod, _ = abi.NewMethod("function execute(" +
 		"bytes32[] proof, " +
-		"tuple(uint256 id, address sender, address receiver, bytes data, bool skip)[] objs)")
+		"tuple(uint256 id, address sender, address receiver, bytes data) stateSync)")
```
Stefan-Ethernal: Are we going to remove this if it wasn't used anymore?

consensus/polybft/consensus_runtime.go
```diff
@@ -224,7 +224,9 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	}
 }
 
-func (c *consensusRuntime) createCommitmentAndBundles(txs []*types.Transaction) error {
+// getCommitmentFromTransaction gets the registered commitment (if any)
+// from a transaction, and builds bundles for it
+func (c *consensusRuntime) getCommitmentFromTransaction(txs []*types.Transaction) error {
```
Stefan-Ethernal: Rename this to `getCommitmentFromTransactions`?

Stefan-Ethernal (commented):
Generally looks good, apart from the minor suggestions.

consensus/polybft/commitment.go
goran-ethernal: Yes, it managed to hide from us while we were removing the unused code :sunglasses:

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1013/commits/1ff7061ac7b16bc09d087c87578b6a4292a157a1

consensus/polybft/commitment.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1013/commits/1ff7061ac7b16bc09d087c87578b6a4292a157a1

goran-ethernal (approved):
Removing 930 lines like a boss :sunglasses: . Apart from waiting for the SC team to remove that `skip` field and change the names of some existing ones, looks great.

Let's wait for them to finish those changes, so we can finally use the `dev` branch of `core-contracts` :muscle: .

consensus/polybft/state.go
```diff
@@ -602,29 +559,21 @@ func (s *State) insertBundles(bundles []*BundleProof) error {
 	})
 }
 
-// getBundles gets bundles that are not executed
-func (s *State) getBundles(stateSyncExecutionIndex, maxNumberOfBundles uint64) ([]*BundleProof, error) {
-	var bundles []*BundleProof
+// getStateSyncProof gets state sync proof that are not executed
+func (s *State) getStateSyncProof(stateSyncExecutionIndex uint64) (*types.StateSyncProof, error) {
```
goran-ethernal: Nitpick: Maybe rename this `stateSyncExecutionIndex` to `stateSyncID`?

e2e-polybft/bridge_test.go
```diff
@@ -96,10 +77,10 @@ func executeStateSync(t *testing.T, client *jsonrpc.Client, txRelayer txrelayer.
 
 	t.Log("State sync proofs:", stateSyncProof)
 
-	input, err := types.ExecuteBundleABIMethod.Encode([2]interface{}{stateSyncProof.Proof, stateSyncEventsToAbiSlice(stateSyncProof.StateSync)})
+	input, err := types.ExecuteStateSyncABIMethod.Encode([2]interface{}{stateSyncProof.Proof, stateSyncProof.StateSync.ToMap()})
```
goran-ethernal: You can use `stateSyncProof.EncodeAbi()` here.

goran-ethernal (approved):
Apart from small comments, as I said before, great job :100:

Nemanja0x (approved):

types/state_transaction.go
```diff
@@ -0,0 +1,137 @@
...
+		"data":     sse.Data,
+	}
+}
+
+// ToABI converts StateSyncEvent to ABI
+func (sse *StateSyncEvent) ToABI() ([]byte, error) {
+	ma := make([]map[string]interface{}, 1)
+	ma[0] = sse.ToMap()
+
+	return StateSyncABIType.Encode(ma)
```
Stefan-Ethernal: Nitpick, maybe try this way?

```suggestion
	return StateSyncABIType.Encode([]interface{}{sse.ToMap()})
```

consensus/polybft/consensus_runtime.go
```diff
@@ -205,9 +205,9 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	// after the block has been written we reset the txpool so that the old transactions are removed
 	c.config.txPool.ResetWithHeaders(block.Header)
 
-	// handle commitment and bundles creation
-	if err := c.createCommitmentAndBundles(block.Transactions); err != nil {
-		c.logger.Error("failed to create commitments and bundles", "error", err)
+	// handle commitment and proofs creation
+	if err := c.getCommitmentFromTransactions(block.Transactions); err != nil {
+		c.logger.Error("on block inserted error", "err", err)
```
Stefan-Ethernal: Should we change this log message to be more specific?

For example, like this (although doesn't need to be exactly like this):
```suggestion
		c.logger.Error("failed to retrieve commitment from transactions, "err", err)
```

consensus/polybft/consensus_runtime.go
```diff
@@ -242,23 +244,11 @@ func (c *consensusRuntime) createCommitmentAndBundles(txs []*types.Transaction)
...
-	if err != nil {
-		return fmt.Errorf("build bundles, get next execution index error: %w", err)
-	}
-
-	if err := c.buildBundles(
-		epoch.Commitment, commitment.Message, nextStateSyncExecutionIdx); err != nil {
-		return fmt.Errorf("build bundles error: %w", err)
+	if err := c.buildProofs(
+		epoch.Commitment, commitment.Message); err != nil {
+		return fmt.Errorf("create commitment error: %w", err)
```
Stefan-Ethernal: WDYT?
```suggestion
		return fmt.Errorf("build commitment proofs error: %w", err)
```

consensus/polybft/state.go
```diff
@@ -471,8 +461,8 @@ func (s *State) insertStateSyncEvent(event *StateSyncEvent) error {
 
 // getStateSyncEventsForCommitment returns state sync events for commitment
 // if there is an event with index that can not be found in db in given range, an error is returned
-func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64) ([]*StateSyncEvent, error) {
-	var events []*StateSyncEvent
+func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64) ([]*types.StateSyncEvent, error) {
+	var events []*types.StateSyncEvent
 
 	err := s.db.Update(func(tx *bolt.Tx) error {
```
Stefan-Ethernal: Shouldn't we use `View` function here instead, since we are basically reading the data from the DB?

```suggestion
	err := s.db.View(func(tx *bolt.Tx) error {
```

consensus/polybft/state.go
```diff
@@ -239,8 +229,8 @@ var (
 	exitEventsBucket = []byte("exitEvent")
 	// bucket to store commitments
 	commitmentsBucket = []byte("commitments")
-	// bucket to store bundles
-	bundlesBucket = []byte("bundles")
+	// bucket to store state sync proofs
+	stateSyncProofsBucket = []byte("stateSyncProofs")
```
Stefan-Ethernal, line 233: Just curious:
are we ever going to remove `stateSyncProofs` from the db or we must keep them forever, because of verification, since we don't know when somebody can query for the proof even if the state sync is fully executed?

consensus/polybft/commitment.go
```diff
@@ -4,33 +4,24 @@ import (
...
 	"math/big"
 	"reflect"
 
 	"github.com/0xPolygon/polygon-edge/crypto"
 	"github.com/0xPolygon/polygon-edge/state/runtime/precompiled"
 	"github.com/0xPolygon/polygon-edge/types"
 	"github.com/mitchellh/mapstructure"
-	"github.com/umbracle/ethgo"
 	"github.com/umbracle/ethgo/abi"
 )
```
Stefan-Ethernal: WDYT about keeping this file named as `state_transaction.go`? Rationale for it is because it not only contains logic for ABI encode/decode of commitments but also uptime transactions.

types/state_transaction.go
ferranbt: Yep, it is better to encode with array instead of map 0 thought both should work.

ferranbt (approved):

Stefan-Ethernal (approved):
Leaving minor comments, otherwise looks good.

e2e-polybft/bridge_test.go
vcastellm: You sure?

consensus/polybft/state.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/8a39e5377aeef0fe0d04fc62daecbd3b01bf3242

types/state_transaction.go
vcastellm: Much better, done in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/49d9ac7c63011e162ac4fa9b2162401a8a105bbf and finished in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/18ad7ad813b4507514a0b9cea34e2d4f5904690f

e2e-polybft/bridge_test.go
vcastellm: You're right done in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/4d99c07b504640117cc9121f63c3b0e065a3e45b

consensus/polybft/state.go
vcastellm: Weird it was like that before 😛

consensus/polybft/state.go
vcastellm: We're going to leave them forever for that very reason.

Stefan-Ethernal: 👌

consensus/polybft/commitment.go
vcastellm: Doing it in https://github.com/0xPolygon/polygon-edge/pull/1013/commits/1fc911037c5f7204aeb4daadac360cfb69f9fcd9 because it's of the scope of this PR but in the line of separating responsibilities in the code base, this deserves a refactor IMO splitting this to files, and even with the contractsapi gen.

PR #1014, opened 2022-12-08
Bundle build fix

goran-ethernal:
# Description

Since our bundle size is now 1, `buildBundles` function had an issue with building bundles for submitted commitments, where the start bundle index was not calculated correctly.

Bundles for a newly submitted commitment should always start from 0.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -586,9 +586,7 @@ func (c *consensusRuntime) buildBundles(commitment *Commitment, commitmentMsg *C
 
 	var bundleProofs []*BundleProof
 
-	startBundleIdx := commitmentMsg.GetBundleIdxFromStateSyncEventIdx(stateSyncExecutionIndex)
-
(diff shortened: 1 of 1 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -586,9 +586,7 @@ func (c *consensusRuntime) buildBundles(commitment *Commitment, commitmentMsg *C
 
 	var bundleProofs []*BundleProof
 
-	startBundleIdx := commitmentMsg.GetBundleIdxFromStateSyncEventIdx(stateSyncExecutionIndex)
```
Stefan-Ethernal, line 589: What do you think about removing `GetBundleIdxFromStateSyncEventIdx`, since it is being used only in the unit tests for assertions.

Stefan-Ethernal (approved):

vcastellm (approved):

ferranbt (approved):

PR #1018, opened 2022-12-08
EVM-302  Save proposer priorities after each block

igorcrevar:
# Description

After each block is inserted proposer priorities has to be stored in order that calculator can be determine for the next block

# Changes include

# Checklist

## Testing


consensus/polybft/state.go
```diff
@@ -840,6 +844,35 @@ func (s *State) bucketStats(bucketName []byte) *bolt.BucketStats {
 	return stats
 }
 
+// getProposerCalculatorSnapshot gets latest proposer calculator snapshot
+// we are keeping two snapshots just in case if latest write mess up snapshot
```
Stefan-Ethernal, line 848: Is this comment entirely correct? I don't see we are preserving two snapshots, since we are constantly storing a given instance by using the same value for `proposerCalcSnapshotKey`?

consensus/polybft/state.go
```diff
@@ -59,6 +59,8 @@ const (
 	stateSyncMainBundleSize = 10
 	// number of stateSyncEvents to be grouped into one StateTransaction
 	stateSyncBundleSize = 1
+	// latest proposer calculator snapshot key
+	proposerCalcSnapshotKey = 7
```
Stefan-Ethernal, line 63: If I'm correct, you are basically storing the latest snapshot? Couldn't you use some of the attributes from the `ProposerSnapshot` (like `Height` or `Height|Round`) itself instead of relying on this constant?

Also you mentioned that you want this storage to be double buffered (preserving two instances), which is not the case currently I'd say. 
If you would rely on some unique attribute value instead, you could achieve that.

Stefan-Ethernal: If I'm correct, you are basically storing the latest snapshot? Couldn't you use some of the attributes from the `ProposerSnapshot` (like `Height` or `Height|Round`) itself instead relying on this constant?

Also you mentioned that you want this storage to be double buffered (preserving two instances), which is not the case currently I'd say. If you would rely on some unique attribute value instead, you could achieve that. Although in that case, we should remove all the unnecessary snapshots (if there are any) when writing a new one into the DB.

consensus/polybft/consensus_runtime.go
```diff
@@ -1328,6 +1315,38 @@ func (c *consensusRuntime) BuildRoundChangeMessage(
...
+	}
+
+	// write snapshot if not already written
+	if (blockNumber-from+1)%saveEveryNIterations != 0 {
+		if err := c.state.writeProposerCalculatorSnapshot(snapshot); err != nil {
+			return fmt.Errorf("cannot save proposer calculator snapshot for block %d: %w", blockNumber, err)
+		}
+	}
+
+	// requires Lock
```
Stefan-Ethernal: Is this comment necessary? Any pending work to be done around it?

consensus/polybft/consensus_runtime.go
```diff
@@ -1328,6 +1315,38 @@ func (c *consensusRuntime) BuildRoundChangeMessage(
...
 
+func (c *consensusRuntime) updateProposerSnapshot(snapshot *ProposerSnapshot, blockNumber uint64) error {
+	const saveEveryNIterations = 5
+
+	c.logger.Info("Update proposal snapshot started", "block", blockNumber)
+
+	from := snapshot.Height
+
+	for height := from; height <= blockNumber; height++ {
+		c.config.proposerCalc.Update(snapshot, height, c.config)
```
Stefan-Ethernal: Handle this error?

consensus/polybft/consensus_runtime.go
```diff
@@ -220,6 +230,13 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	} else {
 		c.lock.Lock()
 		c.lastBuiltBlock = block.Header
+
+		proposalSnapshot := c.proposerSnapshot.Copy()
+		if err := c.updateProposerSnapshot(proposalSnapshot, block.Number()); err != nil {
+			c.logger.Warn("Could not update proposer calculator", "err", err)
```
Stefan-Ethernal:
```suggestion
			c.logger.Warn("Could not update proposer snapshot", "error", err)
```

consensus/polybft/consensus_runtime.go
```diff
@@ -220,6 +230,13 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	} else {
 		c.lock.Lock()
 		c.lastBuiltBlock = block.Header
+
+		proposalSnapshot := c.proposerSnapshot.Copy()
```
Stefan-Ethernal: Rename to `proposerSnapshot` or just `snapshot`? `proposalSnapshot` is a bit misleading.

consensus/polybft/proposer_calculator.go
```diff
@@ -34,238 +33,337 @@ var (
...
-	// IncrementProposerPriority increments priorities number of times
-	IncrementProposerPriority(times uint64) error
-
-	// GetLatestProposer returns latest calculated proposer
-	GetLatestProposer(round uint64) (types.Address, bool)
-}
-
-type validatorCalcMetadata struct {
+// Holds ValidatorMetadata together with priority
+type ProposerValidator struct {
```
Stefan-Ethernal: Thinking out loud is `PrioritizedValidator` a more convenient name maybe?

consensus/polybft/proposer_calculator.go
```diff
@@ -34,238 +33,337 @@ var (
...
+			priority = val.ProposerPriority // + int64(val.Metadata.VotingPower) - int64(x.VotingPower)
+		}
+
+		newValidatorsCalcProposer[i] = &ProposerValidator{
+			Metadata:         x,
+			ProposerPriority: priority,
+		}
+	}
+
+	// TODO: centering
```
Stefan-Ethernal, line 345: This is still to be tackled?

consensus/polybft/proposer_calculator.go
```diff
@@ -34,238 +33,337 @@ var (
...
+		addressOldValidatorMap[x.Metadata.Address] = x
+	}
+
+	// create new validators snapshot
+	for i, x := range newValidatorSet {
+		priority := int64(0)
+
+		// TODO: change priority if validator existed previous
+		if val, exists := addressOldValidatorMap[x.Address]; exists {
+			priority = val.ProposerPriority // + int64(val.Metadata.VotingPower) - int64(x.VotingPower)
```
Stefan-Ethernal, line 336: What about this commented addition and substraction? Do we need it?
Although there is a TODO above this if condition, so I'd say there is still some work left to be done.

consensus/polybft/consensus_runtime.go
igorcrevar: Will be changed with https://github.com/0xPolygon/polygon-edge/pull/1047

consensus/polybft/proposer_calculator.go
igorcrevar: Its in another task, Sasa is working on that one https://github.com/0xPolygon/polygon-edge/pull/1048/files

consensus/polybft/proposer_calculator.go
igorcrevar: It is part where Sasa will do his magic with another PR :) https://github.com/0xPolygon/polygon-edge/pull/1048/files

consensus/polybft/state.go
Stefan-Ethernal: I'd just remove this, since there is only one `ProposerSnapshot` in the db.
```suggestion
```

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/5b5dee9d1e9ef84ccabad386f0d57ec61aa962fe

consensus/polybft/proposer_calculator.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/5b5dee9d1e9ef84ccabad386f0d57ec61aa962fe

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/5b5dee9d1e9ef84ccabad386f0d57ec61aa962fe

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/5b5dee9d1e9ef84ccabad386f0d57ec61aa962fe

consensus/polybft/state.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/5b5dee9d1e9ef84ccabad386f0d57ec61aa962fe

consensus/polybft/proposer_calculator_test.go
```diff
@@ -442,27 +409,103 @@ func TestProposerCalculator_UpdatesForNewValidatorSet(t *testing.T) {
...
+	require.NoError(t, err)
+
+	v1 := &ValidatorMetadata{Address: types.Address{0x1}, BlsKey: keys[0].PublicKey(), VotingPower: 100}
+	v2 := &ValidatorMetadata{Address: types.Address{0x2}, BlsKey: keys[1].PublicKey(), VotingPower: 100}
+	v3 := &ValidatorMetadata{Address: types.Address{0x3}, BlsKey: keys[2].PublicKey(), VotingPower: 100}
+	v4 := &ValidatorMetadata{Address: types.Address{0x4}, BlsKey: keys[3].PublicKey(), VotingPower: 100}
+	v5 := &ValidatorMetadata{Address: types.Address{0x5}, BlsKey: keys[4].PublicKey(), VotingPower: 100}
+
+	accountSet := []*ValidatorMetadata{v1, v2, v3, v4, v5}
+	vs, err := NewValidatorSet(accountSet, hclog.NewNullLogger())
```
Stefan-Ethernal, line 482: Seems like this could have been shortened if `testValidators` abstraction was used...?

Nemanja0x (approved):

0xSasaPrsic (approved):
Looks good!

consensus/polybft/consensus_runtime.go
```diff
@@ -325,7 +342,11 @@ func (c *consensusRuntime) populateFsmIfBridgeEnabled(
 func (c *consensusRuntime) FSM() error {
 	// figure out the parent. At this point this peer has done its best to sync up
 	// to the head of their remote peers.
-	parent, epoch := c.getLastBuiltBlockAndEpoch()
+	c.lock.RLock()
+	proposerSnapshot := c.proposerSnapshot.Copy()
+	parent, epoch := new(types.Header), new(epochMetadata)
```
goran-ethernal: I am not sure about this, because now we are not consistent, since we have that `getEpoch` and `getLastBlockAndEpoch` functions which return copied data. 

What do you think, is it better to have a single function that returns all three fields copied, which is under Read lock, something like `getEpochFields() (*proposerSnapshot, *types.Header, *epochMetadata)`, so we can use it everywhere, since there are places in runtime that require latest block and epoch, some places where we require only block, some where we require all three, some where we require only epoch, or to just use the read lock and copy only what we need in the function that needs it.

Both have their advantages and disadvantages, but my point is we should just choose one approach and be consistent. My vote is for the first option since we don't hold that much data in those fields :thinking:

goran-ethernal (approved):
Looks Good. Left one comment. I think there is room for some Unit Tests for negative cases maybe.

consensus/polybft/consensus_runtime.go
igorcrevar: We need to have something like 
```
type runtimeSyncData struct {
	// epoch is the metadata for the current epoch
	epoch *epochMetadata

	// lastBuiltBlock is the header of the last processed block
	lastBuiltBlock *types.Header

	// proposerSnapshot is the object which calculates new proposer
	proposerSnapshot *ProposerSnapshot
}
```

igorcrevar: Fixed on https://github.com/0xPolygon/polygon-edge/pull/1047 https://github.com/0xPolygon/polygon-edge/pull/1047/commits/751644154d6a0ff5f1d3cd048534f636c45f7cb9

Stefan-Ethernal (approved):

consensus/polybft/state.go
```diff
@@ -841,6 +845,34 @@ func (s *State) bucketStats(bucketName []byte) *bolt.BucketStats {
 	return stats
 }
 
+// getProposerSnapshot gets latest proposer snapshot
+func (s *State) getProposerSnapshot() (*ProposerSnapshot, error) {
```
ferranbt, line 852: This function does not have any test. Please update too the storage layout at the top of the file to understand how is this data stored.

consensus/polybft/extra.go
```diff
@@ -296,7 +296,7 @@ func unmarshalValidators(validatorsRaw []*fastrlp.Value) (AccountSet, error) {
 
 // IsEmpty returns indication whether delta is empty (namely added, updated slices and removed bitmap are empty)
 func (d *ValidatorSetDelta) IsEmpty() bool {
-	return len(d.Added) == 0 &&
+	return d == nil || len(d.Added) == 0 &&
```
ferranbt: It is not a good pattern for me to expect a nil delta like this. Instead, make sure that `IsEmpty` only gets called if delta is not nil.

consensus/polybft/consensus_runtime.go
```diff
@@ -1144,7 +1137,27 @@ func (c *consensusRuntime) HasQuorum(
 			return c.fsm.validators.HasQuorum(signers)
 		}
 
-		return c.fsm.validators.HasQuorumWithoutProposer(signers)
+		if len(messages) == 0 {
```
ferranbt, line 1153: What is going on here? `HasQuorum` is an important part of the consensus. Any big changes like this should have some comments. Besides, I do not like how bulky and complex is this function getting.

consensus/polybft/consensus_runtime.go
```diff
@@ -477,6 +463,13 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) error {
 	c.lock.Lock()
 	c.epoch = epoch
 	c.lastBuiltBlock = header
+
+	proposerSnapshot := c.proposerSnapshot.Copy()
+	if err := c.updateProposerSnapshot(proposerSnapshot, header.Number); err != nil {
```
ferranbt: Do you need to do this here? Why not do it alongside the other part when the block is inserted?

ferranbt: It is confusing to do the same thing in two places.

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,437 @@
...
+	GetLatestProposer(snapshot *ProposerSnapshot, round, height uint64) (types.Address, bool)
+}
+
+type proposerCalculator struct {
+	// logger instance
+	logger hclog.Logger
+}
+
+// NewProposerCalculator creates a new proposer calculator object
+func NewProposerCalculator(logger hclog.Logger) *proposerCalculator {
```
ferranbt: So you have two data structures: `calculator` and `snapshot`. The first one is stateless and its some sort of function pattern, the other one has a lifecycle (store, load, update). 

Why not merge both things on the same place? Something like this (similar to the hook concept of last week)
```
type PriorityProposer struct {
   // snapshot
}

func (p *PriorityProposer) init() {
  snapshot = // load from storage
}

func (p *PriorityProposer) OnBlock(block ...) {
   // update stuff
   snapshot = ...
}

func (p *PriorityProposer) CalcProposer(round uint64) ... {
   // ...
}
```

ferranbt (changes requested):

consensus/polybft/consensus_runtime.go
```diff
@@ -313,7 +330,11 @@ func (c *consensusRuntime) populateFsmIfBridgeEnabled(
 func (c *consensusRuntime) FSM() error {
 	// figure out the parent. At this point this peer has done its best to sync up
 	// to the head of their remote peers.
-	parent, epoch := c.getLastBuiltBlockAndEpoch()
+	c.lock.RLock()
+	proposerSnapshot := c.proposerSnapshot.Copy()
+	parent, epoch := new(types.Header), new(epochMetadata)
+	*parent, *epoch = *c.lastBuiltBlock, *c.epoch
```
b00ris: It's not a safe copy for c.lastBuiltBlock. Could you use c.lastBuiltBlock.Copy?

b00ris: For epoch also, because below you are working the same array as c.epoch.Validators outside the mutex

consensus/polybft/consensus_runtime.go
igorcrevar: This will be changed with EVM-303. The problem is currently we have condition:
- if end of epoch -> call restart epoch and change state of runtime (lastblock, snapshot, epoch)
- otherwise change lastblock, snapshot

consensus/polybft/proposer_calculator_test.go
0xSasaPrsic: Creating random address would lead to inconsistency and unpredictable behavior.

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/4d4fff05b2f2df8263b3a8903d32fe77e2853e17

consensus/polybft/extra.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/4d4fff05b2f2df8263b3a8903d32fe77e2853e17

consensus/polybft/state.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/commit/4d4fff05b2f2df8263b3a8903d32fe77e2853e17 and https://github.com/0xPolygon/polygon-edge/pull/1018/commits/21d401f3f10b521f5fbb656a8d8ca91fe336956e

consensus/polybft/consensus_runtime.go
Nemanja0x: "Besides, I do not like how bulky and complex is this function getting." - completely agree, but we need to make changes on go-ibft side if we want to implement HasQuorum as it should be implemented.

As I mentioned on previous PR, there is a situation when go-ibft is not going to send a message from the proposer but expects  that backend will include voting power of the proposer into the calculation of quorum. Therefore, we need to get the latest proposer and include it into the calculation of quorum.

On a slack discussion couple of weeks ago I have mentioned that this function is going to look terrible.

consensus/polybft/proposer_calculator.go
Nemanja0x: This is maybe the most important issue of the PR and it is good to discuss it.
 So this is the trouble:
1. `Syncer` and `Consensus` work in parallel. It can happen that `Syncer` finishes while `Consensus` is in the middle of a round.

2. `ConsensusRuntime` lives while blockchain lives and, therefore, we have decided to keep latest priories  as a part of `ConsensusRuntime`. The informartion is always up to date since it is modified no matter who ever first inserts a block (`Syncer` or `Consensus`)

There can be a situation when `Syncer` updates priories while `Consensus` is still running. As a consequence, `Consensus` will have wrong information.  What is more, it can happen that `Consensus` accesses priorities while `Syncer` is changing it.  To prevent that, we make a copy of data when `FSM` is created and (of course) store it in `FSM`. That copy is used only while the current instance of `FSM` exists for `Consensus`. That is why we decoupled data (priorities) from the algorithm. Everyone keeps its  own data and uses the algorithm when it is needed.

consensus/polybft/consensus_runtime.go
```diff
@@ -477,6 +463,13 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) error {
 	c.lock.Lock()
 	c.epoch = epoch
 	c.lastBuiltBlock = header
+
+	proposerSnapshot := c.proposerSnapshot.Copy()
+	if err := c.updateProposerSnapshot(proposerSnapshot, header.Number); err != nil {
+		c.logger.Warn("Could not update proposer calculator", "err", err)
```
Stefan-Ethernal, line 467: Nitpick, just to be consistent with other logs.

```suggestion
		c.logger.Warn("could not update proposer calculator", "err", err)
```

consensus/polybft/state.go
```diff
@@ -59,6 +62,8 @@ const (
 	stateSyncMainBundleSize = 10
 	// number of stateSyncEvents to be grouped into one StateTransaction
 	stateSyncBundleSize = 1
+	// static key which is used to save latest proposer snapshot. It will be always one object in bucket
+	proposerSnapshotStaticKey = 7
```
ferranbt: The static key should be a `[]byte` not a number.

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+	return &ProposerSnapshot{
+		Validators: valCopy,
+		Height:     pcs.Height,
+		Round:      pcs.Round,
+		Proposer:   proposer,
+	}
+}
+
+// ProposerCalculator interface - proposer calculator algorithm should implement this interface
+type ProposerCalculator interface {
```
ferranbt: Why do we need this interface here? For testing?

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	return &proposerCalculator{
+		snapshot: snap,
+		logger:   logger,
+	}, nil
+}
+
+// NewProposerCalculator creates a new proposer calculator object
+func NewProposerCalculatorFromSnapshot(pcs *ProposerSnapshot, logger hclog.Logger) *proposerCalculator {
```
ferranbt: So `proposerCalculator` in this scenario does two things? One is to keep the up-to-date snapshot with the calls to `Update` and another one is to serve as the "Snapshot" object that calculates the proposer.

We should not use it for both things at the same time. Move the `CalcProposer` function to the Snapshot itself and you remove the need for this two functionalities.

consensus/polybft/consensus_runtime.go
```diff
@@ -208,6 +217,7 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	} else {
 		c.lock.Lock()
 		c.lastBuiltBlock = block.Header
+		c.proposerCalculator.Update(block.Number(), c.config, c.state)
```
ferranbt: It does not get called at the end of the epoch?

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+
+	pc.snapshot.Proposer = proposer
+	pc.snapshot.Round = round
+
+	pc.logger.Info("New proposer calculated", "height", height, "round", round, "address", proposer.Metadata.Address)
+
+	return proposer.Metadata.Address, nil
+}
+
+func (pc *proposerCalculator) Update(blockNumber uint64, config *runtimeConfig, state *State) error {
```
ferranbt: Move the two static parameters `config` and `state` to the constructor of the `proposerCalculator` instead.

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+	const saveEveryNIterations = 5
+
+	pc.logger.Info("Update proposal snapshot started", "block", blockNumber)
+
+	from := pc.snapshot.Height
+
+	for height := from; height <= blockNumber; height++ {
+		if err := pc.updatePerBlock(height, config); err != nil {
```
ferranbt: This is just a question but could not we ensure that this gets called on a sequence of blocks and we do not have sequences of skipped blocks?

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+	if avg.IsInt64() {
+		return avg.Int64(), nil
+	}
+
+	return 0, fmt.Errorf("cannot represent avg ProposerPriority as an int64 %v", avg)
+}
+
+// rescalePriorities rescales the priorities such that the distance between the
+// maximum and minimum is smaller than `diffMax`.
+func rescalePriorities(snapshot *ProposerSnapshot, totalVotingPower int64) error {
```
ferranbt, line 430:
```suggestion
func (snapshot *ProposerSnapshot) rescalePriorities(totalVotingPower int64) error {
```
If possible, lets not follow a function approach and create internal methods for the type.

consensus/polybft/consensus_runtime.go
```diff
@@ -108,16 +108,25 @@ type consensusRuntime struct {
 	// checkpointManager represents abstraction for checkpoint submission
 	checkpointManager *checkpointManager
 
+	// proposerCalculator is the object which calculates new proposer
+	proposerCalculator ProposerCalculator
```
```suggestion
	proposerCalculator *ProposerCalculator
```

ferranbt (changes requested):

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	pc.logger.Info("Get latest proposer",
+		"height", height, "round", round, "address", pc.snapshot.Proposer.Metadata.Address)
+
+	return pc.snapshot.Proposer.Metadata.Address, true
+}
+
+// CalcProposer calculates next proposer
+func (pc *proposerCalculator) CalcProposer(round, height uint64) (types.Address, error) {
```
ferranbt:
```suggestion
func (pc *proposerCalculator) CalcProposer(round uint64) (types.Address, error) {
```
The Snapshot here should already know what the height of the block is.

consensus/polybft/proposer_calculator_test.go
```diff
@@ -0,0 +1,510 @@
...
+
+	vs, err := validators.toValidatorSet()
+	require.NoError(t, err)
+
+	snapshot := NewProposerSnapshot(0, metadata)
+	pc := NewProposerCalculatorFromSnapshot(snapshot, hclog.NewNullLogger())
+
+	assert.Equal(t, int64(15), vs.totalVotingPower)
+
+	currProposerAddress, err := pc.CalcProposer(0, 0)
```
ferranbt: Could you do a table test here?

consensus/polybft/proposer_calculator_test.go
```diff
@@ -0,0 +1,510 @@
...
+
+	proposerAddressR6, err := pc.CalcProposer(6, 0)
+	require.NoError(t, err)
+	assert.Equal(t, metadata[0].Address, proposerAddressR6)
+}
+
+func TestProposerCalculator_SamePriority(t *testing.T) {
+	t.Parallel()
+
+	keys, err := bls.CreateRandomBlsKeys(5)
```
ferranbt, line 85: Can you use `newTestValidatorsWithAliases` in this scenario?

consensus/polybft/proposer_calculator_test.go
```diff
@@ -0,0 +1,510 @@
...
+	proposerR2, err := pc.CalcProposer(2, 0)
+	require.NoError(t, err)
+	assert.Equal(t, types.Address{0x1}, proposerR2)
+
+	proposerR2, err = pc.CalcProposer(2, 0) // call again same round
+	require.NoError(t, err)
+	assert.Equal(t, types.Address{0x1}, proposerR2)
+}
+
+func TestProposerCalculator_ProposerSelection1(t *testing.T) {
```
ferranbt: How is this test different from `TestProposerCalculator_CalculateProposer`?

consensus/polybft/proposer_calculator_test.go
```diff
@@ -0,0 +1,510 @@
...
+	pc := NewProposerCalculatorFromSnapshot(snapshot, hclog.NewNullLogger())
+
+	var proposers = make([]types.Address, numberOfIteration)
+
+	for i := uint64(0); i < numberOfIteration; i++ {
+		proposers[i], err = pc.CalcProposer(i, 4)
+		require.NoError(t, err)
+	}
+
+	expected := []types.Address{
```
ferranbt: What does this expected array represent?

consensus/polybft/consensus_runtime.go
ferranbt: Please open a ticket, we have to give priority to improve this part.

consensus/polybft/consensus_runtime.go
```diff
@@ -1144,7 +1134,26 @@ func (c *consensusRuntime) HasQuorum(
 			return c.fsm.validators.HasQuorum(signers)
 		}
 
-		return c.fsm.validators.HasQuorumWithoutProposer(signers)
+		if len(messages) == 0 {
+			return false
+		}
+
+		propAddress, exist := c.fsm.proposerCalculator.GetLatestProposer(messages[0].View.Round, blockNumber)
```
ferranbt: This should be the job of the `proposerSnapshot`.

consensus/polybft/proposer_calculator.go
Nemanja0x: Agree. Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/a46c6ec90b94f099b894f1e7aec34258d99b8b26

consensus/polybft/consensus_runtime.go
Nemanja0x: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/a46c6ec90b94f099b894f1e7aec34258d99b8b26

consensus/polybft/proposer_calculator.go
Nemanja0x: The idea was simmilar as for `ValidatorSet`, to have an interface if we want to change algorithm. Since we are moving `CalcProposer` to `ProposerSnapshot`, I'll remove the interface

Nemanja0x: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/48f6b5d8b23ffd0eb1cfc40fe6aa9ccfd9188845

consensus/polybft/consensus_runtime.go
Nemanja0x: This is fixed as ProposerCalculator interface was removed: https://github.com/0xPolygon/polygon-edge/pull/1018/commits/48f6b5d8b23ffd0eb1cfc40fe6aa9ccfd9188845

consensus/polybft/proposer_calculator.go
Nemanja0x: Yep, it should. Since we receive both `round` and `height` when `CalcProposer` is called by `go-ibft`, we use the received `height` just to do an additional validation before actually calculating a proposer. In this way we are sure that `ProposerSnapshot` data is completely aligned with the state of `go-ibft`. For now, it can detect possible bugs/race conditions in saving/loading `ProposerSnapshot` feature.

consensus/polybft/proposer_calculator.go
Nemanja0x: This is actually called whenever a block is inserted (`OnBlockInserted`). That `FOR` loop should do only one iteration each time because the snapshot of the previous block was (hopefully) saved.  
We made this approach with the `FOR` loop just in case that somehow the snapshot of the previous block was not saved (some unexpected situation occurred). That way snapshot would catch up the blockchain during the next call of `OnBlockInserted`.

consensus/polybft/proposer_calculator_test.go
Nemanja0x: I think that `TestProposerCalculator_ProposerSelection` is copied from `Tendermint`. Not sure about this. Sasa has to help us with this. @0xSasaPrsic

consensus/polybft/proposer_calculator_test.go
Nemanja0x: Those are addresses of proposers which are expected to be calculated in that UT. @0xSasaPrsic to confirm.

consensus/polybft/proposer_calculator_test.go
Nemanja0x: It looks like that all the proposer_calculator tests should be switched to `TestValidatorsWithAliases`. Furthermore, some of the tests will be changed by https://github.com/0xPolygon/polygon-edge/pull/1048 which should be rebased on top of this PR. 
Is it ok to modify  the tests after all the PRs related to `VotingPower` are merged?

consensus/polybft/proposer_calculator_test.go
Nemanja0x: https://github.com/0xPolygon/polygon-edge/pull/1018#discussion_r1051650359

consensus/polybft/consensus_runtime.go
Nemanja0x: Here it is https://polygon.atlassian.net/browse/EVM-326

consensus/polybft/proposer_calculator.go
Nemanja0x: We want to have `config` and `state` as parameters of `proposerCalculator`?

consensus/polybft/proposer_calculator.go
Nemanja0x: Can this be a new task? It will be easier after we finish the whole feature.

consensus/polybft/consensus_runtime.go
Nemanja0x: It is called after each block. `OnBlockInserted` will be modified through https://github.com/0xPolygon/polygon-edge/pull/1047 which will address possible race conditions and it is built on top of this PR.

consensus/polybft/proposer_calculator.go
ferranbt: Why not, they are static parameters. Those parameters are not related to the `Update` action.

consensus/polybft/proposer_calculator_test.go
0xSasaPrsic: This is poorly named test, it tests that inverse order of priorities int the array does not affect selection of proposer and doing multiple iteration where frequency of the validator with the highest voting power is higher then the ones with the lower voting power.

consensus/polybft/proposer_calculator_test.go
0xSasaPrsic: Yes, expected addresses in the order in which it is expected for the proposer to be selected.

consensus/polybft/state.go
Stefan-Ethernal: @ferranbt Seems it is impossible to put empty byte slice as a key. Tried it and I'm getting an error from bolt db ("key required").

I agree that it is strange we have 7 as a key, but not sure what to do here? Should we use just `proposerSnapshot.Height` and then fetch it using `cursor.Last()/cursor.First()` (although I guess that would probably have _some_ performance penalty)...

goran-ethernal: Why not just use `proposerSnapshotKey = []byte("oneKeyToRuleThemAll")`. Just some string, if we are always rewriting that one snapshot.

consensus/polybft/proposer_calculator.go
Nemanja0x: Fixed https://github.com/0xPolygon/polygon-edge/pull/1018/commits/f0879b3c9bf91004bf3dd6f1243cc2be6cf960d5

consensus/polybft/proposer_calculator_test.go
ferranbt: Ok, we just need to add that description to the unit tests. It is important to keep this context.

consensus/polybft/state.go
Stefan-Ethernal: Fixed per @goran-ethernal suggestion (https://github.com/0xPolygon/polygon-edge/pull/1018/commits/184be592c0d4fd3e7f18d786152fc1c73a1890ba)

consensus/polybft/proposer_calculator_test.go
0xSasaPrsic: Changes applied https://github.com/0xPolygon/polygon-edge/pull/1018/commits/01cea74b365bd7a4c9bdac169d934a204b77c850

consensus/polybft/proposer_calculator.go
```diff
@@ -0,0 +1,487 @@
...
+	return &ProposerCalculator{
+		snapshot: snap,
+		config:   config,
+		state:    config.State,
+		logger:   logger,
+	}, nil
+}
+
+// NewProposerCalculator creates a new proposer calculator object
+func NewProposerCalculatorFromSnapshot(pcs *ProposerSnapshot, config *runtimeConfig,
```
ferranbt, line 187: Is this need it anymore?

consensus/polybft/proposer_calculator.go
ferranbt: Please add to Jira.

consensus/polybft/proposer_calculator_test.go
ferranbt: Sure, please add to Jira.

consensus/polybft/proposer_calculator.go
Nemanja0x: I think we'll needed it for https://github.com/0xPolygon/polygon-edge/pull/1047. Currently, it used only for tests.

consensus/polybft/proposer_calculator_test.go
Nemanja0x: Here is the task: https://polygon.atlassian.net/browse/EVM-329

consensus/polybft/proposer_calculator.go
Nemanja0x: Here is the task: https://polygon.atlassian.net/browse/EVM-330

ferranbt (approved):

PR #1024, opened 2022-12-09
Integrate L2 exit workflow

b00ris:
# Description
Full e2e L2->L1 test.
Closes https://polygon.atlassian.net/browse/EVM-162 and https://polygon.atlassian.net/browse/EVM-91

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


.github/workflows/e2e.yaml
```diff
@@ -23,6 +23,8 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: 1.18.x
+      - name: Compile core contracts
+        run: make compile-core-contracts
```
Stefan-Ethernal: This is not needed for IBFT E2E tests, but only for the polybft ones. We have a separate GH workflow for polybft e2e tests `e2e-polybft.yml`
```suggestion
```

consensus/polybft/fsm.go
```diff
@@ -628,6 +628,10 @@ func getExitEventsFromReceipts(epoch, block uint64, receipts []*types.Receipt) (
 		}
 
 		for j := 0; j < len(receipts[i].Logs); j++ {
+			if receipts[i].Logs[j].Address != contracts.L2StateSenderContract {
+				continue
+			}
+
 			event, err := decodeExitEvent(convertLog(receipts[i].Logs[j]), epoch, block)
```
Stefan-Ethernal: Although not sure if it makes more sense adding this check inside `decodeExitEvent` itself. WDYT?

```suggestion
                        log := receipts[i].Logs[j]
			if log.Address != contracts.L2StateSenderContract {
				continue
			}

			event, err := decodeExitEvent(convertLog(log), epoch, block)
```

.github/workflows/test.yml
```diff
@@ -24,7 +24,8 @@ jobs:
 
       - name: Install Dependencies
         run: ./setup-ci.sh
-
+      - name: Compile core contracts
+        run: make compile-core-contracts
```
Stefan-Ethernal: This wasn't necessary before. Can you point out for which exact unit test(s) we need to have compiled smart contracts?

command/genesis/polybft_params.go
```diff
@@ -252,12 +254,18 @@ func (p *genesisParams) deployContracts() (map[types.Address]*chain.GenesisAccou
 			relativePath: "common/Merkle.sol",
 			address:      contracts.MerkleContract,
 		},
+		{
+			// Merkle contract
```
Stefan-Ethernal:
```suggestion
			// L2StateSender contract
```

command/rootchain/initcontracts/init_contracts.go
```diff
@@ -279,39 +288,53 @@ func initializeCheckpointManager(txRelayer txrelayer.TxRelayer) error {
...
+	}
+
+	txn := &ethgo.Transaction{
+		To:    &helper.ExitHelperAddress,
+		Input: initCheckpointInput,
+	}
+
+	receipt, err := txRelayer.SendTransaction(txn, helper.GetRootchainAdminKey())
+	if err != nil {
+		return fmt.Errorf("failed to send transaction to CheckpointManager. error: %w", err)
```
Stefan-Ethernal:
```suggestion
		return fmt.Errorf("failed to send transaction to ExitHelper. error: %w", err)
```

command/rootchain/initcontracts/init_contracts.go
```diff
@@ -279,39 +288,53 @@ func initializeCheckpointManager(txRelayer txrelayer.TxRelayer) error {
...
+	}
+
+	if receipt.Status != uint64(types.ReceiptSuccess) {
+		return errors.New("failed to initialize CheckpointManager")
```
Stefan-Ethernal:
```suggestion
		return errors.New("failed to initialize ExitHelper")
```

command/rootchain/initcontracts/init_contracts.go
```diff
@@ -279,39 +288,53 @@ func initializeCheckpointManager(txRelayer txrelayer.TxRelayer) error {
...
 	if err != nil {
 		return nil, err
 	}
 
-	validatorSetMap := make([]map[string]interface{}, len(validatorsInfo))
-
-	sort.Slice(validatorsInfo, func(i, j int) bool {
-		return bytes.Compare(validatorsInfo[i].Address.Bytes(),
-			validatorsInfo[j].Address.Bytes()) < 0
-	})
```
Stefan-Ethernal: This was important point when submitting validator set to the rootchain: they need to be ordered exactly the same way as on client-side.
`ChildValidatorSet` orders validators by `totalStake` and here we are providing validators by name of its respective folders where data are kept... To me, this should be reverted, otherwise signature verification on `CheckpointManager` wouldn't work correctly.

e2e-polybft/bridge_test.go
```diff
@@ -171,30 +179,187 @@ func TestE2E_Bridge_MainWorkflow(t *testing.T) {
...
+		txn := &ethgo.Transaction{
+			To:    &fundAddr,
+			Value: big.NewInt(1000000000000000000),
+		}
+
+		_, err := txRelayer.SendTransactionLocal(txn)
+		if err != nil {
+			t.Error(validator.Address, "error on funding", err)
+		}
+	}
```
Stefan-Ethernal: Remove the function.
Validators are funded as a part of this function in e2e framework:

https://github.com/0xPolygon/polygon-edge/blob/84ef5305c2bf4fd28155987f7690347fd0531a95/e2e-polybft/framework/test-cluster.go#L293

```suggestion
```

consensus/polybft/contractsapi/init.go
```diff
@@ -0,0 +1,155 @@
...
+	L1ExitTestABI *abi.ABI
+)
+
+func init() {
+	_, filename, _, _ := runtime.Caller(0) //nolint: dogsled
+	scpath := path.Join(path.Dir(filename), "../../../core-contracts/artifacts/contracts/")
+
+	var err error
+
+	Rootchain, err = ReadArtifact(scpath, "root/CheckpointManager.sol", "CheckpointManager")
```
Stefan-Ethernal: Rename it to `CheckpointManager`?

consensus/polybft/state.go
```diff
@@ -381,19 +379,22 @@ func insertExitEventToBucket(bucket *bolt.Bucket, exitEvent *ExitEvent) error {
...
 		bucket := tx.Bucket(exitEventsBucket)
 
-		key := bytes.Join([][]byte{itob(epoch), itob(exitEventID), itob(checkpointBlockNumber)}, nil)
-		v := bucket.Get(key)
-		if v == nil {
+		key := bytes.Join([][]byte{itob(epoch), itob(exitEventID)}, nil)
+		k, v := bucket.Cursor().Seek(key)
+		fmt.Println(k, "k")
+		fmt.Println(key, "key")
+		fmt.Println(v, "v")
```
Stefan-Ethernal: Are these leftovers? Should we remove them?

command/rootchain/initcontracts/init_contracts.go
```diff
@@ -279,39 +288,53 @@ func initializeCheckpointManager(txRelayer txrelayer.TxRelayer) error {
 	return nil
 }
 
+func initializeExitHelper(txRelayer txrelayer.TxRelayer, checkpointManagerAddress ethgo.Address) error {
+	initCheckpointInput, err := contractsapi.ExitHelper.Abi.GetMethod("initialize").
```
Stefan-Ethernal: Rename to `initExitHelperInput`?

.github/workflows/test.yml
Stefan-Ethernal: Is it because of this `TestPerformExit`? Could we somehow refactor that test to make it any smaller (and do we need it in the first place, since we have E2E test)?

consensus/polybft/state.go
```diff
@@ -381,19 +379,22 @@ func insertExitEventToBucket(bucket *bolt.Bucket, exitEvent *ExitEvent) error {
...
+func (s *State) getExitEvent(exitEventID, epoch uint64) (*ExitEvent, error) {
 	var exitEvent *ExitEvent
 
 	err := s.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket(exitEventsBucket)
 
-		key := bytes.Join([][]byte{itob(epoch), itob(exitEventID), itob(checkpointBlockNumber)}, nil)
-		v := bucket.Get(key)
-		if v == nil {
+		key := bytes.Join([][]byte{itob(epoch), itob(exitEventID)}, nil)
```
goran-ethernal, line 388: Is there a reason you removed `checkpointBlockNumber` here? Since this function is called on `GenerateExitProof`, where we know the block number, I would leave it as it is, since it was faster before, since it just gets the value by key, without the need to do `Seek`.

consensus/polybft/fsm.go
b00ris: We don't need to decode events not from L2StateSenderContract

consensus/polybft/state.go
b00ris: Yes, there is. We have 2 block numbers there:
* checkpoint block number(when we need to stop create trie)
* exit event block number(that we need there)
In api we are sending only first one.

consensus/polybft/fsm.go
Stefan-Ethernal: I know, I was just suggesting moving that check to `decodeExitEvent` function as well.

However still would like to create a local variable `log` instead of indexing slices twice.

b00ris: Switched to for range. My idea was to filter not relevant logs as early as possible. Because if we move this check to decode function we'll have to convert every log to ethgo format, before filtering.

Stefan-Ethernal: Ok makes sense :wink:

.github/workflows/test.yml
b00ris: I commited artifacts inside codebase, so i got rid of this.

consensus/polybft/contractsapi/gen/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"os"
+	"path"
+	"runtime"
+
+	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi/artifact"
+	"github.com/dave/jennifer/jen"
+)
+
+func main() {
```
goran-ethernal, line 13: Correct me if I'm wrong. If I am manually compiling contracts on my local machine, I would need to run this after that?

consensus/polybft/system_state_test.go
```diff
@@ -203,7 +203,7 @@ func TestStateProvider_Txn_Panics(t *testing.T) {
 	require.Panics(t, func() { _, _ = provider.Txn(ethgo.ZeroAddress, key, []byte{0x1}) })
 }
 
-func newTestTransition(t *testing.T) *state.Transition {
+func newTestTransition(t *testing.T, alloc map[types.Address]*chain.GenesisAccount) *state.Transition {
```
goran-ethernal, line 206: I don't think we are setting this `alloc` map anywhere in tests. It's always `nil`, so we can probably remove it from this test helper function.

goran-ethernal (approved):
Great job. :100: 

It would be great if we could abstract that UT a bit though, it's setup is quite long :confused:

consensus/polybft/contractsapi/gen/main.go
b00ris: If you call compile-core-contracts - not.
From Makefile:
```
compile-core-contracts:
 	cd core-contracts && npm install && npm run compile
 	go run ./consensus/polybft/contractsapi/gen/main.go
```

consensus/polybft/contractsapi/gen_sc_data.go
```diff
@@ -1,8 +1,10 @@
 package contractsapi
```
Stefan-Ethernal, line 1: Consider adding this file to git ignore as it is result of generation process. Rationale is to prevent merge conflicts around it in the future. WDYT?

consensus/polybft/contractsapi/init.go
```diff
@@ -0,0 +1,161 @@
...
...00080fd5b818601915086601f83011261027357600080fd5b813581811115610285576102856101f9565b604051601f8201601f19908116603f011681019083821181831017156102ad576102ad6101f9565b816040528281528960208487010111156102c657600080fd5b8260208601602083013760006020848301015280955050505050509250925092565b600181811c908216806102fc57607f821691505b60208210810361031c57634e487b7160e01b600052602260045260246000fd5b50919050565b601f82111561036c57600081815260208120601f850160051c810160208610156103495750805b601f850160051c820191505b8181101561036857828155600101610355565b5050505b505050565b815167ffffffffffffffff81111561038b5761038b6101f9565b61039f8161039984546102e8565b84610322565b602080601f8311600181146103d457600084156103bc5750858301515b600019600386901b1c1916600185901b178555610368565b600085815260208120601f198616915b82811015610403578886015182559484019460019091019084016103e4565b50858210156104215787850151600019600388901b60f8161c191681555b5050505050600190811b01905550565b60006001820161045157634e487b7160e01b600052601160045260246000fd5b506001019056fea264697066735822122038ac4dcd21701f192702ee4a99e63fef6c90075b018b5338d65c1d477a999c6364736f6c63430008110033")                                                                 //nolint
```
Stefan-Ethernal: There is typo in naming `L11ExitDeployedBytecode` -> `L1ExitDeployedBytecode` :slightly_smiling_face: 

```suggestion
	L1ExitDeployedBytecode = hex.MustDecodeHex("0x608060405234801561001057600080fd5b50600436106100575760003560e01c806361bc221a1461005c57806373d4a13a14610078578063767800de1461008d578063af640d0f146100b8578063f43cda8b146100c1575b600080fd5b61006560035481565b6040519081526020015b60405180910390f35b6100806100d6565b60405161006f91906101ab565b6001546100a0906001600160a01b031681565b6040516001600160a01b03909116815260200161006f565b61006560005481565b6100d46100cf36600461020f565b610164565b005b600280546100e3906102e8565b80601f016020809104026020016040519081016040528092919081815260200182805461010f906102e8565b801561015c5780601f106101315761010080835404028352916020019161015c565b820191906000526020600020905b81548152906001019060200180831161013f57829003601f168201915b505050505081565b6000839055600180546001600160a01b0319166001600160a01b03841617905560026101908282610371565b50600380549060006101a183610431565b9190505550505050565b600060208083528351808285015260005b818110156101d8578581018301518582016040015282016101bc565b506000604082860101526040601f19601f8301168501019250505092915050565b634e487b7160e01b600052604160045260246000fd5b60008060006060848603121561022457600080fd5b8335925060208401356001600160a01b038116811461024257600080fd5b9150604084013567ffffffffffffffff8082111561025f57600080fd5b818601915086601f83011261027357600080fd5b813581811115610285576102856101f9565b604051601f8201601f19908116603f011681019083821181831017156102ad576102ad6101f9565b816040528281528960208487010111156102c657600080fd5b8260208601602083013760006020848301015280955050505050509250925092565b600181811c908216806102fc57607f821691505b60208210810361031c57634e487b7160e01b600052602260045260246000fd5b50919050565b601f82111561036c57600081815260208120601f850160051c810160208610156103495750805b601f850160051c820191505b8181101561036857828155600101610355565b5050505b505050565b815167ffffffffffffffff81111561038b5761038b6101f9565b61039f8161039984546102e8565b84610322565b602080601f8311600181146103d457600084156103bc5750858301515b600019600386901b1c1916600185901b178555610368565b600085815260208120601f198616915b82811015610403578886015182559484019460019091019084016103e4565b50858210156104215787850151600019600388901b60f8161c191681555b5050505050600190811b01905550565b60006001820161045157634e487b7160e01b600052601160045260246000fd5b506001019056fea264697066735822122038ac4dcd21701f192702ee4a99e63fef6c90075b018b5338d65c1d477a999c6364736f6c63430008110033")                                                                 //nolint
```

consensus/polybft/contractsapi/init.go
```diff
@@ -0,0 +1,161 @@
...
+          "name": "_data",
+          "type": "bytes"
+        }
+      ],
+      "name": "onL2StateReceive",
+      "outputs": [],
+      "stateMutability": "nonpayable",
+      "type": "function"
+    }
+  ]
```
Stefan-Ethernal, line 160: Does it make sense to put this part to a separate file as well as `L1ExitTestBytecode` and `L1ExitDeployedBytecode`?

Also it would be great if SC team eventually would implement such SC on their end (they did the same thing with `MRC20.sol` basically) and we can use that one for testing purposes.

.github/workflows/test.yml
```diff
@@ -24,7 +24,6 @@ jobs:
 
       - name: Install Dependencies
         run: ./setup-ci.sh
-
```
Stefan-Ethernal: Nitpick: could you add new line here (as it was before)?

consensus/polybft/contractsapi/gen/main.go
```diff
@@ -0,0 +1,73 @@
...
+	_, filename, _, _ := runtime.Caller(0) //nolint: dogsled
+	currentPath := path.Dir(filename)
+	scpath := path.Join(currentPath, "../../../../core-contracts/artifacts/contracts/")
+
+	f := jen.NewFile("contractsapi")
+	f.Comment("This is generated file don't change it manually.  DO NOT EDIT.")
```
Stefan-Ethernal: Nitpick: shorten this comment (initially I was triggered to change it because of one extra space before "DO NOT EDIT" part, so at least trim that extra space)? :slightly_smiling_face: 

```suggestion
	f.Comment("This is auto-generated file. DO NOT EDIT.")
```

Stefan-Ethernal (approved):
LGTM, apart from the minor suggestions.

consensus/polybft/contractsapi/gen_sc_data.go
b00ris: I thought about it. The idea was to store generated file to be able to compile binary without smart contracts.
Regarding updates - it'll require updating core-contracts branch. If we update it in 2 separated branches(which is strange), we'll have to resolve conflict in core-contracts commit also.

b00ris: I'd prefer to have it commited and update when we update core-contracts branch

consensus/polybft/contractsapi/init.go
b00ris: MRC20 - uses native token precompile it uses not only for testing. I'm afraid we'll have to support some testing contracts on our side. I suppose we can use Burn contracts on L1 for that purpose, but it's not implemented yet.
 The main reason why i decided to put it just in a variable - simplicity. I'd say that we need a separate issue about organising compilation of testing contracts.

go.mod
```diff
@@ -101,6 +101,7 @@ require (
 	github.com/containerd/console v1.0.1 // indirect
 	github.com/containerd/continuity v0.0.0-20191214063359-1097c8bae83b // indirect
 	github.com/coreos/go-systemd/v22 v22.3.2 // indirect
+	github.com/dave/jennifer v1.6.0 // indirect
```
ferranbt, line 104: Please add a JIRA ticket to replace this library.

ferranbt (approved):

vcastellm (commented):
Missing description about motivation for using Jenn code gen

PR #1025, opened 2022-12-09
Add scaling to get validator set

b00ris:
# Description
Fixes a bug that return unscaled stake

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


consensus/polybft/system_state.go
```diff
@@ -99,7 +101,7 @@ func (s *SystemStateImpl) GetValidatorSet() (AccountSet, error) {
 		val := &ValidatorMetadata{
 			Address:     types.Address(addr),
 			BlsKey:      pubKey,
-			VotingPower: stake.Uint64(),
+			VotingPower: chain.ConvertWeiToTokensAmount(stake).Uint64(),
```
Stefan-Ethernal: Are you sure that we need to scale stake here, because we scale it initially when sending initializing `ChildValidatorSet` (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/consensus/polybft/contracts_initializer.go#L62) and as I'm aware, SC doesn't scale anything on its side as it is basically agnostic about units (whether it is expressed in weis, tokens etc.)? 

Let's discuss this next week as a group, because we should finally agree upon strategy. Should we start using WEIs everywhere, except in proposer selection algorithm (before executing it, we should probably just scale it by dividing with 10**18)....

b00ris:

b00ris:
Closed in favor to other idea

PR #1028, opened 2022-12-09
disabling CGO for release builds

epikichi:
# Description

As discussed with @lazartravica , removing CGO to use native bindings.


JekaMas (commented):
Lgtm, although please check the data race detector requirements https://go.dev/doc/articles/race_detector

Issue #845, opened 2022-10-28
`Failed to extract the last proposer` due to IBFT extra unmarshalRLP error 


dankostiuk:
# `Failed to extract the last proposer` due to IBFT extra unmarshalRLP error 
## Description

We were able to run v0.6.0 on our testnet successfully but after a few days one of our validators seemed to get stuck while trying to sync up to a later block from another peer.

This seemed to correlate with a memory leaked observed at around the same time.

Debugging this issue further we found that when initializing the snapshots to catch up to the latest blockchain header, we encounter a header (10411966) which we attempt to derive the proposer from (`signer.EcrecoverFromHeader(header)`).  RLP unmarshaling (UnmarshalRpl in rlp_unmarshal.go) then fails with the message:
```
cannot parse RLP: cannot parse long array: cannot parse array value: cannot parse empty string
```

Restarting does not help (see logs below).

---

Current snapshots on the problematic validator:
```
[{"Number":10278105,"Hash":"0x6ae34c6fc0e94c06604c9eb388c5f2e81556840b5f26c4ad4de3078ad3839587","Votes":[],"Type":"ecdsa","Set":[{"Address":"0xA3a585c6d59CCE6aAe7035e8df48b3327cC8BE54"},{"Address":"0x7EFF16a34e3433182D636488bc97919b10283F37"},{"Address":"0x9cde4F11EA57163023A9a8a5dF499b27813639EB"},{"Address":"0x3e64F88C6C7a1310236B242180c0Ba1409d10F4d"}]}]
```

Current metadata on the problematic validator:
```
{"LastBlock":10283596}
```

Current snapshots on a healthy validator on the same testnet:
```
[{"Number":10283500,"Hash":"0x4d4bba3dddabdcc45ddfa4582763823ff33451c6876130ae72de9b89ea92ea27","Votes":[],"Type":"ecdsa","Set":[{"Address":"0xA3a585c6d59CCE6aAe7035e8df48b3327cC8BE54"},{"Address":"0x7EFF16a34e3433182D636488bc97919b10283F37"},{"Address":"0x9cde4F11EA57163023A9a8a5dF499b27813639EB"},{"Address":"0x3e64F88C6C7a1310236B242180c0Ba1409d10F4d"}]}]
```

Current metadata on a healthy validator on the same testnet:
```
{"LastBlock":10453018}
```

## Your environment
- OS and version. `Ubuntu 20`
- The version of the Polygon Edge.  `v0.6.0`
[...]

PR #1042, opened 2022-12-13
IBFT extra unmarshalRLP error

trimixlover:
# Description

Customers faced an [issue](https://github.com/0xPolygon/polygon-edge/issues/845)
The issue is non existent in versions prior to 0.6.x and is present in all 0.6.x

Upon further inspection notice three things:
- The snapshot is invalid (its missing a new validator that should have been included based on genesis)
- UnmarshalRLP on extra data error when trying to insert a block
- UnmarshalRLP on extra data error when trying to sync snapshot

I haven't been able to pinpoint the exact cause of the issue. This fix can help the code recover **prior** to `i.blockchain.WriteBlock` so it discards it instead of writing and syncs with other valid node.
This doesn't **fix** the code. It **mitigates** the issue. Motivation is not to alter the code if we are not yet sure the exact nature of the corruption.

Assumptions:
- It happens upon insertion of the block
- It happens when packing committed seals
- Write block catches the error after already writing the header
- Updating the snapshot after restart repeats this issue by trying to unmarshal corrupted header

To locate the issue we searched for new code that is introduced when inserting the block. Also it is important to note that while doing the consensus mechanism the code checks for the validity of extra data. So the next natural conclusion would be that something happens after the consensus in the finalisation.

This is a piece of code which alters the extra data and also doesn't handle errors when packing it in.
`header, err := i.currentSigner.WriteCommittedSeals(newBlock.Header, committedSealsMap)`
Unsuccessfully trying to replicate the corruption by forcing return of error, discards the possibility of the fault happening because of the return of the error but rather because of the faulty alteration in the `packFn`

The potentially corrupted extra data is then written into storage, but then tried to be unmarshaled when its too late in `b.consensus.ProcessHeaders([]*types.Header{header})` when updating the snapshot.

To mitigate the issue I introduced a safety net. It tries to unmarshal the extra data after alteration and before write. This way we force the node to drop the block in the early phase of insert and make it fall back to sync but still managing to log out the before and after hex. 

I chose to log these particular values because we know it is still valid before alteration. So it can be unmarshaled. Every piece of data that is needed to pack committed seals is within original extra data so we can replicate the steps locally. The corrupted hex is logged out so we can see if we are dealing with empty array or corrupted rlp.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -77,10 +77,20 @@ func (i *backendIBFT) InsertBlock(
 	// This is a safety net to help us narrow down and also recover before
 	// writing the block
 	if err := i.ValidateExtraDataFormat(newBlock.Header); err != nil {
+
+		//Format committed seals to make them more readable
+		committedSealsStr := make([]string, len(committedSealsMap))
+		for i, seal := range committedSeals {
+			committedSealsStr[i] = fmt.Sprintf("{signer=%v signature=%v}",
+				hex.EncodeToHex(seal.Signer),
+				hex.EncodeToHex(seal.Signature))
+		}
+
 		i.logger.Error("cannot write block: corrupted extra data",
 			"err", err,
-			"\n\tbefore", hex.EncodeToHex(extraDataBackup),
-			"\n\tafter", hex.EncodeToHex(header.ExtraData))
+			"before", hex.EncodeToHex(extraDataBackup),
+			"after", hex.EncodeToHex(header.ExtraData),
+			"committedSeals", committedSealsStr)
 
 		return
 	}
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -77,7 +77,6 @@ func (i *backendIBFT) InsertBlock(
 	// This is a safety net to help us narrow down and also recover before
 	// writing the block
 	if err := i.ValidateExtraDataFormat(newBlock.Header); err != nil {
-
 		//Format committed seals to make them more readable
 		committedSealsStr := make([]string, len(committedSealsMap))
 		for i, seal := range committedSeals {
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -9,6 +9,7 @@ import (
 	"github.com/0xPolygon/go-ibft/messages"
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
+	"github.com/0xPolygon/polygon-edge/helper/hex"
 	"github.com/0xPolygon/polygon-edge/state"
 	"github.com/0xPolygon/polygon-edge/types"
 )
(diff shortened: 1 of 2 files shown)


consensus/ibft/consensus_backend.go
```diff
@@ -64,6 +70,21 @@ func (i *backendIBFT) InsertBlock(
...
 	}
 
+	// WriteCommittedSeals alters the extra data before writing the block
+	// It doesn't handle errors while pushing changes which can result in
+	// corrupted extra data.
+	// We don't know exact circumstance of the unmarshalRLP error
+	// This is a safety net to help us narrow down and also recover before
+	// writing the block
+	if err := i.ValidateExtraDataFormat(newBlock.Header); err != nil {
+		i.logger.Error("cannot write block: corrupted extra data",
```
JekaMas, line 88: could we use fmt.Sprintf?

consensus/ibft/ibft.go
```diff
@@ -626,3 +626,15 @@ func verifyProposerSeal(
 
 	return nil
 }
+
+// ValidateExtraDataFormat Verifies that extra data can be unmarshaled
+func (i *backendIBFT) ValidateExtraDataFormat(header *types.Header) error {
+	blockSigner, _, _, err := getModulesFromForkManager(
```
JekaMas, line 632: shall we check this error here?

consensus/ibft/consensus_backend.go
ferranbt: This is the format of the [logger](https://github.com/hashicorp/go-hclog) that Edge uses.

ferranbt (approved):

consensus/ibft/consensus_backend.go
trimixlover: What I can do is rearrange the args for err to maybe be on the last position and more conform to what @ferranbt linked. If the reasoning for Sprintf is because of the conversion of byte array to hex, as far as i searched, its considerably slower than doing it by hand.

If this is only motivation can we maybe keep it as is?

@ferranbt please let me know if your comment is a remark to my code or to @JekaMas comment

consensus/ibft/ibft.go
trimixlover: Yes I feel like we actually should. Thank you

consensus/ibft/consensus_backend.go
JekaMas: I am more about "\n\t" - that kind of things. Could we either get rid of them or think about formatting?

JekaMas (approved):

consensus/ibft/consensus_backend.go
```diff
@@ -64,6 +70,21 @@ func (i *backendIBFT) InsertBlock(
...
+			"err", err,
+			"\n\tbefore", hex.EncodeToHex(extraDataBackup),
+			"\n\tafter", hex.EncodeToHex(header.ExtraData))
```
lazartravica: @trimixlover i suggest to print the committed seals as received by consensus here as well to have everything needed to debug.

lazartravica (changes requested):
1 requested change in what is being printed, otherwise LGTM.

consensus/ibft/consensus_backend.go
trimixlover: Yes we can. Will be removed.

lazartravica (approved):

igorcrevar (approved):

Nemanja0x (approved):

PR #1044, opened 2022-12-13
Implement the relayer

vcastellm:
# Description

Introducing the `Relayer` in this PR. Related to RFC-139, we're introducing a new process that will be run along the client when using `polybft` consensus, its job is to listen for [Commit events](https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/StateReceiver.sol#L50) (currently `NewBundleCommit`) in `StateReceiver` contract in L2, with the sole purpose of automatically executing the committed `StateSyncs`.
 
Steps as follow:

1. Receive a new event
2. Retrieve the `StateSync` proofs using the bridge endpoint `bridge_getStateSyncProof`
3. Send a transaction using the proof

# Changes include

# Checklist

## Testing

### Manual tests

# Additional comments

This will need update of SCs before working


consensus/polybft/polybft.go
```diff
@@ -108,6 +110,9 @@ type Polybft struct {
 
 	// tx pool as interface
 	txPool txPoolInterface
+
+	// relayer instance
+	relayer *relayer.Relayer
```
ferranbt: Ideally, the relayer is not part of `polybft` in the future.

vcastellm: Yeah, it's enough separated to just move it to one subcommand or a separate binary, but currently we're going to start it on `IsBridgeEnabled` inside polybft.

command/server/init.go
```diff
@@ -316,3 +317,9 @@ func (p *serverParams) initGRPCAddress() error {
 
 	return nil
 }
+
+func (p *serverParams) initRelayer() error {
```
Stefan-Ethernal: Inline this function, as it is basically a one liner (also we have error as return value, but always returning a `nil`)...?

relayer/event_tracker.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/umbracle/ethgo/jsonrpc"
+	"github.com/umbracle/ethgo/tracker"
+	boltdbStore "github.com/umbracle/ethgo/tracker/store/boltdb"
+)
+
+type eventSubscription interface {
+	AddLog(log *ethgo.Log)
+}
+
+type eventTracker struct {
```
Stefan-Ethernal: Since we now have 2 event trackers, which can be confusing from time to time, what do you think about calling this one `sidechainEventTracker` and the one from the `consensus/polybft` `rootchainEventTracker`?

Ideally we should merge these two implementations and subscriber should be able to determine is it `L2StateSync` event or `StateSync` (the one emitted from the rootchain). WDYT?

relayer/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+	client            *jsonrpc.Client
+	txRelayer         txrelayer.TxRelayer
+	key               ethgo.Key
+}
+
+func sanitizeRPCEndpoint(rpcEndpoint string) string {
+	if rpcEndpoint == "" || strings.Contains(rpcEndpoint, "0.0.0.0") {
+		_, port, err := net.SplitHostPort(rpcEndpoint)
+		if err == nil {
+			rpcEndpoint = fmt.Sprintf("http://%s:%s", "localhost", port)
```
Stefan-Ethernal: As far as I can remember, I've tried initializing `jsonrpc.Client` by providing `localhost` and it didn't work. Then I switched back to providing localhost ip address itself (i.e. 127.0.0.1). Did you come to a different conclusion?

relayer/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+		if sid, ok := vals["startId"].(big.Int); ok {
+			startID = sid.Uint64()
+		}
+
+		var endID uint64
+		if eid, ok := vals["endId"].(big.Int); ok {
+			endID = eid.Uint64()
+		}
+
+		r.logger.Info("Commit: Block %d StartID %d EndID %d\n", log.BlockNumber, startID, endID)
```
Stefan-Ethernal: Be mindful here, logger is expecting odd number of arguments in order to format it properly. So either use `fmt.Sprintf` and specify single formatted string or apply these changes:

```suggestion
		r.logger.Info("Commit", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
```

relayer/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+func (r *Relayer) queryStateSyncProof(stateSyncID string) (*types.StateSyncProof, error) {
+	// retrieve state sync proof
+	var stateSyncProof types.StateSyncProof
+
+	err := r.client.Call("bridge_getStateSyncProof", &stateSyncProof, stateSyncID)
+	if err != nil {
+		return nil, err
+	}
+
+	r.logger.Debug("state sync proof:", stateSyncProof)
```
Stefan-Ethernal: Logger expects odd number of arguments:

```suggestion
	r.logger.Debug(fmt.Sprintf("state sync proof: %v", stateSyncProof))
```

relayer/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+	// execute the state sync
+	txn := &ethgo.Transaction{
+		From:     r.key.Address(),
+		To:       (*ethgo.Address)(&contracts.StateReceiverContract),
+		GasPrice: 0,
+		Gas:      types.StateTransactionGasLimit,
+		Input:    input,
+	}
+
+	_, err = r.txRelayer.SendTransaction(txn, r.key)
```
Stefan-Ethernal: Would you like to check receipt status and return an error if its status is failed?

relayer/event_tracker.go
goran-ethernal: I agree with Stefan. Those two `event_trackers` could be merge to one implementation I think. The `subscriber` interface is the same one we use in `consensus_runtime`. The things that differ are the store name, address and endpoint, but that can be configurable.

command/server/init.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1044/commits/833e372b302c61d8de227238700e75b440cea9a4

relayer/relayer.go
vcastellm: I think I haven't checked it just assumed it should work, anyway we can go back to ip.

vcastellm: https://github.com/0xPolygon/polygon-edge/pull/1044/commits/e546e3f7b4455faa439de7e0d907d2b7cf0530b3

relayer/relayer.go
vcastellm: Nice one, done in https://github.com/0xPolygon/polygon-edge/pull/1044/commits/d083e8096a78e2777d4da633efee1d706d273973

relayer/event_tracker.go
vcastellm: I thought on this, but it require a refactor to extract the tracker from polybft, I agree with you in general, let me give it a try.

vcastellm: So far so good https://github.com/0xPolygon/polygon-edge/pull/1044/commits/94b365a259240eb43c601fbe38c9b3332e6191f3

relayer/relayer.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		var endID uint64
+		if eid, ok := vals["endId"].(big.Int); ok {
+			endID = eid.Uint64()
+		}
+
+		r.logger.Info("Commit", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
+
+		for i := startID; i < endID; i++ {
```
goran-ethernal: I think the for loop should go to `i <= endID` since `endID` is the `ID` of the last stateSync in commitment.

relayer/relayer.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	r.logger.Debug(fmt.Sprintf("state sync proof: %v", stateSyncProof))
+
+	return &stateSyncProof, nil
+}
+
+// executeStateSync executes the state sync
+func (r *Relayer) executeStateSync(stateSyncProof *types.StateSyncProof) error {
+	input, err := types.ExecuteStateSyncABIMethod.Encode(
```
goran-ethernal: You can replace this line with `stateSyncProof.EncodeAbi()` like in the previous PR.

server/server.go
```diff
@@ -283,6 +287,13 @@ func NewServer(config *Config) (*Server, error) {
 		return nil, err
 	}
 
+	// start relayer
+	if config.Relayer {
+		if err := m.setupRelayer(); err != nil {
```
ferranbt, line 292: Why is the relayer part of server? The relayer is an entity that in the current scope only lives inside `polybft`.

relayer/relayer.go
```diff
@@ -0,0 +1,166 @@
...
+		Gas:      types.StateTransactionGasLimit,
+		Input:    input,
+	}
+
+	receipt, err := r.txRelayer.SendTransaction(txn, r.key)
+	if err != nil {
+		return fmt.Errorf("failed to send state sync transaction: %w", err)
+	}
+
+	if receipt.Status == 0 {
```
goran-ethernal: Maybe consider replacing this with `if receipt.Status == uint64(ReceiptStatusFailed)`. To me it's more descriptive, and mitigates a chance to make a mistake.

relayer/relayer_test.go
```diff
@@ -0,0 +1,102 @@
...
+			"http://127.0.0.1:8545",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := sanitizeRPCEndpoint(tt.endpoint); got != tt.want {
+				t.Errorf("sanitizeRPCEndpoint() = %v, want %v", got, tt.want)
+			}
+		})
```
Stefan-Ethernal: Capture loop variable?
```suggestion
                tt := tt
		t.Run(tt.name, func(t *testing.T) {
			if got := sanitizeRPCEndpoint(tt.endpoint); got != tt.want {
				t.Errorf("sanitizeRPCEndpoint() = %v, want %v", got, tt.want)
			}
		})
```

relayer/relayer_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	txRelayer.On("SendTransaction", txn, key).Return(&ethgo.Receipt{}, nil)
+	r.executeStateSync(sp)
+
+	txRelayer.AssertExpectations(t)
+}
+
+// Test sanitizeRPCEndpoint
+func Test_sanitizeRPCEndpoint(t *testing.T) {
+	tests := []struct {
```
Stefan-Ethernal: Add t.Parallel()?

```suggestion
        t.Parallel()
	tests := []struct {
```

relayer/relayer_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (t *txRelayerMock) SendTransactionLocal(txn *ethgo.Transaction) (*ethgo.Receipt, error) {
+	args := t.Called(txn)
+
+	return nil, args.Error(1)
+}
+
+func Test_executeStateSync(t *testing.T) {
+	txRelayer := &txRelayerMock{}
```
Stefan-Ethernal: Parallelize execution of test?

tracker/event_tracker.go
```diff
@@ -15,15 +15,32 @@ type eventSubscription interface {
...
+	rpcEndpoint  string
+	contractAddr ethgo.Address
+	subscriber   eventSubscription
+	logger       hcf.Logger
 }
 
-func (e *eventTracker) start() error {
-	provider, err := jsonrpc.NewClient(e.config.Bridge.JSONRPCEndpoint)
+func NewEventTracker(
+	dataDir string,
```
Stefan-Ethernal: Is it intended to have different databases for different events being tracked? 

If that is the case, then we should parameterize name of db file:
`store, err := boltdbStore.New(filepath.Join(e.dataDir, "/deposit.db"))`

relayer/relayer.go
vcastellm: Indeed https://github.com/0xPolygon/polygon-edge/pull/1044/commits/4c6c016cfa246e1caac60a6f78f50cd79e716308

relayer/relayer.go
vcastellm: Yes we can! 🇺🇸 https://github.com/0xPolygon/polygon-edge/pull/1044/commits/4c6c016cfa246e1caac60a6f78f50cd79e716308

relayer/relayer.go
vcastellm: Didn't know it existed in types 👍 https://github.com/0xPolygon/polygon-edge/pull/1044/commits/4c6c016cfa246e1caac60a6f78f50cd79e716308

goran-ethernal: I think you made a mistake, you put the `ReceiptSuccess` in that condition.

relayer/relayer_test.go
vcastellm: Good one

relayer/relayer_test.go
vcastellm: Sure https://github.com/0xPolygon/polygon-edge/pull/1044/commits/90fa5454dd485d427f760d4478549faf087c772c

tracker/event_tracker.go
vcastellm: Yes, so this escaped the refactor https://github.com/0xPolygon/polygon-edge/pull/1044/commits/90fa5454dd485d427f760d4478549faf087c772c

relayer/relayer.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// create the JSON RPC client
+	client, err := jsonrpc.NewClient(endpoint)
+	if err != nil {
+		logger.Error("Failed to create the JSON RPC client", "err", err)
+
+		return nil
+	}
+
+	txRelayer, err := txrelayer.NewTxRelayer(txrelayer.WithClient(client))
```
b00ris: Could you add error log there?

relayer/relayer.go
```diff
@@ -0,0 +1,165 @@
...
+			endID = eid.Uint64()
+		}
+
+		r.logger.Info("Commit", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
+
+		for i := startID; i <= endID; i++ {
+			// query the state sync proof
+			stateSyncProof, err := r.queryStateSyncProof(strconv.Itoa(int(i)))
+			if err != nil {
+				r.logger.Error("Failed to query state sync proof", "err", err)
```
b00ris: Shouldnt we add continue there?

relayer/relayer.go
```diff
@@ -0,0 +1,165 @@
...
+func (r *Relayer) queryStateSyncProof(stateSyncID string) (*types.StateSyncProof, error) {
+	// retrieve state sync proof
+	var stateSyncProof types.StateSyncProof
+
+	err := r.client.Call("bridge_getStateSyncProof", &stateSyncProof, stateSyncID)
+	if err != nil {
+		return nil, err
+	}
+
+	r.logger.Debug(fmt.Sprintf("state sync proof: %v", stateSyncProof))
```
b00ris: Is it replacement for stateSyncProof.String()?

b00ris (approved):
Good job!

goran-ethernal (approved):
LGTM :100:

relayer/relayer.go
vcastellm: Do you prefer that we add the `.String()`?

relayer/relayer.go
vcastellm: 🦅 👁️ https://github.com/0xPolygon/polygon-edge/pull/1044/commits/3b6b5da9392e5895c8602434463edbd621220847

relayer/relayer.go
vcastellm: yes https://github.com/0xPolygon/polygon-edge/pull/1044/commits/3809503103b25708d9c9da547cf0b76bd5becd14

server/server.go
vcastellm: FTR: IMO we need the flag in server, so why don't have it properly confined to server instead of burying it in polybft?

tracker/event_tracker.go
```diff
@@ -15,34 +14,51 @@ type eventSubscription interface {
...
 
-	store, err := boltdbStore.New(filepath.Join(e.dataDir, "/deposit.db"))
+	store, err := boltdbStore.New(e.dbPath)
 
 	if err != nil {
 		return err
 	}
 
-	e.logger.Info("Start tracking events", "bridge", e.config.Bridge.BridgeAddr)
+	e.logger.Info("Start tracking events", "bridge", e.contractAddr)
```
Stefan-Ethernal:
```suggestion
	e.logger.Info("Start tracking events", "contract address", e.contractAddr)
```

server/server.go
Stefan-Ethernal: I'm leaning more towards Ferran's opinion, although it is not really black vs white situation here:
Relayer is polybft specific and is coupled with bridge workflow. So I'd say we should instantiate it in the polybft in case Bridge is enabled (although we can agree if we want to have additional relayer flag separately configured independently if validator is running in bridge mode or not). 

The only challenge I see currently is are we able to easily retrieve all the server configuration required by the relayer in the polybft itself (like data dir, json rpc address etc.).

relayer/relayer.go
```diff
@@ -0,0 +1,170 @@
...
+		if sid, ok := vals["startId"].(big.Int); ok {
+			startID = sid.Uint64()
+		}
+
+		var endID uint64
+		if eid, ok := vals["endId"].(big.Int); ok {
+			endID = eid.Uint64()
+		}
+
+		r.logger.Info("Commit", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
```
Stefan-Ethernal: Nitpick, completely optional, just making log slightly more verbose

```suggestion
		r.logger.Info("NewBundleCommit event captured", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
```

relayer/relayer.go
```diff
@@ -0,0 +1,170 @@
...
+
+	hcf "github.com/hashicorp/go-hclog"
+	"github.com/umbracle/ethgo"
+	"github.com/umbracle/ethgo/abi"
+	"github.com/umbracle/ethgo/jsonrpc"
+)
+
+var commitEvent = abi.MustNewEvent(`event NewBundleCommit(uint256 startId, uint256 endId, bytes32 root)`)
+
+type Relayer struct {
```
Stefan-Ethernal: Would `StateSyncExecutor` be semantically more adequate name?

Stefan-Ethernal: We have landed to `StateSyncRelayer` name eventually 🙂

vcastellm:
As agreed in extra convo with @ferranbt @goran-ethernal and @Stefan-Ethernal We're renaming `relayer` to `statesyncrelayer` and moving the package inside `polybft`, the motivation is that the new name is more semantically correct with the specific type of Relayer that it is, and as it only be usable for PolyBFT, the package won't be available in the root path of the project.

relayer/relayer.go
vcastellm: https://github.com/0xPolygon/polygon-edge/pull/1044#issuecomment-1363895539

relayer/relayer.go
vcastellm: https://github.com/0xPolygon/polygon-edge/pull/1044/commits/9cc552109e99770e8763a7b7aab95828655bc8f2

Stefan-Ethernal (approved):

server/server.go
vcastellm: Agreement FTR https://github.com/0xPolygon/polygon-edge/pull/1044#issuecomment-1363895539

PR #1047, opened 2022-12-13
Evm 303: Fix race conditions after a block is inserted

Nemanja0x:
# Description

To prevent race conditions after a block is inserted a lock and additional verification has been introduced into `OnBlockInserted` function:
1. The lock is added in order to guard the function execution (and put stop to invalid data modifications)  which can be called from both syncer and consensus during their simultaneous work.
2. Additional verification implies checking if a block was previously already inserted. If the block is already a part of the blockchain further execution of the function is skipped.

Furthermore,  a function named `getGuardedData` has been introduced in order to provide proper access to the data which is modified during `OnBlockInserted`. `getGuardedData` groupes the data in `guardedDataDTO struct` and provides it to any other entities like `FSM`.  `getGuardedData` uses the same lock as `OnBlockInserted`.

# Changes include

# Checklist

## Testing


consensus/polybft/consensus_runtime.go
```diff
@@ -1085,13 +1085,19 @@ func (c *consensusRuntime) BuildProposal(view *proto.View) []byte {
...
-	block, err := fsm.Insert(proposal, committedSeals)
+	alreadyInserted, block, err := fsm.Insert(proposal, committedSeals)
 	if err != nil {
 		c.logger.Error("cannot insert proposal", "error", err)
 
 		return
 	}
 
+	if alreadyInserted {
+		c.logger.Info("cannot insert proposal it was alredy previously inserted")
```
Stefan-Ethernal, line 1096:
```suggestion
		c.logger.Info("cannot insert proposal it was already previously inserted")
```

Stefan-Ethernal: I would completely remove this log, as we already have similar one in the [blockchain.WriteBlock](https://github.com/0xPolygon/polygon-edge/blob/develop/blockchain/blockchain.go#L884)

```suggestion
```

Nemanja0x: It is completely removed.

consensus/polybft/consensus_runtime.go
```diff
@@ -214,34 +221,52 @@ func (c *consensusRuntime) AddLog(eventLog *ethgo.Log) {
 
 // OnBlockInserted is called whenever fsm or syncer inserts new block
 func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
+	c.lock.Lock()
```
goran-ethernal, line 228: Could we just use the `getLastBuiltBlockAndEpoch` function to read the last built block and `epochMetadata`, and then just lock the lines where we set the, `c.epoch`, `c.lastBuiltBlock` and `c.proposerSnapshot`?

consensus/polybft/consensus_runtime.go
```diff
@@ -417,81 +440,68 @@ func (c *consensusRuntime) FSM() error {
 }
 
 // restartEpoch resets the previously run epoch and moves to the next one
-func (c *consensusRuntime) restartEpoch(header *types.Header) error {
+// returns *epochMetadata different from nil if the lastEpoch is not the current one and everything was successful
+func (c *consensusRuntime) restartEpoch(header *types.Header, lastEpoch *epochMetadata) (*epochMetadata, error) {
```
goran-ethernal: I would say we don't need this `lastEpoch` parameter, since it's already in the runtime object, and the function is under lock.

consensus/polybft/consensus_runtime.go
```diff
@@ -417,81 +440,68 @@ func (c *consensusRuntime) FSM() error {
...
 		"restartEpoch",
 		"block number", header.Number,
 		"epoch", epochNumber,
 		"validators", validatorSet.Len(),
 	)
 
-	return nil
+	return &epochMetadata{
+		Number:         epochNumber,
+		LastCheckpoint: 0,
```
goran-ethernal: You don't need to set this `LastCheckpoint` and `Blocks` fields, they are removed throuygh another PR, plus, they are not used anywhere.

consensus/polybft/consensus_runtime.go
igorcrevar: No, we need everything to be executed in exactly one transaction. Everything must go under the lock :(

consensus/polybft/consensus_runtime.go
Nemanja0x: fixed https://github.com/0xPolygon/polygon-edge/pull/1047/commits/7a42499fef66c5a3aeae3a49869b75595f958a44

consensus/polybft/consensus_runtime.go
Nemanja0x: fixed https://github.com/0xPolygon/polygon-edge/pull/1047/commits/7a42499fef66c5a3aeae3a49869b75595f958a44

consensus/polybft/consensus_runtime.go
```diff
@@ -222,26 +239,35 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 		c.logger.Error("on block inserted error", "err", err)
 	}
 
+	var (
+		proposerSnapshot = c.proposerSnapshot.Copy()
```
Stefan-Ethernal: Move this line above this one, because otherwise we would copy and assign proposerSnapshot even when error occurs in epoch-ending blocks but it is not required to have it in that use case.
`if err := c.updateProposerSnapshot(proposerSnapshot, block.Number()); err != nil {`

consensus/polybft/consensus_runtime.go
```diff
@@ -222,26 +239,35 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
...
 	}
+
+	// finally update runtime state (lastBuiltBlock, epoch, proposerSnapshot)
+	c.epoch = epoch
+	c.lastBuiltBlock = block.Header
+	c.proposerSnapshot = proposerSnapshot
 }
 
 func (c *consensusRuntime) createCommitmentAndBundles(txs []*types.Transaction) error {
+	parentBlock, epoch := c.lastBuiltBlock, c.epoch
```
Stefan-Ethernal, line 274: Is this safe to do this way, retreiving a shallow copy, instead of acquiring a copy via `getLastBuiltBlockAndEpoch`?

Btw it is not entirely clear to me when we should return a copy and call  `getLastBuiltBlockAndEpoch` and when it is safe just doing shallow copy without even locking it.

Stefan-Ethernal (approved):

consensus/polybft/consensus_runtime.go
igorcrevar: this method is called under the lock. It would be better if this method receives all the necessary parameters and not pick them from the object state, but for now we gonna leave them like this

goran-ethernal (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -405,81 +423,73 @@ func (c *consensusRuntime) FSM() error {
...
 		if lastEpoch.Number == epochNumber {
-			return nil
+			return lastEpoch, nil
 		}
 	}
 
 	validatorSet, err := c.config.polybftBackend.GetValidators(header.Number, nil)
 	if err != nil {
-		return err
+		return nil, err
```
0xSasaPrsic: can we wrap this error with a comment to have better insights of the error?
Something similar to **"cannot get validators: %w, err"**

0xSasaPrsic (approved):
Left you minor comment, otherwise looks good.

consensus/polybft/consensus_runtime.go
```diff
@@ -156,15 +163,16 @@ func (c *consensusRuntime) getEpoch() *epochMetadata {
...
 
-// getLastBuiltBlockAndEpoch returns last build block and current epochMetadata in a thread-safe manner.
-func (c *consensusRuntime) getLastBuiltBlockAndEpoch() (*types.Header, *epochMetadata) {
+// getSyncData returns last build block, proposer snapshot and current epochMetadata in a thread-safe manner.
+func (c *consensusRuntime) getSyncData() (*types.Header, *epochMetadata, *ProposerSnapshot) {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
 	lastBuiltBlock, epoch := new(types.Header), new(epochMetadata)
 	*lastBuiltBlock, *epoch = *c.lastBuiltBlock, *c.epoch
```
b00ris: it's not safe copy. Could you use c.lastBuiltBlock.Copy?

b00ris (approved):

b00ris (commented):
Good job!

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1047/commits/70d5daed07acc17e48bd989c49a2921c8dc76d2a

consensus/polybft/consensus_runtime.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1047/commits/70d5daed07acc17e48bd989c49a2921c8dc76d2a

consensus/polybft/proposer_calculator.go
```diff
@@ -45,7 +45,6 @@ type ProposerSnapshot struct {
 	Round      uint64
 	Proposer   *PrioritizedValidator
 	Validators []*PrioritizedValidator
-	//logger     hclog.Logger
```
0xSasaPrsic, line 48: can this be removed?

consensus/polybft/consensus_runtime.go
```diff
@@ -67,6 +67,17 @@ type epochMetadata struct {
 	Commitment *Commitment
 }
 
+type guardedDataDTO struct {
```
0xSasaPrsic, line 70: can we have better naming for this, it seems like a generic name?

Nemanja0x: It has already been  merged, we can change it via another PR.

Stefan-Ethernal: > can we have better naming for this, it seems like a generic name?

:+1: `runtimeSharedData` or something similar...

PR #1051, opened 2022-12-14
Add polybft consensus support to edge

epikichi:
# Description
* Adding PolyBFT Consensus option to Docker compose
* Adding `--docker` modes to the `scripts/cluster` script
* Adding `README` to `docker` and `scripts` folder
* Changed `log-level` to `debug` for all containers

## How to use
* `scripts/cluster ibft --docker`  - run docker environment with `ibft` consensus
* `scripts/cluster polybft --docker` run docker environment with `polybft` consensus
* `scripts/cluster (ibft or polybft) --docker destroy` - destroy docker environment (volumes will also be deleted)
* `scripts/cluster (ibft or polybft) --docker stop` - stop docker environment


Makefile
```diff
@@ -58,12 +58,29 @@ test-e2e-polybft:
...
+	EDGE_CONSENSUS=polybft docker-compose -f ./docker/local/docker-compose.yml up -d --build
+
+.PHONY: stop-local-polybft
+stop-local-polybft:
+	EDGE_CONSENSUS=polybft docker-compose -f ./docker/local/docker-compose.yml stop
+
+.PHONY: destroy-local-polybft
+destroy-local-polybft:
+	EDGE_CONSENSUS=polybft docker-compose -f ./docker/local/docker-compose.yml down -v
+
```
vcastellm: IMO Makefile is not the place to manage the lifecycle of local env, it's not building any aspect of the binary, this belongs to a `docker` script in `scripts` folder

vcastellm (changes requested):

docker/README.md
```diff
@@ -0,0 +1,48 @@
+# Docker deployment
```
vcastellm: Even though this is a nice README it's hidden in this directory, could be of great help for users if documented in Edge docs instead of here https://wiki.polygon.technology/docs/edge/get-started/set-up-ibft-locally, wdyt?

Makefile
ZeljkoBenovic: This local env var determines which consensus mechanism is to be used. It acts as a kind of a switch statement.   
The `docker-compose` file passes it to the `Dockerfile` as a second argument, which is used by the `entrypoint` script to create genesis for a specific consensus.
It must be set before the `init` container spins up, as it needs to know which consensus is being used.

The alternative to this approach is to have separate files for each consensus.    
We would than need to have dedicated `docker-compose`, `Dockerfile` and `entrypoint` script for each consensus.

IMHO, this is much simpler approach, as we're using the same files and the users can easily make consensus selection just by setting `EDGE_CONSENSUS` env var.

But if you disagree with this, we can easily split this logic.

docker/README.md
ZeljkoBenovic: Great idea, and the users would definitely benefit from having something similar to this `README` in the public docs.    
We should probably do that, once `feature/v3-parity` is merged into develop and we have a release including this feature.    

I don't think that we have any information regarding `polybft` in our public docs yet.

consensus/polybft/polybft.go
```diff
@@ -131,7 +131,7 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 		}
 
 		if err = initContract(contracts.ValidatorSetContract, input, "ChildValidatorSet", transition); err != nil {
-			return err
+			return fmt.Errorf("ChildValidatorSet SC call failed! error: %w", err)
```
Stefan-Ethernal: Nitpick: I would consider slightly changing this error message. Also similar should be done for `NativeTokenContract`. https://github.com/0xPolygon/polygon-edge/blob/df4e5dae55b2f32c6d7080d43f0c469b3248907d/consensus/polybft/polybft.go#L143

```suggestion
			return fmt.Errorf("ChildValidatorSet SC initialization failed: %w", err)
```

docker/local/polygon-edge.sh
```diff
@@ -3,35 +3,42 @@
...
+              "$POLYGON_EDGE_BIN" genesis \
+                --dir /genesis/genesis.json \
+                --consensus ibft \
+                --ibft-validators-prefix-path data- \
+                --bootnode /dns4/node-1/tcp/1478/p2p/$(echo $secrets | jq -r '.[0] | .node_id') \
+                --bootnode /dns4/node-2/tcp/1478/p2p/$(echo $secrets | jq -r '.[1] | .node_id')
+              echo "Genesis file has been successfully generated"
+              ;;
+          "polybft")
+              echo "Generating PolyBFT secrets..."
```
Stefan-Ethernal, line 34: Workflow has slightly changed with #1005...

Now as a prerequisite for  running a `genesis` command is `manifest` command. Manifest command is polybft specific and it creates json with validator public information as well as rootchain configuration (in case we are running polybft in bridge mode).

For additional information, you can either check `scripts/cluster` (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/scripts/cluster#L22) script and [README](https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/consensus/polybft/README.md) or ping me. :slightly_smiling_face:

command/genesis/utils.go
```diff
@@ -98,43 +105,61 @@ func parsePremineInfo(premineInfoRaw string) (*premineInfo, error) {
...
+		match, _ := regexp.MatchString(`^.*\D-\d{1}$`, strings.TrimSpace(path))
+		if match {
+			v.paths = append(v.paths, path)
+		}
+	}
+
+	return nil
+}
+
+// ReadValidatorsByRegexp trys to find validator secrets when local secrets manager is used.
```
Stefan-Ethernal:
```suggestion
// ReadValidatorsByRegexp tries to find validator secrets when local secrets manager is used.
```

command/genesis/utils.go
```diff
@@ -98,43 +105,61 @@ func parsePremineInfo(premineInfoRaw string) (*premineInfo, error) {
...
+		return nil, fmt.Errorf("could not find validator on the designated path error=%w", err)
+	}
 
 	// we must sort files by number after the prefix not by name string
-	sort.Slice(files, func(i, j int) bool {
-		f := strings.TrimPrefix(files[i].Name(), prefix)
-		s := strings.TrimPrefix(files[j].Name(), prefix)
-		num1, _ := strconv.Atoi(strings.TrimLeft(f, "-"))
-		num2, _ := strconv.Atoi(strings.TrimLeft(s, "-"))
+	// the last character is always a digit
```
Stefan-Ethernal: Hm not sure is it safe to assume this will always be the case?
Especially now, when we have make possible that validator secrets are located on the same hierarchy levels, which means that directory names don't need to be unique.

command/genesis/utils.go
```diff
@@ -98,43 +105,61 @@ func parsePremineInfo(premineInfoRaw string) (*premineInfo, error) {
...
 
-	validCnt := 0
-
-	for _, file := range files {
-		if file.IsDir() && strings.HasPrefix(file.Name(), prefix) {
-			files[validCnt] = file
-			validCnt++
-		}
+	// check the root folder of genesis file for validator data
+	if err := filepath.Walk(dir, validatorPaths.matchValidatorDataDirs); err != nil {
```
Stefan-Ethernal: Just thinking out loud: is it a requirement to have validator secrets placed within different root folder(s)? Appreciate the effort, but not sure is it reasonable to make logic more complex to support this use case (especially if it is not an essential one)?

command/genesis/utils.go
```diff
@@ -98,43 +105,61 @@ func parsePremineInfo(premineInfoRaw string) (*premineInfo, error) {
...
 }
 
+// matchValidatorDataDirs uses regexp to match validator directories which must have the last
+// three digits: non-digit, hyphen(-), digit.
+// Found paths are appended to validatorDataPaths.paths
+func (v *validatorDataPaths) matchValidatorDataDirs(path string, info fs.FileInfo, _ error) error {
+	if info.IsDir() && strings.HasPrefix(info.Name(), filepath.Base(v.pathPrefix)) {
+		// match paths that have a number as the last digit /tmp/data-1, /tmp/test-data-2, etc.
+		// matches any string that has last 3 chars: not-digit, - , digit
+		match, _ := regexp.MatchString(`^.*\D-\d{1}$`, strings.TrimSpace(path))
```
Stefan-Ethernal: Two things:
- `^.*\D-\d{1}$` with this regexp, I haven't managed to capture e.g.test-chain, so my question is should we even match non-digit here, since seems the result is the same as running this regexp `^.*-\d{1}$`? Can you give a counter-example?
- consider putting this regexp as const?

Stefan-Ethernal (commented):
Tnx for effort. Leaving some comments and doubts to consider.

consensus/polybft/polybft.go
ZeljkoBenovic: Reworked the docker scripts, so that these code changes are not needed. 
Reverted all the code changes, so now only docker related files are changed.

Makefile
ZeljkoBenovic: @vcastellm updated readme file with the instruction on how to run with `make` or directly with `docker-compose`.
Kept the same logic, as we're using existing docker files, but just switching the consensus with `EDGE_CONSENSUS` env var. This keeps the scripts simple and lean.     
Also added in some customization options for the chain.

docker/local/polygon-edge.sh
ZeljkoBenovic: Reworked scripts to include the latest changes to the workflow.
It works as intended now.

Stefan-Ethernal (approved):
LGTM :rocket:

Makefile
vcastellm: My proposal is to remove all the `run-local` stuff out from Makefile to a `docker` script in `scripts`, the fact that it's driven by env variables it's not related to using Makefile, it can be also improved by adding an option to the `cluster` script for running under docker, `scripts/cluster polybft --docker` should be the UI

vcastellm (dismissed):
See https://github.com/0xPolygon/polygon-edge/pull/1051#discussion_r1056125288

vcastellm (commented):
I don't want to block this for just the scripts thing, LGTM if we agree as a team that Makefile is the place

Makefile
ZeljkoBenovic: @vcastellm I see now that we already have dedicated `scripts` folder. Edited the `cluster` script to accommodate `ibft` and `polybft` docker environments.     
Let me know if this works for you and the rest of the team.

Will change the PR description, but there are `README.md` files in both `docker` and `scripts` folder with some explanation how to use the script.

vcastellm (approved):
LGTM

PR #1052, opened 2022-12-15
Pending checkpoints submission E2E test

Stefan-Ethernal:
# Description

This PR updates `TestE2E_CheckpointSubmission` so that it stops and restarts rootchain server and makes sure that pending checkpoints are sent successfully.

# Changes include

# Checklist

## Testing


b00ris (approved):

goran-ethernal (approved):
It would be better if we had an event from smart contract that checkpoint is submitted, and we would just check the logs (we would be able to check if every checkpoint from block 5 to block 20 was submitted), but, this will do. :100:

PR #1053, opened 2022-12-15
Fix loop closure violations in unit tests

Stefan-Ethernal:
# Description

This PR fixes:
- loop closure violations in polybft unit tests and 
- makes `TestExtra_CreateValidatorSetDelta_Cases` runs deterministically (by fixing randomization of voting power updates)

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/extra_test.go
+++ b/consensus/polybft/extra_test.go
@@ -369,46 +369,47 @@ func TestExtra_CreateValidatorSetDelta_Cases(t *testing.T) {
 	}
 
 	for _, c := range cases {
-		t.Run(c.name, func(t *testing.T) {
+		localCase := c
+		t.Run(localCase.name, func(t *testing.T) {
 			t.Parallel()
 
 			vals := newTestValidatorsWithAliases([]string{})
 
-			for _, name := range c.oldSet {
+			for _, name := range localCase.oldSet {
 				vals.create(name, 1)
 			}
-			for _, name := range c.newSet {
+			for _, name := range localCase.newSet {
 				vals.create(name, 1)
 			}
 
-			oldValidatorSet := vals.getPublicIdentities(c.oldSet...)
+			oldValidatorSet := vals.getPublicIdentities(localCase.oldSet...)
 			// update voting power to random value
 			maxVotingPower := big.NewInt(100)
-			for _, name := range c.updated {
+			for _, name := range localCase.updated {
 				v := vals.getValidator(name)
 				vp, err := rand.Int(rand.Reader, maxVotingPower)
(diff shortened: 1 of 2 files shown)


consensus/polybft/extra_test.go
```diff
@@ -369,46 +369,52 @@ func TestExtra_CreateValidatorSetDelta_Cases(t *testing.T) {
 	}
 
 	for _, c := range cases {
-		t.Run(c.name, func(t *testing.T) {
+		localCase := c
```
ferranbt: Is this an issue because the tests run in Parallel?

Stefan-Ethernal: Yes and I fixed it because VSC reported violation (although in this case we have only single test case to run). However without having `t.Parallel()` I'd say sub-tests would be executed in parallel because of `t.Run` which spawns a go routine.

vcastellm (approved):
A bit weird because we only have 1 test case but LGTM if it fixes the issue

Stefan-Ethernal:
> A bit weird because we only have 1 test case but LGTM if it fixes the issue

Let me give you a bit more context, because this PR fixes two things:
1. VS Code reported warning for loop closures and it perfectly makes sense, especially for `Test_buildLogsFromReceipts`, because there we have multiple test cases and tests weren't executed correctly without capturing a loop variable (although it is not easily noticeable, without adding logs to the test, because both test cases are running succesfully). For more info, read [this](https://www.gopherguides.com/articles/table-driven-testing-in-parallel#:~:text=gopherguides/training%20%20%20%20%20%20%20%200.156s-,Bug%20Time!,-How%20closely%20did).
On the other hand, it doesn't make any difference at the moment fixing it for `TestExtra_CreateValidatorSetDelta_Cases`, since currently there is only single sub test. Though we could easily decide to expand it in the future and we have landmine there and it is good to address it right now.
2. `TestExtra_CreateValidatorSetDelta_Cases` was flaky, because of assigning same random value for voting power as the old one (which is 1). That part was fixed here: https://github.com/0xPolygon/polygon-edge/pull/1053/files#diff-2f601f47fe8d02bd0d03ad2bbb4c245778e316affb10eb5497b9c4d1b58d11d1R393

consensus/polybft/extra_test.go
Stefan-Ethernal: @ferranbt more detailed explanation [here](https://github.com/0xPolygon/polygon-edge/pull/1053#issuecomment-1353257905)

goran-ethernal (approved):

b00ris (approved):

PR #1054, opened 2022-12-16
Fix segmentation fault panic on `ValidateSender`

Stefan-Ethernal:
# Description
This PR fixes segmentation fault in case IBFT message arrives before `FSM` is being initialized, which can happen as IBFT messages delivery and initialization of `FSM` object happen in a separate go routines.

Image below explains what are the execution paths for initialization of `FSM` object vs IBFT message delivery in order to get better understanding of the problem itself.
image

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -1028,18 +1028,34 @@ func (c *consensusRuntime) IsValidBlock(proposal []byte) bool {
 }
 
 func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
+	msgNoSig, err := msg.PayloadNoSig()
+	if err != nil {
+		c.logger.Error("IsValidSender failed", "error", err)
 
-	// IsValidSender is called whenever a message is received from the peer.
-	// Since it is triggered from a separate go routine, it can happen that fsm instance is not yet initialized.
-	if c.fsm == nil {
 		return false
 	}
 
-	err := c.fsm.ValidateSender(msg)
+	// recover ECDSA address
+	signerAddress, err := wallet.RecoverAddressFromSignature(msg.Signature, msgNoSig)
 	if err != nil {
-		c.logger.Error("invalid sender", "error", err)
+		c.logger.Error("IsValidSender failed", "error", fmt.Errorf("failed to recover address from message: %w", err))
+
+		return false
+	}
+
+	// verify that message signature and sender address match
+	if !bytes.Equal(msg.From, signerAddress.Bytes()) {
+		c.logger.Error("IsValidSender failed", "error",
+			fmt.Errorf("signer address %s doesn't match From field", signerAddress.String()))
+
+		return false
+	}
+
+	epoch := c.getEpoch()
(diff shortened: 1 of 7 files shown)


goran-ethernal:
What do you think about moving this `ValidateSender` to `consensus_runtime`? It only requires the validator set from `fsm` and that is gotten from the `epochMetadata.validators` field. That way, we won't have this check, nor race conditions, nor a possibility that something is `nil`.

goran-ethernal (approved):
Yep, this is looking better. :100:

consensus/polybft/polybft.go
```diff
@@ -220,6 +216,11 @@ func (p *Polybft) Start() error {
 		return fmt.Errorf("consensus runtime start - restart epoch failed: %w", err)
 	}
 
+	// subscribe to consensus engine messages
+	if err := p.subscribeToIbftTopic(); err != nil {
+		return fmt.Errorf("topic subscription failed: %w", err)
+	}
```
Stefan-Ethernal: This change is the most important here:
prevent IBFT messages receiving before we are able to process it (consensus runtime, epoch metadata etc. is fully initialized)

consensus/polybft/consensus_runtime.go
```diff
@@ -991,12 +991,34 @@ func (c *consensusRuntime) IsValidBlock(proposal []byte) bool {
 }
 
 func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
```
Stefan-Ethernal, line 993: We can discuss whether sender validation belongs to the `consensusRuntime` or the `FSM`. This suggestion came from @goran-ethernal and I agree, because `consensusRuntime` already has enough knowledge to validate sender on its end. No need to have indirection to the `FSM`

consensus/polybft/polybft.go
Stefan-Ethernal: Update:
This is most likely not necessary when #1047 gets merged, because it calls `restartEpoch` when creating a `consensusRuntime` instance (https://github.com/0xPolygon/polygon-edge/pull/1047/files#diff-e5b2e93010ff84c3777aef136e0834d710dd22861ca62ffbab9d5baa12a52b3cR145)

vcastellm (approved):
LGTM

consensus/polybft/consensus_runtime.go
```diff
@@ -1031,9 +1031,12 @@ func (c *consensusRuntime) IsValidSender(msg *proto.Message) bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
-	err := c.fsm.ValidateSender(msg)
-	if err != nil {
-		c.logger.Error("invalid sender", "error", err)
+	if c.fsm == nil {
+		return false
```
b00ris, line 1037: Could you add a log there?

b00ris (approved):

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1054/commits/66ced16002fd71851da727dd3a734e9ef76a02b4

PR #1055, opened 2022-12-16
Polybft data storages

ferranbt:
# Description

This PR implements the `datastore` RFC. Note that the added/deleted lines are a lot but it is only because of splitting `state.go` in the different stores. Functionality has not changed.

# Changes include

# Checklist

## Testing


vcastellm (commented):
this https://github.com/0xPolygon/polygon-edge/pull/1055/files#diff-514a0d26a7bee06b50bd0d4d72613aa918eb7ae096be6ea973ea6e2680d4dd75R210-R225 can be moved to its corresponding Datastore

consensus/polybft/state_store_epoch.go
```diff
@@ -0,0 +1,159 @@
...
+// cleanValidatorSnapshotsFromDB cleans the validator snapshots bucket if a limit is reached,
+// but it leaves the latest (n) number of snapshots
+func (s *EpochStore) cleanValidatorSnapshotsFromDB(epoch uint64) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(validatorSnapshotsBucket)
+
+		// paired list
+		keys := make([][]byte, 0)
+		values := make([][]byte, 0)
+		if numberOfSnapshotsToLeaveInDB > 0 { // TODO this is always true?!
```
Stefan-Ethernal: I'm aware that you don't like to change not explicitly related to the PR itself, but it is a good opportunity to remove this unnecessary condition :slightly_smiling_face:

Stefan-Ethernal (approved):

vcastellm (changes requested):
FTR Move the vars for each datastore to the corresponding file

```

var (
	// ABI
	stateTransferEventABI = abi.MustNewEvent("event StateSynced(uint256 indexed id, address indexed sender, address indexed receiver, bytes data)")   //nolint:lll
	exitEventABI          = abi.MustNewEvent("event L2StateSynced(uint256 indexed id, address indexed sender, address indexed receiver, bytes data)") //nolint:lll
	ExitEventABIType      = abi.MustNewType("tuple(uint256 id, address sender, address receiver, bytes data)")
)
const (
	// validatorSnapshotLimit defines a maximum number of validator snapshots
	// that can be stored in cache (both memory and db)
	validatorSnapshotLimit = 100
	// numberOfSnapshotsToLeaveInMemory defines a number of validator snapshots to leave in memory
	numberOfSnapshotsToLeaveInMemory = 2
	// numberOfSnapshotsToLeaveInMemory defines a number of validator snapshots to leave in db
	numberOfSnapshotsToLeaveInDB = 10
	// number of stateSyncEvents to be processed before a commitment message can be created and gossiped
	stateSyncMainBundleSize = 10
	// number of stateSyncEvents to be grouped into one StateTransaction
	stateSyncBundleSize = 1
)
```

consensus/polybft/state.go
```diff
@@ -1,152 +1,20 @@
...
-exit events/
-|--> (id+epoch+blockNumber) -> *ExitEvent (json marshalled)
-
-proposer snapshot/
-|--> staticKey - only current one snapshot is preserved -> *ProposerSnapshot (json marshalled)
-*/
-
 var (
 	// ABI
 	stateTransferEventABI = contractsapi.StateSender.Abi.Events["StateSynced"]
```
goran-ethernal: I would move this to StateSync store.

consensus/polybft/state.go
```diff
@@ -1,152 +1,20 @@
...
 	exitEventABI          = contractsapi.L2StateSender.Abi.Events["L2StateSynced"]
```
goran-ethernal: I would move these exit event related variables to checkpoint store.

consensus/polybft/state.go
```diff
@@ -183,48 +51,25 @@ type TransportMessage struct {
...
-	// bucket to store epochs and all its nested buckets (message votes and message pool events)
-	epochsBucket = []byte("epochs")
-	// bucket to store message votes (signatures)
-	messageVotesBucket = []byte("votes")
-	// bucket to store validator snapshots
-	validatorSnapshotsBucket = []byte("validatorSnapshots")
-	// bucket to store proposer calculator snapshot
-	proposerCalcSnapshotBucket = []byte("proposerCalculatorSnapshot")
 	// array of all parent buckets
 	parentBuckets = [][]byte{syncStateEventsBucket, exitEventsBucket, commitmentsBucket, stateSyncProofsBucket,
```
goran-ethernal: We could move bucket creation to each store, and let it create its own bucket. It would mean, every store would create a db transaction on node startup, but we would have a more separated code. What do you think?

consensus/polybft/state.go
```diff
@@ -260,528 +109,6 @@ func initMainDBBuckets(db *bolt.DB) error {
...
-		if err != nil {
-			return err
-		}
-
-		return nil
-	})
-}
-
 // epochsDBStats returns stats of epochs bucket in db
 func (s *State) epochsDBStats() *bolt.BucketStats {
```
goran-ethernal: You can move this to Epoch store.

consensus/polybft/state_test.go
```diff
@@ -1,23 +1,13 @@
...
-	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
-	"github.com/0xPolygon/polygon-edge/types"
 	"github.com/hashicorp/go-hclog"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	"github.com/umbracle/ethgo"
-	"github.com/umbracle/ethgo/abi"
 )
 
 func newTestState(t *testing.T) *State {
```
goran-ethernal: You can move this function to `helpers_test.go` and remove the `state_test.go` completely.

goran-ethernal (changes requested):
I think we should move these data stores, as well as `state.go` to a separate data layer package (`polybft/state`)

consensus/polybft/state_test.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1055/commits/744c97ae877890d400757a57cdfc3b71611ae49e

consensus/polybft/state.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1055/commits/744c97ae877890d400757a57cdfc3b71611ae49e

consensus/polybft/state.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1055/commits/744c97ae877890d400757a57cdfc3b71611ae49e. Hope the proposed solution makes sense (motivation was to run all buckets creation in a single transaction).

consensus/polybft/state.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1055/commits/744c97ae877890d400757a57cdfc3b71611ae49e

consensus/polybft/state.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1055/commits/744c97ae877890d400757a57cdfc3b71611ae49e

Stefan-Ethernal:
> I think we should move these data stores, as well as `state.go` to a separate data layer package (`polybft/state`)

Agree. However, this deserves a follow-up PR (already started working on it), because it is not as simple as moving storage files into separate packages, but also involves updating caller code as well.

goran-ethernal (approved):
Looks good :100: :rocket: .

igorcrevar (approved):

vcastellm (approved):
LGTM

PR #1058, opened 2022-12-18
Consolidate `ExtendByteSlice` implementations

Stefan-Ethernal:
# Description

This PR merges all the implementations of `extendByteSlice` helper functions into a single one, since they are the same.

# Changes include

# Checklist

## Testing

### DIFF
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -175,13 +175,20 @@ func PadLeftOrTrim(bb []byte, size int) []byte {
 	return tmp
 }
 
-// ExtendByteSlice extends given byte slice by needLen parameter.
-// It puts hard limit on length and capacity to needLen value.
-func ExtendByteSlice(b []byte, needLen int) []byte {
-	b = b[:cap(b)]
-	if n := needLen - len(b); n > 0 {
+// ExtendByteSlice extends given byte slice by needLength parameter.
+// If trim set to true, it puts hard limit on length and capacity to needLength value.
+func ExtendByteSlice(b []byte, needLength int, trim bool) []byte {
+	if trim {
+		b = b[:cap(b)]
+	}
+
+	if n := needLength - len(b); n > 0 {
 		b = append(b, make([]byte, n)...)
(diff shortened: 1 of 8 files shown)


helper/common/common.go
```diff
@@ -174,3 +174,21 @@ func PadLeftOrTrim(bb []byte, size int) []byte {
 
 	return tmp
 }
+
+// ExtendByteSlice extends given byte slice by needLength parameter.
+// If trim set to true, it puts hard limit on length and capacity to needLength value.
+func ExtendByteSlice(b []byte, needLength int, trim bool) []byte {
```
ferranbt: I would prefer that we do two functions for the extension instead, one for the trim and one without or even keep the one of the bitmap as it is. The reason is that `extendByteSlice` (as it is) is a useful function for Go high-performance programming and I do not want to have to worry about misusage.

ferranbt (changes requested):

helper/common/common.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1058/commits/f4fe15458c48131f939419d67a60747946d15d4e

helper/common/common.go
```diff
@@ -174,3 +174,14 @@ func PadLeftOrTrim(bb []byte, size int) []byte {
 
 	return tmp
 }
+
+// ExtendByteSlice extends given byte slice by needLength parameter and trims it
+func ExtendByteSlice(b []byte, needLength int) []byte {
```
ferranbt, line 179: Could you add a couple of unit tests for the function?

ferranbt (approved):

helper/common/common.go
Stefan-Ethernal: :heavy_check_mark: https://github.com/0xPolygon/polygon-edge/pull/1058/commits/b776177b72378fdc6a2497a769b4f6c4fafd1058

PR #1060, opened 2022-12-19
Remove voting power scaling

Stefan-Ethernal:
# Description

This PR removes scaling of voting power and uses genesis balances as is instead for smart contracts initialization and proposer selection algorithm. It means that instead of relying on `uint64`/`int64` data types for voting power and proposer priority attributes, those are now represented as `big.Int` in order to be able to hold large numbers.
Since proposer selection relies on `big.Int`, hard limit for max total voting power is removed as well.

# Changes include

# Checklist

## Testing


consensus/polybft/validator_set.go
```diff
@@ -96,6 +90,10 @@ func (vs validatorSet) Len() int {
...
 }
 
-func (vs validatorSet) getQuorumSize() int64 {
-	return int64(math.Ceil((2 * float64(vs.totalVotingPower)) / 3))
+// getQuorumSize calculates quorum size as 2/3 super-majority of provided total voting power
+func getQuorumSize(totalVotingPower *big.Int) *big.Int {
+	// TODO: @Stefan-Ethernal math.Ceil for big.Int?
+	quorum := new(big.Int)
+
+	return quorum.Mul(totalVotingPower, big.NewInt(2)).Div(quorum, big.NewInt(3))
```
Stefan-Ethernal: We are lacking `math.Ceil` here in order to have 2/3 super-majority. Not sure how to accomplish ceil for `big.Int` though.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1060/commits/5de8ba548d9ddc7ef5ceca003554710d449d5142

igorcrevar (approved):

goran-ethernal (approved):

consensus/polybft/validator_set.go
```diff
@@ -97,6 +91,10 @@ func (vs validatorSet) Len() int {
...
 }
 
-func (vs validatorSet) getQuorumSize() int64 {
-	return int64(math.Ceil((2 * float64(vs.totalVotingPower)) / 3))
+// getQuorumSize calculates quorum size as 2/3 super-majority of provided total voting power
+func getQuorumSize(totalVotingPower *big.Int) *big.Int {
+	quorum := new(big.Int)
+	quorum.Mul(totalVotingPower, big.NewInt(2))
+	// divide and round up result to next bigger integer value
+	return quorum.Add(quorum, big.NewInt(3)).Sub(quorum, big.NewInt(1)).Div(quorum, big.NewInt(3))
```
vcastellm: Isn't this the same?: 
```suggestion
	return quorum.Add(quorum, big.NewInt(2)).Div(quorum, big.NewInt(3))
```

vcastellm (approved):
LGTM

0xSasaPrsic (approved):

consensus/polybft/consensus_runtime_test.go
```diff
@@ -34,28 +34,27 @@ func TestConsensusRuntime_GetVotes(t *testing.T) {
...
 	}
 
 	commitment, _, _ := buildCommitmentAndStateSyncs(t, stateSyncsCount, epoch, 0)
 
-	quorumSize := validatorAccounts.toValidatorSetWithError(t).getQuorumSize()
-
 	require.NoError(t, state.insertEpoch(epoch))
 
-	votesCount := quorumSize + 1
+	votesCount := int(getQuorumSize(validatorAccs.GetTotalVotingPower()).Int64())
```
b00ris, line 52: Why has it changed?

b00ris (approved):

consensus/polybft/consensus_runtime_test.go
Stefan-Ethernal: It is changed because of changes made to the `validatorSet`. I haven't really felt like `getQuroumSize` should be tight with validator set instance, because it requires only total voting power as a parameter (and you see on the example of this test, that previously we needed validatorSet instance just to be able to call `getQuorumSize`).

Btw entire `validatorSet` struct feels like a bit redundant and is there only because `AccountSet` is a type alias rather than a proper struct. Maybe we can discuss it in the future to merge `validatorSet` and `AccountSet` into a single implementation.

PR #1062, opened 2022-12-19
Feature contractsapi gen

ferranbt:
# Description

This PR generates client Go code stubs for the smart contracts API.

To generate Go code from smart contracts API, run this command:
`go run ./consensus/polybft/contractsapi/gen2/main.go`

The generated code will be written to `/consensus/polybft/contractsapi/contractsapi.go` file.

# Changes include

# Checklist

## Testing


consensus/polybft/contractsapi/gen2/main.go
```diff
@@ -0,0 +1,262 @@
...
+		for _, method := range c.methods {
+			res = append(res, rr.Gen(ar.Abi.Methods[method]))
+		}
+
+		for _, event := range c.events {
+			res = append(res, rr.GenEvent(ar.Abi.Events[event]))
+		}
+	}
+
+	str := `// Code generated by scapi/gen. DO NOT EDIT.
```
JekaMas: does the generated file was done with this template? it seems a bit different. could we add a command to regenerate mocks to the Makefile and CI, which must fail if there is a difference between committed and generated code?

ferranbt: Which parts do you see different from the template?

vcastellm (commented):
Just notice that this PR will get outdated soon because of the upcoming changes in bundles, etc

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,91 @@
+// Code generated by scapi/gen. DO NOT EDIT.
```
vcastellm, line 1: Would be great to generate the code in multiple files, one for each type

consensus/polybft/contractsapi/decoder.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return err
+	}
+
+	return decodeImpl(val, out)
+}
+
+func decodeMethod(method *abi.Method, input []byte, out interface{}) error {
+	if len(input) < 4 {
+		return fmt.Errorf("invalid bundle data, len = %d", len(input))
```
vcastellm, line 27: Is this only specific for bundles?

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,91 @@
...
+	Root    [32]byte `abi:"root"`
+}
+
+type Commit struct {
+	Bundle    Bundle `abi:"bundle"`
+	Signature []byte `abi:"signature"`
+	Bitmap    []byte `abi:"bitmap"`
+}
+
+func (c *Commit) EncodeAbi() ([]byte, error) {
```
vcastellm: Why not all types have `Encode/DecodeAbi`?

ferranbt:
> Just notice that this PR will get outdated soon because of the upcoming changes in bundles, etc

Yep, I am aware. Right now I have not touched anything out of the `contractsapi` repo.

consensus/polybft/contractsapi/decoder.go
ferranbt: Nope, wrong error comment.

consensus/polybft/contractsapi/contractsapi.go
ferranbt: I am not fond of that, you might end up with many different small files.

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	_ = big.NewInt
+	_ = types.Address{}
+	_ = ethgo.Log{}
+)
+
+var (
+	commitType = abi.MustNewMethod("function commit(tuple(uint256 startId,uint256 endId,uint256 leaves,bytes32 root) bundle,bytes signature,bytes bitmap)") //nolint:all
```
b00ris: I suggest to add an exception for generated files instead of holing

ferranbt: Not sure I follow what do you mean?

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+func (c *Checkpoint) DecodeAbi(buf []byte) error {
+	return decodeStruct(CheckpointABIType, buf, &c)
+}
+
+var NewValidatorSetABIType = abi.MustNewType("tuple(address _address,uint256[4] blsKey,uint256 votingPower)")
+
+type NewValidatorSet struct {
+	_address    types.Address `abi:"_address"`
```
ferranbt: We need to handle this

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1062/commits/d47f9485d8bba2ca81d07006c241fd7b392ff0f7

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,285 @@
...
+	return EpochABIType.Encode(e)
+}
+
+func (e *Epoch) DecodeAbi(buf []byte) error {
+	return decodeStruct(EpochABIType, buf, &e)
+}
+
+var UptimeABIType = abi.MustNewType("tuple(uint256 epochId,tuple(address validator,uint256 signedBlocks)[] uptimeData,uint256 totalBlocks)")
+
+var UptimeDataABIType = abi.MustNewType("tuple(address validator,uint256 signedBlocks)")
```
Stefan-Ethernal: Maybe consider changing the generator, so that it groups variables in var block instead (if there is more than one variable)?

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,285 @@
...
+	StateSyncResultEventType = abi.MustNewEvent("event StateSyncResult(uint256 indexed counter,bool indexed status,bytes message)") //nolint:all
+)
+
+type StateSyncResultEvent struct {
+	Counter *big.Int `abi:"counter"`
+	Status  bool     `abi:"status"`
+	Message []byte   `abi:"message"`
+}
+
+func (S *StateSyncResultEvent) ParseLog(log *ethgo.Log) error {
```
Stefan-Ethernal: Change the generator so that it changes the first letter of the event name to lowercase.

consensus/polybft/contractsapi/gen2/main.go
```diff
@@ -0,0 +1,347 @@
...
+		contractStruct := renderTmpl(tmpl, inputs)
+
+		res = append(res, contractStruct)
+
+		for _, method := range c.methods {
+			res = append(res, rr.GenMethod(c.contractName, c.artifact.Abi.Methods[method]))
+		}
+
+		for _, event := range c.events {
+			res = append(res, rr.GenEvent(c.artifact.Abi.Events[event]))
```
Stefan-Ethernal: I believe here we should lowercase the first character... (related to https://github.com/0xPolygon/polygon-edge/pull/1062#discussion_r1085012870)?

Stefan-Ethernal (commented):
WDYT about renaming folders:

`gen` -> `artifacts-gen`
`gen2` -> `bindings-gen`

consensus/polybft/contractsapi/gen2/main.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1062/commits/b5b79ab945f3a4be11047cf3ea7261782bfc177b

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1062/commits/b5b79ab945f3a4be11047cf3ea7261782bfc177b

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: It complicates the generator quite a bit, since we don't know how many nested structs we can have (that generate function is a recursion).

scripts/cluster
```diff
@@ -41,7 +41,7 @@ function startServerFromDockerCompose() {
         ;;
       "polybft")
         cd core-contracts && npm install && npm run compile && cd -
-        go run ./consensus/polybft/contractsapi/gen/main.go
+        go run ./consensus/polybft/contractsapi/artifacts-gen/main.go
```
Stefan-Ethernal, line 44: Would be very useful if we could create a new command in Makefile with both artifacts and bindings generation.

Such command should look like this (double check if there are some typos):
```make
.PHONY: generate-smart-contract-bindings
generate-smart-contract-bindings:
	go run ./consensus/polybft/contractsapi/artifacts-gen/main.go
	go run ./consensus/polybft/contractsapi/bindings-gen/main.go
```

Also we should update `compile-core-contracts` target as well

```make
.PHONY: compile-core-contracts
compile-core-contracts:
	cd core-contracts && npm install && npm run compile
	$(MAKE) generate-smart-contract-bindings
```

Stefan-Ethernal (approved):
Nice! :rocket: 

Leaving a small suggestion to add a new target in makefile, so that we don't need to remember by heart commands which are needed to be run in order to (re)generate SC artifacts and bindings.

scripts/cluster
vcastellm: I think it's not necessary to compile and gen code on every cluster run, with the Make commands should be enough?

consensus/polybft/contractsapi/bindings-gen/main.go
```diff
@@ -0,0 +1,293 @@
...
+const (
+	contractVariableName   = "%sContract"
+	contractStructName     = "%sContractImpl"
+	contractVariableFormat = "%s = &%s{Artifact: %s}"
+	abiTypeNameFormat      = "var %sABIType = abi.MustNewType(\"%s\")"
+	eventNameFormat        = "%sEvent"
+)
+
+func main() {
+	cases := []struct {
```
vcastellm, line 27: Can this be dynamic instead of hardcoded, based on the contract?

scripts/cluster
Stefan-Ethernal: > I think it's not necessary to compile and gen code on every cluster run, with the Make commands should be enough?

Contracts are compiled in case polybft is started in a Docker container. It probably makes sense to do so, if we are going to share this script with the end users eventually. 
During development, we are most commonly starting polybft protocol from the binary (and there we don't have smart contract compilation as a step, it is assumed those are already compiled).

Anyhow, my comment here just relates to adding new make command, so that it makes our lives a bit easier (I don't want to bother and memorize which commands I need to run in order to generate artifacts and bindings, therefore it is much easier with make commands approach). Any objections? :slightly_smiling_face:

vcastellm: I didn't realize it was for Docker only, it makes more sense, but even that, it's the same as with local binary, why not have in the docs and remove it from the script? It'll improve startup time.

@Stefan-Ethernal you mean a make command to start the cluster with docker? I asked Platform team to remove the Docker stuff from Makefile I know but this will be a combination of make + script, best of both worlds so LGTM.

consensus/polybft/contractsapi/bindings-gen/main.go
goran-ethernal: It can, but, we don't need all the functions and events from contracts, plus we don't need to generate stubs from all contracts as well. I would say this is ok, because that artifact generator that Boris did also has a similar thing.

vcastellm: I won't be selective on this, TBH I prefer that it generates everything, we use only what we need.

goran-ethernal: I am fine with both. @ferranbt what do you think?

scripts/cluster
Stefan-Ethernal: > @Stefan-Ethernal you mean a make command to start the cluster with docker?

My suggestion is to provide a make command for creating both artifacts and bindings at the same time. It is a separate topic whether we need it for starting cluster or not, but having such make cmd would make our lives easier during development. That's my main goal. :slightly_smiling_face:

consensus/polybft/contractsapi/bindings-gen/main.go
Stefan-Ethernal: I'm more with Victor's suggestion here and generate everything if that's not going to overcomplicate logic. 
For example, it is a question of why we don't have generated bindings for staking/unstaking and delegation/undelegation. We could use those bindings from the helper commands we have. 

Therefore it is probably better to generate everything and use whatever subset we need.

b00ris (approved):

goran-ethernal (approved):

consensus/polybft/contractsapi/bindings-gen/main.go
```diff
@@ -0,0 +1,308 @@
...
+
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/umbracle/ethgo/abi"
+	"github.com/umbracle/ethgo"
+)
+
+`
+	str += strings.Join(res, "\n")
+
+	output, err := format.Source([]byte(str))
```
igorcrevar, line 100: I think that it would be nice if  `cobra.Command` is used in this `main` and if panics are not raised but errors handled

goran-ethernal: Not sure if we would get anything useful, since this is called manually, and on CI.

consensus/polybft/contractsapi/bindings-gen/main.go
ferranbt: It is described in the RFC section for this change. The bindings for staking/unstaking methods for example interact with the contracts over JsonRPC while the ones we are covering here interact with the contract as `Transaction` objects inside `Polybft`, that is why we only need a small subset of them.

The other Go bindings should be generated with [`abigen`](https://github.com/umbracle/ethgo/tree/main/cmd/abigen).

consensus/polybft/contractsapi/bindings-gen/main.go
ferranbt: I agree with Goran, this is only a small generator, it does not have any problem if it panics, it only worsens our own DX and it should not happen that often.

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,240 @@
...
+func (E *Epoch) DecodeAbi(buf []byte) error {
+	return decodeStruct(EpochABIType, buf, &E)
+}
+
+type UptimeData struct {
+	Validator    types.Address `abi:"validator"`
+	SignedBlocks *big.Int      `abi:"signedBlocks"`
+}
+
+var UptimeDataABIType = abi.MustNewType("tuple(address validator,uint256 signedBlocks)")
```
ferranbt, line 113: Does this needs to be public?

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (U *Uptime) EncodeAbi() ([]byte, error) {
+	return UptimeABIType.Encode(U)
+}
+
+func (U *Uptime) DecodeAbi(buf []byte) error {
+	return decodeStruct(UptimeABIType, buf, &U)
+}
+
+type CommitEpoch struct {
```
ferranbt: Would it be a problem to suffix "Method" at the end of the methods?

consensus/polybft/contractsapi/contractsapi.go
```diff
@@ -0,0 +1,240 @@
...
+	Signature  []byte               `abi:"signature"`
+	Bitmap     []byte               `abi:"bitmap"`
+}
+
+func (c *Commit) EncodeAbi() ([]byte, error) {
+	return StateReceiver.Abi.Methods["commit"].Encode(c)
+}
+
+func (c *Commit) DecodeAbi(buf []byte) error {
+	return decodeMethod(StateReceiver.Abi.Methods["commit"], buf, c)
```
ferranbt, line 39: It would be nice as a followup to remove the `Abi` artifacts generated which are big (at least for the JSON part) and do multi-line declarations like [this][(https://github.com/umbracle/ethgo/blob/main/abi/abi_test.go#L227) for all the artifacts.

The bytecode we can generate at the bottom of the go file for the types that we require.

ferranbt (commented):
Cannot approve because I opened the PR, but I approve the changes from @goran-ethernal.

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: No problem. Will change it.

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: We need these public, because we use some of them to create hashes for certain things. For example, commitment hash.

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: I would recommend doing that through a separate PR.

consensus/polybft/contractsapi/contractsapi.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1062/commits/36db7ca35f9a04332c90133fc96e9c191f00608f

consensus/polybft/contractsapi/contractsapi.go
ferranbt: Also, I am not sure to what extend it would be possible, if we introduce `abigen` we remove the need for much of the artifacts.

goran-ethernal: Maybe we can dedicate a day to have some POC. Introduce `abigen` and see what we can use.

scripts/cluster
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1062/commits/54a4a29ba22aec3c8e3620a6d060448200867977

igorcrevar (approved):

Nemanja0x (approved):

consensus/polybft/contractsapi/bindings-gen/main.go
vcastellm: @ferranbt This will require maintenance in the future, adding or removing new parts, if we generate everything regardless of what we use, it's just gen and forget. I would prefer to generate everything.

consensus/polybft/contractsapi/contractsapi.go
vcastellm: This sounds interesting

PR #1063, opened 2022-12-20
Create a full block object during polybft

ferranbt:
# Description

This PR introduces the concept of `FullBlock`. A `FullBlock` is a `Block` object including its `Receipts`. Before, there was a bug because `polybft` and `blockchain` would only communicate over `Block` objects. However, at some point over that coordination, the `Receipt` objects would be needed it and the Block was being re-executed again.

The important nuance of the PR is that inside the `fsm` we keep the target `FullBlock` block that is going to be inserted and that it was processed either during `BlockProposal` or `Validate`. If not, and we rely on the `proposal []byte` from `go-ibft` (which only includes `Block`) we would have to do extra tricks (like caching receipts in `blockchain`) to ensure we do no recompute `Receipts` and `From` addresses from the transactions.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -874,6 +874,68 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
 	}, nil
 }
 
+// WriteFullBlock writes a single block to the local blockchain.
+// It doesn't do any kind of verification, only commits the block to the DB
+// This function is a copy of WriteBlock but with a full block which does not
+// require to compute again the Receipts.
+func (b *Blockchain) WriteFullBlock(fblock *types.FullBlock, source string) error {
+	block := fblock.Block
+
+	b.writeLock.Lock()
+	defer b.writeLock.Unlock()
+
+	if block.Number() <= b.Header().Number {
+		b.logger.Info("block already inserted", "block", block.Number(), "source", source)
+
+		return nil
+	}
+
+	header := block.Header
+
+	if err := b.writeBody(block); err != nil {
+		return err
+	}
+
+	// Write the header to the chain
+	evnt := &Event{Source: source}
+	if err := b.writeHeaderImpl(evnt, header); err != nil {
+		return err
+	}
+
+	// write the receipts, do it only after the header has been written.
+	// Otherwise, a client might ask for a header once the receipt is valid,
+	// but before it is written into the storage
+	if err := b.db.WriteReceipts(block.Hash(), fblock.Receipts); err != nil {
+		return err
+	}
+
+	// update snapshot
+	if err := b.consensus.ProcessHeaders([]*types.Header{header}); err != nil {
+		return err
+	}
+
+	b.dispatchEvent(evnt)
+
+	// Update the average gas price
+	b.updateGasPriceAvgWithBlock(block)
+
+	logArgs := []interface{}{
+		"number", header.Number,
+		"txs", len(block.Transactions),
+		"hash", header.Hash,
+		"parent", header.ParentHash,
+	}
+
+	if prevHeader, ok := b.GetHeaderByNumber(header.Number - 1); ok {
+		diff := header.Timestamp - prevHeader.Timestamp
+		logArgs = append(logArgs, "generation_time_in_seconds", diff)
+	}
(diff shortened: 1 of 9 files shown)


consensus/polybft/blockchain_wrapper.go
```diff
@@ -26,7 +26,7 @@ type blockchainBackend interface {
 	CurrentHeader() *types.Header
 
 	// CommitBlock commits a block to the chain.
-	CommitBlock(block *types.Block) ([]*types.Receipt, error)
+	CommitBlock(block *types.FullBlock) error
```
Stefan-Ethernal, line 29: What I would do here is just remove `State` field from the `StateBlock` (it isn't used) and use it, because it is the same as `FullBlock` (or other way around: remove `StateBlock` and use `FullBlock` everywhere where `StateBlock` needed instead).

consensus/polybft/polybft.go
```diff
@@ -359,6 +361,8 @@ func (p *Polybft) startPbftProcess() {
 
 			return
 		}
+
+		p.logger.Debug("time to run the sequence", "seconds", time.Since(now))
```
Stefan-Ethernal, line 365: Are we going to keep this log?

Btw it would be very nice if we could establish open telemetry tracing in IBFT some day. :slightly_smiling_face:

ferranbt: It was helpful, I think eventually these logs should be in the engine directly, so it would show directly which stage takes longer.

consensus/polybft/blockchain_wrapper.go
ferranbt: Fixed https://github.com/0xPolygon/polygon-edge/pull/1063/commits/8f1ed3b53aeeb4a3ee5f23c309c954501bd526fa

blockchain/blockchain.go
```diff
@@ -874,6 +874,68 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
 	}, nil
 }
 
+// WriteFullBlock writes a single block to the local blockchain.
+// It doesn't do any kind of verification, only commits the block to the DB
+// This function is a copy of WriteBlock but with a full block which does not
+// require to compute again the Receipts.
+func (b *Blockchain) WriteFullBlock(fblock *types.FullBlock, source string) error {
```
ferranbt, line 881: I wrote this function instead of changing the other one (`WriteBlock`) in order to avoid any issues with `ibft` consensus protocol.

consensus/polybft/polybft.go
Stefan-Ethernal: > should be in the engine directly, so it would show directly which stage takes longer.

Yes exactly, that's why I think open telemetry is useful (at least I find it useful back when we were running pbft).

Stefan-Ethernal (approved):

goran-ethernal (approved):

consensus/polybft/fsm.go
```diff
@@ -442,10 +450,8 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 
 // Insert inserts the sealed proposal
 func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) (*types.Block, error) {
-	newBlock := &types.Block{}
-	if err := newBlock.UnmarshalRLP(proposal); err != nil {
-		return nil, fmt.Errorf("cannot unmarshal proposal: %w", err)
-	}
+	newBlock := f.target.Block
```
vcastellm, line 453: Nitpick: can we create the FullBlock here and work with it forward?

consensus/polybft/polybft.go
vcastellm: Aside from tracing, agree with the log, but also emit a metric?

vcastellm (approved):
LGTM aside from the comments

consensus/polybft/fsm.go
ferranbt: I do not think so because the full block was created already as part of Insert and Validate.

PR #1064, opened 2022-12-21
RFC173 Checkpoint consensus issue

goran-ethernal:
# Description

Through this PR we removed the exit events that happened in the currently built block from the exit event root hash, because they are not finalized. This way we are mitigating the risk of a potential malicious exits added by a malicious proposer.

Now, exit events that happened in the currently built block (which can be a checkpoint block) are moved to the next checkpoint block (this way we are sure they are finalized and agreed upon by the validator consensus).

**Important note** here is that exit events that happened in the epoch ending block (which is a checkpoint block by default), will be moved to the exit root of the next epoch.

See the image below for an illustrated process:

v3 checkpointing - Trie workflow_new

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/fsm_test.go
+++ b/consensus/polybft/fsm_test.go
@@ -149,22 +149,9 @@ func TestFSM_BuildProposal_WithExitEvents(t *testing.T) {
 	stateBlock := createDummyStateBlock(parentBlockNumber+1, parent.Hash, extra)
 	receipts := make([]*types.Receipt, numOfReceipts)
 
-	personType := abi.MustNewType("tuple(string company, string address)")
-	encodedData, err := personType.Encode(map[string]string{"company": "Polygon", "address": "Planet Earth"})
-	require.NoError(t, err)
-
 	for i := 0; i < numOfReceipts; i++ {
 		receipts[i] = &types.Receipt{Logs: []*types.Log{
-			{
-				Address: contracts.L2StateSenderContract,
-				Data:    encodedData,
-				Topics: []types.Hash{
-					types.Hash(exitEventABI.ID()),
-					types.BytesToHash([]byte{uint8(i)}),
-					types.BytesToHash(types.StringToAddress("0x1111").Bytes()),
-					types.BytesToHash(types.StringToAddress("0x2222").Bytes()),
-				},
-			},
+			createTestLogForExitEvent(t, uint64(i)),
 		}}
 	}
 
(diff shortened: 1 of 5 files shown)


igorcrevar (approved):

consensus/polybft/fsm_test.go
```diff
@@ -1597,3 +1626,24 @@ func generateValidatorDelta(validatorCount int, allAccounts, previousValidatorSe
...
+	encodedData, err := someType.Encode(map[string]string{"firstName": "John", "lastName": "Doe"})
+	require.NoError(t, err)
+
+	log := &types.Log{
+		Address: contracts.L2StateSenderContract,
+		Topics:  topics,
+		Data:    encodedData,
+	}
+
+	return log
```
igorcrevar: Just small thing, it would be nicer if this code is just
```
return &types.Log{
		Address: contracts.L2StateSenderContract,
		Topics:  topics,
		Data:    encodedData,
	}
```

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1064/commits/65f0dd47f4ce915d25db419433c7c0bc7d1d13a9

consensus/polybft/fsm_test.go
```diff
@@ -1433,6 +1365,107 @@ func TestFSM_Validate_FailToVerifySignatures(t *testing.T) {
...
+	mBlockBuilder := newBlockBuilderMock(buildBlock)
+
+	mBackendMock.On("CommitBlock", mock.MatchedBy(func(i interface{}) bool {
+		stateBlock, ok := i.(*types.FullBlock)
+		require.True(t, ok)
+
+		return stateBlock.Block.Number() == buildBlock.Block.Number() && stateBlock.Block.Hash() == buildBlock.Block.Hash()
+	})).Return(error(nil)).Once()
+
+	validatorSet, err := NewValidatorSet(validatorsMetadata[0:len(validatorsMetadata)-1], hclog.NewNullLogger())
```
Stefan-Ethernal: Why don't provide entire validator set here (in that case, you could just use `validators.toValidatorSetWithError(t)`)?

b00ris (approved):

Stefan-Ethernal (approved):
Minor comment to consider, otherwise 🚀

consensus/polybft/fsm_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1064/commits/32fb514276f79cedaf8b4bcde9c8d717a0857710

PR #1065, opened 2022-12-21
Fix hash calculation for proposal for go-ibft's update

### LINKED ISSUE #1169: Upgrade go-ibft package version to 0.1.0
Upgrade go-ibft package version to 0.1.0
## Description
Projects which want to use Polygon Edge to bootstrap their blockchains can't modify the quorum logic because in go-ibft@0.0.0 part of the quorum logic is handled there instead of in the client. This is changed in version 0.1.0.

## Proposed solution
Upgrading the go-ibft version to 0.1.0 will allow for blockchain projects using Polygon Edge to apply changes to the quorum logic when needed (applying staking amount weight for example) without the need to fork and apply changes in the go-ibft package.

Kourin1996:
# Description

As described in https://github.com/0xPolygon/go-ibft/pull/55, hash calculation needs to update for IBFT2.0 consensus.
In particular, IBFT backend must generate hash for IBFT consensus from tuple of `(EthereumBlock, round)`.

This PR changes hash calculation add adds new field for round in IBFT Extra.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -3,10 +3,10 @@ package ibft
 import (
 	"context"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/0xPolygon/go-ibft/messages"
+	"github.com/0xPolygon/go-ibft/messages/proto"
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
 	"github.com/0xPolygon/polygon-edge/helper/hex"
@@ -142,7 +142,11 @@ func (i *backendIBFT) MaximumFaultyNodes() uint64 {
 	return uint64(CalcMaxFaultyNodes(i.currentValidators))
 }
 
-func (i *backendIBFT) Quorum(blockNumber uint64) uint64 {
+func (i *backendIBFT) HasQuorum(
+	blockNumber uint64,
+	messages []*proto.Message,
+	msgType proto.MessageType,
+) bool {
 	validators, err := i.forkManager.GetValidators(blockNumber)
 	if err != nil {
 		i.logger.Error(
@@ -151,13 +155,16 @@ func (i *backendIBFT) Quorum(blockNumber uint64) uint64 {
 			"err", err,
 		)
 
-		// return Math.MaxInt32 to prevent overflow when casting to int in go-ibft package
-		return math.MaxInt32
+		return false
 	}
 
-	quorumFn := i.quorumSize(blockNumber)
+	if msgType == proto.MessageType_PREPREPARE {
+		return len(messages) > 0
+	}
 
-	return uint64(quorumFn(validators))
+	quorum := i.quorumSize(blockNumber)(validators)
+
+	return len(messages) >= quorum
 }
 
 // buildBlock builds the block, based on the passed in snapshot and parent header
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -144,7 +144,7 @@ func (i *backendIBFT) MaximumFaultyNodes() uint64 {
 
 func (i *backendIBFT) HasQuorum(
 	blockNumber uint64,
-	messages []*proto.Message,
+	msgs []*proto.Message,
 	msgType proto.MessageType,
 ) bool {
 	validators, err := i.forkManager.GetValidators(blockNumber)
@@ -158,13 +158,28 @@ func (i *backendIBFT) HasQuorum(
 		return false
 	}
 
-	if msgType == proto.MessageType_PREPREPARE {
-		return len(messages) > 0
+	var (
+		numMsgs = len(msgs)
+		quorum  = i.quorumSize(blockNumber)(validators)
+	)
+
+	switch msgType {
+	case proto.MessageType_PREPREPARE:
+		return numMsgs >= 1
+	case proto.MessageType_PREPARE:
+		return numMsgs >= quorum-1
+	case proto.MessageType_COMMIT, proto.MessageType_ROUND_CHANGE:
+		return numMsgs >= quorum
 	}
 
-	quorum := i.quorumSize(blockNumber)(validators)
+	// should not reach here
+	i.logger.Warn(
+		"invalid message type when calculation quorum",
+		"height", blockNumber,
+		"messageType", msgType,
+	)
 
-	return len(messages) >= quorum
+	return false
 }
 
 // buildBlock builds the block, based on the passed in snapshot and parent header
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -45,10 +45,11 @@ func (i *backendIBFT) BuildEthereumBlock(blockNumber uint64) []byte {
 // the ProposedBlock is hashed as proto.Marshal(ProposedBlock) and then keccakk hashed.
 func (i *backendIBFT) InsertBlock(
 	ethereumBlock []byte,
+	roundNumber uint64,
 	committedSeals []*messages.CommittedSeal,
 ) {
 	newBlock := &types.Block{}
-	if err := newBlock.UnmarshalRLP(proposal); err != nil {
+	if err := newBlock.UnmarshalRLP(ethereumBlock); err != nil {
 		i.logger.Error("cannot unmarshal proposal", "err", err)
 
 		return
@@ -66,7 +67,7 @@ func (i *backendIBFT) InsertBlock(
 	copy(extraDataBackup, extraDataOriginal)
 
 	// Push the committed seals to the header
-	header, err := i.currentSigner.WriteCommittedSeals(newBlock.Header, committedSealsMap)
+	header, err := i.currentSigner.WriteCommittedSeals(newBlock.Header, roundNumber, committedSealsMap)
 	if err != nil {
 		i.logger.Error("cannot write committed seals", "err", err)
 
--- a/consensus/ibft/consensus_backend.go
+++ b/consensus/ibft/consensus_backend.go
@@ -144,7 +144,7 @@ func (i *backendIBFT) MaximumFaultyNodes() uint64 {
 
 func (i *backendIBFT) HasQuorum(
 	blockNumber uint64,
-	msgs []*proto.Message,
+	messages []*proto.Message,
 	msgType proto.MessageType,
 ) bool {
 	validators, err := i.forkManager.GetValidators(blockNumber)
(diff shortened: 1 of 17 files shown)


consensus/ibft/signer/extra.go
```diff
@@ -62,6 +84,12 @@ func (i *IstanbulExtra) MarshalRLPWith(ar *fastrlp.Arena) *fastrlp.Value {
 	// ParentCommittedSeal
 	if i.ParentCommittedSeals != nil {
 		vv.Set(i.ParentCommittedSeals.MarshalRLPWith(ar))
+	} else if i.RoundNumber != nil {
```
igorcrevar: Why is there  `else if i.RoundNumber != nil` and bellow again if with same condition

consensus/ibft/signer/extra.go
```diff
@@ -37,6 +42,23 @@ type Seals interface {
 	UnmarshalRLPFrom(*fastrlp.Parser, *fastrlp.Value) error
 }
 
+func parseRoundNumber(v *fastrlp.Value) (*uint64, error) {
```
igorcrevar: Can we use  `num := binary.LittleEndian.Uint64(array)` or `num := binary.BigEndian.Uint64(array)` here?

consensus/ibft/consensus_backend.go
```diff
@@ -40,12 +40,16 @@ func (i *backendIBFT) BuildProposal(blockNumber uint64) []byte {
 	return block.MarshalRLP()
 }
 
+// TODO: Yoshiki please do this change, we need to make sure that on validating (when syncing) we make sure that
+// the committed seals are verifying against ProposedBlock (structure found in go-ibft/messages/proto/messages.pb.go
+// the ProposedBlock is hashed as proto.Marshal(ProposedBlock) and then keccakk hashed.
```
vcastellm:
```suggestion
// the ProposedBlock is hashed as proto.Marshal(ProposedBlock) and then keccak hashed.
```

consensus/ibft/consensus_backend.go
```diff
@@ -3,18 +3,18 @@ package ibft
...
+	"github.com/0xPolygon/go-ibft/messages/proto"
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
 	"github.com/0xPolygon/polygon-edge/helper/hex"
 	"github.com/0xPolygon/polygon-edge/state"
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
-func (i *backendIBFT) BuildProposal(blockNumber uint64) []byte {
+func (i *backendIBFT) BuildEthereumBlock(blockNumber uint64) []byte {
```
vcastellm: I remember that we called it `EthereumBlock` but what's the motivation for this name? why not call it simply `Block`?

Kourin1996: Renamed to BuildProposal at f3cd3dc5c2a57f3820b5362e50b650689a512c43

consensus/ibft/consensus_backend.go
Kourin1996: This comment has been removed

consensus/ibft/signer/extra.go
Kourin1996: Not it uses GetUint64, fastrlp provides originally

consensus/ibft/signer/extra.go
Kourin1996: Fixed 4f5ec060938dfd3d015f7d3631f0ace119caa4a9

Kourin1996: Fixed 4f5ec060938dfd3d015f7d3631f0ace119caa4a9

Kourin1996: To put RoundNumber in correct position

consensus/ibft/verifier.go
```diff
@@ -184,3 +182,43 @@ func (i *backendIBFT) extractProposer(header *types.Header) (types.Address, erro
 
 	return proposer, nil
 }
+
+func (i *backendIBFT) getProposalHashFromBlockBytes(ethereumBlock []byte, round *uint64) (types.Hash, error) {
```
b00ris: Attached document about compatibility of this change with checkpointing/slashing for 0.7.x. https://www.notion.so/polygontechnology/Signing-in-goibft-6a3ba4a50deb4efd85c1e2d71f401a6d

Kourin1996: As discussed yesterday, the hash for Committed Seals is calculated by keccak256(block hash + round bytes) from commit c0d1d00f0dbfd7da255fa6b7756013596660970b

trimixlover (approved):

Stefan-Ethernal:
Does it make sense to address these errors before the review process?

```go
Error: consensus/ibft/messages.go:27:30: undefined: proto.ProposedBlock
Error: consensus/ibft/messages.go:93:22: undefined: proto.ProposedBlock
Error: consensus/ibft/verifier.go:185:63: undefined: proto.ProposedBlock
Error: consensus/ibft/ibft.go:199:3: cannot use i (variable of type *backendIBFT) as type core.Backend in argument to newIBFT:
```

Kourin1996:
> Does it make sense to address these errors before the review process?

Need a release in go-ibft side because this PR includes a change based on a open PR in go-ibft
https://github.com/0xPolygon/go-ibft/pull/55

vcastellm (approved):

igorcrevar (approved):

b00ris (approved):

consensus/ibft/signer/extra.go
igorcrevar: I understand, resolved

consensus/ibft/signer/extra.go
```diff
@@ -62,6 +92,14 @@ func (i *IstanbulExtra) MarshalRLPWith(ar *fastrlp.Arena) *fastrlp.Value {
 	// ParentCommittedSeal
 	if i.ParentCommittedSeals != nil {
 		vv.Set(i.ParentCommittedSeals.MarshalRLPWith(ar))
+	} else if i.RoundNumber != nil {
+		vv.Set(ar.NewNull())
+	}
```
Stefan-Ethernal: Isn't it enough to remove this part (if condition below is anyways going to set round number if it is not nil)?

This way it seems like `ParentCommittedSeals` are in some relation with `RoundNumber`, but I don't see how they relate.

```suggestion
	}
```

consensus/ibft/signer/extra.go
```diff
@@ -242,6 +306,14 @@ func packCommittedSealsIntoExtra(
 			// ParentCommittedSeal
 			if len(oldValues) >= 4 {
 				newArrayValue.Set(oldValues[3])
+			} else if roundNumber != nil {
+				newArrayValue.Set(ar.NewNull())
+			}
```
Stefan-Ethernal: Again not sure why do we need this else if?

consensus/ibft/signer/extra.go
Kourin1996: Fixed 0fd2657058e238256b737ad5dcafc250f0f4c813

consensus/ibft/signer/extra.go
```diff
@@ -37,6 +42,35 @@ type Seals interface {
 	UnmarshalRLPFrom(*fastrlp.Parser, *fastrlp.Value) error
 }
 
+// parseRoundNumber parses RLP-encoded bytes into round
+// FYI, Extra has 8 bytes space for round in order to distinguish between null and 0
```
Stefan-Ethernal: If I'm correct this comment is now redundant?

```suggestion
```

consensus/ibft/signer/extra.go
```diff
@@ -20,6 +22,8 @@ var (
 	IstanbulExtraSeal = 65
 
 	zeroBytes = make([]byte, 32)
+
+	ErrRoundNumberOverflow = errors.New("round number is out of range for 64bit")
```
Stefan-Ethernal: Should we make this local (unexport it)?

consensus/ibft/signer/extra.go
```diff
@@ -37,6 +42,35 @@ type Seals interface {
...
+		return nil, err
+	}
+
+	if len(roundBytes) == 0 {
+		return nil, nil
+	}
+
+	if len(roundBytes) > 8 {
+		return nil, ErrRoundNumberOverflow
+	}
```
Stefan-Ethernal: Nitpick, maybe just reverse order of these two if statements, so that negative incomes are grouped together. WDYT?

```suggestion
	if len(roundBytes) > 8 {
		return nil, ErrRoundNumberOverflow
	}

	if len(roundBytes) == 0 {
		return nil, nil
	}
```

Stefan-Ethernal (approved):
LGTM

consensus/ibft/signer/extra.go
Kourin1996: No

consensus/ibft/signer/extra.go
Kourin1996: Fixed c2fee48aa934d3210d2de82b9de7d339769ea35e

consensus/ibft/signer/extra.go
Kourin1996: Fixed c2fee48aa934d3210d2de82b9de7d339769ea35e

consensus/ibft/signer/extra.go
Kourin1996: @Stefan-Ethernal

Stefan-Ethernal: I was referring that round is not nil and in that case it is 8 bytes length or if it is `nil` it's length is 0. With having that on mind I'm not sure if comment was trying to outline the same thing, because it is a bit confusing to me.

Kourin1996: Round in IBFT Extra can be nil in order to support backward compatibility

Kourin1996: In case of that Extra has Round, bytes data would have non-zero size always.

Kourin1996: Fixed comment 63f4cc01137fd2e71b99d270de89bd232a6b05cd. Is that fine with you @Stefan-Ethernal ?

Stefan-Ethernal: > Round in IBFT Extra can be nil in order to support backward compatibility

Yes aware of it and it is completely fine 👍 . However I was referring that comment is somewhat imprecise, since we would have either 8 bytes or 0 (in case round is `nil`). 

Don't mind keeping it, but just wanted to make this clear.

Stefan-Ethernal: > Fixed comment [63f4cc0](https://github.com/0xPolygon/polygon-edge/commit/63f4cc01137fd2e71b99d270de89bd232a6b05cd). Is that fine with you @Stefan-Ethernal ?

Yep this is better. Tnx 🙏

Kourin1996: @Stefan-Ethernal ping

PR #1067, opened 2022-12-21
Feature - State Sync manager

ferranbt:
# Description

Introduced a concept of `StateSyncManager`, a struct that handles state sync saving workflow, and building and submitting commitments to the `StateReceiver` contract.

`StateSyncManager` is one of the building blocks in `consensus_runtime`, but it encapsulates the state sync logic mentioned above.

This makes testing easier (as we can mock the state sync workflow, in `consensus_runtime` tests, much more easily), makes code more granular and separates logic to concerning parts.

# Changes include

# Checklist

## Testing


consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,492 @@
...
+							if err := s.addLog(log); err != nil {
+								s.logger.Error("failed to decode state sync event", "hash", log.TransactionHash, "error", err)
+							}
+						}
+					}
+				case <-tt.DoneCh:
+					s.logger.Info("Historical sync done")
+				}
+			}
+		}()
```
Stefan-Ethernal: Bear in mind that this part of the PR (or possibly entire PR) is going to clash with @vcastellm, at least because of capturing `event_tracker` in a separate package: https://github.com/0xPolygon/polygon-edge/pull/1044/files#diff-4cd7f6ea48e9e952f59f2be59cf5142240f7d7337766f99b2ac04bb337775a82R17

consensus/polybft/consensus_runtime.go
```diff
@@ -391,10 +303,12 @@ func (c *consensusRuntime) FSM() error {
 	}
 
 	if c.IsBridgeEnabled() {
-		err := c.populateFsmIfBridgeEnabled(ff, epoch, parent, isEndOfEpoch, isEndOfSprint)
+		commitment, err := c.stateSyncManager.Commitment()
```
goran-ethernal: This means that you can register commitment in any block in an epoch, but be aware, if you do not have quorum (which is a valid case), you will return error here, which is not something we want to do in that case, right?

Another issue is, I don't see anywhere that you reset that commitment field in `stateSyncManager` after you successfully submit a commitment after you reach quorum. Basically, you can submit the same commitment more than once in epoch.

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,494 @@
...
+			return
+		}
+
+		if err := s.deliverMessage(transportMsg); err != nil {
+			s.logger.Warn("failed to deliver message", "error", err)
+		}
+	})
+}
+
+func (s *StateSyncManager) deliverMessage(msg *TransportMessage) error {
```
vcastellm: This method name is semantically confusing for me, this method is saving the message vote in the state.

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,494 @@
...
+		"deliver message",
+		"hash", hex.EncodeToString(msg.Hash),
+		"sender", msg.NodeID,
+		"signatures", numSignatures,
+	)
+
+	return nil
+}
+
+func (s *StateSyncManager) initTracker() error {
```
vcastellm: Good for initial approach but I think it's better to use the new `tracker` package introduced in https://github.com/0xPolygon/polygon-edge/pull/1044, depending on what PR goes first

vcastellm (commented):
I like it overall aside from the comments

consensus/polybft/state_sync_manager.go
goran-ethernal: Used `eventTracker` from `tracker` package:

Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/503224004886fb793d1d63d972144d5323a9e546

consensus/polybft/state_sync_manager.go
goran-ethernal: Changed the function name to `saveVote` to closely describe what the function does.

Commit: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/503224004886fb793d1d63d972144d5323a9e546

consensus/polybft/state_sync_manager.go
goran-ethernal: Used the tracker from `tracker` package.

Commit: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/503224004886fb793d1d63d972144d5323a9e546

consensus/polybft/consensus_runtime.go
goran-ethernal: Resolved comment in: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/503224004886fb793d1d63d972144d5323a9e546

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+		// Epoch metadata is undefined
+		return nil
+	}
+
+	if msg.EpochNumber < epoch {
+		// or received message for some of the older epochs.
+		return nil
+	}
+
+	if !valSet.validators.ContainsNodeID(msg.NodeID) {
```
Stefan-Ethernal: Nitpick (doesn't harm if it stays like it is currently, but maybe I like any of these two approaches slightly more):

```suggestion
	if !valSet.Includes(types.StringToAddress(msg.NodeID)) {
```
or

```suggestion
	if !valSet.validators.ContainsAddress(types.StringToAddress(msg.NodeID)) {
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+	"github.com/0xPolygon/polygon-edge/tracker"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/hashicorp/go-hclog"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/umbracle/ethgo"
+	"google.golang.org/protobuf/proto"
+)
+
+// stateSyncConfig holds the configuration data of state sync manager
+type stateSyncConfig struct {
```
Stefan-Ethernal, line 48: Would it harm to rely on `runtimeConfig` (and remove `stateSyncConfig`) instead? `stateSyncConfig` has subset of fields which `runtimeConfig` already has (although I'm for separation of concerns generally speaking, these are the "plain" DTO structs, without any logic)...

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+
+	if valSet == nil {
+		// Epoch metadata is undefined
+		return nil
+	}
+
+	if msg.EpochNumber < epoch {
+		// or received message for some of the older epochs.
+		return nil
+	}
```
Stefan-Ethernal: Condition doesn't add much complexity written this way, but we spare a few lines of code :wink: 

```suggestion
	if valSet == nil || msg.EpochNumber < epoch {
		// Epoch metadata is undefined or received message for some of the older epochs.
		return nil
	}
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+	config *stateSyncConfig
+
+	// per epoch fields
+	lock         sync.Mutex
+	commitment   *Commitment
+	epoch        uint64
+	validatorSet *validatorSet
+}
+
+// topic is an interface for a gossip p2p message
```
```suggestion
// topic is an interface for p2p message gossiping
```

```suggestion
// topic is an interface for p2p message gossiping
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+	jsonrpcAddr     string
+	dataDir         string
+	topic           topic
+	key             *wallet.Key
+}
+
+// StateSyncManager is a struct that manages the workflow of
+// saving and querying state sync events, and creating, and submitting new commitments
+type StateSyncManager struct {
+	// configuration fields
```
Stefan-Ethernal: Is this comment necessary?
```suggestion
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+		lock:   sync.Mutex{},
+	}
+
+	return s, nil
+}
+
+// init subscribes to bridge topics (getting votes) and start the event tracker routine
+func (s *StateSyncManager) init() error {
+	if err := s.initTracker(); err != nil {
+		return err
```
Stefan-Ethernal: Add more context to error message, like for example:
```suggestion
		return fmt.Errorf("failed to init event tracker", "error", err)
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	if err := s.initTransport(); err != nil {
+		return err
```
Stefan-Ethernal: Add more context to error message, like for example:
```suggestion
		return fmt.Errorf("failed to initialize state sync transport layer", "error", err)
```

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+// saving and querying state sync events, and creating, and submitting new commitments
+type StateSyncManager struct {
+	// configuration fields
+	logger hclog.Logger
+	state  *State
+
+	config *stateSyncConfig
+
+	// per epoch fields
+	lock         sync.Mutex
```
Stefan-Ethernal, line 67: Could this be `sync.RWMutex`? It is there only because of validator set basically if I'm correct?

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	commitment, err := s.buildCommitment(req.NewEpochID, nextCommittedIndex)
+	if err != nil {
+		return err
+	}
+
+	s.commitment = commitment
+	s.validatorSet = req.ValidatorSet
+	s.epoch = req.NewEpochID
```
Stefan-Ethernal, line 337: Isn't it enough just to add `Lock` for these two lines (if we are going to switch to `RWMutex`)? Basically we need lock just because messages are arriving from different goroutines?

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,469 @@
...
+		}
+	})
+}
+
+// saveVote saves the gotten vote to boltDb for later quorum check and signature aggregation
+func (s *StateSyncManager) saveVote(msg *TransportMessage) error {
+	s.lock.Lock()
+	epoch := s.epoch
+	valSet := s.validatorSet
+	s.lock.Unlock()
```
Stefan-Ethernal, line 145: Add `RLock` if we are going to switch to `RWMutex` (so that we allow multiple messages to be processed in parallel if they arrive in the very same time, although chances are pretty low for that, but anyways)?

consensus/polybft/consensus_runtime.go
```diff
@@ -126,6 +114,9 @@ type consensusRuntime struct {
 	// proposerCalculator is the object which manipulates with ProposerSnapshot
 	proposerCalculator *ProposerCalculator
 
+	// manager for state sync bridge transactions
+	stateSyncManager *StateSyncManager
```
Stefan-Ethernal: Personally I'm not very keen we are constantly checking `IsBridgeEnabled` on multiple places in consensus runtime...

Therefore, what do you think about introducing null object pattern:
1. abstract `StateSyncManager` to some interface (expose only necessary functions: `PostBlock`, `PostEpoch` etc. which looks very much like hook system)
2. have `NullStateSyncManager` implementation by default (if bridge not enabled) and no need to check constantly is bridge enabled. Only downside I see currently is that we would need to create those `request` objects in any way...

Stefan-Ethernal:
This function can be safely removed from consensusRuntime (used only in the unit test)

`func validateVote(vote *MessageSignature, epoch *epochMetadata) error {`

Stefan-Ethernal (commented):
Looking much better :fire:

consensus/polybft/state_sync_manager.go
goran-ethernal: It will be changed in another PR, for multiple commitments,

consensus/polybft/state_sync_manager.go
goran-ethernal: RW lock will be added through another PR, for multiple commitments.

consensus/polybft/state_sync_manager.go
goran-ethernal: RW lock will be added through another PR, for multiple commitments.

consensus/polybft/state_sync_manager.go
Stefan-Ethernal: Good for me

consensus/polybft/consensus_runtime.go
goran-ethernal: Agreed. Will change it.

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/state_sync_manager.go
goran-ethernal: I would keep them separate, as this is only done once, and as you said, separates concerns.

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/acaaa3d20726fc29d0361e1b8472d936ceba4e67

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/c693688956d77512f89b400048ba503057396c78

vcastellm (commented):
`stateSyncCommitmentSize` can be moved to the manager

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,483 @@
...
+	Init() error
+	Commitment() (*CommitmentMessageSigned, error)
+	PostBlock(req *PostBlockRequest) error
+	PostEpoch(req *PostEpochRequest) error
+}
+
+var _ StateSyncManager = (*dummyStateSyncManager)(nil)
+
+// dummyStateSyncManager is used when bridge is not enabled
+type dummyStateSyncManager struct{}
```
vcastellm, line 40: Can this be `nil` instead of dummy?

vcastellm (commented):
All functions that are not used in the state anymore and are used only in the manager, for example `decodeStateSyncEvent`, would be great to move them to the manager, so we isolate all statesync things in it.

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/7ad92ae33af92531583957b17d1a9ffdc8645521

vcastellm: :D no, I mean if we can not assign `nil` value to the var? In case we can not, `dummy` is good name hahah

vcastellm (approved):
LGTM aside from comment

consensus/polybft/state_sync_manager.go
goran-ethernal: Aaaahh, I see your point :smile: . By default (when bridge is not configured), we will always use this `nil(dummy)StateSyncManager` in the `runtime` constructor. So basically, it will always be either this dummy version, or the full state sync manager version.

consensus/polybft/consensus_runtime.go
```diff
@@ -134,6 +134,7 @@ func newConsensusRuntime(log hcf.Logger, config *runtimeConfig) (*consensusRunti
 		lastBuiltBlock:     config.blockchain.CurrentHeader(),
 		proposerCalculator: proposerCalculator,
 		logger:             log.Named("consensus_runtime"),
+		stateSyncManager:   &dummyStateSyncManager{},
```
Stefan-Ethernal: Let's instantiate stateSyncManager only once. Create some function which would check if bridge is enabled and based on it would return either `nilStateSyncManager` or (proper) `StateSyncManager` instance.

Stefan-Ethernal (approved):

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1067/commits/562b685df4cc5a3b75eb492d9ee85d1e524c1f8b

goran-ethernal (approved):

consensus/polybft/state_sync_manager.go
```diff
@@ -0,0 +1,526 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	// number of stateSyncEvents to be processed before a commitment message can be created and gossiped
+	stateSyncCommitmentSize = 10
+)
+
+// StateSyncManager is an interface that defines functions for state sync workflow
+type StateSyncManager interface {
```
ferranbt, line 30: I do not see the use case to do this abstraction here. The only reason would be to put it inside `consensus_runtime` but even that I think this is lightweight enough to not mock it and use it directly.

consensus/polybft/consensus_runtime.go
```diff
@@ -167,15 +157,41 @@ func newConsensusRuntime(log hcf.Logger, config *runtimeConfig) (*consensusRunti
...
+		)
 
-	return c.epoch
+		if err != nil {
+			return err
+		}
+
+		c.stateSyncManager = stateSyncManager
+	} else {
+		c.stateSyncManager = &dummyStateSyncManager{}
```
ferranbt, line 190: If the `stateSyncManager` is not enabled because the bridge is not enabled, it should never be called. I understand this is a safety measure in case it is ever called. But, I would prefer if we ensure on tests that never happens instead of having extra code to ensure that.

Note that once we have hooks (if we ever do it though) we do not need this check anymore.

ferranbt (commented):
Cannot Approve because I made the PR. LGTM except for the comments about the abstraction.

consensus/polybft/consensus_runtime.go
goran-ethernal: While I do understand your point, I must agree with Stefan here. `Nullable` pattern is a good practice, since it reduces the unnecessary checks and conditions in code, makes testing easier, and it doesn't really add any more complexity (since that `nil` or `dummy` implementation doesn't do anything). Plus, as you said, it keeps as out of harms way of nil exceptions. To me it makes sense to have it here, and if we ever implement that hook system, then I would agree, we can remove this `nil` implementation since we would not need it.

consensus/polybft/state_sync_manager.go
goran-ethernal: I agree, it's really lightweight to mock it. This interface is added only because of the `nil` implementation of `stateSyncManager`, which we discuss in comment: https://github.com/0xPolygon/polygon-edge/pull/1067#discussion_r1058790320

consensus/polybft/consensus_runtime.go
ferranbt: The other option would be to make `StateSyncManager` aware that he is not active and do nothing when `PostEpoch` and `PostBlock` get called.

goran-ethernal: Yes, but that means we will just be moving those checks to the `stateSyncManager` :confused: . That hook system is looking better by the minute :smile: .

PR #1069, opened 2022-12-22
Fix uptime total blocks calculation

Stefan-Ethernal:
# Description

Validator rewards are being calculated using this formula:

```solidity
uint256 totalStake = (validator.stake + _validators.getDelegationPool(uptimeData.validator).supply)
uint256 validatorReward = (reward * totalStake * uptimeData.signedBlocks) / (activeStake * uptime.totalBlocks);
```
However there is an issue on the client side, as uptime total blocks is calculated as sum of how many blocks was signed by each validator during the epoch, instead of just denoting for how many blocks uptime information is being submitted. Therefore reward calculation is not giving expected results.

Fix consists of providing total blocks for uptime calculation, instead of aggregating number of signatures for each block in given epoch.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -33,6 +33,8 @@ var (
 	errNotAValidator = errors.New("node is not a validator")
 	// errQuorumNotReached represents "quorum not reached for commitment message" error message
 	errQuorumNotReached = errors.New("quorum not reached for commitment message")
+	// errEpochToSmall is error message denoting that epoch doesn't contain enough blocks
+	errEpochTooSmal = errors.New("epoch size not large enough to calculate uptime")
 )
 
 // txPoolInterface is an abstraction of transaction pool
@@ -733,7 +735,7 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
 		// this means that epoch size must at least be 3 blocks,
 		// since we are not calculating uptime for lastBlockInEpoch and lastBlockInEpoch-1
 		// they will be included in the uptime calculation of next epoch
-		return nil, errors.New("epoch size not large enough to calculate uptime")
+		return nil, errEpochTooSmal
 	}
 
 	calculateUptimeForBlock := func(header *types.Header, validators AccountSet) error {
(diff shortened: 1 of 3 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -760,17 +765,15 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
 	endBlock := getEndEpochBlockNumber(epoch.Number, c.config.PolyBFTConfig.EpochSize)
 
 	blockHeader := currentBlock
-	validators := epoch.Validators
-
-	var found bool
+	blockExists := false
```
goran-ethernal, line 768: Why do you need this outside the loop?

goran-ethernal (approved):

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: I have just renamed that variable and initialized it. The reason for having it outside from the loop is `blockHeader` instance, because we need that one to be initialized outside of for loop and therefore we need to move that `blockExists` outside as well.

FTR: https://github.com/0xPolygon/polygon-edge/pull/1069/files#diff-e5b2e93010ff84c3777aef136e0834d710dd22861ca62ffbab9d5baa12a52b3cR795

consensus/polybft/consensus_runtime.go
```diff
@@ -727,12 +729,13 @@ func (c *consensusRuntime) deliverMessage(msg *TransportMessage) error {
...
 func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *epochMetadata) (*CommitEpoch, error) {
 	uptimeCounter := map[types.Address]uint64{}
+	totalBlocks := uint64(0)
 
 	if c.config.PolyBFTConfig.EpochSize < (uptimeLookbackSize + 1) {
 		// this means that epoch size must at least be 3 blocks,
 		// since we are not calculating uptime for lastBlockInEpoch and lastBlockInEpoch-1
 		// they will be included in the uptime calculation of next epoch
-		return nil, errors.New("epoch size not large enough to calculate uptime")
+		return nil, errEpochTooSmal
```
```suggestion
		return nil, errEpochTooSmall
```

Stefan-Ethernal: Good catch 👍 Fixed https://github.com/0xPolygon/polygon-edge/pull/1069/commits/7f45a1f7c7880b80350f4ed76153c74ee69277a6

b00ris (approved):

igorcrevar (approved):

ferranbt (approved):

PR #1077, opened 2022-12-26
Remove unnecessary panics from polybft

Stefan-Ethernal:
# Description

This PR removes panics from the polybft package and adds error handling instead.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -201,5 +201,5 @@ func (s *stateProvider) Call(addr ethgo.Address, input []byte, opts *contract.Ca
 // Txn is part of the contract.Provider interface to make Ethereum transactions. We disable this function
 // since the system state does not make any transaction
 func (s *stateProvider) Txn(ethgo.Address, ethgo.Key, []byte) (contract.Txn, error) {
-	return nil, errSendTxnUnsupported
+	panic(errSendTxnUnsupported)
 }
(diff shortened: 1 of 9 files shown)


0xSasaPrsic (approved):
Looks good.

consensus/polybft/blockchain_wrapper.go
```diff
@@ -196,5 +201,5 @@ func (s *stateProvider) Call(addr ethgo.Address, input []byte, opts *contract.Ca
 // Txn is part of the contract.Provider interface to make Ethereum transactions. We disable this function
 // since the system state does not make any transaction
 func (s *stateProvider) Txn(ethgo.Address, ethgo.Key, []byte) (contract.Txn, error) {
```
igorcrevar, line 203: I would leave panic here, this is just to notify developer that this object should not be used for something for which is not meant

igorcrevar (approved):
I would not remove panic from Txn method but otherwise good -> approved

Nemanja0x (approved):

consensus/polybft/blockchain_wrapper.go
ferranbt: I agree with @igorcrevar. This can never happen.

ferranbt (approved):

consensus/polybft/blockchain_wrapper.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1077/commits/9f8f1c2022511ec0bc37851484a86adc3bd311ca

PR #1080, opened 2022-12-27
Handled go routines in event tracker

Stefan-Ethernal:
# Description

This PR adds context to `event_tracker` in order to control termination of child go routines it spawns and removes redundant goroutine from it.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -119,9 +119,6 @@ type consensusRuntime struct {
 
 	// logger instance
 	logger hcf.Logger
-
-	// ctx is a context instance used for goroutines shutdown
-	ctx context.Context
 }
 
 // newConsensusRuntime creates and starts a new consensus runtime instance with event tracking
@@ -136,11 +133,10 @@ func newConsensusRuntime(ctx context.Context, log hcf.Logger, config *runtimeCon
 		config:             config,
 		lastBuiltBlock:     config.blockchain.CurrentHeader(),
 		proposerCalculator: proposerCalculator,
-		ctx:                ctx,
 		logger:             log.Named("consensus_runtime"),
 	}
 
-	if err := runtime.initStateSyncManager(log); err != nil {
+	if err := runtime.initStateSyncManager(ctx, log); err != nil {
 		return nil, err
 	}
 
@@ -172,10 +168,10 @@ func newConsensusRuntime(ctx context.Context, log hcf.Logger, config *runtimeCon
 
 // initStateSyncManager initializes state sync manager
 // if bridge is not enabled, then a dummy state sync manager will be used
-func (c *consensusRuntime) initStateSyncManager(log hcf.Logger) error {
+func (c *consensusRuntime) initStateSyncManager(ctx context.Context, logger hcf.Logger) error {
 	if c.IsBridgeEnabled() {
 		stateSyncManager, err := NewStateSyncManager(
-			log,
+			logger,
 			c.config.State,
 			&stateSyncConfig{
 				key:             c.config.Key,
@@ -183,7 +179,6 @@ func (c *consensusRuntime) initStateSyncManager(log hcf.Logger) error {
 				jsonrpcAddr:     c.config.PolyBFTConfig.Bridge.JSONRPCEndpoint,
 				dataDir:         c.config.DataDir,
 				topic:           c.config.bridgeTopic,
-				ctx:             c.ctx,
 			},
 		)
 
@@ -196,7 +191,7 @@ func (c *consensusRuntime) initStateSyncManager(log hcf.Logger) error {
 		c.stateSyncManager = &dummyStateSyncManager{}
 	}
 
-	return c.stateSyncManager.Init()
+	return c.stateSyncManager.Init(ctx)
 }
 
 // getGuardedData returns last build block, proposer snapshot and current epochMetadata in a thread-safe manner.
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -2,6 +2,7 @@ package polybft
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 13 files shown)


tracker/event_tracker.go
```diff
@@ -67,28 +67,30 @@ func (e *EventTracker) Start() error {
...
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+
+			case evnt := <-tt.EventCh:
+				if len(evnt.Removed) != 0 {
+					panic("this will not happen anymore after tracker v2")
```
Stefan-Ethernal, line 83: This panic should ideally vanish, although we must agree how to resolve it properly.

`Removed` field denotes there is a reorg, and therefore log is eventually revoked from the chain. However not sure what's the correct approach to change it, since it denotes that a reorg happened on a target chain, which can be either L1/Ethereum or L2/Polygon (depending which events we are tracking).
I'm aware that on Polygon chain, there shouldn't ever be reorgs and forks because of instant finality.
On the other hand, post-merge Ethereum shouldn't [face reorgs as before](https://hackmd.io/@vbuterin/reorg_wtf_slides#/), if any (?) (although it seems there is at least theoretical possibility to have it, since it still runs "weighted" [fork choice](https://ethereum.org/en/developers/docs/consensus-mechanisms/#:~:text=however%2C%20now%20that%20Ethereum%20runs%20on%20proof%2Dof%2Dstake%20it%20adopted%20an%20updated%20fork%2Dchoice%20algorithm%20that%20measures%20the%20%27weight%27%20of%20the%20chain.%20The%20weight%20is%20the%20accumulated%20sum%20of%20validator%20votes%2C%20weighted%20by%20validator%20staked%2Dether%20balances.) algorithm).
So the question boils down to should we worry about reorgs on the Ethereum Beacon chain?
Based on it, we have several options to consider what to do with that panic in event_tracker:
- remove entirely handling of `Removed` logs together with `panic` (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/tracker/event_tracker.go#L80)
- just add logging, but ignore the existence of `Removed` logs (consider only `Added` ones)
- preserve mentioned `if` and `panic` until further notice
- change logic how we are treating logs (such that we remove logs from polybft internal state in this case, because it basically denotes that a reorg happened and given log is no longer a valid). If there can be reorgs, this is probably the best approach.

consensus/polybft/polybft_test.go
```diff
@@ -178,26 +178,32 @@ func TestPolybft_Close(t *testing.T) {
...
 	syncer.On("Close", mock.Anything).Return(errExpected).Once()
 
-	polybft.closeCh = make(chan struct{})
-
 	assert.Error(t, errExpected, polybft.Close())
 
 	select {
-	case <-polybft.closeCh:
+	case <-polybft.ctx.Done():
 		assert.Fail(t, "channel closing is invoked")
```
vcastellm:
```suggestion
		assert.Fail(t, "context done")
```

tracker/event_tracker.go
```diff
@@ -34,11 +34,11 @@ func NewEventTracker(
 		rpcEndpoint:  rpcEndpoint,
 		contractAddr: contractAddr,
 		subscriber:   subscriber,
-		logger:       logger,
+		logger:       logger.Named("event_tracker"),
```
vcastellm, line 37: All event trackers will have the same logger prefix?

vcastellm (commented):
🎩 aside from comments and the reorgs issue

consensus/polybft/polybft_test.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1080/commits/a814c6d53b93bc4f84d9daddd5eb1e69833db23a

tracker/event_tracker.go
Stefan-Ethernal: In fact no. The one instantiated by `StateSyncManager` will have `state-sync` prefix, followed by `event_tracker` suffix. On the other hand, the one from `StateSyncRelayer` is going to have `relayer` prefix (again followed by `event_tracker`). So to me it seems ok. Do you feel the same?

vcastellm (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -118,10 +119,13 @@ type consensusRuntime struct {
 
 	// logger instance
 	logger hcf.Logger
+
+	// ctx is a context instance used for goroutines shutdown
+	ctx context.Context
```
JekaMas: Could we find a solution without storing context? It's not very recommended practice.

Stefan-Ethernal: Is this approach any better: https://github.com/0xPolygon/polygon-edge/pull/1080/commits/221ea2f0d9e0580dfbe3b7b5df93931f802649f4?

Unfortunately, I don't see it is feasible to remove it from `Polybft` state (https://github.com/0xPolygon/polygon-edge/pull/1080/files#diff-f66bb89ba6547fbe9286cab6d8318c1df7257099c5139d82b03955e2f4fc5462R68-R69).

JekaMas: It is better. It might be also make sense to store done channel instead of context itself. How do you think?

Stefan-Ethernal: > It might be also make sense to store done channel instead of context itself. How do you think?

Not sure about it, because I need `context` instance in the tracker itself (https://github.com/0xPolygon/polygon-edge/blob/EVM-126-handle-go-routines-in-event-tracker/tracker/event_tracker.go#L70). So I need to pass in context to the `consensusRuntime`, which propagates it to `EventTracker` (abstraction around `ethgo.Tracker`). If I'm correct, that would mean that we need both `context` and `doneChannel` stored in the polybft as well as `cancelFn` callback (to signal context about cancellation). IMO that would introduce confusion. WDYT?

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1080/commits/65734416be50f46c4e9bee510c17904ccdb27d9e

begmaroman (approved):

ferranbt (approved):

PR #1082, opened 2022-12-28
EVM-337 mcl bls integration

igorcrevar:
# Description

Integrate mcl crypto library for bn254/bls signing/verifying

The wrapper around `mcl` and all privatekey/publickey/signature logic is implemented in separated repo: https://github.com/0xPolygon/bnsnark1

# Changes include

# Checklist

## Testing

# Additional comments

In order to make everything work with our smart contracts, custom hash to `g1` function is implemented - code is copied and adapted from previous `kliic/bn254` library. Current implementations in `mcl` uses this method https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-03, but we needed this https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-07 and that is why this custom code was necessary.  

Also, serialization and de-serialization in `mcl` was not compatible with smart contract needs, so there is custom implementation analog to the serialization/de-serialization implementation which is used in smart contracts.


consensus/polybft/checkpoint_manager.go
```diff
@@ -5,7 +5,7 @@ import (
 	"math/big"
 	"strconv"
 
-	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	bls "github.com/0xPolygon/bnsnark1/core"
```
b00ris: Is it worth to rename core to bls? There is an alias for every import.

consensus/polybft/signer/common.go
```diff
@@ -1,25 +0,0 @@
-package bls
-
-import (
-	"encoding/hex"
-)
-
-var (
-	domain, _ = hex.DecodeString("508e30424791cb9a71683381558c3da1979b6fa423b2d6db1396b1d94d7c4a78")
```
b00ris, line 8: Is it worth to keep domain there?

igorcrevar: That file is deleted. Everything is inside bnsnark1 repo

consensus/polybft/checkpoint_manager.go
igorcrevar: I am not sure :( And I am not sure that bls is good name. Because bn and bls are different curves. I kept bls as import name just to prevent too much changes in original files

consensus/polybft/signer/common.go
ferranbt: Is the Domain inside bnsnark1? I thought the domain is something application specific.

state/runtime/precompiled/bls_agg_sigs_verification.go
```diff
@@ -3,8 +3,8 @@ package precompiled
 import (
 	"errors"
 
+	bls "github.com/0xPolygon/bnsnark1/core"
 	"github.com/0xPolygon/polygon-edge/chain"
-	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
```
ferranbt: When I mentioned that we should move bnsnark out in a separate repo, I did not mean to remove `signer` from the repo. Could we keep `signer` and make it so its a wrapper around `bnsnark1` repo?

.github/workflows/lint.yml
```diff
@@ -23,6 +23,21 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
```
ferranbt: I do not know enough about cgo right now, but I have used cgo libraries in the past before in other projects, how is it that I never had to setup these CI compilation workarounds?

ferranbt (changes requested):

state/runtime/precompiled/bls_agg_sigs_verification.go
igorcrevar: It does not make any sense than. it would be just one file in that case

.github/workflows/lint.yml
igorcrevar: Because `mcl` is complex project it must be build before usage and built c libraries should be linked with `CGO_LDFLAGS`. If you look at https://github.com/alinush/go-mcl his workflow is to manually download `mcl` project to /usr/local, compile it there and than to setup all those 4 environment variables to point to /usr/local/mcl...

consensus/polybft/signer/common.go
igorcrevar: Yes, I have provided method to setup domain if needed something different for it. But I put value we currently use as a default value

state/runtime/precompiled/bls_agg_sigs_verification.go
ferranbt: I do not follow. If you were to replace `killic` with `bnsnark1` libraries you would not have to remove the `signer` package.

.github/workflows/lint.yml
ferranbt: I have tested [bls-go-binary](https://github.com/herumi/bls-go-binary) which depends also on `mcl` and it does not require any build steps.

state/runtime/precompiled/bls_agg_sigs_verification.go
ferranbt: My point is that `polybft` should not be aware that we are using `bsnark1` under the hood. The `signer` was already a good abstraction around `Private/Public Keys` and `Signatures`.

.github/workflows/lint.yml
igorcrevar: Yes, it seams that they are using prebuilt library files. I had a problem because github used different glibc than the one i used to built `.a` files. If we use prebuilt libraries it would be a lot less code (submodule will not be required also) but i am not sure it will work on every machine

JekaMas (approved):

.github/workflows/lint.yml
igorcrevar: We have decided to use prebuilt `c static library` `*.a` files. Currently he have builds for linux/amd64 and darwin/arm64 os/architecture. Build process is as simple as it was before https://github.com/0xPolygon/polygon-edge/commit/e031ba77119ffc612da0841f3487540b5cac7219

state/runtime/precompiled/bls_agg_sigs_verification.go
igorcrevar: I have added wrapper around everything bnsnark1 provided inside `bls.go` file https://github.com/0xPolygon/polygon-edge/commit/e031ba77119ffc612da0841f3487540b5cac7219

consensus/polybft/signer/common.go
igorcrevar: domain value is now set in our `bls.go` `init` func https://github.com/0xPolygon/polygon-edge/commit/e031ba77119ffc612da0841f3487540b5cac7219

consensus/polybft/signer/serialization.go
```diff
@@ -0,0 +1,70 @@
...
+		return [2]*big.Int{}, fmt.Errorf("expect length 64 but got %d", len(bytes))
+	}
+
+	return [2]*big.Int{
+		new(big.Int).SetBytes(reverse(bytes[:32])),
+		new(big.Int).SetBytes(reverse(bytes[32:])),
+	}, nil
+}
+
+func BytesToBigInt4(bytes []byte) ([4]*big.Int, error) {
```
ferranbt: Do these functions need to be public?

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1082/commits/9d94ae1b96a471650e94a9732f1c44245a7cdc57

consensus/polybft/signer/bls.go
```diff
@@ -0,0 +1,173 @@
+package bls
```
Stefan-Ethernal: Consider renaming package to `signer` or folder name to `bls`?

Stefan-Ethernal (approved):
Like it. :+1:

goran-ethernal (approved):

vcastellm (changes requested):
Are you sure the library is using the pre-build libs? I don't believe that Go can do that, trying to build from an Alpine docker images gives:
```
# github.com/0xPolygon/bnsnark1/core
cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
```

Also notice that Alpine uses `musl` no `libc`, and we use Alpine as the release image, so the resulting binary won't work.

igorcrevar:
> Are you sure the library is using the pre-build libs? I don't believe that Go can do that, trying to build from an Alpine docker images gives:
> 
> ```
> # github.com/0xPolygon/bnsnark1/core
> cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
> ```
> 
> Also notice that Alpine uses `musl` no `libc`, and we use Alpine as the release image, so the resulting binary won't work.

You need to have gcc. Yes, it can work with prebuilt static library (`*.a`) files but Ferran has problem for example because his `libstdc++` is older than the one I used to build `*a` files. We should use the oldest gcc/libstdc++ we want to support for each os/architecture

vcastellm:
> > Are you sure the library is using the pre-build libs? I don't believe that Go can do that, trying to build from an Alpine docker images gives:
> > ```
> > # github.com/0xPolygon/bnsnark1/core
> > cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also notice that Alpine uses `musl` no `libc`, and we use Alpine as the release image, so the resulting binary won't work.
> 
> You need to have gcc. Yes, it can work with prebuilt static library (`*.a`) files but Ferran has problem for example because his `libstdc++` is older than the one I used to build `*a` files. We should use the oldest gcc/libstdc++ we want to support for each os/architecture

Yes, you're right I got confused by the wrapper itself, so I thought the libs are the wrapper.

I tried with Alpine amd64 with all build stack installed and my CPU is skyrocketing but worked.

My main concern is that from now on, we'll be limiting the usage of Edge to these specific combination of os/archs. 

People won't be able to run under BSD, the most stable OS 😛, where to put the limit? :/

igorcrevar:
> > > Are you sure the library is using the pre-build libs? I don't believe that Go can do that, trying to build from an Alpine docker images gives:
> > > ```
> > > # github.com/0xPolygon/bnsnark1/core
> > > cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also notice that Alpine uses `musl` no `libc`, and we use Alpine as the release image, so the resulting binary won't work.
> > 
> > 
> > You need to have gcc. Yes, it can work with prebuilt static library (`*.a`) files but Ferran has problem for example because his `libstdc++` is older than the one I used to build `*a` files. We should use the oldest gcc/libstdc++ we want to support for each os/architecture
> 
> Yes, you're right I got confused by the wrapper itself, so I thought the libs are the wrapper.
> 
> I tried with Alpine amd64 with all build stack installed and my CPU is skyrocketing but worked.
> 
> My main concern is that from now on, we'll be limiting the usage of Edge to these specific combination of os/archs.
> 
> People won't be able to run under BSD, the most stable OS stuck_out_tongue, where to put the limit? :/

They can run on BSD or any other os but someone with that system must build `mcl` and place two `*.a` files in iib folder of bnsark1 repo :D Also one line in `mcl.go` is needed regarding `#cgo LDFLAGS` `-L` library location path. 

I have built those files for linux (but probably we need to build them again with older `gcc/libstdc++` - someone with older Ubuntu can do that for example) and Stefan have built for macos/arm64

vcastellm:
> They can run on BSD or any other os but someone with that system must build `mcl` and place two `*.a` files in iib folder of bnsark1 repo :D Also one line in `mcl.go` is needed regarding `#cgo LDFLAGS` `-L` library location path.

I won't call it user friendly :/

> I have built those files for linux (but probably we need to build them again with older gcc/libstdc++ - someone with older Ubuntu can do that for example) and Stefan have built for macos/arm64

I think we have a problem if we can not provide a proper build of the libs ourselves, also this should be automated in CI and be cross-platform, maybe that's a bit exaggerated, but at least we need a method to build the selected platforms, correctly, deterministically from any machine, probably using docker is the easiest way.

Stefan-Ethernal:
Should we close this PR in favour of #1161, since polybft is integrated with github.com/umbracle/go-eth-bn256?

PR #1083, opened 2022-12-28
Multiple state sync commitments per epoch

goran-ethernal:
# Description

Implemented multiple commitments per epoch feature, as per RFC-178. Wrote appropriate UTs, and a `e2e` test (`TestE2E_Bridge_MultipleCommitmentsPerEpoch`).

Process of building commitments is as follows:

1. Using the event tracker, listen for state sync events, and store them in db.
2. At the beginning of an epoch, check if we have enough events to build a commitment (minimum size of commitment at least, which is 2 events).
    a. If we have, build a commitment, sign it, and gossip a vote.
    b. If we do not have enough state sync events, wait until we have enough.
3. If we gossiped a commitment which did not have a full size (10 events), on every state sync event that was received in the meantime, create a new commitment (with number of events of the old gossiped commitment, plus the new state syncs), sign it and gossip it.
4. At every sprint block, check which of the gossiped commitments have quorum of votes, find the largest commitment that has quorum, and submit it.
5. Once the block is finalized, check it to see if any was submit commitment transaction occurred.
     a. If there is, create a merkle tree, build state sync proofs, and save them to `boltDb`. Also, discard any cached commitment, 
    since a new one will be built in the next block.
     b. If there isn’t, wait until some proposer submits the largest commitment with quorum.
6. Once the epoch ends, and a new epoch is started, discard any pending commitments (that were built in the previous epoch, and are waiting for quorum of messages), since the validator set changes (we need to rebuild the last pending checkpoint in the new epoch).

commitmentsBuild

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -242,11 +242,6 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 	// after the block has been written we reset the txpool so that the old transactions are removed
 	c.config.txPool.ResetWithHeaders(block.Header)
 
-	// handle commitment and proofs creation
-	if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
-		c.logger.Error("failed to post block state sync", "err", err)
-	}
-
 	var (
 		epoch = c.epoch
 		err   error
@@ -259,6 +254,11 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 			c.logger.Error("failed to restart epoch after block inserted", "error", err)
 
 			return
+		}
+	} else {
+		// handle commitment and proofs creation
+		if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
+			c.logger.Error("failed to post block state sync", "err", err)
 		}
 	}
 
@@ -320,12 +320,14 @@ func (c *consensusRuntime) FSM() error {
 		logger:            c.logger.Named("fsm"),
 	}
 
-	commitment, err := c.stateSyncManager.Commitment()
-	if err != nil {
-		return err
-	}
+	if isEndOfSprint {
+		commitment, err := c.stateSyncManager.Commitment()
+		if err != nil {
+			return err
+		}
 
-	ff.proposerCommitmentToRegister = commitment
+		ff.proposerCommitmentToRegister = commitment
+	}
 
 	if isEndOfEpoch {
 		ff.uptimeCounter, err = c.calculateUptime(parent, epoch)
@@ -403,7 +405,6 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) (*epochMetadata, e
 	)
 
 	reqObj := &PostEpochRequest{
-		BlockNumber:  header.Number,
 		SystemState:  systemState,
 		NewEpochID:   epochNumber,
 		ValidatorSet: NewValidatorSet(validatorSet, c.logger),
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -247,6 +247,11 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 		err   error
 	)
 
+	// handle commitment and proofs creation
+	if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
+		c.logger.Error("failed to post block state sync", "err", err)
+	}
+
 	// TODO - this condition will need to be changed to recognize that either slashing happened
 	// or epoch reached its fixed size
 	if c.isFixedSizeOfEpochMet(block.Header.Number, c.epoch) {
@@ -254,11 +259,6 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 			c.logger.Error("failed to restart epoch after block inserted", "error", err)
 
 			return
-		}
-	} else {
-		// handle commitment and proofs creation
-		if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
-			c.logger.Error("failed to post block state sync", "err", err)
 		}
 	}
 
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -254,7 +254,7 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
 
 	// TODO - this condition will need to be changed to recognize that either slashing happened
 	// or epoch reached its fixed size
-	if c.isFixedSizeOfEpochMet(block.Header.Number, c.epoch) {
+	if c.isFixedSizeOfEpochMet(block.Header.Number, epoch) {
 		if epoch, err = c.restartEpoch(block.Header); err != nil {
 			c.logger.Error("failed to restart epoch after block inserted", "error", err)
 
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	eventsBufferSize   = 10
+	maxCommitmentSize  = 10
 	stateFileName      = "consensusState.db"
 	uptimeLookbackSize = 2 // number of blocks to calculate uptime from the previous epoch
 )
@@ -178,11 +178,12 @@ func (c *consensusRuntime) initStateSyncManager(logger hcf.Logger) error {
 			logger,
 			c.config.State,
 			&stateSyncConfig{
-				key:             c.config.Key,
-				stateSenderAddr: c.config.PolyBFTConfig.Bridge.BridgeAddr,
-				jsonrpcAddr:     c.config.PolyBFTConfig.Bridge.JSONRPCEndpoint,
-				dataDir:         c.config.DataDir,
-				topic:           c.config.bridgeTopic,
+				key:               c.config.Key,
+				stateSenderAddr:   c.config.PolyBFTConfig.Bridge.BridgeAddr,
+				jsonrpcAddr:       c.config.PolyBFTConfig.Bridge.JSONRPCEndpoint,
+				dataDir:           c.config.DataDir,
+				topic:             c.config.bridgeTopic,
+				maxCommitmentSize: maxCommitmentSize,
 			},
 		)
 
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -443,16 +443,23 @@ func (s *State) insertStateSyncEvent(event *types.StateSyncEvent) error {
 }
 
 // getStateSyncEventsForCommitment returns state sync events for commitment
-// if there is an event with index that can not be found in db in given range, an error is returned
-func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64) ([]*types.StateSyncEvent, error) {
+// based on the getAll flag, the function will try to get all events in range, and if it fails,
+// it will either return an error, or return all the events it can without returning an error
+func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64,
+	getAll bool) ([]*types.StateSyncEvent, error) {
 	var events []*types.StateSyncEvent
+
+	var notEnoughEventsErr error
+	if getAll {
+		notEnoughEventsErr = errNotEnoughStateSyncs
+	}
 
 	err := s.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket(syncStateEventsBucket)
 		for i := fromIndex; i <= toIndex; i++ {
 			v := bucket.Get(itob(i))
 			if v == nil {
-				return errNotEnoughStateSyncs
+				return notEnoughEventsErr
 			}
 
 			var event *types.StateSyncEvent
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -443,10 +443,7 @@ func (s *State) insertStateSyncEvent(event *types.StateSyncEvent) error {
 }
 
 // getStateSyncEventsForCommitment returns state sync events for commitment
-// based on the getAll flag, the function will try to get all events in range, and if it fails,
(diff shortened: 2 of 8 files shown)


consensus/polybft/state.go
```diff
@@ -428,16 +428,23 @@ func (s *State) insertStateSyncEvent(event *types.StateSyncEvent) error {
...
+		notEnoughEventsErr = errNotEnoughStateSyncs
+	}
+
 	err := s.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket(syncStateEventsBucket)
 		for i := fromIndex; i <= toIndex; i++ {
 			v := bucket.Get(itob(i))
 			if v == nil {
-				return errNotEnoughStateSyncs
+				return notEnoughEventsErr
```
igorcrevar: can we just do:
```
if getAll {
   return errNotEnoughStateSyncs
}

return nil
```
here without variable?

consensus/polybft/state_sync_manager.go
```diff
@@ -409,38 +423,49 @@ func (s *stateSyncManager) buildProofs(commitmentMsg *CommitmentMessage) error {
...
-			// this is a valid case, there is not enough state syncs
-			return nil, nil
-		}
+	if len(stateSyncEvents) < minCommitmentSize {
+		// there is not enough state sync events to build at least the minimum commitment
+		return nil
+	}
 
-		return nil, err
+	if largestPendingCommitment != nil && largestPendingCommitment.ToIndex >= stateSyncEvents[len(stateSyncEvents)-1].ID {
```
igorcrevar: If `largestPendingCommitment` is used only for this check, is it possible toget read of variable on stack and do something like:
```
	if len(s.pendingCommitments) > 0 &&
            s.pendingCommitments[len(s.pendingCommitments)-1].ToIndex >= stateSyncEvents[len(stateSyncEvents)-1].ID { 
            ...
	}
```

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/7473df4bf9f724b43e552512a25260e8ead65fa2

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/7473df4bf9f724b43e552512a25260e8ead65fa2

igorcrevar (approved):

consensus/polybft/state_sync_manager.go
```diff
@@ -319,24 +326,22 @@ type PostEpochRequest struct {
...
-	defer s.lock.Unlock()
+
+	s.pendingCommitments = nil
+	s.validatorSet = req.ValidatorSet
+	s.epoch = req.NewEpochID
 
 	// build a new commitment at the end of the epoch
 	nextCommittedIndex, err := req.SystemState.GetNextCommittedIndex()
 	if err != nil {
 		return err
```
JekaMas, line 354: is it intensional to keep holding lock in case of that return?

goran-ethernal: That's a mistake from my side. Good catch :muscle:

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/a26ea1dc1a24ce4b625357b5d53237f5a149808e

consensus/polybft/consensus_runtime.go
```diff
@@ -237,16 +237,16 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
...
 	)
 
+	// handle commitment and proofs creation
+	if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
+		c.logger.Error("failed to post block state sync", "err", err)
+	}
+
 	// TODO - this condition will need to be changed to recognize that either slashing happened
 	// or epoch reached its fixed size
 	if c.isFixedSizeOfEpochMet(block.Header.Number, c.epoch) {
```
Stefan-Ethernal: Would you like to use local variable instead?
```suggestion
	if c.isFixedSizeOfEpochMet(block.Header.Number, epoch) {
```

consensus/polybft/consensus_runtime.go
```diff
@@ -237,16 +237,16 @@ func (c *consensusRuntime) OnBlockInserted(block *types.Block) {
...
 
-	// handle commitment and proofs creation
-	if err := c.stateSyncManager.PostBlock(&PostBlockRequest{Block: block}); err != nil {
-		c.logger.Error("failed to post block state sync", "err", err)
-	}
-
 	var (
 		epoch = c.epoch
 		err   error
 	)
```
Stefan-Ethernal, line 249: Don't see any benefit in moving this var block. If we are going to move it, I'd just put it at the very beginning of function (below locking).

consensus/polybft/state_sync_manager.go
```diff
@@ -82,7 +84,7 @@ func NewStateSyncManager(logger hclog.Logger, state *State, config *stateSyncCon
 		logger: logger.Named("state-sync"),
```
Stefan-Ethernal: Nitpick
```suggestion
		logger: logger.Named("state_sync_manager"),
```

consensus/polybft/state_sync_manager.go
```diff
@@ -60,14 +62,14 @@ var _ StateSyncManager = (*stateSyncManager)(nil)
...
 	config *stateSyncConfig
 
 	// per epoch fields
-	lock         sync.Mutex
-	commitment   *Commitment
-	epoch        uint64
-	validatorSet *validatorSet
+	lock               sync.RWMutex
+	pendingCommitments []*Commitment
+	validatorSet       *validatorSet
```
Stefan-Ethernal: Nitpick
```suggestion
	validatorSet       ValidatorSet
```

consensus/polybft/state_sync_manager.go
```diff
@@ -82,7 +84,7 @@ func NewStateSyncManager(logger hclog.Logger, state *State, config *stateSyncCon
 		logger: logger.Named("state-sync"),
 		state:  state,
 		config: config,
-		lock:   sync.Mutex{},
+		lock:   sync.RWMutex{},
```
Stefan-Ethernal: Nitpick, mutexes doesn't need to be initialized, but as you wish :slightly_smiling_face: 
```suggestion
```

consensus/polybft/state_sync_manager.go
```diff
@@ -198,42 +200,50 @@ func (s *stateSyncManager) AddLog(eventLog *ethgo.Log) {
 
 		return
 	}
+
+	if err := s.buildCommitment(); err != nil {
+		s.logger.Error("could not build a commitment on arrival of new state sync", "err", err)
```
Stefan-Ethernal: Do you see benefit in logging id of captured state sync event?

consensus/polybft/state.go
```diff
@@ -443,16 +443,22 @@ func (s *State) insertStateSyncEvent(event *types.StateSyncEvent) error {
 }
 
 // getStateSyncEventsForCommitment returns state sync events for commitment
-// if there is an event with index that can not be found in db in given range, an error is returned
-func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64) ([]*types.StateSyncEvent, error) {
+// based on the getAll flag, the function will try to get all events in range, and if it fails,
+// it will either return an error, or return all the events it can without returning an error
+func (s *State) getStateSyncEventsForCommitment(fromIndex, toIndex uint64,
+	getAll bool) ([]*types.StateSyncEvent, error) {
```
Stefan-Ethernal: Consider returning an error in any case and handle it properly from the caller. It would imply removing `getAll` parameter entirely.

If you don't think that removing `getAll` is a good idea, then its naming is a bit misleading, since it denotes whether an error should be returned or not. In that case consider renaming it somehow (`propagateError`, not sure?)...

Stefan-Ethernal (commented):
Generally looks good, but leaving a couple of comments to consider.

Stefan-Ethernal (commented):
Generally looks good, but leaving a couple of comments to consider.

consensus/polybft/state_sync_manager.go
```diff
@@ -22,8 +22,10 @@ import (
 )
 
 const (
-	// number of stateSyncEvents to be processed before a commitment message can be created and gossiped
-	stateSyncCommitmentSize = 10
+	// maximum number of stateSyncEvents that a commitment can have
+	maxCommitmentSize = 10
+	// minimum number of stateSyncEvents that a commitment can have
+	minCommitmentSize = 2
```
ferranbt, line 28: Can you add a comment on why did we choose this number?

consensus/polybft/consensus_runtime.go
goran-ethernal: I don't think this PR changed anything about this var block :confused: . I think it was added at this place because we don't need those variables before the given lines.

consensus/polybft/state_sync_manager.go
```diff
@@ -22,8 +22,10 @@ import (
 )
 
 const (
-	// number of stateSyncEvents to be processed before a commitment message can be created and gossiped
-	stateSyncCommitmentSize = 10
+	// maximum number of stateSyncEvents that a commitment can have
+	maxCommitmentSize = 10
```
ferranbt: Can you make this value a param of the manager? This will likely not be a global variable in the future so it would be better if we do not forget about that (by not making it a global variable).

consensus/polybft/state.go
goran-ethernal: I agree.

consensus/polybft/state_sync_manager.go
```diff
@@ -60,14 +62,14 @@ var _ StateSyncManager = (*stateSyncManager)(nil)
...
-
 	config *stateSyncConfig
 
 	// per epoch fields
-	lock         sync.Mutex
-	commitment   *Commitment
-	epoch        uint64
-	validatorSet *validatorSet
+	lock               sync.RWMutex
+	pendingCommitments []*Commitment
```
ferranbt, line 74: What is the lifecycle of this field? It seems to me that if the process gets shutdown, the system would not recover pending commitments?

ferranbt (commented):
Some general comments.

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/37d13b5aeda743254f04235b861ccd87e76e247f

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/37d13b5aeda743254f04235b861ccd87e76e247f

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/37d13b5aeda743254f04235b861ccd87e76e247f

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/37d13b5aeda743254f04235b861ccd87e76e247f

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/37d13b5aeda743254f04235b861ccd87e76e247f

consensus/polybft/state_sync_manager.go
goran-ethernal: We do not save this anywhere but in memory. When we get back up, we will build new pending commitments on restart epoch (if we get back up, and the chain is in a newer epoch), or we will build new pending commitments on every state sync event that the event tracker gets. In both cases, we might not build the smallest commitments that the other validators have, depending on the state sync events we have and get when event tracker pulls everything we missed.

But I think that is ok, because, we will build larger commitments, and send our votes for them.

ferranbt: Ok, from what I see on the RFC too. If the node gets shutdown it "resets" the `pendingCommitments` in two ways: it either waits for the next log or for the next epoch. Am I right? If that is correct, it is good for me.

goran-ethernal: That is right.

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/5536eee6702ed7cf4159803120aa141c8d1bfce7

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/5536eee6702ed7cf4159803120aa141c8d1bfce7

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1083/commits/9087bc8c572b7cb12fe8259878552b38a814f786

consensus/polybft/state_sync_manager.go
```diff
@@ -22,8 +22,9 @@ import (
 )
 
 const (
-	// number of stateSyncEvents to be processed before a commitment message can be created and gossiped
-	stateSyncCommitmentSize = 10
+	// minimum number of stateSyncEvents that a commitment can have
+	// minimum number is 2 because smart contract expects that the merkle tree has at least two leafs
```
```suggestion
	// (minimum number is 2 because smart contract expects that the merkle tree has at least two leaves)
```

Stefan-Ethernal (approved):
LGTM

ferranbt (approved):

PR #1087, opened 2022-12-30
Fix gasCost field in return of debug trace

Kourin1996:
# Description

Fix wrong `gasCost` field in return value of debug tracing. Previously this value indicated the total consumed gas after the execution of operation but the expected value is the used gas for the execution of the operation. This PR fixes this issue and adds nil check for config argument in debug tracing API.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/state/runtime/evm/state.go
+++ b/state/runtime/evm/state.go
@@ -237,7 +237,7 @@ func (c *state) Run() ([]byte, error) {
 		inst := dispatchTable[op]
 		if inst.inst == nil {
 			c.exit(errOpCodeNotFound)
-			c.captureExecutionError(op.String(), c.ip, gasCopy)
+			c.captureExecutionError(op.String(), c.ip, gasCopy, 0)
 
 			break
 		}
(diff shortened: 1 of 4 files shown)


trimixlover (approved):

state/runtime/evm/state.go
```diff
@@ -237,28 +237,28 @@ func (c *state) Run() ([]byte, error) {
...
+			c.captureExecutionError(op.String(), c.ip, gasCopy, inst.gas)
 
 			break
 		}
 
-		c.captureSuccessfulExecution(op.String(), gasCopy)
+		c.captureSuccessfulExecution(op.String(), gasCopy, gasCopy-c.gas)
 
 		// execute the instruction
 		inst.inst(c)
```
lazartravica: @Kourin1996 are you sure that this order is correct? (:261-:264)

Some opcodes have dynamic gas cost (e.g. [LOG](https://github.com/0xPolygon/polygon-edge/blob/316ce9b408796667e533cdca8871401d833fae81/state/runtime/evm/instructions.go#L1036-L1042)), and the correct gas cost won't be known before executing the instruction itself.

lazartravica (changes requested):
You even managed the fix the missing configuration parameter pretty error, awesome stuff, just please in the future open two separate PRs for this. 

Take a look at my comment for dynamic gas calculation

state/runtime/evm/state.go
Kourin1996: Fixed 51e065166d8c161dcc99b16ab69ab70b8433fb35

lazartravica (approved):
LGTM!

PR #1099, opened 2023-01-06
EVM-338 fix race condition on account initialization

yaziciahmet:
# Description

When a new account sent multiple transactions at the same time, a possible race condition was occurring as follows (using 2 threads example):

1. thread-1 stores an empty account inside `accountsMap`
2. thread-2 checks whether account with address exists, and sees that it does, so it continues without creating an account itself
3. thread-2 continues it's execution thinking account is valid, and calls a function on one of its fields, which causes a nil dereference exception since thread-1 did not set the account's fields yet

This fix carries the field initialization into the store phase. Since `LoadOrStore` function is already thread-safe, there won't be a state where an invalid account exists inside the `accountMap`. And even if multiple threads try to create an account, it is OK since `LoadOrStore` only allows storing once, and `account.init.Do` is executed only once and is thread safe as well.

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -19,20 +19,15 @@ type accountsMap struct {
 
 // Intializes an account for the given address.
 func (m *accountsMap) initOnce(addr types.Address, nonce uint64) *account {
-	a, _ := m.LoadOrStore(addr, &account{})
+	a, _ := m.LoadOrStore(addr, &account{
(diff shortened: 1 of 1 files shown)


yaziciahmet (commented):

yaziciahmet (commented):
recheck

PR #1103, opened 2023-01-09
Implement `Stringer` interface in `AccountSet`

Stefan-Ethernal:
# Description

This PR implements `fmt.Stringer` interface by the `AccountSet`, so that we have prettified logging output for validator accounts.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/validator_metadata.go
+++ b/consensus/polybft/validator_metadata.go
@@ -1,6 +1,7 @@
 package polybft
 
 import (
+	"bytes"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


goran-ethernal (approved):

igorcrevar (approved):

consensus/polybft/validator_metadata.go
```diff
@@ -116,13 +117,24 @@ func (v *ValidatorMetadata) UnmarshalRLPWith(val *fastrlp.Value) error {
...
 
+// fmt.Stringer implementation
+func (as AccountSet) String() string {
+	var buf bytes.Buffer
+	for _, v := range as {
+		buf.WriteString(v.String())
+		buf.WriteString("\n")
+	}
+
+	return buf.String()
```
JekaMas: Could we just use `strings.Join(as, "\n")` ? It also uses strings.Builder with buffer inside.

JekaMas (approved):

consensus/polybft/validator_metadata.go
ferranbt: It also would remove the `\n` at the end of the string.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1103/commits/126ff2ec8265d399c3b9637f8dab164a488a9b38

PR #1104, opened 2023-01-09
EVM-338 & EVM-340 fix race condition on account initialization

yaziciahmet:
# Description

When a new account sent multiple transactions at the same time, a possible race condition was occurring as follows (using 2 threads example):

- thread-1 stores an empty account inside accountsMap
- thread-2 checks whether account with address exists, and sees that it does, so it continues without creating an account itself
- thread-2 continues it's execution thinking account is valid, and calls a function on one of its fields, which causes a nil dereference exception since thread-1 did not set the account's fields yet

This fix carries the field initialization into the store phase. Since LoadOrStore function is already thread-safe, there won't be a state where an invalid account exists inside the accountMap. And even if multiple threads try to create an account, it is OK since LoadOrStore only allows storing once, and account.init.Do is executed only once and is thread safe as well.

# Changes include

# Checklist

## Testing

## Manual Testing

I ran the test which is added in this PR multiple times (more than 50) in both debug mode and with `go test`. Debug mode is done with breakpoints around the lines where data race occurs. 

## Additional Comments

Since this is a data race bug, it is relatively hard to fully encapsulate this case in a unit test. I believe the best that can be done is to send batch of transactions from single account in parallel with the hope of encountering this specific case, which seems pretty high-likely according to the testing I did manually. I added such test to PR.

One weird behavior I observed was that when I run the test in the unchanged version with `go test`, it's always success. Same goes for testing in debug mode with no breakpoints. But when I debug the test with breakpoints around the lines which data race occurs, it almost always fails.

This could be expected to some extent since breakpoints changes behavior and might cause priority changes on threads. But it is weird that I literally got error 0 times when I did it with `go test` and got error 99% of the time when I did it in debug mode with breakpoints.

PS: I ran `go clean -testcache` after every `go test` as well.

For clarification, I also tested with the updated version, everything worked fine with all above mentioned cases.

### DIFF
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -19,18 +19,18 @@ type accountsMap struct {
 
 // Intializes an account for the given address.
 func (m *accountsMap) initOnce(addr types.Address, nonce uint64) *account {
-	a, _ := m.LoadOrStore(addr, &account{
+	a, loaded := m.LoadOrStore(addr, &account{
 		enqueued:    newAccountQueue(),
 		promoted:    newAccountQueue(),
 		maxEnqueued: m.maxEnqueuedLimit,
 		nextNonce:   nonce,
 	})
 	newAccount := a.(*account) //nolint:forcetypeassert
-	// run only once
-	newAccount.init.Do(func() {
-		// update global count
+
+	if !loaded {
+		// update global count if it was a store
 		atomic.AddUint64(&m.count, 1)
-	})
+	}
 
 	return newAccount
 }
@@ -148,7 +148,6 @@ func (m *accountsMap) allTxs(includeEnqueued bool) (
 // indicating the account's enqueued transaction(s)
 // are ready to be moved to the promoted queue.
 type account struct {
-	init               sync.Once
 	enqueued, promoted *accountQueue
 	nextNonce          uint64
 	demotions          uint64
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -19,20 +19,15 @@ type accountsMap struct {
 
 // Intializes an account for the given address.
 func (m *accountsMap) initOnce(addr types.Address, nonce uint64) *account {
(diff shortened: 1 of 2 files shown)


txpool/account.go
```diff
@@ -19,20 +19,15 @@ type accountsMap struct {
 
 // Intializes an account for the given address.
 func (m *accountsMap) initOnce(addr types.Address, nonce uint64) *account {
-	a, _ := m.LoadOrStore(addr, &account{})
+	a, _ := m.LoadOrStore(addr, &account{
+		enqueued:    newAccountQueue(),
```
JekaMas, line 23: Do we have a test to check that the fix helped?

yaziciahmet:

yaziciahmet:
recheck

txpool/account.go
yaziciahmet: Added it 9d62607b5734071734f241fff023aab89c7a9068

Please also check the *Additional Comments* section in the description.

txpool/account.go
```diff
@@ -19,20 +19,15 @@ type accountsMap struct {
 
 // Intializes an account for the given address.
 func (m *accountsMap) initOnce(addr types.Address, nonce uint64) *account {
-	a, _ := m.LoadOrStore(addr, &account{})
+	a, _ := m.LoadOrStore(addr, &account{
```
Kourin1996: In order to increment count only, you can see 2nd returns of `LoadOrStore` instead of using `init.Do`

yaziciahmet: Done https://github.com/0xPolygon/polygon-edge/pull/1104/commits/3f2cba1bf1a10aa113cf4647ee9714a93d1b7754

Kourin1996 (approved):

trimixlover (approved):

PR #1107, opened 2023-01-09
Add disclaimer message when running polybft consensus protocol

Stefan-Ethernal:
# Description

PR adds disclaimer message when running polybft consensus protocol (both when running a polybft server and generating polybft genesis).

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/polybft.go
+++ b/consensus/polybft/polybft.go
@@ -213,7 +213,7 @@ func (p *Polybft) Initialize() error {
 
 // Start starts the consensus and servers
 func (p *Polybft) Start() error {
-	fmt.Println(DisclaimerMessage)
+	p.logger.Warn(DisclaimerMessage)
 	p.logger.Info("starting polybft consensus", "signer", p.key.String())
 
 	// start syncer (also initializes peer map)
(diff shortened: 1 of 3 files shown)


consensus/polybft/polybft.go
```diff
@@ -211,6 +213,7 @@ func (p *Polybft) Initialize() error {
 
 // Start starts the consensus and servers
 func (p *Polybft) Start() error {
+	fmt.Println(DisclaimerMessage)
```
vcastellm: Should we print using logger?

Stefan-Ethernal: Had same doubt to be honest, but if you have it too, then probably the logger it is :slightly_smiling_face:

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1107/commits/d09c14a53978821b32e51817a3d429e276620d76
Since genesis cmd doesn't have a logger instance, disclaimer is printed as part of CLI output.

goran-ethernal (approved):

vcastellm (approved):
LGTM I think it's not necessary to show it on every command.

ferranbt (approved):

PR #1110, opened 2023-01-11
Return FullBlock from syncer

goran-ethernal:
# Description

This PR modifies the `OnBlockInserted` callback to `consensus engine` from `syncer` so that it returns a `types.FullBlock` (which contains transaction receipts as well, not just the block).

This is needed so that the `consensus engine` has all the necessary data for `bridge` to properly function, specifically exit workflow, since currently we are only extracting exit events when a block is inserted through consensus.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -679,44 +679,48 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
 // outside the method call
 func (b *Blockchain) VerifyPotentialBlock(block *types.Block) error {
 	// Do just the initial block verification
-	return b.verifyBlock(block)
+	_, err := b.verifyBlock(block)
+
+	return err
 }
 
 // VerifyFinalizedBlock verifies that the block is valid by performing a series of checks.
 // It is assumed that the block status is sealed (committed)
-func (b *Blockchain) VerifyFinalizedBlock(block *types.Block) error {
+func (b *Blockchain) VerifyFinalizedBlock(block *types.Block) (*types.FullBlock, error) {
 	// Make sure the consensus layer verifies this block header
 	if err := b.consensus.VerifyHeader(block.Header); err != nil {
-		return fmt.Errorf("failed to verify the header: %w", err)
+		return nil, fmt.Errorf("failed to verify the header: %w", err)
 	}
 
 	// Do the initial block verification
-	if err := b.verifyBlock(block); err != nil {
-		return err
+	receipts, err := b.verifyBlock(block)
+	if err != nil {
+		return nil, err
 	}
 
-	return nil
+	return &types.FullBlock{Block: block, Receipts: receipts}, nil
 }
 
 // verifyBlock does the base (common) block verification steps by
 // verifying the block body as well as the parent information
-func (b *Blockchain) verifyBlock(block *types.Block) error {
+func (b *Blockchain) verifyBlock(block *types.Block) ([]*types.Receipt, error) {
 	// Make sure the block is present
 	if block == nil {
-		return ErrNoBlock
+		return nil, ErrNoBlock
 	}
 
 	// Make sure the block is in line with the parent block
 	if err := b.verifyBlockParent(block); err != nil {
-		return err
+		return nil, err
 	}
 
 	// Make sure the block body data is valid
-	if err := b.verifyBlockBody(block); err != nil {
-		return err
+	receipts, err := b.verifyBlockBody(block)
+	if err != nil {
+		return nil, err
 	}
 
-	return nil
+	return receipts, nil
 }
 
 // verifyBlockParent makes sure that the child block is in line
@@ -774,7 +778,7 @@ func (b *Blockchain) verifyBlockParent(childBlock *types.Block) error {
 // - The trie roots match up (state, transactions, receipts, uncles)
 // - The receipts match up
 // - The execution result matches up
-func (b *Blockchain) verifyBlockBody(block *types.Block) error {
+func (b *Blockchain) verifyBlockBody(block *types.Block) ([]*types.Receipt, error) {
 	// Make sure the Uncles root matches up
 	if hash := buildroot.CalculateUncleRoot(block.Uncles); hash != block.Header.Sha3Uncles {
 		b.logger.Error(fmt.Sprintf(
@@ -783,7 +787,7 @@ func (b *Blockchain) verifyBlockBody(block *types.Block) error {
 			block.Header.Sha3Uncles,
 		))
 
-		return ErrInvalidSha3Uncles
+		return nil, ErrInvalidSha3Uncles
 	}
 
 	// Make sure the transactions root matches up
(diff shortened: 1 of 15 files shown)


consensus/polybft/fsm.go
```diff
@@ -406,14 +406,13 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 }
 
 // Insert inserts the sealed proposal
-func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) (*types.Block, error) {
-	newBlock := f.target.Block
-	receipts := f.target.Receipts
+func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) (*types.FullBlock, error) {
+	newBlock := &types.FullBlock{Block: f.target.Block, Receipts: f.target.Receipts}
```
ferranbt:
```suggestion
	newBlock := f.target
```

ferranbt: Are any of the changes in this function necessary or just code style?

goran-ethernal: Since `OnBlockInserted` is called both from `consensus` and `syncer`, I changed the `fsm.Insert` function to return a `FullBlock`, since we will need receipts from both sources in the `OnBlockInserted` (to get the exit events).

b00ris (approved):

consensus/polybft/fsm.go
goran-ethernal: Fix for the suggested change: https://github.com/0xPolygon/polygon-edge/pull/1110/commits/e8ebe7f15c2119f613e8be4f2e463362977c3d1a

Stefan-Ethernal (approved):

blockchain/blockchain.go
```diff
@@ -902,6 +902,15 @@ func (b *Blockchain) WriteFullBlock(fblock *types.FullBlock, source string) erro
 		return err
 	}
 
+	if fblock.Receipts == nil { // Fetch the block receipts from cache if not already set
+		blockReceipts, receiptsErr := b.extractBlockReceipts(block)
```
ferranbt: I do not want to do this. The last performance bug was due to not being able to keep track of when was `extractBlockReceipts` being used, which is why `FullBlock` and this method were introduced.

The `extractBlockReceipts` method will eventually get deprecated once `ibft` is removed.

Also, the initial logic of the function changes a lot, from inserting only into the storage the block data entries to maybe also compute the full block. When is this necessary?

syncer/syncer.go
```diff
@@ -242,11 +242,13 @@ func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.B
 				return lastReceivedNumber, false, fmt.Errorf("unable to verify block, %w", err)
 			}
 
-			if err := s.blockchain.WriteBlock(block, syncerName); err != nil {
+			fullBlock := &types.FullBlock{Block: block}
+
+			if err := s.blockchain.WriteFullBlock(fullBlock, syncerName); err != nil {
```
ferranbt, line 246: Could we execute the block here explicitly, return a `FullBlock` and then call `WriteFullBlock`?

blockchain/blockchain.go
goran-ethernal: I didn't want to change the `blockchain.go` a lot just because of the `syncer`, so that function you added looked as the easiest place to change, also with the smallest amount of changes as well.

I can do it another way, but it will require changing `blockchain.go` other functions.

ferranbt (changes requested):
Overall looks good, only major thing is to remove the functionality from `WriteFullBlock` of executing the `Block` itself and move that to a custom function (this is, split what `WriteBlock` does in two steps).

blockchain/blockchain.go
goran-ethernal: Ok, changed this a bit, as per our discussion. Check it out: https://github.com/0xPolygon/polygon-edge/pull/1110/commits/51ac590f0b5f1539a72ed8f9f4127c0bbcbcdec7

syncer/syncer.go
goran-ethernal: Ok, changed this a bit, as per our discussion. Check it out: https://github.com/0xPolygon/polygon-edge/pull/1110/commits/51ac590f0b5f1539a72ed8f9f4127c0bbcbcdec7

igorcrevar (approved):

blockchain/blockchain.go
```diff
@@ -881,64 +885,27 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
...
-
-	if prevHeader, ok := b.GetHeaderByNumber(header.Number - 1); ok {
-		diff := header.Timestamp - prevHeader.Timestamp
-		logArgs = append(logArgs, "generation_time_in_seconds", diff)
-	}
-
-	b.logger.Info("new block", logArgs...)
-
-	return nil
+	return b.writeBlockInternal(block, source,
```
ferranbt: What is the benefit of having this with the callback? Having fewer lines of code and merging both functions? As I mentioned, we are transitioning from ibft (in prod) to polybft (in dev).

In the near future, `WriteBlock` is going to go away, it is only here right now because `ibft` relies on that (and its internal cache to some extend). This callback method overcomplicates the logic to simplify a code that will be legacy pretty soon.

Were these changes required for these two functions at all?

ferranbt: I see this was part of this commit 62166e6877f5a6a6f4a780d7047511caa530b8df. Could we revert it?

Stefan-Ethernal: If it is going to be legacy soon or not, that is a separate topic. However having ~90% same code in two functions isn't good coding practice at all (hope we agree on that). That's why I opted to change it anyway, because I don't see it makes code any harder for maintenance and following.

Anyhow, your call, we can drop that commit easily if you don't like it :slightly_smiling_face:

ferranbt: Yes, I would prefer if we can drop it. I wrote the `WriteFullBlock` function and I was aware that I copied much of that function, my mindset for that was similar to what I described in the comment:

- `WriteBlock` uses a tricky cache (`receiptsCache`) that might have some hidden flows. Being unaware of those is what created some of the performance issues over the holidays that were fixed with `WriteFullBlock`.
- I was wary that changing `WriteBlock` would create issues on the `ibft` consensus protocol in "production", also because of hidden flows that we might not be aware.
- I wanted to have as few changes as possible on `ibft` consensus protocol (I thought about merging the functionality of `WriteFullBlock` inside `WriteBlock` but I would need some flags and changing the function calls).
- Once `ibft` gets deprecated, we can remove it and make `WriteFullBlock` the correct workflow (and remove the cache). See RFC-144 for more details on that.

Stefan-Ethernal: Ok hopefully we are going to deprecate `WriteBlock` any time soon, because my concern is that it is going to be there forever and it is not nice having two (almost) same implementations :slightly_smiling_face:

goran-ethernal: Reverted: https://github.com/0xPolygon/polygon-edge/pull/1110/commits/c7d6bdb11957cae8d473049af496a17dc223ca71

ferranbt (approved):

vcastellm (commented):
Late LGTM

PR #1111, opened 2023-01-11
Introduce dynamic fee tx type

begmaroman:
# Description

As part of EIP-1559 implementation, the dynamic fee t type should be introduced. This PR does not update any business logic, but just introduces a new tx type and updates rlp marshaling and unmarshaling logic accordingly. 

- Added dynamic fee tx type
- Added gas fee cap and gas tip cap
- Updated rlp marshaling and unmarshaling logic by adding these two fields for the new tx type 
- Made small code cleanup 

# Changes include

# Checklist

## Testing


types/tx.go
```diff
@@ -0,0 +1,102 @@
...
+	default:
+		return "UnknownTX"
+	}
+}
+
+// Tx represents the general transaction behavior interface.
+// It contains common transaction field getters and setters as
+// well as additional functions for managing transaction.
+// Any modification function should return modified transaction object.
+type Tx interface {
```
begmaroman: gasFeeCap and gasTipCap will be introduced later

vcastellm (commented):
I like it, but why not include the set of common methods in the `txBase` struct?

begmaroman:
> I like it, but why not include the set of common methods in the `txBase` struct?

They are included already. Basically, some methods should be specific per tx type so `txBase` does not implement `Tx` interface, and setters could not be placed within `txBase` because they return Tx object. All possible common methods are defined per `txBase` type.

types/tx_dynamic_fee.go
```diff
@@ -0,0 +1,93 @@
...
+
+// SetGas sets the given transaction gas.
+func (t *txDynamicFee) SetGas(gas uint64) Tx {
+	t.gas = gas
+
+	return t
+}
+
+// SetTo sets the given transaction receiver address.
+func (t *txDynamicFee) SetTo(addr Address) Tx {
```
b00ris, line 32: tx.Addr can be nil. It's using to create a contract. Using this signature we cant set To to nil.

types/tx.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// SetGas sets the given transaction gas.
+	SetGas(gas uint64) Tx
+
+	// To returns transaction receiver address.
+	// It could be nil in case of the state transaction.
+	To() *Address
+
+	// SetTo sets the given transaction receiver address.
+	SetTo(addr Address) Tx
```
b00ris, line 57: I'm trying to finish reviews for all rfc/pr's that required my review in december. It is my tech dept, because i was overloaded and exhausted in december.

b00ris: Setting addr to nil is a requirement for transaction. We can't set addr to nil using this signature.

begmaroman: But it is nil by default. Does that not mean if we want to set an address, it should have some value? Could you point me to some place in the code where we explicitly set address to nil please?

Stefan-Ethernal: Well whenever we deploy smart contract, `To` field should be nil, no? 

However I see your point @begmaroman, `to` field in the transaction itself is reference type and therefore will be `nil` by default if unset. The question which arises is: is there any use case where we would want to set it to nil after we have created transaction instance. I cannot think of it atm.

begmaroman:
As discussed per slack, we would go without a transaction interface atm and would be able to introduce it later if it is really needed. Currently, it might bring more complexity to the code. Let me share more thoughts:
- A common transaction object has the same fields regardless of transaction type. There could be some small differences (gas tip cap, gas fee cap) but overall behavior and model are the same.
- All transaction types have almost the same behavior. There could be some minor differences but it could be easily handled by checking tx type. For example, from address should be nil for state tx; dynamic gas fee tx must have gas fee cap and gas tip cap.
- Most of the places in the code expects specific transaction type so having an interface is not really needed in most of these places.
- 
So it seems like it would be more difficult to support interface because it would bring more complexity to the code and maintenance. Even if we need to have some transaction-type-specific logic, it would be wrapped into a separate function within 'types' package.

types/tx.go
begmaroman: Update: https://github.com/0xPolygon/polygon-edge/pull/1111#issuecomment-1380101310

types/rlp_unmarshal.go
```diff
@@ -301,86 +301,131 @@ func (t *Transaction) UnmarshalRLP(input []byte) error {
...
+	t.GasTipCap = new(big.Int)
+	t.Value = new(big.Int)
+	t.V = new(big.Int)
+	t.R = new(big.Int)
+	t.S = new(big.Int)
+	t.From = ZeroAddress
+
+	// Type
+	i := 0
+	if t.Type, err = ReadRlpTxType(v.Get(i)); err != nil {
```
Stefan-Ethernal: I feel that it is better to be explicit here and don't rely on variable which gets incremented (but provide exact index, as it was before). That way it is easier to follow code, since it must follow order from marshal.
Also it is easier to make a bug in case we add new field but forget to increment `i` some day... :slightly_smiling_face:  

WDYT?

types/types.go
```diff
@@ -9,14 +9,24 @@ import (
...
+	ZeroHash    = Hash{}
+)
+
+var (
+	// EmptyRootHash is the root when there are no transactions
+	EmptyRootHash = StringToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
+
+	// EmptyUncleHash is the root when there are no uncles
+	EmptyUncleHash = StringToHash("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347")
+)
```
Stefan-Ethernal: Does it make sense to merge these two blocks? (though I'm ok with having it separated as well)

```suggestion
var (
	ZeroAddress = Address{}
	ZeroHash    = Hash{}

	// EmptyRootHash is the root when there are no transactions
	EmptyRootHash = StringToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")

	// EmptyUncleHash is the root when there are no uncles
	EmptyUncleHash = StringToHash("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347")
)
```

types/rlp_unmarshal.go
```diff
@@ -301,86 +301,131 @@ func (t *Transaction) UnmarshalRLP(input []byte) error {
...
+			// address
+			addr := BytesToAddress(vv)
+			t.From = addr
 		}
 	}
 
 	return nil
 }
+
+func ReadRlpTxType(rlpValue *fastrlp.Value) (TxType, error) {
```
Stefan-Ethernal: Do we need this exported?

types/rlp_unmarshal.go
Stefan-Ethernal: EDIT:
Now I realize it is not as convenient to rely on constant values, because of this if statement https://github.com/0xPolygon/polygon-edge/pull/1111/files#diff-efc5999399a6ff1082b29a52ce57fd7bfd6ebe4789d686a44c70523d2a61c684R340-R352 :thinking:

begmaroman: Basically, for non-dynamic-gas-fee tx type we won't have gas fee cap and gas tip cap fields in the raw message so that's why indices could be different depending on the tx type. It won't work with the way when we explicitly specify an index until we add these two fields for any tx type, but it would make the message bigger so which would affect the network performance. Maybe not big impact, but still.

begmaroman: Also, if someone forget to increment this variable, most likely some tests would fail.

Stefan-Ethernal: Also another point:
is it safe to introduce new fields in the middle of existing ones? It is not backward compatible (e.g. we have introduced `Type` as the first field and moved all the other existing fields by one).

begmaroman: Another option is to place these two fields at the end of the message. It could work for us. Let me do that.

types/rlp_unmarshal.go
```diff
@@ -301,86 +301,131 @@ func (t *Transaction) UnmarshalRLP(input []byte) error {
 
 // UnmarshalRLPFrom unmarshals a Transaction in RLP format
 func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
-	elems, err := v.GetElems()
-	if err != nil {
-		return err
+	if v.Len() < 10 {
+		return fmt.Errorf("incorrect number of elements to decode transaction, expected 10 but found %d", v.Len())
 	}
```
Stefan-Ethernal: Should we cover situation where `Type` isn't set (and assume it is `LegacyTx`)? 

I am not entirely sure should we take care about backward compatibility or not, that's why questions like this arise? :slightly_smiling_face:

begmaroman: I've updated marshaling logic so we set tx type in any way now in order to be more accurate

types/rlp_unmarshal.go
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/1111/commits/6d99243429dec0ab1bf3c1a3f63e4a8145941d5c

types/types.go
begmaroman: Done

types/rlp_unmarshal.go
begmaroman: Done

types/rlp_unmarshal.go
Stefan-Ethernal: Yeah, but I envision situation where we would have some block which is created before having EIP-1559 transactions and such transactions wouldn't have marshaled `Type` field.

You are returning [here](https://github.com/0xPolygon/polygon-edge/pull/1111/files#diff-efc5999399a6ff1082b29a52ce57fd7bfd6ebe4789d686a44c70523d2a61c684R405-R411) `LegacyTx` and it is correct, but also an error (and we shouldn't treat it as error in case of "legacy" transactions). So I'd say we should consider `Type` field optional. However as I said: if we are going to run everything from scratch and not worry about backward compatibility, then we are good to go with this.

JekaMas (approved):

types/rlp_unmarshal.go
```diff
@@ -301,86 +301,121 @@ func (t *Transaction) UnmarshalRLP(input []byte) error {
 
 // UnmarshalRLPFrom unmarshals a Transaction in RLP format
 func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
-	elems, err := v.GetElems()
-	if err != nil {
-		return err
+	if v.Len() < 10 {
```
b00ris: Why 10? Could you add more context there?

Stefan-Ethernal: Because `Transaction` struct has (at least) 10 fields marshaled :slightly_smiling_face: 
DynamicFeeTx has even more than 10, 12 if I'm correct (all the fields as `LegacyTx` + `GasTipCap` and `GasFeeCap`).

begmaroman: Correct

types/rlp_unmarshal.go
begmaroman: Good point. According to the clack conv, we have to be backward compatible. Done.

vcastellm (commented):
`DynamicGeeTx` => `DynamicFeeTx`? 😄

begmaroman:
@vcastellm fixed :)

vcastellm (approved):
LGTM
G
T
M

b00ris (approved):

types/rlp_unmarshal.go
```diff
@@ -312,75 +312,120 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
+			t.GasFeeCap = new(big.Int)
+			t.GasTipCap = new(big.Int)
+
+			// gasFeeCap
+			if err = elems[10].GetBigInt(t.GasFeeCap); err != nil {
+				return err
+			}
+
+			// gasTipCap
+			if err = elems[11].GetBigInt(t.GasTipCap); err != nil {
```
Stefan-Ethernal: You didn't want to simplify like you did it for other `*big.Int` fields?

```suggestion
			// gasFeeCap
			if err = elems[10].GetBigInt(t.GasFeeCap); err != nil {
				return err
			}

			// gasTipCap
			if err = elems[11].GetBigInt(t.GasTipCap); err != nil {
```

Stefan-Ethernal (approved):
LGTM, just leaving minor suggestion

types/rlp_unmarshal.go
```diff
@@ -312,75 +312,120 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
 		// address
 		addr := BytesToAddress(vv)
 		t.To = &addr
-	} else {
-		// reset To
-		t.To = nil
 	}
+
 	// value
-	t.Value = new(big.Int)
```
ferranbt, line 338: I would prefer to keep the declaration before the RLP set like this unless there is a problem with that.

types/rlp_unmarshal.go
```diff
@@ -312,75 +312,120 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
-	t.S = new(big.Int)
 	if err = elems[8].GetBigInt(t.S); err != nil {
 		return err
 	}
 
-	if len(elems) >= 10 {
+	if len(elems) > 9 {
 		// Type
-		if t.Type, err = ReadRlpTxType(elems[9]); err != nil {
+		if t.Type, err = readRlpTxType(elems[9]); err != nil {
```
ferranbt, line 369: I do not think this is correct. The type of the transaction is not on element 9 as far as I know but at the beginning of the transaction.

Stefan-Ethernal: It is correct, because Roman changed marshalling logic, to be at index=9, so that we are backward compatible:
https://github.com/0xPolygon/polygon-edge/pull/1111/files#diff-7f00e74f75ee3a971d40d6f05e54168503eb91a38717d7ed211dafab18ef478cR194.

Although since from what I can see, we are not explicitly setting `LegacyTx`, so in that case, maybe we can put `Type` back to index=0?

ferranbt: By backward compatible, we meant with the previous polybft versions which have this enabled? Ok, what I mean is not that but that there is an official specification to encode an Ethereum transaction with a type (https://eips.ethereum.org/EIPS/eip-2718). I do not think this encoding (or for that matter) or the previous one follows that format.

Stefan-Ethernal: > By backward compatible, we meant with the previous polybft versions which have this enabled?

By backward compatibility I was thinking about transactions which were created and are present in blocks before we've introduced this change and `Type` field in the first place (doesn't matter is it IBFT or PolyBFT). I would imagine we would encounter a use case where some end-users would like to still run IBFT and keep running with "old" transaction data format. Am I correct there?

begmaroman: @ferranbt i didn't change the order of elements in rlp marshaling. Just added gas fee cap and gas tip cap to the end. I guess it would be better to update these things within a separate PR just to keep things more accurate then. Wdyt? 
And yes, I though the same as @Stefan-Ethernal regarding backward compatibility. Does that mean that we can change the order then @ferranbt ?

types/rlp_unmarshal.go
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/1111/commits/e4bba66b1a9a52de5025843186abd58460d72146

types/rlp_unmarshal.go
begmaroman: Done: https://github.com/0xPolygon/polygon-edge/pull/1111/commits/e4bba66b1a9a52de5025843186abd58460d72146

begmaroman:
Blocked by https://github.com/0xPolygon/polygon-edge/pull/1114

PR #1113, opened 2023-01-12
EVM-356 Update checkpoint out of fsm

goran-ethernal:
# Description

Moved the `exitEvent` handling from `fsm` to `OnBlockInserted` callback in `consensusRuntime`.

In `OnBlockInserted` `checkpointManager` will handle new exit events that happened in the newly finalized block inside the `PostBlock` function. This will insert all the exit events from the given block to `boltDb`.

If the block is epoch ending block, in the `PostEpoch` function, `checkpointManager` will update the exit events to point to the next epoch (as per RFC-30). This is only done for the exit events that happened in the epoch ending blocks, since we can not ensure that they were finalized, and to prevent malicious attacks (exits).

**Importan note**: This PR solves the issue where exit events were not gotten from the blocks gotten from `syncer`. This way we will not miss any exit events that happened if node was down or not synced with network for a given amount of time.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -35,6 +35,30 @@ var (
 	defaultCheckpointsOffset = uint64(900)
 )
 
+type CheckpointManager interface {
+	PostBlock(req *PostBlockRequest) error
+	PostEpoch(req *PostEpochRequest) error
+	IsCheckpointBlock(blockNumber uint64, isEpochEndingBlock bool) bool
+	SubmitCheckpoint(header *types.Header, isEndOfEpoch bool) error
+	SetLatestCheckpointID(checkpointNumber uint64)
+}
+
+var _ CheckpointManager = (*dummyCheckpointManager)(nil)
+
+type dummyCheckpointManager struct{}
+
+func (d *dummyCheckpointManager) PostBlock(req *PostBlockRequest) error { return nil }
+func (d *dummyCheckpointManager) PostEpoch(req *PostEpochRequest) error { return nil }
+func (d *dummyCheckpointManager) IsCheckpointBlock(blockNumber uint64, isEpochEndingBlock bool) bool {
+	return false
+}
+func (d *dummyCheckpointManager) SubmitCheckpoint(header *types.Header, isEndOfEpoch bool) error {
+	return nil
+}
+func (d *dummyCheckpointManager) SetLatestCheckpointID(checkpointNumber uint64) {}
+
+var _ CheckpointManager = (*checkpointManager)(nil)
+
 // checkpointManager encapsulates logic for checkpoint data submission
 type checkpointManager struct {
 	// key is the identity of the node submitting a checkpoint
@@ -98,8 +122,13 @@ func (c *checkpointManager) getLatestCheckpointBlock() (uint64, error) {
 	return latestCheckpointBlockNum, nil
 }
 
-// submitCheckpoint sends a transaction with checkpoint data to the rootchain
-func (c *checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfEpoch bool) error {
+// SetLatestCheckpointID sets the latestCheckpointID field
+func (c *checkpointManager) SetLatestCheckpointID(checkpointNumber uint64) {
+	c.latestCheckpointID = checkpointNumber
+}
+
+// SubmitCheckpoint sends a transaction with checkpoint data to the rootchain
+func (c *checkpointManager) SubmitCheckpoint(latestHeader *types.Header, isEndOfEpoch bool) error {
 	lastCheckpointBlockNumber, err := c.getLatestCheckpointBlock()
 	if err != nil {
 		return err
@@ -158,7 +187,7 @@ func (c *checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfE
 			continue
 		}
 
-		if err = c.encodeAndSendCheckpoint(txn, *parentHeader, *parentExtra, true); err != nil {
+		if err = c.encodeAndSendCheckpoint(txn, parentHeader, parentExtra, true); err != nil {
 			return err
 		}
 
@@ -176,13 +205,13 @@ func (c *checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfE
 		}
 	}
 
-	return c.encodeAndSendCheckpoint(txn, latestHeader, *currentExtra, isEndOfEpoch)
+	return c.encodeAndSendCheckpoint(txn, latestHeader, currentExtra, isEndOfEpoch)
 }
 
 // encodeAndSendCheckpoint encodes checkpoint data for the given block and
 // sends a transaction to the CheckpointManager rootchain contract
 func (c *checkpointManager) encodeAndSendCheckpoint(txn *ethgo.Transaction,
-	header types.Header, extra Extra, isEndOfEpoch bool) error {
+	header *types.Header, extra *Extra, isEndOfEpoch bool) error {
 	c.logger.Debug("send checkpoint txn...", "block number", header.Number)
 
 	nextEpochValidators := AccountSet{}
@@ -220,7 +249,7 @@ func (c *checkpointManager) encodeAndSendCheckpoint(txn *ethgo.Transaction,
 }
 
 // abiEncodeCheckpointBlock encodes checkpoint data into ABI format for a given header
-func (c *checkpointManager) abiEncodeCheckpointBlock(blockNumber uint64, blockHash types.Hash, extra Extra,
+func (c *checkpointManager) abiEncodeCheckpointBlock(blockNumber uint64, blockHash types.Hash, extra *Extra,
 	nextValidators AccountSet) ([]byte, error) {
 	aggs, err := bls.UnmarshalSignature(extra.Committed.AggregatedSignature)
 	if err != nil {
@@ -252,10 +281,11 @@ func (c *checkpointManager) abiEncodeCheckpointBlock(blockNumber uint64, blockHa
 	return submitCheckpointMethod.Encode(params)
 }
 
-// isCheckpointBlock returns true for blocks in the middle of the epoch
+// IsCheckpointBlock returns true for blocks in the middle of the epoch
 // which are offset by predefined count of blocks
-func (c *checkpointManager) isCheckpointBlock(blockNumber uint64) bool {
-	return blockNumber == c.latestCheckpointID+c.checkpointsOffset
+// or if given block is an epoch ending block
+func (c *checkpointManager) IsCheckpointBlock(blockNumber uint64, isEpochEndingBlock bool) bool {
+	return isEpochEndingBlock || blockNumber == c.latestCheckpointID+c.checkpointsOffset
 }
 
 // PostBlock is called on every insert of finalized block (either from consensus or syncer)
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -261,15 +261,8 @@ func (c *checkpointManager) isCheckpointBlock(blockNumber uint64) bool {
 // PostBlock is called on every insert of finalized block (either from consensus or syncer)
 // It will read any exit event that happened in block and insert it to state boltDb
 func (c *checkpointManager) PostBlock(req *PostBlockRequest) error {
-	epoch := req.Epoch
-	if req.IsEpochEndingBlock {
-		// exit events that happened in epoch ending blocks,
-		// should be added to the tree of the next epoch
-		epoch++
-	}
-
 	// commit exit events only when we finalize a block
-	events, err := getExitEventsFromReceipts(epoch, req.FullBlock.Block.Number(), req.FullBlock.Receipts)
+	events, err := getExitEventsFromReceipts(req.Epoch, req.FullBlock.Block.Number(), req.FullBlock.Receipts)
 	if err != nil {
 		return err
 	}
@@ -281,6 +274,14 @@ func (c *checkpointManager) PostBlock(req *PostBlockRequest) error {
 	}
 
 	return nil
+}
+
+// PostEpoch notifies the checkpoint manager that an epoch has changed,
+// so that it can discard any previous epoch commitments, and build a new one (since validator set changed)
+func (c *checkpointManager) PostEpoch(req *PostEpochRequest) error {
+	// update any exit events that happened in the last block of previous epoch and move them to the current one
+	// this is done to make sure we know they are finalized properly
+	return c.state.updateExitEvents(req.FirstBlockOfEpoch-1, req.NewEpochID-1, req.NewEpochID)
 }
 
 // getExitEventsFromReceipts parses logs from receipts to find exit events
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -3,9 +3,11 @@ package polybft
 import (
 	"fmt"
 	"math/big"
+	"sort"
 	"strconv"
 
 	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	"github.com/0xPolygon/polygon-edge/contracts"
 	"github.com/0xPolygon/polygon-edge/txrelayer"
 	"github.com/0xPolygon/polygon-edge/types"
 	metrics "github.com/armon/go-metrics"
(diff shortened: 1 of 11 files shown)


consensus/polybft/state_sync_manager.go
```diff
@@ -362,14 +362,18 @@ func (s *stateSyncManager) PostEpoch(req *PostEpochRequest) error {
 }
 
 type PostBlockRequest struct {
-	// Block is a reference of the executed block
-	Block *types.Block
+	// FullBlock is a reference of the executed block
```
ferranbt: Early comment, can you move these functions (i.e. `PostX` to a new `handlers.go` file? Now they are going to be generic for different modules.

I am open to a better name. We can leave that for the PR review.

consensus/polybft/state_sync_manager.go
```diff
@@ -362,14 +362,18 @@ func (s *stateSyncManager) PostEpoch(req *PostEpochRequest) error {
...
+	FullBlock *types.FullBlock
+	// Epoch is the epoch number of the executed block
+	Epoch uint64
+	// IsEpochEndingBlock indicates if this is the last block in epoch
+	IsEpochEndingBlock bool
```
ferranbt: Note that per the RFC doc, there is already another handler for `PostEpoch`. Would that work for this use case?

ferranbt: From what I understand about the problem, a checkpoint is active in two parts: for each block it has to read the receipts and at the end of the epoch it has to send the checkpoint.

goran-ethernal: That is something I wanted to write in the PR description. I opted this for simplicity and to avoid two writes to `boltDb`. Basically if an exit event happens in the last block in epoch, we won't know if it was finalized, so to avoid errors, and possible malicious exits, we will include all exits that happened at the epoch ending blocks in the next epoch (they will be part of exit root of the first block in next epoch). Since `consensus_runtime` already knows on `OnBlockInserted` if the block is an epoch ending block, I just added this field, so that I can save the exit events right away in `boltDb` as part of the next epoch.

Alternative to this is to remove this field, and on `PostBlock` just save these events in the epoch they occurred. But, on the `PostEpoch`, `checkpointManager` would need to see if there are any exit events saved in the last block of previous epoch, and update them in `boltDb` to the new epoch. This is doable, but it means we would have additional update transactions to `boltDb`.

consensus/polybft/state_sync_manager.go
goran-ethernal: Sure. New file for these structs is definitely better.

consensus/polybft/state_sync_manager.go
ferranbt: My main concern is not writing multiple times to boltdb. We kind of do it right now already and later on we are going to optimize that with a boltdb transaction.

I am wary that we are going to make some general assumptions about how the modules work (i.e. they update the epoch only with PostEpoch functions) and optimize for that on `consensus_runtime` but we cannot because each module uses the handles in a different way which is what we wanted to avoid since the beginning with RFC-176.

goran-ethernal: I agree, it should be uniformed. I'll change this and implement the alternative approach I mentioned in previous comment, where we will handle this in `PostEpoch` implementation in `checkpointManager`.

ferranbt: Let's do that and If you feel it gets tricky let's revisit the assumptions. I wrote them from a high-level perspective.

goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/817c6c54e162ac9dffe91e92805331333f69587d

ferranbt (approved):

consensus/polybft/consensus_runtime.go
```diff
@@ -743,20 +755,18 @@ func (c *consensusRuntime) InsertBlock(proposal []byte, committedSeals []*messag
...
-					c.logger.Warn("failed to submit checkpoint",
-						"checkpoint block", header.Number,
-						"epoch number", epochNumber,
-						"error", err)
-				}
-			}(*fullBlock.Block.Header, fsm.epochNumber)
-
-			c.checkpointManager.latestCheckpointID = fullBlock.Block.Number()
-		}
+	if c.checkpointManager.IsCheckpointBlock(fullBlock.Block.Header.Number, fsm.isEndOfEpoch) &&
```
ferranbt, line 755: This can be done on a follow-up PR but I assume we are also going to move this also on a PostBlock op right? That will simplify the `checkpointManager` function to only have the handles as far as I can see.

ferranbt (approved):

consensus/polybft/consensus_runtime.go
goran-ethernal: Indeed. Didn't want to do all the work through this PR, since we have 2 more tasks for checkpoints for RFC-176 prep.

ferranbt: Cool. That is what I thought too.

consensus/polybft/state.go
```diff
@@ -398,34 +398,82 @@ func (s *State) getExitEventsForProof(epoch, checkpointBlock uint64) ([]*ExitEve
...
 }
 
-// getExitEvents returns exit events for given epoch and provided filter
-func (s *State) getExitEvents(epoch uint64, filter func(exitEvent *ExitEvent) bool) ([]*ExitEvent, error) {
-	var events []*ExitEvent
+// updateExitEvents updates the exit events that happened in given block and old epoch to the provided new epoch
+// that means their key in db changes, so we have to remove the old event and reinsert the new one
+func (s *State) updateExitEvents(blockNumber, oldEpoch, newEpoch uint64) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		exitEvents, err := getExitEvents(tx, oldEpoch, func(exitEvent *ExitEvent) bool {
```
b00ris: Correct me, if I'm wrong. We are writing events in PostBlock hook, but if it is the last block of the epoch we overwrite event keys in a separate transaction.

b00ris (changes requested):
Generally - good change. But i don't like the idea of writing(or overriting) exit events in 2 transactions

consensus/polybft/state.go
```diff
@@ -398,34 +398,82 @@ func (s *State) getExitEventsForProof(epoch, checkpointBlock uint64) ([]*ExitEve
...
+		exitEvents, err := getExitEvents(tx, epoch, filter)
+
+		if err != nil {
+			return err
+		}
+
+		events = exitEvents
+
+		return nil
+	})
```
Stefan-Ethernal: Small simplification proposal (using single reference for `exitEvents`, instead of two), if you find it correct:

```suggestion
	var (
		exitEvents []*ExitEvent
		err        error
	)

	err = s.db.View(func(tx *bolt.Tx) error {
		exitEvents, err = getExitEvents(tx, epoch, filter)
		if err != nil {
			return err
		}

		return nil
	})
```

consensus/polybft/state.go
```diff
@@ -398,34 +398,82 @@ func (s *State) getExitEventsForProof(epoch, checkpointBlock uint64) ([]*ExitEve
...
+
+func getExitEvents(tx *bolt.Tx, epoch uint64, filter func(exitEvent *ExitEvent) bool) ([]*ExitEvent, error) {
+	c := tx.Bucket(exitEventsBucket).Cursor()
+	prefix := itob(epoch)
+
+	var events []*ExitEvent
+
+	for k, v := c.Seek(prefix); bytes.HasPrefix(k, prefix); k, v = c.Next() {
+		var event *ExitEvent
+		if err := json.Unmarshal(v, &event); err != nil {
```
Stefan-Ethernal: Not sure if this gets optimized by the compiler itself, but maybe consider declaring variable before for loop (just in case):

```suggestion
        var event *ExitEvent
	for k, v := c.Seek(prefix); bytes.HasPrefix(k, prefix); k, v = c.Next() {
		if err := json.Unmarshal(v, &event); err != nil {
```

consensus/polybft/checkpoint_manager.go
```diff
@@ -170,13 +205,13 @@ func (c *checkpointManager) submitCheckpoint(latestHeader types.Header, isEndOfE
...
 	}
 
-	return c.encodeAndSendCheckpoint(txn, latestHeader, *currentExtra, isEndOfEpoch)
+	return c.encodeAndSendCheckpoint(txn, latestHeader, currentExtra, isEndOfEpoch)
 }
 
 // encodeAndSendCheckpoint encodes checkpoint data for the given block and
 // sends a transaction to the CheckpointManager rootchain contract
 func (c *checkpointManager) encodeAndSendCheckpoint(txn *ethgo.Transaction,
-	header types.Header, extra Extra, isEndOfEpoch bool) error {
```
Stefan-Ethernal, line 179: The reason I didn't pass pointers but values is because I wanted to explicitly suggest that those header parameters shouldn't be mutated in this function (but rather read-only parameters).
Not against the change though (it gets annoying dereferencing pointers all the time).

consensus/polybft/consensus_runtime.go
```diff
@@ -509,11 +531,6 @@ func (c *consensusRuntime) calculateUptime(currentBlock *types.Header, epoch *ep
 	return commitEpoch, nil
 }
 
-// InsertExitEvents is an implementation of checkpointBackend interface
-func (c *consensusRuntime) InsertExitEvents(exitEvents []*ExitEvent) error {
-	return c.state.insertExitEvents(exitEvents)
-}
-
 // BuildEventRoot is an implementation of checkpointBackend interface
 func (c *consensusRuntime) BuildEventRoot(epoch uint64) (types.Hash, error) {
```
Stefan-Ethernal, line 532: Why this isn't part of `checkpointManager`?

consensus/polybft/consensus_runtime.go
```diff
@@ -46,8 +46,6 @@ type txPoolInterface interface {
 type checkpointBackend interface {
 	// BuildEventRoot generates an event root hash from exit events in given epoch
 	BuildEventRoot(epoch uint64) (types.Hash, error)
-	// InsertExitEvents inserts provided exit events to persistence storage
-	InsertExitEvents(exitEvents []*ExitEvent) error
 }
```
Stefan-Ethernal, line 49: If `BuildEventRoot` becomes part of `checkpointManager`, then move this interface to `checkpoint_manager.go` as well.

consensus/polybft/consensus_runtime.go
Stefan-Ethernal: Ok I was wondering the same: why this part isn't moved to `checkpointManager`. When it is moved away to `PostBlock`, then we can remove `IsCheckpointBlock` function from `CheckpointManager` interface.

Stefan-Ethernal (approved):
Looks good :rocket:

consensus/polybft/state.go
ferranbt: I think this was also discussed here https://github.com/0xPolygon/polygon-edge/pull/1113#discussion_r1068034498

consensus/polybft/consensus_runtime.go
goran-ethernal: It will be. There are two more tasks for `checkpointManager`. Didn't want to have all the changes in one big PR, so I will break them into smaller PRs.

consensus/polybft/consensus_runtime.go
goran-ethernal: Then we won't need this interface at all.

consensus/polybft/checkpoint_manager.go
goran-ethernal: But that is also what we are trying to do with these `components` (`handlers`) in RFC-176. You are getting immutable data, so you can create a snapshot of your state. I don't think there is a need to change a state of any object passed to these components, nor it should be.

consensus/polybft/state.go
goran-ethernal: @b00ris In this case `PostBlock` and `PostEpoch` are both called sequentially in the same `OnBlockInserted` call. First the `PostBlock` will insert them in the current epoch, but if it is the end of epoch, `PostEpoch` will update them to new epoch, so we won't be starting a new epoch before doing this (insures that security about finalizing exit events), but yes, we have two `db` transactions, because we didn't want to misuse `PostBlock` to do work of `PostEpoch`. It misses the idea of those two requests that should be separated.

As Ferran mention we discussed this in detail in https://github.com/0xPolygon/polygon-edge/pull/1113#discussion_r1068034498.

consensus/polybft/state.go
goran-ethernal: FIx: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/203d13118b19e21ebdbe4b33216c1fc0170d6a7c

consensus/polybft/state.go
goran-ethernal: Ok, so reverted to my first commit, as we discussed on slack: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/a25085b5b42d6b03aeee49817e4dba246299e3ad

consensus/polybft/checkpoint_manager.go
```diff
@@ -246,8 +279,70 @@ func (c *checkpointManager) abiEncodeCheckpointBlock(blockNumber uint64, blockHa
...
+
+	return nil
+}
+
+// getExitEventsFromReceipts parses logs from receipts to find exit events
+func getExitEventsFromReceipts(epoch, block uint64, receipts []*types.Receipt) ([]*ExitEvent, error) {
+	events := make([]*ExitEvent, 0)
+
+	for i := 0; i < len(receipts); i++ {
+		if len(receipts[i].Logs) == 0 {
```
igorcrevar: You don need this check, range will work even with `nil` values

consensus/polybft/consensus_runtime.go
```diff
@@ -738,20 +752,18 @@ func (c *consensusRuntime) InsertBlock(proposal []byte, committedSeals []*messag
...
+		go func(header *types.Header, epochNumber uint64) {
+			if err := c.checkpointManager.SubmitCheckpoint(header, fsm.isEndOfEpoch); err != nil {
+				c.logger.Warn("failed to submit checkpoint",
+					"checkpoint block", header.Number,
+					"epoch number", epochNumber,
+					"error", err)
+			}
+		}(fullBlock.Block.Header, fsm.epochNumber)
+
+		c.checkpointManager.SetLatestCheckpointID(fullBlock.Block.Number())
```
igorcrevar: Is it ok to update latest checkpoint id even if submit checkpoint fails (which is called from another routine)?

goran-ethernal: Actually we only use that field in `checkpointManager` to know on which blocks to submit a new checkpoint. We do not care if it was successful or not. It means that if we sent a checkpoint transaction on block 1100, we will wait for another `offset` of 30 mins or the end of epoch to send another checkpoint or resend the last one if it was not successful. 

The name of this function and field is missleading, so I'll change it.

consensus/polybft/checkpoint_manager.go
```diff
@@ -246,8 +279,70 @@ func (c *checkpointManager) abiEncodeCheckpointBlock(blockNumber uint64, blockHa
...
+		epoch++
+	}
+
+	// commit exit events only when we finalize a block
+	events, err := getExitEventsFromReceipts(epoch, req.FullBlock.Block.Number(), req.FullBlock.Receipts)
+	if err != nil {
+		return err
+	}
+
+	if len(events) > 0 {
```
igorcrevar: Maybe (i am not 100% sure) would be better if this check is performed in `insertExitEvents` itself. That way anyone who calls `insertExitEvents` does not need to take care of this check, and also it will be less code in consensus runtime :)

consensus/polybft/checkpoint_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/be68a84e56c50260e2d3fcf66bcc559212cc3375

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/be68a84e56c50260e2d3fcf66bcc559212cc3375

consensus/polybft/checkpoint_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1113/commits/be68a84e56c50260e2d3fcf66bcc559212cc3375

b00ris (approved):

igorcrevar (approved):

PR #724, opened 2022-09-09
[Rootnet Staking] State transactions

dbrajovic:
# Description

This PR introduces a new transaction type (not a new `struct`) for applying events from the rootchain contract. 

RLP marshalling/unmarshalling and `executor` flows have been modified to suit this particular use case. 

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [ ] I have tested this code with the official test suite
- [x] I have tested this code manually

### Manual tests

1. Build `develop` binary and start a 4-node cluster
2. Send a transaction
3. Stop the cluster and build a binary from this branch
4. Start the cluster again
5. Send a transaction

# Additional comments

Fixes EDGE-776

PR #1114, opened 2023-01-13
Align RLP marshaling logic according to eip-2718

begmaroman:
# Description

Align RLP marshaling logic according to eip-2718.
Example: https://github.com/0xPolygon/polygon-edge/pull/724/files and https://github.com/umbracle/ethgo/blob/main/structs_marshal_rlp.go#L84
EIP: https://eips.ethereum.org/EIPS/eip-2718

**NEED TO ADD CHAIN ID AND ACCESS LIST**

# Changes include

# Checklist

## Testing


types/tests/transaction_test.go
```diff
@@ -0,0 +1,154 @@
...
+		Name              string         `json:"name"`
+		AccountAddress    ethgo.Address  `json:"accountAddress"`
+		PrivateKey        ethgo.ArgBytes `json:"privateKey"`
+		SignedTransaction ethgo.ArgBytes `json:"signedTransactionChainId5"`
+
+		Data     *ethgo.ArgBytes `json:"data,omitempty"`
+		Value    *ethgo.ArgBig   `json:"value,omitempty"`
+		To       *ethgo.Address  `json:"to,omitempty"`
+		GasLimit *ethgo.ArgBig   `json:"gasLimit,omitempty"`
+		Nonce    uint64          `json:"nonce,omitempty"`
```
JekaMas: Do we need to omit zero value Nonce and Name?

types/rlp_unmarshal.go
```diff
@@ -60,12 +80,24 @@ func (b *Block) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
 		return err
 	}
 
-	for _, txn := range txns {
```
vcastellm, line 63: Why not `range`?

types/rlp_unmarshal.go
```diff
@@ -355,30 +433,24 @@ func (t *Transaction) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
-				// reset From
-				t.From = ZeroAddress
-			}
+		// We need to set From field for state transaction,
+		// because we are using unique, predefined address, for sending such transactions
+		// From
+		if vv, err := v.Get(9).Bytes(); err == nil && len(vv) == AddressLength {
+			// address
+			addr := BytesToAddress(vv)
+			t.From = addr
```
vcastellm: Nitpick: Inline?

types/rlp_unmarshal_storage.go
```diff
@@ -30,12 +31,25 @@ func (b *Body) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
 		return err
 	}
 
-	for _, txn := range txns {
+	for i := 0; i < len(txns); i++ {
```
vcastellm: Duplicate code? https://github.com/0xPolygon/polygon-edge/pull/1114/files#diff-efc5999399a6ff1082b29a52ce57fd7bfd6ebe4789d686a44c70523d2a61c684R83

types/rlp_unmarshal_storage.go
```diff
@@ -67,8 +96,17 @@ func (t *Transaction) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value)
 		return err
 	}
 
-	if len(elems) < 2 {
-		return fmt.Errorf("incorrect number of elements to decode transaction, expected 2 but found %d", len(elems))
+	// come TransactionType first if exist
+	if len(elems) != 2 && len(elems) != 3 {
+		return errors.New("expected 2 or 3 elements")
```
vcastellm: I find this error more precise but would be helpful to maintain the elems count.

types/tests/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
vcastellm: Why are `npm` files included in the PR?

types/rlp_unmarshal_storage.go
```diff
@@ -93,20 +133,45 @@ func (r *Receipts) UnmarshalStoreRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) er
 		return err
 	}
 
-	for _, elem := range elems {
+	for i := 0; i < len(elems); i++ {
```
vcastellm: Looks like dup code with `rlp_unmarshal.go`, a func woul be great.

vcastellm (commented):
There are some linting issue too

types/rlp_unmarshal.go
begmaroman: Because it has a little bit complicated logic, look at line 90

types/rlp_unmarshal_storage.go
begmaroman: They are similar but use different objects. Generalising this func would look overcomplicated IMO. Feel free to suggest something if you want.

types/rlp_unmarshal.go
begmaroman: Fixed

types/tests/package.json
begmaroman: These tests are not suitable until we add chain ID and access list

types/rlp_unmarshal_storage.go
begmaroman: Done

types/rlp_unmarshal_storage.go
begmaroman: They are similar but use different objects. Generalising this func would look overcomplicated IMO. Feel free to suggest something if you want.

types/tests/transaction_test.go
begmaroman: No needed atm. Removed these tests.

types/rlp_unmarshal.go
vcastellm: We are incrementing the index manually to skip the type right?

types/rlp_unmarshal.go
```diff
@@ -60,12 +80,26 @@ func (b *Block) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
...
-		bTxn := &Transaction{}
-		if err := bTxn.UnmarshalRLPFrom(p, txn); err != nil {
+	for i := 0; i < len(txns); i++ {
+		txType := LegacyTx
+		if txns[i].Type() == fastrlp.TypeBytes {
+			if err = txType.UnmarshalRLPFrom(p, txns[i]); err != nil {
+				return err
+			}
+
+			i++
```
vcastellm: Why are we doing this? can you add code comment?

types/rlp_unmarshal.go
```diff
@@ -60,12 +80,26 @@ func (b *Block) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
...
+
+		bTxn := &Transaction{
+			Type: txType,
+		}
+
+		if err = bTxn.UnmarshalRLPFrom(p, txns[i]); err != nil {
 			return err
 		}
 
+		bTxn.ComputeHash()
```
```suggestion
		_ := bTxn.ComputeHash()
```

or prefer to obtain the result and work with it explicitly

begmaroman: But it just returns the same object. Do you think we really need it?

types/rlp_unmarshal.go
begmaroman: Done

types/rlp_unmarshal.go
vcastellm: Then why it modifies the object inside the function and returns the same object? It's a function with hidden side effects, a code smell, https://weblogs.asp.net/podwysocki/side-effecting-functions-are-code-smells, preferably return the new object, I'd refactor that function if possible.

types/rlp_unmarshal.go
```diff
@@ -34,6 +34,26 @@ func UnmarshalRlp(obj unmarshalRLPFunc, input []byte) error {
 	return nil
 }
 
+func (t *TxType) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
```
Stefan-Ethernal: Does it make sense to cover this function with a unit test?

types/rlp_unmarshal.go
```diff
@@ -60,12 +80,29 @@ func (b *Block) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
...
+		}
+
+		if err = bTxn.UnmarshalRLPFrom(p, txns[i]); err != nil {
 			return err
 		}
 
+		bTxn.ComputeHash()
+
 		b.Transactions = append(b.Transactions, bTxn)
 	}
```
Stefan-Ethernal: Should we extract this function to some helper one, as it is (almost) the same as the one in `rlp_unmarshal_storage.go` (https://github.com/0xPolygon/polygon-edge/pull/1114/files#diff-885da5543ea77aa838b1210b3eb06bf214b06f2f77e0c2134051e4628efcbb15R34)?

Only this line differs (maybe we could provide it as a callback argument)
`if err = bTxn.UnmarshalRLPFrom(p, txns[i]); err != nil {`

WDYT?

types/rlp_unmarshal.go
```diff
@@ -181,23 +218,50 @@ func (r *Receipts) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error {
...
+			TransactionType: txType,
+		}
+
+		if err = rr.UnmarshalRLPFrom(p, elems[i]); err != nil {
 			return err
 		}
 
-		(*r) = append(*r, rr)
+		*r = append(*r, rr)
 	}
```
Stefan-Ethernal: Extract to helper function (same as with `Transaction` unmarshal logic)? The only line which differs between `rlp_unmarshal.go` and `rlp_unmarshal_storage.go` is this one: https://github.com/0xPolygon/polygon-edge/pull/1114/files#diff-885da5543ea77aa838b1210b3eb06bf214b06f2f77e0c2134051e4628efcbb15R152. WDYT?

Stefan-Ethernal (commented):
Generally looks good, but leaving a couple of suggestions to consider.

types/rlp_unmarshal.go
begmaroman: We already have tests for the high-level function. Don't really see a reason why we should duplicate that test.

types/rlp_unmarshal.go
begmaroman: Done

types/rlp_unmarshal.go
begmaroman: Done

types/rlp_unmarshal.go
begmaroman: Done

vcastellm (approved):
LGTM

types/rlp_unmarshal.go
Stefan-Ethernal: I've executed tests for the `types` package, and as you can this function is not covered at all (neither explicitly nor implicitly):
image

I believe `func txTypeFromByte(b byte) (TxType, error)` is covered, but I wouldn't say that's the case for 
`func (t *TxType) UnmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) error`. Am I missing out on something? Is it unit tested outside of the `types` package perhaps?

types/rlp_unmarshal.go
```diff
@@ -13,6 +13,8 @@ type RLPUnmarshaler interface {
 
 type unmarshalRLPFunc func(p *fastrlp.Parser, v *fastrlp.Value) error
 
+type unmarshalRLPFromCbFunc func(TxType, *fastrlp.Parser, *fastrlp.Value) error
```
Stefan-Ethernal: Nitpick (so that we are consistent with naming, since we have `unmarshalRLPFunc` from before). WDYT?

```suggestion
type unmarshalRLPFromFunc func(TxType, *fastrlp.Parser, *fastrlp.Value) error
```

Stefan-Ethernal (approved):
LGTM :heavy_check_mark:

types/rlp_unmarshal.go
begmaroman: Done

types/rlp_unmarshal.go
begmaroman: For example, `TestRLPMarshall_And_Unmarshall_TypedTransaction` calls `UnmarshalRLP` that uses `unmarshalRLPFrom` inside.

Stefan-Ethernal: Nope, it extracts `TxType` using `txTypeFromByte`
(https://github.com/0xPolygon/polygon-edge/pull/1114/files#diff-efc5999399a6ff1082b29a52ce57fd7bfd6ebe4789d686a44c70523d2a61c684R253) and it sets offset so that first byte is skipped and `txType` doesn't get unmarshaled in `unmarshalRLPFrom` function...

begmaroman: Added test

PR #1130, opened 2023-01-16
Update state transaction logic to use base fee and tips

begmaroman:
# Description

- Updated state transition according to London's requirements. Meaning different logic of miner fees calculation and burning mechanism.
- Updated tests by providing the zero burn contract address for those where London fork is enabled. One of the next PRs would update some tests by including a real burn contract in order to test a whole flow. 
- Added a logic to prefill (base tip cap and base fee cap) or (gas price) depending on the provided values. Meaning if the gas price is provided, the other two fields should be calculated, and vice versa. 
- Added base fee calculation function.

# Changes include

# Checklist

## Testing


chain/params.go
```diff
@@ -1,18 +1,54 @@
...
+}
+
+// CalculateBurntContract calculates burn contract address for the given block number
+func (p *Params) CalculateBurntContract(number uint64) (types.Address, error) {
+	keys := make([]string, 0, len(p.BurntContract))
+	for k := range p.BurntContract {
+		keys = append(keys, k)
+	}
+
+	if len(keys) == 0 {
```
igorcrevar, line 36: Can we check this as the first thing in function? like `if len(p.BurntContract ) == 0 { ...`

chain/params.go
```diff
@@ -1,18 +1,54 @@
...
+		return types.ZeroAddress, ErrBurntContractAddressMissing
+	}
+
+	sort.Strings(keys)
+
+	for i := 0; i < len(keys)-1; i++ {
+		valUint, _ := strconv.ParseUint(keys[i], 10, 64)
+		valUintNext, _ := strconv.ParseUint(keys[i+1], 10, 64)
```
igorcrevar, line 44: Is maybe better to use temp `prev` variable which is firstly initialized with `prev, _ := strconv.ParseUint(keys[0], 10, 64)` and then update that variable in loop in order to not convert twice?

vcastellm (commented):
Can you indicate in the PR description, how is this related to https://github.com/0xPolygon/polygon-edge/pull/1176?

PR #1133, opened 2023-01-17
EVM-360 Invalid state root error on v0.7.1-alpha2

igorcrevar:
# Description

In block builder, header timestamp was not set and transition was created with timestamp 0 before any tx writing.
Because of that opcode `TIMESTAMP = 0x42` (`opTimstamp`) returned different results when tx is executed while building and verifying proposal

The solution is to change `block builder` `Reset` to include timestamp for current block and to call `Reset` each time before proposal is created and not just once when we are creating `fsm`.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -4,7 +4,6 @@ import (
 	"time"
 
 	"github.com/0xPolygon/polygon-edge/consensus"
-	"github.com/0xPolygon/polygon-edge/consensus/ibft/signer"
 	"github.com/0xPolygon/polygon-edge/state"
 	"github.com/0xPolygon/polygon-edge/txpool"
 	"github.com/0xPolygon/polygon-edge/types"
@@ -24,9 +23,6 @@ type BlockBuilderParams struct {
 	// Coinbase that is signing the block
 	Coinbase types.Address
 
-	// Vanity extra for the block
-	Extra []byte
-
 	// GasLimit is the gas limit for the block
 	GasLimit uint64
 
@@ -40,25 +36,14 @@ type BlockBuilderParams struct {
 	TxPool txPoolInterface
 }
 
-func NewBlockBuilder(params *BlockBuilderParams) (*BlockBuilder, error) {
-	// extra can only be 32 size max. it is better to trim that to return
-	// an error that we have to propagate. It should be up to higher level
-	// code to error if the extra supplied by the user is too big.
-	if len(params.Extra) > signer.IstanbulExtraVanity {
-		params.Extra = params.Extra[:signer.IstanbulExtraVanity]
-	}
-
-	if params.Extra == nil {
-		params.Extra = make([]byte, 0)
-	}
-
+func NewBlockBuilder(params *BlockBuilderParams) *BlockBuilder {
 	builder := &BlockBuilder{
 		params: params,
 	}
 
 	builder.Reset()
 
-	return builder, nil
+	return builder
 }
 
 var _ blockBuilder = &BlockBuilder{}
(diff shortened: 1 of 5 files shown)


consensus/polybft/fsm.go
```diff
@@ -19,7 +18,8 @@ import (
 )
 
 type blockBuilder interface {
-	Reset() error
+	Reset()
```
goran-ethernal: I think we are not using this `Reset` function anywhere but in `blockBuilder` constructor, so maybe remove it from the interface?

igorcrevar: Fixed cf149bdd68be46cd6c649fa3438ee51abc9f5ca4

goran-ethernal (approved):

vcastellm (approved):
LGTM

Stefan-Ethernal (approved):

consensus/polybft/fsm.go
```diff
@@ -19,7 +18,7 @@ import (
 )
 
 type blockBuilder interface {
-	Reset() error
+	Init() error
```
ferranbt: Was it necessary to change this name?

consensus/polybft/block_builder.go
```diff
@@ -82,31 +61,37 @@ type BlockBuilder struct {
 	state *state.Transition
 }
 
-// Reset is used to indicate that the current block building has been interrupted
-// and it has to clean any data
-func (b *BlockBuilder) Reset() error {
+// Init initializes block builder before adding transactions and actual block building
+func (b *BlockBuilder) Init() error {
+	// set the timestamp
+	parentTime := time.Unix(int64(b.params.Parent.Timestamp), 0)
```
ferranbt, line 67: Can we check what other chains do with this? I am not sure what is the consensus, should the timestamp be the current one or the time when the block was meant to be created?

ferranbt:
Could you please be more descriptive in the PR description about the solution? it is only describing the problem scope.

consensus/polybft/block_builder.go
igorcrevar: This is just copy pasted code as it was before. The only "subtle" difference is that timestamp was set after all transactions have been written and now it is before

consensus/polybft/fsm.go
igorcrevar: name has been reverted https://github.com/0xPolygon/polygon-edge/pull/1133/commits/4ccd5429fc4e2c8a2d3648f4b04e8aa43e20d007

consensus/polybft/block_builder.go
igorcrevar: The full logic is this (as it was before):
1. take the timestamp from the parent
2. add block time
3. if that timestamp is lower than current time update it to current time (` time.Now()`)
4. write that timestamp to header  as uint64 (`uint64(headerTime.Unix())`)

igorcrevar: I think (if I understand correctly - I was looking at `miner.go` in go-eth), this is how setting timestamp working in `go-eth`:
1. Before transactions are filed the timestamp is created as:
```
	timestamp := genParams.timestamp
	if parent.Time() >= timestamp {
		if genParams.forceTime {
			return nil, fmt.Errorf("invalid timestamp, parent %d given %d", parent.Time(), timestamp)
		}
		timestamp = parent.Time() + 1
	}
```
2. `genParams.timestamp` is set to `time.Now().Unix()`

I think we are good to go with our timestamp setting

consensus/polybft/block_builder.go
```diff
@@ -82,31 +61,37 @@ type BlockBuilder struct {
...
 		StateRoot:    types.EmptyRootHash, // this avoids needing state for now
 		TxRoot:       types.EmptyRootHash,
 		ReceiptsRoot: types.EmptyRootHash, // this avoids needing state for now
 		Sha3Uncles:   types.EmptyUncleHash,
 		GasLimit:     b.params.GasLimit,
+		Timestamp:    uint64(headerTime.Unix()),
 	}
 
-	b.block = nil
-	b.txns = []*types.Transaction{}
```
ferranbt, line 102: Is there any reason to move these variables?

ferranbt (approved):

consensus/polybft/block_builder.go
igorcrevar: just because i wanted to group all `b.*` in same place

PR #1141, opened 2023-01-18
BuildProofs for missed commitments

goran-ethernal:
# Description

When a node was offline (or out of sync) for some time, it can miss on some commitments that might have happened in the meantime.

Since `syncer` is calling the `OnBlockInserted` callback in `consensus_runtime` for each block, the `state_sync_manager` will extract all the commitments that happened in those blocks, and save them to `boltDb`.
In the meantime, in parallel, `event_tracker` will get all the `stateSync` events that the node missed while out of sync, and save them to `boltDb` as well.

But, this process does not build the `state sync proofs` for the commitments that happened while the node was out of sync.

That is why, through this PR, when a user (or relayer) calls `GetStateSyncProof` `RPC` function, `state_sync_manager` will first check if it has a proof built for given `state sync event`. If it does have it, it will simply return it.
If it does not have it, it probably means, we were out of sync, and we need to build those proofs, so it will get the `commitment` from `db` (which was extracted from blocks while syncing), and build the proofs for it and save them in `db`.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state_sync_manager.go
+++ b/consensus/polybft/state_sync_manager.go
@@ -406,9 +406,12 @@ func (s *stateSyncManager) GetStateSyncProof(stateSyncID uint64) (*types.StateSy
 		}
 
 		proof, err = s.state.getStateSyncProof(stateSyncID)
+		if err != nil {
+			return nil, fmt.Errorf("cannot get state sync proof for StateSync id %d: %w", stateSyncID, err)
+		}
 	}
 
-	return proof, err
+	return proof, nil
 }
 
 // buildProofs builds state sync proofs for the submitted commitment and saves them in boltDb for later execution
--- a/consensus/polybft/state_sync_manager.go
+++ b/consensus/polybft/state_sync_manager.go
@@ -33,6 +33,7 @@ type StateSyncManager interface {
 	Init() error
 	Close()
 	Commitment() (*CommitmentMessageSigned, error)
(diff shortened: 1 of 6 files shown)


igorcrevar (approved):

consensus/polybft/state_sync_manager.go
```diff
@@ -381,6 +385,32 @@ func (s *stateSyncManager) PostBlock(req *PostBlockRequest) error {
...
+		commitment, err := s.state.getCommitmentForStateSync(stateSyncID)
+		if err != nil {
+			return nil, fmt.Errorf("cannot find commitment for StateSync id %d: %w", stateSyncID, err)
+		}
+
+		if err := s.buildProofs(commitment.Message); err != nil {
+			return nil, fmt.Errorf("cannot build proofs for commitment for StateSync id %d: %w", stateSyncID, err)
+		}
+
+		proof, err = s.state.getStateSyncProof(stateSyncID)
```
vcastellm, line 408: Maybe check the error here to return a descriptive one, and return the proof if there's no error?

vcastellm (approved):
LGTM aside from the nitpick

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1141/commits/b58c1f2b790131f7c395ce98c05d19935f7b20b8

ferranbt (approved):

PR #1143, opened 2023-01-19
EVM-373 overbroad permissioned files

yaziciahmet:
# Description

This PR addresses EVM-373, local files which are created within the application having too much permissions on themselves.

What is changed is that, all the permissions are given to only the owner and its group. The reasoning behind this decision is to allow the owners to have the flexibility to decide whether the files can be accessed from different users or not. If they want multiple users to access the files, they can put them into the same group, if not, they can put different users into different groups.

Affected files:
- genesis -> Read + Write -> genesis can be overwritten (without being deleted and re-created) with the `polygon-edge genesis` command, hence it should have write access as well
- default-config -> Read + Write -> same reason with genesis for the command `polygon-edge server export`
- snapshots and metadata -> Read + Write -> both are created from the function `writeDataStore`, and are updated on different runs (rather than being deleted and re-created)
- manifest -> Read + Write -> same reason with genesis for the command `polygon-edge manifest`
- private key and secrets -> Read only -> They are generated, and if they exist, they are never tried to be overridden, which allows them to be read only
- secret manager config -> Read + Write -> same reason with genesis for the command `polygon-edge secrets generate`
- node folders -> Read + Write + Exec -> The node directories are created with the function `createDir`. It should be ok to give all permissions to a folder, since it means the files can be seen and new files can be created within this directory.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/helper/helper.go
+++ b/command/helper/helper.go
@@ -238,8 +238,8 @@ func WriteGenesisConfigToDisk(genesisConfig *chain.Chain, genesisPath string) er
 	if err != nil {
 		return fmt.Errorf("failed to generate genesis: %w", err)
 	}
-
-	if err := os.WriteFile(genesisPath, data, os.ModePerm); err != nil {
+	/* #nosec */
+	if err := os.WriteFile(genesisPath, data, 0660); err != nil {
 		return fmt.Errorf("failed to write genesis: %w", err)
 	}
 
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -107,7 +107,7 @@ func createDir(path string) error {
 	}
 
 	if os.IsNotExist(err) {
-		if err := os.MkdirAll(path, os.ModePerm); err != nil {
+		if err := os.MkdirAll(path, 0770); err != nil {
 			return err
 		}
 	}
--- a/command/server/export/export.go
+++ b/command/server/export/export.go
@@ -70,12 +70,12 @@ func generateConfig(config config.Config) error {
 	if err != nil {
 		return fmt.Errorf("could not marshal config struct, %w", err)
 	}
-
+	/* #nosec */
 	if err := os.WriteFile(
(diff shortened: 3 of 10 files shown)


Stefan-Ethernal (approved):
LGTM

b00ris (approved):

command/helper/helper.go
```diff
@@ -238,8 +238,8 @@ func WriteGenesisConfigToDisk(genesisConfig *chain.Chain, genesisPath string) er
 	if err != nil {
 		return fmt.Errorf("failed to generate genesis: %w", err)
 	}
-
-	if err := os.WriteFile(genesisPath, data, os.ModePerm); err != nil {
+	/* #nosec */
```
JekaMas, line 241: why we need that comment?

command/helper/helper.go
```diff
@@ -238,8 +238,8 @@ func WriteGenesisConfigToDisk(genesisConfig *chain.Chain, genesisPath string) er
...
+	if err := os.WriteFile(genesisPath, data, 0660); err != nil {
```
JekaMas, line 242: could we have one constant for that value in a separate package?

helper/common/common.go
```diff
@@ -107,7 +107,7 @@ func createDir(path string) error {
 	}
 
 	if os.IsNotExist(err) {
-		if err := os.MkdirAll(path, os.ModePerm); err != nil {
+		if err := os.MkdirAll(path, 0770); err != nil {
```
JekaMas, line 110: we still need 0770 here?

Stefan-Ethernal:
Hey @JekaMas , I think it is better to review #1164, because that one is continuation of this PR. This would be eventually closed in favour of #1164.

PR #1148, opened 2023-01-20
Support Syncer.Close() on running node

tuommaki:
# Description

Sometimes there can be situations where it's useful to be able to stop P2P syncing of blockchain, e.g. in some modular blockchain architectures where external DA solution is used for block distribution, but P2P syncing is used in the program startup for quick update syncing.

Existing `SyncPeerClient` and `network.Topic` had partial support for graceful shutdown of the messaging system and this change makes couple small additions that were necessary to avoid situations where e.g. peer change events are written into closed channel.

# Changes include

# Checklist

## Testing

### Manual tests

We have built Polygon Avail Settlement Layer blockchain on top of Edge, by using several parts of it as a library / framework. Our blockchain is based on optimistic protocol and the source of truth for blockchain block distribution is Avail. Our watchtower node implementation does its initial syncing via P2P network, but after the syncing is done and the staking completed, it stops the syncer. This resulted in `panic` due to write to closed channel on peer event changes. This change fixes the problem for us.


begmaroman (approved):

network/gossip.go
```diff
@@ -34,6 +34,15 @@ func (t *Topic) createObj() proto.Message {
 	return message
 }
 
+func (t *Topic) Close() {
```
JekaMas, line 39: Is it possible that this method will be called multiple times in parallel? Maybe with graceful shutdown and multiple signals?

Stefan-Ethernal (approved):

goran-ethernal (approved):

b00ris (approved):

vcastellm (approved):
LGTM

network/gossip.go
tuommaki: Right now nothing in Edge calls this and in Avail SL it's called only once. It can be safeguarded by wrapping internal implementation inside a `sync.Once`, if that seems necessary. WDYT?

JekaMas (approved):

network/gossip.go
JekaMas: If possible, add sync.Once or an atomic with state of the node. any would work

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/0xPolygon/polygon-edge
 
-go 1.18
+go 1.19
```
tuommaki: Can we bump up the minimum required Go version into `1.19`? This would simplify the use of atomics with `atomic.Bool`. If this is not feasible, I'll change the atomics to use the older API.

network/gossip.go
tuommaki: I added `atomic.Bool` both to `network.Topic` and `syncer.syncPeerClient`.

JekaMas (approved):

go.mod
vcastellm: @tuommaki We'll need to conduct a series of tests to upgrade the Go version, that will take some time, if you need this merged ASAP consider to make it 1.18 compatible.

tuommaki: I changed the code to use older `atomic` functions and `*uint64` . The CI is somehow requesting for a token now though :thinking:

vcastellm (approved):
LGTM

vcastellm:
Hey @tuommaki can you please rebase/merge from `develop`? CI should be fixed

PR #1149, opened 2023-01-20
EVM-353 Different voting power property based test

igorcrevar:
# Description

We need a property based test that allows us to check that we can create different cluster with different voting power.

# Changes include

# Checklist

## Testing

# Additional comments

In order to run the test `env` variable `export E2E_PROPERTY_TESTS=true` must be set

### DIFF
--- a/e2e-polybft/property_test.go
+++ b/e2e-polybft/property_test.go
@@ -3,6 +3,8 @@ package e2e
 import (
 	"fmt"
 	"math"
+	"os"
+	"strings"
 	"testing"
 	"time"
 
@@ -11,7 +13,14 @@ import (
 	"pgregory.net/rapid"
 )
 
+// property based tests enabled
+const envPropertyBaseTestEnabled = "E2E_PROPERTY_TESTS"
+
 func TestProperty_DifferentVotingPower(t *testing.T) {
+	if strings.ToLower(os.Getenv(envPropertyBaseTestEnabled)) != "true" {
+		t.Skip("Property based tests are disabled.")
+	}
+
 	t.Parallel()
 
 	const (
@@ -21,7 +30,7 @@ func TestProperty_DifferentVotingPower(t *testing.T) {
 
 	rapid.Check(t, func(tt *rapid.T) {
 		var (
-			numNodes  = rapid.Uint64Range(4, 30).Draw(tt, "number of cluster nodes")
+			numNodes  = rapid.Uint64Range(4, 8).Draw(tt, "number of cluster nodes")
 			epochSize = rapid.OneOf(rapid.Just(4), rapid.Just(10)).Draw(tt, "Size of epoch")
 			numBlocks = rapid.Uint64Range(2, 5).Draw(tt, "number of blocks the cluster should mine")
 		)
--- a/e2e-polybft/property_test.go
+++ b/e2e-polybft/property_test.go
@@ -3,8 +3,6 @@ package e2e
 import (
 	"fmt"
 	"math"
-	"os"
-	"strings"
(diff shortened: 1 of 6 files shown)


vcastellm (approved):

b00ris (approved):

e2e-polybft/property_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+	"time"
+
+	"github.com/0xPolygon/polygon-edge/e2e-polybft/framework"
+	"github.com/0xPolygon/polygon-edge/types"
+	"pgregory.net/rapid"
+)
+
+// property based tests enabled
+const envPropertyBaseTestEnabled = "E2E_PROPERTY_TESTS"
```
Stefan-Ethernal: Should we move this to TestCluster ([here](https://github.com/0xPolygon/polygon-edge/blob/feature/EVM-353/e2e-polybft/framework/test-cluster.go#L26) for example), so that all E2E testing environment vars are kept on a single place?

e2e-polybft/property_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestProperty_DifferentVotingPower(t *testing.T) {
+	if strings.ToLower(os.Getenv(envPropertyBaseTestEnabled)) != "true" {
```
Stefan-Ethernal: Should we export the function `isTrueEnv` from `test-cluster` and use that one instead?

e2e-polybft/property_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	const (
+		blockTime  = time.Second * 6
+		maxPremine = math.MaxUint64
+	)
+
+	rapid.Check(t, func(tt *rapid.T) {
+		var (
+			numNodes  = rapid.Uint64Range(4, 8).Draw(tt, "number of cluster nodes")
+			epochSize = rapid.OneOf(rapid.Just(4), rapid.Just(10)).Draw(tt, "Size of epoch")
```
Stefan-Ethernal: Nitpick (consistency-wise):
```suggestion
			epochSize = rapid.OneOf(rapid.Just(4), rapid.Just(10)).Draw(tt, "epoch size")
```

Stefan-Ethernal (approved):
LGTM with some minor suggestions.

Btw should we create an additional GH workflow, so that we have property-based tests established on GH as well? Or we should aim to run it along with e2e tests? WDYT?

cc @vcastellm, @b00ris

e2e-polybft/property_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1149/commits/47178fb01acb738db85dab15f573b2ceebc50c0a

e2e-polybft/property_test.go
igorcrevar: Did something different but solved this https://github.com/0xPolygon/polygon-edge/pull/1149/commits/47178fb01acb738db85dab15f573b2ceebc50c0a

e2e-polybft/property_test.go
igorcrevar: https://github.com/0xPolygon/polygon-edge/pull/1149/commits/47178fb01acb738db85dab15f573b2ceebc50c0a

b00ris (approved):

vcastellm (commented):
I don't like at all the solution using ENV vars. IMHO it should be predictable that `go test ./...` runs all tests by default, and use the already existing form of excluding tests by pattern name in the make file like `go list ./... | grep -v e2e`

PR #1159, opened 2023-01-24
EVM-370 Added insecure flag for local store. 

ZeljkoBenovic:
# Description

When initializing the chain and using a local storage, private keys are stored as a plaintext format by default, in a local folder the operator has defined during the `init` stage.    

This PR adds the `--insecure` flag as mandatory in order to store private keys unencrypted.    
If the operator tries to run `secrets init` command without `--insecure` flag, the error `use a secrets backend, or supply an --insecure flag...` will be presented.    
When `--insecure` flag is used for storing the secrets locally, the following warning will be presented:  
`[WARNING: INSECURE LOCAL SECRETS - SHOULD NOT BE RUN IN PRODUCTION]`    
as storing secrets using this method, presents a high security risk and should not be used in production environments.

# Changes include

# Checklist

## Testing

### Manual tests

Run `secrets init --data-dir <folder>` - it should fail with an error `use a secrets backend, or supply an --insecure flag...`.     
Run `secrets init --data-dir <folder> --insecure` - it should succeed with a warning `[WARNING: INSECURE LOCAL SECRETS - SHOULD NOT BE RUN IN PRODUCTION]`

# Additional comments

Fixes EVM-370
Fixes EDGE-1024

### DIFF
--- a/command/secrets/init/params.go
+++ b/command/secrets/init/params.go
@@ -9,26 +9,29 @@ import (
 )
 
 const (
-	dataDirFlag = "data-dir"
-	configFlag  = "config"
-	ecdsaFlag   = "ecdsa"
-	blsFlag     = "bls"
-	networkFlag = "network"
-	numFlag     = "num"
+	dataDirFlag        = "data-dir"
+	configFlag         = "config"
+	ecdsaFlag          = "ecdsa"
+	blsFlag            = "bls"
+	networkFlag        = "network"
+	numFlag            = "num"
+	insecureLocalStore = "insecure"
 )
 
 var (
-	errInvalidConfig   = errors.New("invalid secrets configuration")
-	errInvalidParams   = errors.New("no config file or data directory passed in")
-	errUnsupportedType = errors.New("unsupported secrets manager")
+	errInvalidConfig                  = errors.New("invalid secrets configuration")
+	errInvalidParams                  = errors.New("no config file or data directory passed in")
+	errUnsupportedType                = errors.New("unsupported secrets manager")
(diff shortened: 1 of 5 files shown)


trimixlover (approved):

Stefan-Ethernal (approved):
LGTM. 

We should accommodate this change to `polybft-secrets` command on `v3-parity` branch (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/command/polybftsecrets/params.go).
Therefore would it be worth considering rebasing this PR on top of v3-parity and updating both init secrets for Edge and polybft?

command/secrets/init/params.go
```diff
@@ -9,26 +9,32 @@ import (
...
-	blsFlag     = "bls"
-	networkFlag = "network"
-	numFlag     = "num"
+	dataDirFlag        = "data-dir"
+	configFlag         = "config"
+	ecdsaFlag          = "ecdsa"
+	blsFlag            = "bls"
+	networkFlag        = "network"
+	numFlag            = "num"
+	insecureLocalStore = "insecure"
```
vcastellm: Nitpick: For the sake of consistency this should be suffixed with `Flag`

vcastellm (approved):
LGTM aside from the nitpick, and the cluster script should be adapted when merged to v3-parity

command/secrets/init/params.go
ZeljkoBenovic: Good catch 🎖️ Fixed in f5e00eb2a7031ed58278d18adcfb8bce4e15b812

PR #1160, opened 2023-01-24
Checkpoint data validations

Stefan-Ethernal:
# Description

This PR implements validations for checkpoint data. Validations are categorized as basic and full validations, depending if it is being validated from consensus or synchronization protocol (note that in synchronization protocol, only headers are available and therefore there are certain limitations about validation scope).

## Epoch Number
- When epoch number changes (epoch-beginning blocks), it must be incremented by one compared to the parent block’s epoch number. This validation is checked for both consensus and synchronization protocol.
- In case the current validators and next validators are not the same, it is required that the epoch number remains the same. This is due to the fact that if the validator set changes, it is happening on the epoch-ending block, which belongs to the current epoch. This is being validated only during consensus protocol.

## Current/Next Validators Hash
- The current and next validators hash must not be empty (checked for both consensus and synchronization protocol)
- Current validators hash is calculated based on the current validators, whereas the next validators hash is calculated based on the next validators. (checked only during consensus protocol, since we are not able to determine actual validator sets when running synchronization protocol. It is due to not being able to execute block, since only headers are verified)

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/extra.go
+++ b/consensus/polybft/extra.go
@@ -142,9 +142,17 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
 	return nil
 }
 
+// ValidateBasic contains extra data basic set of validations
+func (i *Extra) ValidateBasic(parentExtra *Extra) error {
+	if err := i.Checkpoint.ValidateBasic(parentExtra.Checkpoint); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // Validate contains extra data validation logic
-func (i *Extra) Validate(header *types.Header, parentExtra *Extra,
-	currentValidators AccountSet, nextValidators AccountSet) error {
+func (i *Extra) Validate(parentExtra *Extra, currentValidators AccountSet, nextValidators AccountSet) error {
 	if err := i.Checkpoint.Validate(parentExtra.Checkpoint, currentValidators, nextValidators); err != nil {
 		return err
 	}
@@ -523,9 +531,9 @@ func (c *CheckpointData) Hash(chainID uint64, blockNumber uint64, blockHash type
 	return types.BytesToHash(crypto.Keccak256(abiEncoded)), nil
 }
 
-// Validate encapsulates validation logic for checkpoint data
-func (c *CheckpointData) Validate(parentCheckpoint *CheckpointData,
-	currentValidators AccountSet, nextValidators AccountSet) error {
+// ValidateBasic encapsulates basic validation logic for checkpoint data.
+// It only checks epoch numbers validity and whether validators hashes are non-empty.
+func (c *CheckpointData) ValidateBasic(parentCheckpoint *CheckpointData) error {
 	if c.EpochNumber != parentCheckpoint.EpochNumber {
 		if c.EpochNumber != parentCheckpoint.EpochNumber+1 {
 			// epoch-beginning block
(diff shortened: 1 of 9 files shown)


consensus/polybft/fsm.go
```diff
@@ -242,13 +237,47 @@ func (f *fsm) Validate(proposal []byte) error {
...
+	nextValidators := f.validators.Accounts()
+
+	if f.isEndOfEpoch {
+		if err := f.blockBuilder.Reset(); err != nil {
+			return fmt.Errorf("failed to initialize block builder: %w", err)
+		}
+
+		// we need to write commit epoch transaction to intermediate state,
+		// in order to be able to query next validator set
+		if err := f.blockBuilder.WriteTx(block.Transactions[0]); err != nil {
```
Stefan-Ethernal: We are writing `commitEpoch` transaction in the state, so that we are able to query validator set for next epoch. Don't know for alternative solution atm.

consensus/polybft/fsm.go
```diff
@@ -242,13 +237,47 @@ func (f *fsm) Validate(proposal []byte) error {
...
+			return err
+		}
+
+		nextValidators, err = f.getCurrentValidators(f.blockBuilder.GetState())
```
goran-ethernal: Since `ProcessBlock` that is called at the end of the `fsm.Validate` function already executes the transaction, it would be nice to use that to validate this as well. Maybe expand `ProcessBlock` to return a new object:
```
ProcessedBlock {
  FulLBlock *types.FullBlock
  Transition *Transition 
}
```
That way you can use it to have this validation as well, without the need to double apply the same transaction.

igorcrevar (approved):
Approved. Optimal tx writing as Goran suggested can be done in separate pr I think.
Just cover all ifs with tests (excluding GetExtra errors)

Stefan-Ethernal:
> Just cover all ifs with tests (excluding GetExtra errors)

I have already covered all the branches of checkpoint data validation logic in [TestCheckpointData_Validate](https://github.com/0xPolygon/polygon-edge/pull/1160/files#diff-2f601f47fe8d02bd0d03ad2bbb4c245778e316affb10eb5497b9c4d1b58d11d1R674).
As per `fsm.Validate`, things are a bit trickier and I've tried to implement a unit test, but it got convoluted very fast, so I've given up (at least for now).

consensus/polybft/extra.go
```diff
@@ -142,6 +142,24 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
 	return nil
 }
 
+// ValidateBasic contains extra data basic set of validations
+func (i *Extra) ValidateBasic(parentExtra *Extra) error {
+	if err := i.Checkpoint.ValidateBasic(parentExtra.Checkpoint); err != nil {
```
b00ris: why do you need err check there?

b00ris (approved):
LGTM

consensus/polybft/extra.go
Stefan-Ethernal: It was a leftover, good point, tnx!
Fixed https://github.com/0xPolygon/polygon-edge/pull/1160/commits/ae0ae2eb02daa81f5e7d4ad6f6ab2d68f7ee0b96

consensus/polybft/fsm.go
Stefan-Ethernal: I've chosen to take different approach: instead of providing `Transition` object, I've isolated scope of transition instance through callback function. Otherwise we would enable client code to alter the state, which could be dangerous.

Fixed https://github.com/0xPolygon/polygon-edge/pull/1160/commits/fbe0c20c52382b2b0b6528d5dd68f5b25ec7a971.

goran-ethernal (approved):
LGTM :100:

PR #938, opened 2022-11-20
Remove `json.Marshaler` implementation from `PublicKey`

Stefan-Ethernal:
# Description

Cleanup of `PublicKey` from polybft, `json.Marshaler` implementation is unused therefore can be safely removed.

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [ ] I have tested this code with the official test suite
- [ ] I have tested this code manually

PR #1161, opened 2023-01-24
EVM-395 Implement RFC-175 BLS integration

igorcrevar:
# Description

Implement `RFC-175` conclusion:
bls implementation for `polybft` should be based on `github.com/umbracle/go-eth-bn256`

Currently used `kilic` `bls` library is not well supported and that is why we are replacing it.

# Changes include

# Checklist

## Testing


consensus/polybft/signer/private.go
```diff
@@ -4,50 +4,56 @@ import (
...
+	g2 := new(bn256.G2)
+	g2 = g2.ScalarMult(g2Point, p.s)
 
-	g2.MulScalar(public, g2.One(), p.p)
+	return &PublicKey{g2: g2}
+}
+
+// MarshalJSON marshal the key to json bytes.
+func (p *PrivateKey) MarshalJSON() ([]byte, error) {
+	if p.s == nil {
```
ferranbt: If this happens, someone is not using correctly the library and it should fail during tests. Note that you are not even making the same check during `Marshal`. So, I do not think we need this defensive code.

consensus/polybft/signer/public.go
```diff
@@ -2,45 +2,25 @@ package bls
...
-	}
-
-	return &PublicKey{p: newp}
+	g2 *bn256.G2
 }
 
 // Marshal marshal the key to bytes.
 func (p *PublicKey) Marshal() []byte {
-	if p.p == nil {
+	if p.g2 == nil {
```
ferranbt: Same as in PrivateKey.

consensus/polybft/signer/public.go
```diff
@@ -49,82 +29,86 @@ func (p *PublicKey) MarshalJSON() ([]byte, error) {
...
+
+	return &PublicKey{g2: g2}, nil
 }
 
 // UnmarshalPublicKeyFromBigInt unmarshals public key from 4 big ints
 // Order of coordinates is [A.Y, A.X, B.Y, B.X]
 func UnmarshalPublicKeyFromBigInt(b [4]*big.Int) (*PublicKey, error) {
 	const size = 32
 
-	var pubKeyBuf []byte
```
ferranbt: Was this change necessary?

consensus/polybft/signer/public.go
```diff
@@ -49,82 +29,86 @@ func (p *PublicKey) MarshalJSON() ([]byte, error) {
...
-	newp.Set(g.Zero())
+// Aggregate aggregates all public keys into one
+func (pks PublicKeys) Aggregate() *PublicKey {
+	newp := new(bn256.G2)
 
-	for _, x := range pubs {
-		if x.p != nil {
-			g.Add(newp, newp, x.p)
+	for _, x := range pks {
+		if x.g2 != nil {
```
ferranbt: Same as in PrivateKey, is there any situation when the public key is going to be empty?

consensus/polybft/signer/signature.go
```diff
@@ -2,65 +2,52 @@ package bls
...
-		if next.p != nil {
-			g.Add(newp, s.p, next.p)
-		} else {
-			newp.Set(s.p)
-		}
-	} else if next.p != nil {
-		newp.Set(next.p)
+// Marshal the signature to bytes.
+func (s *Signature) Marshal() ([]byte, error) {
+	if s.g1 == nil {
```
ferranbt: Same as in PrivateKey.

consensus/polybft/signer/signature.go
```diff
@@ -69,43 +56,36 @@ func UnmarshalSignature(raw []byte) (*Signature, error) {
...
-	newp.Set(g.Zero())
+// Aggregate aggregates all signatures into one
+func (sigs Signatures) Aggregate() *Signature {
+	g1 := new(bn256.G1)
 
-	for _, x := range s {
-		if x.p != nil {
-			g.Add(newp, newp, x.p)
+	for _, sig := range sigs {
+		if sig.g1 != nil {
```
ferranbt: Same as in the PublicKey aggregation.

igorcrevar: In the situation when someone manually creates `pk := PublicKey{}`

consensus/polybft/signer/public.go
```diff
@@ -49,82 +29,86 @@ func (p *PublicKey) MarshalJSON() ([]byte, error) {
...
-
-	return &PublicKey{p: p}, nil
-}
-
 // ToBigInt converts public key to 4 big ints
-func (p PublicKey) ToBigInt() [4]*big.Int {
+func (p *PublicKey) ToBigInt() [4]*big.Int {
 	blsKey := p.Marshal()
-	res := [4]*big.Int{
+	if len(blsKey) != 128 {
```
ferranbt: Can this ever happen? Note that you are not doing the same check in the other functions that convert an array to bytes. We should check he `Marshal` function down to the library and figure out the size of the output and if that size can change.

consensus/polybft/signer/signature.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/signature.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/public.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/public.go
igorcrevar: reverted https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/public.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/private.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/c39a660fed18b2128cee9bc0329b13d278d6c8b5

consensus/polybft/signer/public.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/2eaa53c3fd8ce7ee85eecbe9582617aa0d1b1a89

consensus/polybft/signer/public.go
```diff
@@ -49,49 +25,55 @@ func (p *PublicKey) MarshalJSON() ([]byte, error) {
 }
 
 // UnmarshalJSON implements the json.Marshaler interface.
-func (p *PublicKey) UnmarshalJSON(b []byte) error {
-	var d []byte
+func (p *PublicKey) UnmarshalJSON(jsonBytes []byte) error {
```
ferranbt: Why do we need the `UnmarshalJSON` here? I think it should suffice to use `UnmarshalText` here, the same goes for `MarshalJSON` and `MarshalText`.

Stefan-Ethernal: It is implemented because of validators snapshots (because we are persisting snapshots as JSON representation in bolt database): https://github.com/0xPolygon/polygon-edge/pull/938#issuecomment-1323486276.

On the other hand, it is worth trying to switch to `(Un)MarshalText` and whether is it interchangeable with `json.Marshaler`, as I haven't tried it.

Btw is there any benefit in implementing `UnmarshalText/MarshalText` compared to `Unmarshal/MarshalJSON`?

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1161/commits/fa7eba62eebaae0527dd0cb54840e4d4dc53134e

consensus/polybft/signer/hash2point.go
```diff
@@ -0,0 +1,301 @@
...
+
+	modulus, _ = new(big.Int).SetString("30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47", 16)
+
+	zero = new(big.Int).SetInt64(0)
+)
+
+// hashToPoint maps hash message and domain to G1 point
+// Based on https://github.com/thehubbleproject/ and kilic library
+// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-07
+func hashToPoint(msg, domain []byte) (*bn256.G1, error) {
```
vcastellm, line 28: Wouldn't be better to move these functions to the package?

ferranbt (approved):

goran-ethernal (approved):

Nemanja0x (approved):
Just a comment: We do not check input data in some places like Aggregate (do not check if a signature is nil), Marshall, Unmarshall... Maybe that should be done to have a clear log if a bug occurs in code which is calling BLS functionalities and does not submit valid data.
Not a blocker to me.

ferranbt:
> Just a comment: We do not check input data in some places like Aggregate (do not check if a signature is nil), Marshall, Unmarshall... Maybe that should be done to have a clear log if a bug occurs in code which is calling BLS functionalities and does not submit valid data.

We do check for `Unmarshal` functions for both `Signature` and `PublicKey`. I do not think it is necessary to check `Aggregate` since we assume `Polybft` uses `Unmarshal` to decode the points and the check has been performed already.

PR #1162, opened 2023-01-25
Expose `ValidatorInfo` struct

Stefan-Ethernal:
# Description

This PR exposes and utilizes the `ValidatorInfo` struct. It represents a data transfer object, which is used to hold all validator data retrieved from the smart contract.
Note: It is worth to consider merging it with the `ValidatorMetadata` object at some point.

# Changes include

- [ ] Breaking change (change that is not backward-compatible and/or changes current functionality)

# Checklist

- [ ] I have added sufficient documentation in the code

## Testing

### DIFF
--- a/consensus/polybft/system_state.go
+++ b/consensus/polybft/system_state.go
@@ -14,6 +14,17 @@ import (
 	"github.com/umbracle/ethgo/contract"
 )
 
+// ValidatorInfo is data transfer object which holds validator information,
+// provided by smart contract
+type ValidatorInfo struct {
+	Address             ethgo.Address
+	Stake               *big.Int
+	TotalStake          *big.Int
+	Commission          *big.Int
+	WithdrawableRewards *big.Int
+	Active              bool
+}
+
 // SystemState is an interface to interact with the consensus system contracts in the chain
 type SystemState interface {
 	// GetValidatorSet retrieves current validator set from the smart contract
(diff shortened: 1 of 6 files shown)


consensus/polybft/system_state.go
```diff
@@ -14,6 +14,17 @@ import (
 	"github.com/umbracle/ethgo/contract"
 )
 
+// ValidatorInfo is data transfer object which holds validator information,
+// provided by smart contract
+type ValidatorInfo struct {
```
Stefan-Ethernal, line 19: Should it be merged with `ValidatorMetadata` at some point?

ferranbt:
What is the problem being solved here?

Stefan-Ethernal:
> What is the problem being solved here?

What we want to achieve here is returning an object representation of validator info instead of map (https://github.com/0xPolygon/polygon-edge/pull/1162/files#diff-090d11924b8e0865632ddb3e8da73ceb5b944c6ee26ee8be367f8e626097480aR46), which is more error prone when extracting values and it has some repetitions in code.

So far this was mostly used in e2e tests, so instead of remembering the keys for extracting values for map (https://github.com/0xPolygon/polygon-edge/pull/1162/files#diff-35f7a314332bee5e2f8472071db16a9916b25e85662af4cca846deb81b78800cL635-L636), we already have object representation which is more straight-forward to use.

b00ris (approved):

goran-ethernal (approved):

igorcrevar (approved):

PR #1163, opened 2023-01-25
Introduce test smart contracts as artifacts

Stefan-Ethernal:
# Description

We currently have two smart contracts for testing purposes: `TestL1StateReceiver` (which is used as dummy implementation for transferring assets from L2 to L1 chain) and `TestWriteBlockMetadata` (which writes header metadata in order to make sure that there are no issues with hashing state of blocks, resolved via #1133).

We are handling these test smart contracts the same way as we are dealing with the rest of the contracts, which is by creating an `Artifact` object for it. The only difference is that since mentioned contracts are exclusively for testing purposes, those are not being auto-generated, based on implementation from `core-contracts` repo, but instead are manually maintained.

# Changes include

# Checklist

## Testing


goran-ethernal (approved):

consensus/polybft/contractsapi/init.go
```diff
@@ -20,10 +18,9 @@ var (
...
-
-	L1ExitTestABI *abi.ABI
+	//go:embed test-contracts/*
+	testContracts       embed.FS
+	TestL1StateReceiver *artifact.Artifact
```
ferranbt: We are not using `embed` with the other artifacts right? I would prefer to standardize that process and not introduce a different workflow. Note @goran-ethernal is also working on parallel on some of this work.

ferranbt: Note that I would prefer to use `embed` too.

Stefan-Ethernal: > We are not using embed with the other artifacts right?

Note that there is one big difference between `TestL1StateReceiver` contract and all the others: `TestL1StateReceiver` is not part of `core-contracts` repo and it is only there to implement missing part in logic of smart contracts (https://github.com/0xPolygon/core-contracts/blob/dev/contracts/root/ExitHelper.sol#L88, as we currently don't have "production" SC counter-part which implements `onL2StateReceived` function). Therefore Boris decided to implement this part as this dummy contract which is going to be used only for testing purposes. Also that contract is not expected to be changed and it is not part of auto generation process and `gen_sc_data.go` (as other smart contracts).

So what we could do (although workflow would still not be the same for test SC comparing to the others) is to automatically copy all the JSON artifacts we have in the `core-contracts` into this `test-contracts` folder (although probably rename it to simply `contracts`) and load appropriate JSONs the same way we are doing with `TestL1StateReceiver`. Does that make sense?

Stefan-Ethernal: > Note @goran-ethernal is also working on parallel on some of this work.

Yep I'm aware of it. I think these changes aren't in collision with his work 🙂

b00ris (approved):

b00ris:
Generally, good. Need to mention that we have one more testing contract https://github.com/0xPolygon/polygon-edge/blob/28f106ac5eba74f05eae3a86720c3c0c103b3360/e2e-polybft/txpool_test.go#L165

Stefan-Ethernal:
> Generally, good. Need to mention that we have one more testing contract
> 
> https://github.com/0xPolygon/polygon-edge/blob/28f106ac5eba74f05eae3a86720c3c0c103b3360/e2e-polybft/txpool_test.go#L165

Oh yeah. For the moment I forgot about that one. I'll consider to add it here as well. Tnx for mentioning it 👍

igorcrevar (approved):

PR #1164, opened 2023-01-25
Fix EVM-372 & EVM-373 already existing directories and files are not checked for ownership and permissions

yaziciahmet:
# Description

This fix addresses the EVM-372 ticket, which states the:

*While creating new file or folder, already existing files and directories are not checked for permissions and ownership. A malicious user might create the file/folder beforehand with high permissions which could cause a security issue.*

This PR brings 2 functions called `SaveFileSafe` and `CreateDirSafe`.

- They check whether the file or directory exists at the given path,
  - if it doesn't, they simply create it with the given permissions (and data in case of file).
  - If it exists, they check whether the owner is current user,
    - if so, just return
    - if not, check whether the owner is in the same group as current user (check this PR for the reason of this decision #1143)
      - if not in the same group, then return error
      - if in the same group, then check whether the permissions are set correctly
        - if permissions are set correctly, continue with creation
        - if not set correctly, return error (since permissions can only be changed by owner, and we are not sure whether it is safe)

**NOTE:** This fix inherently contains EVM-373 as well, since 373 contains only file permissions change.

# Changes include

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/rootchain/server/server.go
+++ b/command/rootchain/server/server.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/0xPolygon/polygon-edge/command/rootchain/helper"
+	"github.com/0xPolygon/polygon-edge/helper/common"
 )
 
 const (
@@ -133,7 +134,7 @@ func runRootchain(ctx context.Context, outputter command.OutputFormatter, closeC
 	}
 
 	// target directory for the chain
-	if err = os.MkdirAll(params.dataDir, 0700); err != nil {
+	if err = common.CreateDirSafe(params.dataDir, 0700); err != nil {
 		return err
 	}
 
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -155,9 +155,9 @@ func CreateDirSafe(path string, perms fs.FileMode) error {
 	return nil
 }
 
-// Creates a file at path and with perms level permissions or updates the file.
-// If file already exists, owner and permissions are verified.
-// If shouldNotExist is true, an error is returned if file exists
+// Creates a file at path and with perms level permissions.
+// If file already exists, owner and permissions are
+// verified, and the file is overwritten.
 func SaveFileSafe(path string, data []byte, perms fs.FileMode) error {
 	info, err := os.Stat(path)
 	// check if an error occurred other than path not exists
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -201,7 +201,7 @@ func verifyFileOwnerAndPermissions(path string, info fs.FileInfo, expectedPerms 
 
 	// check if permissions are set correctly by the owner
 	if info.Mode() != expectedPerms {
-		return fmt.Errorf("permissions of the file/directory is set incorrectly by another user: %s", path)
+		return fmt.Errorf("permissions of the file/directory '%s' are set incorrectly by another user", path)
 	}
 
 	return nil
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -181,9 +181,9 @@ func SaveFileSafe(path string, data []byte, perms fs.FileMode) error {
 	return nil
 }
 
-// Verifies that the file owner is the current user, or the file owner is in
-// the same group as current user and permissions are set correctly by the owner.
-// Returns an error if occurred, and a bool indicating whether the current user is the owner.
+// Verifies that the file owner is the current user,
+// or the file owner is in the same group as current user
+// and permissions are set correctly by the owner.
 func verifyFileOwnerAndPermissions(path string, info fs.FileInfo, expectedPerms fs.FileMode) error {
 	// get stats
 	stat, ok := info.Sys().(*syscall.Stat_t)
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -4,11 +4,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"math"
 	"math/big"
 	"os"
 	"os/signal"
+	"os/user"
 	"path/filepath"
+	"strconv"
 	"syscall"
 
 	"github.com/0xPolygon/polygon-edge/helper/hex"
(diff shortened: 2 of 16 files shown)


helper/common/common.go
```diff
@@ -99,22 +102,137 @@ func DirectoryExists(directoryPath string) bool {
...
+	info, err := os.Stat(path)
+	// check if an error occurred other than path not exists
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	fileExists := FileExists(path)
+	if fileExists && shouldNotExist {
+		return fmt.Errorf("%s already initialized", path)
+	}
```
trimixlover: I would remove this logic and call it where its needed. Most frequent usage of third param for this kind of functions is also bool but should deal with the logic if the file should be created if it doesn't exist. If you still feel like you need a function that does both, maybe its better idea to create a wrapper function instead, something along the line checkOrCreate.

helper/common/common.go
```diff
@@ -99,22 +102,137 @@ func DirectoryExists(directoryPath string) bool {
...
+		return err
+	}
+
+	// update permissions (just in case) if current user is the owner of the directory
+	if isCurrUserOwner {
+		err := os.Chmod(path, expectedPerms)
+		if err != nil {
+			return err
+		}
+	}
```
trimixlover: Is there any reason why you fell you need this just in case block? Sound more risky to update if everything is fine rather than leaving just as it is.

yaziciahmet: If the user wants to upgrade polygon-edge version, already created files and folders will exist with higher permissions. Or maybe the current user just wanted to play around or whatever and created folders manually with higher or lower permissions. In such cases the app should fix it if possible.

trimixlover: Makes sense but still the logic here should not be concerned with upgrades. That should be handled somewhere else clearly stating that its an upgrade logic. And for the user "played with" case, that user maybe changed the permissions on purpose and like them to remain that way.

yaziciahmet: done

helper/common/common.go
yaziciahmet: done

helper/common/common.go
```diff
@@ -99,17 +107,111 @@ func DirectoryExists(directoryPath string) bool {
...
+		return nil
+	}
+
+	// verify that existing directory's owner and permissions are safe
+	err = verifyFileOwnerAndPermissions(path, info, perms)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Stefan-Ethernal:
```suggestion
	return verifyFileOwnerAndPermissions(path, info, perms)
```

helper/common/common.go
```diff
@@ -99,17 +107,111 @@ func DirectoryExists(directoryPath string) bool {
...
+			return err
+		}
+	}
+
+	// create or overwrite the file
+	if err := os.WriteFile(path, data, perms); err != nil {
+		return err
+	}
+
+	return nil
```
Stefan-Ethernal:
```suggestion
	return os.WriteFile(path, data, perms)
```

helper/common/common.go
```diff
@@ -99,17 +107,111 @@ func DirectoryExists(directoryPath string) bool {
...
+	info, err := os.Stat(path)
+	// check if an error occurred other than path not exists
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	if FileExists(path) {
+		// verify that existing file's owner and permissions are safe
+		err := verifyFileOwnerAndPermissions(path, info, perms)
+		if err != nil {
```
Stefan-Ethernal:
```suggestion
		if err := verifyFileOwnerAndPermissions(path, info, perms); err != nil {
```

command/rootchain/server/server.go
```diff
@@ -133,7 +134,7 @@ func runRootchain(ctx context.Context, outputter command.OutputFormatter, closeC
 	}
 
 	// target directory for the chain
-	if err = os.MkdirAll(params.dataDir, 0700); err != nil {
+	if err = common.CreateDirSafe(params.dataDir, 0700); err != nil {
```
Stefan-Ethernal, line 137: Nitpick, maybe it is worth checking is it enough setting permissions to 0600 (I don't see that we need executable permission for this folder, since it just contains rootchain data, so I'd say it is enough that owner of folder is able to read and write data).

```suggestion
	if err = common.CreateDirSafe(params.dataDir, 0600); err != nil {
```

helper/common/common.go
```diff
@@ -99,17 +107,111 @@ func DirectoryExists(directoryPath string) bool {
...
+
+	// get group id of the owner
+	ownerGID := strconv.FormatUint(uint64(stat.Gid), 10)
+	if currUser.Gid != ownerGID {
+		return fmt.Errorf("file/directory created by a user from a different group: %s", path)
+	}
+
+	// check if permissions are set correctly by the owner
+	if info.Mode() != expectedPerms {
+		return fmt.Errorf("permissions of the file/directory is set incorrectly by another user: %s", path)
```
Stefan-Ethernal:
```suggestion
		return fmt.Errorf("permissions of the file/directory '%s' are set incorrectly by another user", path)
```

Stefan-Ethernal (commented):
Leaving some minor suggestions.

Bringing additional two points to think through:
- Should we close #1143 in favour of this PR, since this one addresses that one as well?
- Not sure do we need executable permissions set anywhere? 🤔 I can not remember that we have anywhere a file or folder, created by Edge itself, which contains some executable within it.

helper/common/common.go
yaziciahmet: Done

helper/common/common.go
yaziciahmet: Done

helper/common/common.go
yaziciahmet: Done

command/rootchain/server/server.go
yaziciahmet: Executable permission on the folder does not mean you can execute executables within this folder. It allows entering into the directory (`cd`), and access inside of the folder. And to remove or add files to a directory both write and execute permissions are needed.

yaziciahmet:
We can close #1143 @Stefan-Ethernal

helper/common/common.go
yaziciahmet: Done

command/rootchain/server/server.go
Stefan-Ethernal: My bad.

Stefan-Ethernal:
> We can close #1143 @Stefan-Ethernal

Cool. Close it whenever you feel it is the moment. 🙂

Stefan-Ethernal (approved):

Kourin1996 (approved):

trimixlover (approved):

PR #1170, opened 2023-01-27
EVM-416 Use one struct for Exit and StateSync proofs in jsonrpc

goran-ethernal:
# Description

This PR uses one placeholder struct for both `ExitProof` and `StateSyncProof` for jsonrpc calls.

```
type Proof struct {
	Data     []Hash // the proof himself
	Metadata map[string]interface{}
}
```

# Changes include

# Checklist

## Testing


consensus/polybft/state_sync_manager.go
```diff
@@ -29,12 +28,41 @@ const (
 	minCommitmentSize = 2
 )
 
+type StateSyncProof struct {
+	Proof     []types.Hash
+	StateSync *contractsapi.StateSyncedEvent
+}
+
+// EncodeAbi contains logic for encoding given ABI data
+func (ssp *StateSyncProof) EncodeAbi() ([]byte, error) {
```
ferranbt: Why does this data structure need to `EncodeAbi`? My logic is that anything that requires abi encoding should be inside contractsapi because its a contract call.

goran-ethernal: Might be some legacy. I'm still working on this, so I will check if we can remove this.

goran-ethernal: Removed it

igorcrevar (approved):
Approved. Couldn't we use generics for this?

consensus/polybft/statesyncrelayer/state_sync_relayer_test.go
```diff
@@ -68,7 +68,20 @@ func Test_executeStateSync(t *testing.T) {
...
+			"StateSync": map[string]interface{}{
+				"ID":       sp.Obj.ID,
+				"Sender":   sp.Obj.Sender,
+				"Receiver": sp.Obj.Receiver,
+				"Data":     sp.Obj.Data,
+			},
+		},
+	}
+
+	r.executeStateSync(proof)
```
Stefan-Ethernal: I know this is a unit test from before, but consider adding assertion which would check that there is no error.

Stefan-Ethernal (approved):

consensus/polybft/statesyncrelayer/state_sync_relayer_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1170/commits/b8e716f22d228277b2b8f0cb757657e7ef82cfa6

vcastellm (commented):
Can you remind me the main motivation for this change?

In principle I prefer not to introduce this change, as this needs introduction of type castings, and less specific types hiding different underlaying structs.

goran-ethernal:
> Can you remind me the main motivation for this change?
> 
> In principle I prefer not to introduce this change, as this needs introduction of type castings, and less specific types hiding different underlaying structs.

It's something @ferranbt proposed because of that cycle import between `types` package and `contractsapi`.

Nemanja0x (approved):

ferranbt (approved):

PR #1175, opened 2023-01-30
EVM-369 fix architecture dependent `int` type

yaziciahmet:
# Description

`int` type in golang is architecture dependent, and usage of it in an application where there is a network of computers which shares information, it could be problematic to have 32-bit architecture send information to a 64-bit computer, and vice-verse, since `int` type won't be interpreted the same.

This fix introduces the replacement of `int` type (as much as possible), especially on the bits of the application where information exchange occurs. It is replaced by `int64`, since 32-bit architectures can also use it, even though it adds some overhead, it is negligible.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -42,7 +42,7 @@ func setFlags(cmd *cobra.Command) {
 		"the directory for the Polygon Edge genesis data",
 	)
 
-	cmd.Flags().Int64Var(
+	cmd.Flags().Uint64Var(
 		&params.chainID,
 		chainIDFlag,
 		command.DefaultChainID,
@@ -207,7 +207,7 @@ func setFlags(cmd *cobra.Command) {
 // with running partners
 func setLegacyFlags(cmd *cobra.Command) {
 	// Legacy chainid flag
-	cmd.Flags().Int64Var(
+	cmd.Flags().Uint64Var(
 		&params.chainID,
 		chainIDFlagLEGACY,
 		command.DefaultChainID,
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -42,7 +42,7 @@ func setFlags(cmd *cobra.Command) {
 		"the directory for the Polygon Edge genesis data",
 	)
 
-	cmd.Flags().Uint64Var(
+	cmd.Flags().Int64Var(
 		&params.chainID,
 		chainIDFlag,
 		command.DefaultChainID,
@@ -207,7 +207,7 @@ func setFlags(cmd *cobra.Command) {
 // with running partners
 func setLegacyFlags(cmd *cobra.Command) {
 	// Legacy chainid flag
-	cmd.Flags().Uint64Var(
+	cmd.Flags().Int64Var(
 		&params.chainID,
 		chainIDFlagLEGACY,
 		command.DefaultChainID,
--- a/command/ibft/snapshot/ibft_snapshot.go
+++ b/command/ibft/snapshot/ibft_snapshot.go
@@ -19,7 +19,7 @@ func GetCommand() *cobra.Command {
 }
 
 func setFlags(cmd *cobra.Command) {
-	cmd.Flags().IntVar(
+	cmd.Flags().Int64Var(
 		&params.blockNumber,
 		numberFlag,
 		-1,
--- a/command/ibft/snapshot/ibft_snapshot.go
+++ b/command/ibft/snapshot/ibft_snapshot.go
@@ -1,6 +1,8 @@
 package snapshot
 
 import (
+	"math"
+
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/0xPolygon/polygon-edge/command/helper"
 	"github.com/spf13/cobra"
@@ -19,10 +21,10 @@ func GetCommand() *cobra.Command {
 }
 
 func setFlags(cmd *cobra.Command) {
-	cmd.Flags().Int64Var(
+	cmd.Flags().Uint64Var(
 		&params.blockNumber,
 		numberFlag,
-		-1,
+		math.MaxUint64,
 		"the block height (number) for the snapshot",
 	)
 }
--- a/consensus/polybft/statesyncrelayer/state_sync_relayer.go
+++ b/consensus/polybft/statesyncrelayer/state_sync_relayer.go
@@ -135,7 +135,7 @@ func (r *StateSyncRelayer) AddLog(log *ethgo.Log) {
 
 		r.logger.Info("Commit", "Block", log.BlockNumber, "StartID", startID, "EndID", endID)
 
-		for i := startID.Int64(); i <= endID.Int64(); i++ {
+		for i := startID.Uint64(); i <= endID.Uint64(); i++ {
 			// query the state sync proof
 			stateSyncProof, err := r.queryStateSyncProof(fmt.Sprintf("0x%x", i))
 			if err != nil {
--- a/consensus/polybft/statesyncrelayer/state_sync_relayer.go
+++ b/consensus/polybft/statesyncrelayer/state_sync_relayer.go
@@ -137,7 +137,7 @@ func (r *StateSyncRelayer) AddLog(log *ethgo.Log) {
 
 		for i := startID.Int64(); i <= endID.Int64(); i++ {
 			// query the state sync proof
-			stateSyncProof, err := r.queryStateSyncProof(fmt.Sprintf("0x%x", int(i)))
+			stateSyncProof, err := r.queryStateSyncProof(fmt.Sprintf("0x%x", i))
 			if err != nil {
 				r.logger.Error("Failed to query state sync proof", "err", err)
 
--- a/validators/store/snapshot/snapshot.go
+++ b/validators/store/snapshot/snapshot.go
@@ -520,7 +520,7 @@ func (s *SnapshotValidatorStore) removeLowerSnapshots(
 	currentHeight uint64,
 ) {
 	// remove in-memory snapshots from two epochs before this one
-	lowerEpoch := int(currentHeight/s.epochSize) - 2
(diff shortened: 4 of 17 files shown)


command/genesis/genesis.go
```diff
@@ -207,7 +207,7 @@ func setFlags(cmd *cobra.Command) {
 // with running partners
 func setLegacyFlags(cmd *cobra.Command) {
 	// Legacy chainid flag
-	cmd.Flags().Uint64Var(
+	cmd.Flags().Int64Var(
```
Kourin1996: Is there any reason to change uint64 to int64?

Kourin1996: I just thought chain ID could be unsigned int

yaziciahmet: I thought the same at first, but after our discussions with Vuk, we decided that user should be able to give negative chain id if they feel like it.
But now I realized that if that is the case, `GetChainID` function inside `blockchain_wrapper.go` should also return int64 instead of uint64. It would be weird if user entered a negative number as chain id, but retrieved a very big uint64 number.

Kourin1996: I think there is no problem even if chain ID is negative but I'm not sure interface using hex value can adopt that.

validators/store/snapshot/snapshot.go
```diff
@@ -520,7 +520,7 @@ func (s *SnapshotValidatorStore) removeLowerSnapshots(
 	currentHeight uint64,
 ) {
 	// remove in-memory snapshots from two epochs before this one
-	lowerEpoch := int(currentHeight/s.epochSize) - 2
+	lowerEpoch := int64(currentHeight/s.epochSize) - 2
```
Stefan-Ethernal: Motivation for this change is not to limit ourselves only to `int64` range. WDYT?
Also this constant `preserveEpochs` should be placed in the const block at the top of the file.

```suggestion
	const preservedEpochs = 2

	currentEpoch := currentHeight / s.epochSize
	if currentEpoch < preservedEpochs {
		return
	}

	// remove in-memory snapshots from two epochs before this one
	lowerEpoch := currentEpoch - preservedEpochs
	purgeBlock := lowerEpoch * s.epochSize
	s.store.deleteLower(purgeBlock)
```

command/ibft/snapshot/ibft_snapshot.go
```diff
@@ -19,7 +19,7 @@ func GetCommand() *cobra.Command {
 }
 
 func setFlags(cmd *cobra.Command) {
-	cmd.Flags().IntVar(
+	cmd.Flags().Int64Var(
```
Stefan-Ethernal: Thinking out loud:
should we change this to be `Uint64` type, but treat it like non-mandatory flag. If not provided, we are assuming that latest block snapshot should be retrieved. WDYT, is it feasible to do so?

consensus/polybft/statesyncrelayer/state_sync_relayer.go
```diff
@@ -135,7 +135,7 @@ func (r *StateSyncRelayer) AddLog(log *ethgo.Log) {
 
 		for i := startID.Int64(); i <= endID.Int64(); i++ {
```
Stefan-Ethernal: AFAIK, state sync ids cannot be negative numbers (https://github.com/0xPolygon/core-contracts/blob/dev/contracts/root/StateSender.sol#L8), so it is better to treat it as `Uint64` instead:

```suggestion
                for i := startID.Uint64(); i <= endID.Uint64(); i++ {
```

Stefan-Ethernal (commented):
Also audit report mentioned these parts to consider for refactoring as well:
- bitmap, used by the EVM state: https://github.com/0xPolygon/polygon-edge/blob/4c5d0d930e32c1c894f9016f82002a51625ad45d/state/runtime/evm/state.go#L111-L118
- ConvertUnmarshalledInt function is not supporting `uint64`, but only `int64` (however it is tricky because of `float64` case and `roundFloat` function): https://github.com/0xPolygon/polygon-edge/blob/4c5d0d930e32c1c894f9016f82002a51625ad45d/helper/common/common.go#L49-L54

command/ibft/snapshot/ibft_snapshot.go
yaziciahmet: In `getSnapshotRequest` function, the negative value of block number is used to determine whether the latest block is requested, or the block with the given number is requested.
Maybe it can be converted to `uint64` and the default value for requesting latest block could be set to `MAX_UINT64`, so that possible block number values to be requested with `snapshot` command would be doubled.

Stefan-Ethernal: Sounds reasonable and the most straightforward thing to do (I was even thinking about `*uint64`, and somehow setting `nil` if that flag wasn't supplied, but probably it would overcomplicate stuff).

yaziciahmet:
@Stefan-Ethernal about https://github.com/0xPolygon/polygon-edge/pull/1175#pullrequestreview-1282854698
- bitmap should be no problem at first sight. Because the size of an array is bounded by the architecture internally. Meaning that a 32-bits arch will panic anyways if you try to have an array size of higher the `int32` (or `uint32`). Since bitmap is `[]byte`, it has an internal bound caused by the architecture, hence converting it or indexing it with `uint64` does not make much sense. But the thing is, `isSet` method on `bitmap` does the following calculation `b.buf[i/bitmapSize]`, so the index bounded by the architecture, is actually not `i`, it is `i/bitmapSize`. Since `bitmapSize = 8`, `i` could actually go up to smh. like 2^35. So in the end, it could make sense to make the `i` `uint64`. WDYT?
- the issue with `ConvertUnmarshalledInt` is that. how to handle negative numbers. With the current version, if there is a negative number, then the number has to written in hex since it uses `strconv.ParseUint`. So `-10` can't be written, you have to write `0xfffff...` smh. But if it is meant to be a very big `uint64` number (big meaning that it is above the positive bound of `int64`), then internally it will be interpreted as negative number due to `int64()` conversion at the end. So the question is do we want to accept negative numbers? If we do, it would be better to read them as `int64` directly, if we don't, then we can remove the `int64()` conversion. WDYT?

validators/store/snapshot/snapshot.go
yaziciahmet: Done

command/ibft/snapshot/ibft_snapshot.go
yaziciahmet: Done

consensus/polybft/statesyncrelayer/state_sync_relayer.go
yaziciahmet: Done

command/genesis/genesis.go
yaziciahmet: Done

Stefan-Ethernal:
> the issue with ConvertUnmarshalledInt is that. how to handle negative numbers. With the current version, if there is a negative number, then the number has to written in hex since it uses strconv.ParseUint. So -10 can't be written, you have to write 0xfffff... smh. But if it is meant to be a very big uint64 number (big meaning that it is above the positive bound of int64), then internally it will be interpreted as negative number due to int64() conversion at the end. So the question is do we want to accept negative numbers? If we do, it would be better to read them as int64 directly, if we don't, then we can remove the int64() conversion. WDYT?

Thinking out loud:
- `ConvertUnmarshalledInt` is currently called only when `JSONNumber` is being unmarshaled (https://github.com/0xPolygon/polygon-edge/blob/develop/helper/common/common.go#L219L237). 
- `JSONNumber` internally holds `uint64` value. Therefore I'd say we are not interested in negative values in this particular case. 
- However, it is worth considering whether should we have two functions for converting raw bytes into: `uint64` and `int64` (e.g. `ConvertUnmarshalledInt` and `ConvertUnmarshalledUint`)...? Alternatively maybe we don't even need `ConvertUnmarshalledInt` atm and we can add it in the future when needed.

Stefan-Ethernal:
> bitmap should be no problem at first sight. Because the size of an array is bounded by the architecture internally. Meaning that a 32-bits arch will panic anyways if you try to have an array size of higher the int32 (or uint32). Since bitmap is []byte, it has an internal bound caused by the architecture, hence converting it or indexing it with uint64 does not make much sense. But the thing is, isSet method on bitmap does the following calculation b.buf[i/bitmapSize], so the index bounded by the architecture, is actually not i, it is i/bitmapSize. Since bitmapSize = 8, i could actually go up to smh. like 2^35. So in the end, it could make sense to make the i uint64. WDYT?

Makes sense, at least that is something we have done in this bitmap implementation as well (https://github.com/0xPolygon/polygon-edge/blob/develop/consensus/polybft/bitmap/bitmap.go).

Though I'm wondering should we change `setCode` function like this (and also don't cast `bitmapSize` constant to `uint64`). WDYT?

```go
func (b *bitmap) setCode(code []byte) {
	codeSize := len(code)
	b.buf = common.ExtendByteSlice(b.buf, codeSize/bitmapSize+1)

	for i := 0; i < codeSize; {
		c := code[i]

		if isPushOp(c) {
			// push op
			i += int(c) - 0x60 + 2
		} else {
			if c == 0x5B {
				// jumpdest
				b.set(uint64(i))
			}
			i++
		}
	}
}
```

yaziciahmet:
@Stefan-Ethernal 
I decided to go with only handling uint64, and leaving int64 for the future. The reasoning behind this decision is that the `UnmarshalJSON` method of `JSONNumber` is the only place that uses `ConvertUnmarshalledInt`, and `UnmarshalJSON` is not used even in a single place. So in the future if the use of `UnmarshalJSON` is somehow desirable, int64 vs uint64 case can be dealt later based on the requirements.

Stefan-Ethernal (approved):

Kourin1996 (approved):

trimixlover (approved):

PR #1176, opened 2023-01-30
Introduce base fee

begmaroman:
# Description

- Introduced base fee field which is part of EIP-1559 feature
- Minor code style changes  

# Changes include

# Checklist

## Testing


blockchain/blockchain_test.go
```diff
@@ -1335,3 +1335,48 @@ func TestBlockchain_VerifyBlockBody(t *testing.T) {
...
+	}{
+		{6, chain.GenesisBaseFee, 20000000, 10000000, chain.GenesisBaseFee}, // usage == target
+		{6, chain.GenesisBaseFee, 20000000, 9000000, 987500000},             // usage below target
+		{6, chain.GenesisBaseFee, 20000000, 11000000, 1012500000},           // usage above target
+		{6, chain.GenesisBaseFee, 20000000, 20000000, 1125000000},           // usage full
+		{6, chain.GenesisBaseFee, 20000000, 0, 875000000},                   // usage 0
+	}
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
```
Stefan-Ethernal: Don't forget to capture loop variable, because of parallel execution of table sub tests

```suggestion
	for i, test := range tests {
	        test := test
		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
```

chain/chain.go
```diff
@@ -12,12 +12,15 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-var (
+const (
+	// GenesisBaseFee is the initial base fee for EIP-1559 blocks.
+	GenesisBaseFee = 1000000000
```
Stefan-Ethernal: Would be great if we could use scientific notation here. 
```suggestion
	GenesisBaseFee = uint64(1e9)
```
However, I think it wouldn't work, because of `types.EncodeUint64`, which doesn't handle scientific notation, but only decimal and hexadecimal numbers. 

So I'd suggest just add a comment suggesting `GenesisBaseFee` is 1 billion by default (so that we don't need to count zero every time we want to check what it genesis base fee 😄 ).

types/rlp_marshal.go
```diff
@@ -91,6 +91,8 @@ func (h *Header) MarshalRLPWith(arena *fastrlp.Arena) *fastrlp.Value {
 	vv.Set(arena.NewBytes(h.MixHash.Bytes()))
 	vv.Set(arena.NewCopyBytes(h.Nonce[:]))
 
+	vv.Set(arena.NewUint(h.BaseFee))
```
ferranbt, line 95: How is this backward compatible? Chains running the old version will not have this field.

consensus/polybft/consensus_runtime.go
```diff
@@ -258,17 +258,17 @@ func (c *consensusRuntime) OnBlockInserted(fullBlock *types.FullBlock) {
 	postBlock := &PostBlockRequest{FullBlock: fullBlock, Epoch: epoch.Number, IsEpochEndingBlock: isEndOfEpoch}
 
 	// handle commitment and proofs creation
-	if err := c.stateSyncManager.PostBlock(postBlock); err != nil {
```
ferranbt: Please revert all the code style changes like this. It gets tricky to review this alongside business logic.

blockchain/blockchain.go
```diff
@@ -1434,3 +1443,33 @@ func (b *Blockchain) GetBlockByNumber(blockNumber uint64, full bool) (*types.Blo
 func (b *Blockchain) Close() error {
 	return b.db.Close()
 }
+
+// CalculateBaseFee calculates the basefee of the header.
+func (b *Blockchain) CalculateBaseFee(parent *types.Header) uint64 {
```
ferranbt, line 1448: There is not explanation here nor in the RFC on how is this base fee being calculated. Please provide comments here and in the RFC.

blockchain/blockchain.go
```diff
@@ -1434,3 +1443,33 @@ func (b *Blockchain) GetBlockByNumber(blockNumber uint64, full bool) (*types.Blo
...
+		y := parent.BaseFee * gasUsedDelta / parentGasTarget
+		baseFeeDelta := y / defaultBaseFeeChangeDenom
+
+		return parent.BaseFee + common.Max(baseFeeDelta, 1)
+	}
+
+	// Otherwise, if the parent block used less gas than its target, the baseFee should decrease.
+	gasUsedDelta := parentGasTarget - parent.GasUsed
+	y := parent.BaseFee * gasUsedDelta / parentGasTarget
+	baseFeeDelta := y / defaultBaseFeeChangeDenom
```
vcastellm: This code is duplicated, would be great to maintain these formulas in a function

chain/chain.go
vcastellm: Can we use `1_000_000_000`? much much 👓

command/genesis/polybft_params.go
```diff
@@ -30,11 +30,9 @@ const (
 	bridgeFlag                 = "bridge-json-rpc"
 
 	defaultManifestPath     = "./manifest.json"
-	defaultEpochSize        = uint64(10)
```
vcastellm, line 33: Where this gone?

jsonrpc/testsuite/block-empty.json
```diff
@@ -18,5 +18,6 @@
     "nonce": "0x0a00000000000000",
     "hash": "0x0800000000000000000000000000000000000000000000000000000000000000",
     "transactions": null,
-    "uncles": null
+    "uncles": null,
+    "baseFee": "0xf"
 }
```
vcastellm:
```suggestion
}

```

jsonrpc/testsuite/block-with-txn-bodies.json
```diff
@@ -35,5 +35,6 @@
             "transactionIndex": "0x2"
         }
     ],
-    "uncles": null
+    "uncles": null,
+    "baseFee": "0xf"
 }
```
vcastellm:
```suggestion
}

```

jsonrpc/testsuite/block-with-txn-hashes.json
```diff
@@ -20,5 +20,6 @@
     "transactions": [
         "0x0800000000000000000000000000000000000000000000000000000000000000"
     ],
-    "uncles": null
+    "uncles": null,
+    "baseFee": "0xf"
 }
```
vcastellm:
```suggestion
}

```

types/rlp_unmarshal.go
```diff
@@ -376,44 +381,64 @@ func (t *Transaction) unmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
+
+	if numElems := len(elems); numElems != num {
+		return fmt.Errorf("incorrect number of transaction elements, expected %d but found %d", num, numElems)
 	}
 
 	p.Hash(t.Hash[:0], v)
 
 	// nonce
-	if t.Nonce, err = elems[0].GetUint64(); err != nil {
+	if t.Nonce, err = getElem().GetUint64(); err != nil {
```
vcastellm, line 412: Am I missing something? is the `getElem` dynamic? how it knows the element index?

blockchain/blockchain.go
begmaroman: There is an explanation under Blockchain section in RFC with the link to original source.

command/genesis/polybft_params.go
begmaroman: Unused

consensus/polybft/consensus_runtime.go
begmaroman: Done

types/rlp_unmarshal.go
begmaroman: Kind of. Dynamic base fee tx has two more fields instead of gas price. Basically, take a look at the code above to find an elements number depending on tx type.

blockchain/blockchain.go
begmaroman: Done

chain/chain.go
begmaroman: Done

types/rlp_marshal.go
begmaroman: In the marshaling function, it would just set it to 0 which is completely fine. In the unmarshaling function (which is more backward-compatible-sensitive), I added a len check so it won't panic for existing headers.

jsonrpc/types.go
```diff
@@ -102,6 +102,7 @@ type block struct {
 	Hash            types.Hash          `json:"hash"`
 	Transactions    []transactionOrHash `json:"transactions"`
 	Uncles          []types.Hash        `json:"uncles"`
+	BaseFee         argUint64           `json:"baseFee"`
```
ferranbt: Blocks without `London` fork enabled should not have the `BaseFee`.

types/rlp_marshal.go
ferranbt: We should discuss this point a bit more because it seems important the backward compatibility. What geth does is that this field is a point and it is only set (marshalled) if `London` is enabled. I do not have a clear answer to this right now.

types/rlp_unmarshal.go
```diff
@@ -376,44 +381,64 @@ func (t *Transaction) unmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
+
+	var num int
+
+	switch t.Type {
+	case LegacyTx:
+		num = 9
+	case DynamicFeeTx:
+		num = 10
+	case StateTx:
+		num = 10
```
```suggestion
	case DynamicFeeTx, StateTx:
		num = 10
```

types/rlp_unmarshal.go
vcastellm: I see, LGTM

vcastellm (approved):
LGTM

types/rlp_unmarshal.go
```diff
@@ -376,44 +384,62 @@ func (t *Transaction) unmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
+
+	if numElems := len(elems); numElems != num {
+		return fmt.Errorf("incorrect number of transaction elements, expected %d but found %d", num, numElems)
 	}
 
 	p.Hash(t.Hash[:0], v)
 
 	// nonce
-	if t.Nonce, err = elems[0].GetUint64(); err != nil {
+	if t.Nonce, err = getElem().GetUint64(); err != nil {
```
goran-ethernal, line 412: Correct me if I'm wrong, but this getElem will not return the correct values if this t.Type == DynamicFeeTx condition is met, because it goes from the first element in array and decreases the array size by each call, and we were extracting the 10th and 11th index from the array before.

types/rlp_unmarshal.go
```diff
@@ -376,44 +384,62 @@ func (t *Transaction) unmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
...
 		return err
 	}
 
 	if t.Type == DynamicFeeTx {
 		// gasFeeCap
 		t.GasFeeCap = new(big.Int)
-		if err = elems[10].GetBigInt(t.GasFeeCap); err != nil {
+		if err = getElem().GetBigInt(t.GasFeeCap); err != nil {
```
goran-ethernal, line 419: Correct me if I'm wrong, but this `getElem` will not return the correct values if this `t.Type == DynamicFeeTx` condition is met, because it goes from the first element in array and decreases the array size by each call, and we were extracting the 10th and 11th index from the array before.

jsonrpc/types.go
begmaroman: Added omitempty option

types/rlp_unmarshal.go
begmaroman: It returns elements with the same indecies basically. Dynamic base fee tx has these two fields instead of the gas price one (according to eth spec). Could you please provide more details if I'm missing something? Maybe a specific test case when it would not work?

types/rlp_marshal.go
begmaroman: Added TODO before this line so we could discuss it later.

types/rlp_unmarshal.go
goran-ethernal: Yeah, it confused me, because I am looking the code comparison, where on branch `feature/EVM-262-EIP-1559` on which this PR is created, you are taking these `GasFeeCap` and `GasTipCap` from 10th and 11th index, but I didn't see the change in `MarshalRLP` when you insert this data in array.

But I see you changed in some earlier PR the `MarshalRLP` function, that got merged to `v3-parity`, so those fields get marshaled on 1st and 2nd index if it is a `DynamicFeeTx`, so it looks like everything is fine in this PR.

Do you need to rebase this PR to `v3-parity` perhaps?

begmaroman: Yea, I've changed ordering according to another eip so it should be: (gas price) OR (base tip cap, base fee cap). Both marshaling and unmarshaling functions were updated accordingly. This branch should be against `feature/EVM-262-EIP-1559` since that one is the parent branch that should be reviewed after everything related to 1559 is done.

jsonrpc/types.go
```diff
@@ -102,6 +102,7 @@ type block struct {
 	Hash            types.Hash          `json:"hash"`
 	Transactions    []transactionOrHash `json:"transactions"`
 	Uncles          []types.Hash        `json:"uncles"`
+	BaseFee         argUint64           `json:"baseFee,omitempty"`
```
ferranbt, line 105: Can we test this with a use case in the `testsuite`?

ferranbt: I mean, test that it is not marshaled if zero.

begmaroman: Done

ferranbt (approved):

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #1180, opened 2023-01-31
opReturnCopyData panic fix

stana-miric:
# Description

The problem with the opReturnDataCopy, as stated in the audit, was when length was set to 0 and memoryOffset was greater than memory length. Length being 0 caused no data allocation. In normal scenario when length (data size) is greater than 0, allocateMemory fn would allocate enough space so that data can be copied to memory starting from the memoryOffset. In the case of 0 length, where no memory was allocated, panic occurred while trying to copy data accessing the memory out of its bounds.
The fix includes:
 - checking memory length against sum of the length and memOffset (as suggested in the audit report) 
 - also, before even allocating the memory there is a check if length equals 0, because it that case there is no need for memory allocation nor for data copying and the fn returns. This excludes further unnecessary calculations, which will result in no data copying anyway. This means that the added check for valid memory length should never return an error because in the case of either unsuccessful memory allocation or length size 0 the fn returned before reaching this point. Putting length check before allocating memory will give precedence to the length in the case when 	length is 0 and memoryOffset is invalid (negative), because in that case we won't have error return (which is expected for negative memOffset) instead the fn will just return without error (which 	should be the case for the 0 length as stated in the audit that is behavior of go-ethereum). If we want to give precedence to the invalid memOffset the check for the length can be put after allocation.
 	
All other places in the code which writes data to the memory after calling allocate(opExtCodeCopy, opCallDataCopy, opCodeCopy, get2), have a check if size is greater than 0 except opCall (allocate is called from buildCallContract). The fix for opCall is included in this PR as well, and tests for this specific scenario are added for both opReturnDataCopy and opCall.

An alternative fix could be removing check for size==0 in allocateMemory fn, which will allocated additional space if memory offset is greater than memory length, but that would be unnecessary allocation since the data length is 0 and no data would be copied.

# Changes include

# Checklist

## Testing

- [x] I have tested this code with the unit test

### DIFF
--- a/state/runtime/evm/instructions.go
+++ b/state/runtime/evm/instructions.go
@@ -813,7 +813,8 @@ func opReturnDataCopy(c *state) {
 	dataOffset := c.pop()
 	length := c.pop()
 
-	if !c.allocateMemory(memOffset, length) {
+	// if length is 0, return immediately since no need for the data copying nor memory allocation
+	if length.Sign() == 0 || !c.allocateMemory(memOffset, length) {
 		return
 	}
 
@@ -842,6 +843,14 @@ func opReturnDataCopy(c *state) {
 	}
 
 	data := c.returnData[dataOffset.Uint64():dataEndIndex]
+	// this should not happen because the memory is allocated considering both memOffset and data length and
+	// in the case of length==0 the fn returns immediately, not reaching this point
+	if uint64(len(c.memory)) < memOffset.Uint64()+uint64(len(data)) {
+		c.exit(errReturnDataOutOfBounds)
+
+		return
+	}
+
 	copy(c.memory[memOffset.Uint64():], data)
 }
 
@@ -1173,7 +1182,7 @@ func opCall(op OpCode) instruction {
 		}
 
 		if result.Succeeded() || result.Reverted() {
-			if len(result.ReturnValue) != 0 {
+			if len(result.ReturnValue) != 0 && offset+size <= uint64(len(c.memory)) {
 				copy(c.memory[offset:offset+size], result.ReturnValue)
 			}
 		}
--- a/state/runtime/evm/instructions.go
+++ b/state/runtime/evm/instructions.go
@@ -813,18 +813,19 @@ func opReturnDataCopy(c *state) {
 	dataOffset := c.pop()
 	length := c.pop()
 
+	if !dataOffset.IsUint64() {
+		c.exit(errReturnDataOutOfBounds)
+
+		return
+	}
+
 	// if length is 0, return immediately since no need for the data copying nor memory allocation
 	if length.Sign() == 0 || !c.allocateMemory(memOffset, length) {
 		return
 	}
 
-	if !dataOffset.IsUint64() {
-		c.exit(errGasUintOverflow)
-
-		return
-	}
-
-	if ulength := length.Uint64(); !c.consumeGas(((ulength + 31) / 32) * copyGas) {
+	ulength := length.Uint64()
+	if !c.consumeGas(((ulength + 31) / 32) * copyGas) {
 		return
 	}
 
@@ -843,15 +844,7 @@ func opReturnDataCopy(c *state) {
 	}
 
 	data := c.returnData[dataOffset.Uint64():dataEndIndex]
-	// this should not happen because the memory is allocated considering both memOffset and data length and
-	// in the case of length==0 the fn returns immediately, not reaching this point
-	if uint64(len(c.memory)) < memOffset.Uint64()+uint64(len(data)) {
-		c.exit(errReturnDataOutOfBounds)
-
-		return
-	}
-
-	copy(c.memory[memOffset.Uint64():], data)
+	copy(c.memory[memOffset.Uint64():memOffset.Uint64()+ulength], data)
 }
 
 func opCodeCopy(c *state) {
@@ -1182,7 +1175,7 @@ func opCall(op OpCode) instruction {
 		}
 
 		if result.Succeeded() || result.Reverted() {
-			if len(result.ReturnValue) != 0 && offset+size <= uint64(len(c.memory)) {
+			if len(result.ReturnValue) != 0 && size > 0 {
 				copy(c.memory[offset:offset+size], result.ReturnValue)
 			}
 		}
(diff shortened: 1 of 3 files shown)


stana-miric:

state/runtime/evm/instructions_test.go
```diff
@@ -101,20 +101,25 @@ func TestMStore(t *testing.T) {
 	assert.Len(t, s.memory, 1024+32)
 }
 
-type mockHostForCreate struct {
+type mockHostForInstrustions struct {
```
goran-ethernal:
```suggestion
type mockHostForInstructions struct {
```

state/runtime/evm/instructions.go
```diff
@@ -1173,7 +1182,7 @@ func opCall(op OpCode) instruction {
 		}
 
 		if result.Succeeded() || result.Reverted() {
-			if len(result.ReturnValue) != 0 {
+			if len(result.ReturnValue) != 0 && offset+size <= uint64(len(c.memory)) {
```
goran-ethernal: What happens if the `offset+size` is larger than the length of memory? Should we exit if this is true, like we do in `opReturnDataCopy`?

state/runtime/evm/instructions.go
```diff
@@ -842,6 +843,14 @@ func opReturnDataCopy(c *state) {
...
 
 	data := c.returnData[dataOffset.Uint64():dataEndIndex]
+	// this should not happen because the memory is allocated considering both memOffset and data length and
+	// in the case of length==0 the fn returns immediately, not reaching this point
+	if uint64(len(c.memory)) < memOffset.Uint64()+uint64(len(data)) {
+		c.exit(errReturnDataOutOfBounds)
+
+		return
+	}
+
```
vcastellm: If this code path is never reached why are we checking it? how are we going to test it?

state/runtime/evm/instructions.go
stana-miric: offset+size will be  larger than the length of memory only if the size is 0 because only in that case memory would not be allocated. according to the behavior in go-ethereum(code sample given in the audit) if size is not greater than 0 then nothing should be done. But probably would be better if I put condition size>0 as it is done in other places (opExtCodeCopy, opCallDataCopy, opCodeCopy, get2) when memory is accessed after allocate is called. 
I can change this condition to: if len(result.ReturnValue) != 0 && size>0  to be aligned with other places

state/runtime/evm/instructions.go
stana-miric: I agree that this is not needed, because if the execution got to this point it means that size is greater then 0 and allocate fn allocated enough memory taking into consideration both memOffset and size. 
I'll remove this check.

state/runtime/evm/instructions.go
```diff
@@ -813,7 +813,8 @@ func opReturnDataCopy(c *state) {
 	dataOffset := c.pop()
 	length := c.pop()
 
-	if !c.allocateMemory(memOffset, length) {
+	// if length is 0, return immediately since no need for the data copying nor memory allocation
+	if length.Sign() == 0 || !c.allocateMemory(memOffset, length) {
```
Stefan-Ethernal: A couple of points:
1. This is interesting as `allocateMemory` already checks if `length == 0`, but for some (strange) reason, it returns `true`. It is a potential landmine changing it to return `false`, because of all the other places the `allocateMemory` is being called... But again, maybe it is worth considering returning `false` in that case and carefully checking all the other places where it is being called.

2. Another topic, unrelated to this PR (but it is a good opportunity to bring it out now, since I've compared this implementation with go ethereum one). We are exiting with wrong error [here](https://github.com/0xPolygon/polygon-edge/pull/1180/files#diff-18194ed7729676269d52dc687dde3cb1d55201d1e44ad585e550aa938f6ccb15R821):

```go
if !dataOffset.IsUint64() {
		c.exit(errGasUintOverflow)
		return
	}
```
instead of it, we should exit with this error:
```go
if !dataOffset.IsUint64() {
		c.exit(errReturnDataOutOfBounds)

		return
	}
```
Finally, it is worth considering checking this condition at the very beginning of the `opReturnDataCopy` function, as it is cheaper to check it instead of calling `allocateMemory`.

state/runtime/evm/instructions.go
```diff
@@ -842,6 +843,14 @@ func opReturnDataCopy(c *state) {
...
 	data := c.returnData[dataOffset.Uint64():dataEndIndex]
+	// this should not happen because the memory is allocated considering both memOffset and data length and
+	// in the case of length==0 the fn returns immediately, not reaching this point
+	if uint64(len(c.memory)) < memOffset.Uint64()+uint64(len(data)) {
+		c.exit(errReturnDataOutOfBounds)
+
+		return
+	}
+
 	copy(c.memory[memOffset.Uint64():], data)
```
Stefan-Ethernal: Compared with go ethereum implementation, it seems like we should copy data limited only to this range:
```go
copy(m.store[offset:offset+size], value)
```

Which in our case means that this line would be turned into: 
```go
copy(c.memory[memOffset.Uint64():memOffset.Uint64()+length.Uint64()], data)
```

Though it seems pretty similar to what is currently implemented, not sure if we could rewrite some data by accident because we haven't specified the upper bound.

Stefan-Ethernal (commented):
Generally looks good and I don't find anything wrong with the PR itself. 

However, thinking out loud about the wider context and whether we have some additional things to be fixed/simplified.

Stefan-Ethernal (approved):

state/runtime/evm/instructions.go
```diff
@@ -813,17 +813,19 @@ func opReturnDataCopy(c *state) {
...
+		c.exit(errReturnDataOutOfBounds)
+
 		return
 	}
 
-	if !dataOffset.IsUint64() {
-		c.exit(errGasUintOverflow)
-
+	// if length is 0, return immediately since no need for the data copying nor memory allocation
+	if length.Sign() == 0 || !c.allocateMemory(memOffset, length) {
```
Stefan-Ethernal, line 823: Not to forget: maybe we should check is it safe to remove size check from `allocateMemory` function (https://github.com/0xPolygon/polygon-edge/blob/feature/v3-parity/state/runtime/evm/state.go#L306)? Would be good if it is safe to remove it from there.

Nemanja0x (approved):

igorcrevar (approved):

goran-ethernal (approved):

begmaroman (approved):

b00ris (approved):

ferranbt (approved):

PR #1181, opened 2023-01-31
EVM-375 Reject contract creation transactions with large init data

goran-ethernal:
# Description

This PR fixes the issue when we send a bunch of handcrafted transactions to node, that create some large smart contracts, which are too large and can cause a decrease in network performance, making the nodes stuck from time to time. Before, these transactions got rejected by the nodes, but only after they were executed.

A check is introduced in transaction pool as well, where if a transaction is a contract creation transaction, we will check the size of its input, and if it is larger than the `MaxCodeInitSize` we will not add the given transaction to the pool. 

_Additional info_: _`MaxInitCodeSize` is a limit in the `Ethereum` blockchain that sets a maximum size for the initial code that is executed when a smart contract is created. This is done to prevent the deployment of smart contracts that are too large, as this could lead to increased costs and decreased network performance. The current value of `MaxInitCodeSize` in `Ethereum` is 24576 bytes._

# Changes include

# Checklist

## Testing

### DIFF
--- a/state/executor.go
+++ b/state/executor.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	spuriousDragonMaxCodeSize = 24576
+	SpuriousDragonMaxCodeSize = 24576
 
 	TxGas                 uint64 = 21000 // Per transaction not creating a contract
 	TxGasContractCreation uint64 = 53000 // Per transaction that creates a contract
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal (approved):

Nemanja0x (approved):

begmaroman (approved):

vcastellm (approved):
LGTM 🎩

PR #1184, opened 2023-02-01
EVM-396 TOB-EDGE-29: BLS Rogue attack + align Polybft with smart contracts dev branch

igorcrevar:
# Description

Initial validators created in genesis block should have signature and that signature should be passed to `initialize` function of `ChildValidatorSet` smart contract. The signature is used as `Prove knowledge of the secret key (KOSK)`, because by providing this signature validator proves that his public key is good (not created specially for rogue attack) and corresponds to his private key.

New switch `chain-id` is added to the `manifest initialization` and `register validator` commands. It is defaulted to `100` in the same way as in the `genesis` command

While I was working on initial task one bug was noticed and fixed (otherwise tests will fail). Pr for that bug https://github.com/0xPolygon/polygon-edge/pull/1191

Smart contract team introduced two different domains for `bls` `hashToMap`, so changes needed to align client with that are also implemented here

# Changes include

# Checklist

## Testing


vcastellm:
Can you provide more context on what's a Rogue attach and the technique used in this PR to fix it?

igorcrevar:
> Can you provide more context on what's a Rogue attach and the technique used in this PR to fix it?

There is explanation here https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html . Also on https://www.notion.so/polygontechnology/RFC-175-b91e1f5b37de4e8b91b41f08a38491e1 (appendix section provides test for rogue attack)

The gist is that there is two ways to prevent it:
1. Change Aggregation of public keys/signatures. This will require changing code on both sc and client side
2. Implementing `Prove knowledge of the secret key (KOSK)`

We are using second approach. The actual verification is done on smart contract side but client must provide sc with all the data - KOSK signature. That special signature is created by signing encoded tuple `(validator address, chain id)`. The private key used for signing is validator private key. Verification is done by using validator public key. In that way we are sure that validator public key is actually good (not manually created) and corresponds to his private key.

b00ris (approved):

begmaroman (approved):

consensus/polybft/wallet/key.go
```diff
@@ -28,8 +29,14 @@ func (k *Key) Address() ethgo.Address {
 	return k.raw.Ecdsa.Address()
 }
 
-func (k *Key) Sign(b []byte) ([]byte, error) {
-	s, err := k.raw.Bls.Sign(b)
+// SignWithDomain signs message with bls and use bls.DomainCheckpointManager domain for mapping message to G1
+func (k *Key) Sign(hash []byte) ([]byte, error) {
```
ferranbt: I would either:
- Rename this function to represent that it is signing with a `bls.DomainCheckpointManager`.
- Remove the function and make sure we do not miss use it.

command/genesis/utils.go
```diff
@@ -126,7 +127,7 @@ func GetValidatorKeyFiles(rootDir, filePrefix string) ([]string, error) {
 }
 
 // ReadValidatorsByPrefix reads validators secrets on a given root directory and with given folder prefix
-func ReadValidatorsByPrefix(dir, prefix string) ([]*polybft.Validator, error) {
+func ReadValidatorsByPrefix(dir, prefix string, chainID int64) ([]*polybft.Validator, error) {
```
ferranbt: This function is doing more stuff than its name represents now. It not only reads the validators by prefix but also initializes them for the manifest with the `kosk`. I would either:
- Rename this function to represent that.
- Keep it as it was and do the `kosk` in another step after this function is called.

igorcrevar: Code would be much more complicated if KOSK signature is generated outside, because in that case we should introduce another structure which contains everything `polybft.Validator` + `private key` - `signature`.

Can we just do first thing you have suggested - rename function? Do you have some suggestions for the new name?

consensus/polybft/wallet/key.go
igorcrevar: The function is needed because of `Key` interface
```
type Key interface {
	Address() Address
	Sign(hash []byte) ([]byte, error)
}
```

ferranbt: Where is that function being used?

igorcrevar: I have realized that for Key implementation we only need signing with `bls.DomainCheckpointManager`.
So I have removed `func (k *Key) SignWithDomain`.  
  https://github.com/0xPolygon/polygon-edge/pull/1184/commits/de66a2c0e858b522f017b2550c91371d141c90f9

igorcrevar: Or maybe it would be even better if we have `Key` structure as
```
type Key struct {
	raw    *Account
	domain []byte
}

func NewKey(raw *Account, domain []byte) *Key {
	return &Key{
		raw:    raw,
		domain: domain,
	}
}

...

// Sign signs message with bls
func (k *Key) Sign(hash []byte) ([]byte, error) {
	s, err := k.raw.Bls.Sign(hash, k.domain)
	if err != nil {
		return nil, err
	}

	return s.Marshal()
}
```

igorcrevar: I did implementation like I suggested in comment above https://github.com/0xPolygon/polygon-edge/pull/1184/commits/d0f708cdae4bd47d1a4ae0e2ce37b86e53f6599a

command/genesis/utils.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/da5bf28e62a21f569b8c14919c21f083cecc2c12

consensus/polybft/extra.go
```diff
@@ -398,7 +398,7 @@ func (s *Signature) VerifyCommittedFields(validators AccountSet, hash types.Hash
 		return err
 	}
 
-	if !aggs.VerifyAggregated(blsPublicKeys, hash[:]) {
+	if !aggs.VerifyAggregated(blsPublicKeys, hash[:], bls.DomainCheckpointManager) {
```
goran-ethernal: Could we pass the domain to the `VerifyCommittedFields` function, so that the caller specifies it, and not hardcoding it like this? Plus the name `VerifyCommittedFields` is not exactly true, maybe rename it just to `Verify` or `VerifySignature`, or `VerifyAggregatedSignature`.

consensus/polybft/polybft_config.go
```diff
@@ -75,21 +77,40 @@ func (p *PolyBFTConfig) IsBridgeEnabled() bool {
...
+	Balance       *big.Int
+	NodeID        string
 }
 
 type validatorRaw struct {
-	Address types.Address `json:"address"`
-	BlsKey  string        `json:"blsKey"`
-	Balance *string       `json:"balance"`
-	NodeID  string        `json:"nodeId"`
+	Address      types.Address `json:"address"`
```
goran-ethernal, line 89: Why do we need this struct? Can't we just have json annotations on the `Validator` struct, and use it to marshal and unmarshal? We can ignore some of the fields if we don't need to marshal them.

consensus/polybft/polybft_config.go
```diff
@@ -75,21 +77,40 @@ func (p *PolyBFTConfig) IsBridgeEnabled() bool {
...
+	BlsKey       string        `json:"blsKey"`
+	BlsSignature string        `json:"blsSignature"`
+	Balance      *string       `json:"balance"`
+	NodeID       string        `json:"nodeId"`
+}
+
+func (v *Validator) MakeKoskSignature(chainID int64) error {
```
goran-ethernal: We already have a function with this name. Could we rename this to `InitKoskSignature`, since this function only changes state of the `Validator` object.

consensus/polybft/polybft_config.go
igorcrevar: This is done by @Stefan-Ethernal because of `big.Int` serialization problem. We can change that in some other PR if we found better solution

consensus/polybft/polybft_config.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/0683391e4b1b15ecfe0f2e80572d0444a795fbbe

consensus/polybft/extra.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/0683391e4b1b15ecfe0f2e80572d0444a795fbbe

goran-ethernal (approved):

server/server.go
```diff
@@ -462,7 +463,7 @@ func (s *Server) setupRelayer() error {
 		s.config.JSONRPC.JSONRPCAddr.String(),
 		ethgo.Address(contracts.StateReceiverContract),
 		s.logger.Named("relayer"),
-		wallet.NewEcdsaSigner(wallet.NewKey(account)),
+		wallet.NewEcdsaSigner(wallet.NewKey(account, bls.DomainCheckpointManager)),
```
Stefan-Ethernal, line 466: Is it correct using `DomainCheckpointManager`, since this transaction is being sent to `StateReceiver` SC, which is deployed on L2?

Generally speaking, when we are using `DomainCheckpointManager` and when `DomainValidatorSet`?

consensus/polybft/polybft_test.go
```diff
@@ -264,3 +266,29 @@ func Test_Factory(t *testing.T) {
 	assert.Equal(t, epochSize, polybft.consensusConfig.EpochSize)
 	assert.Equal(t, params, polybft.config)
 }
+
+func Test_MakeKoskSignature(t *testing.T) {
```
Stefan-Ethernal:
```suggestion
func Test_MakeKoskSignature(t *testing.T) {
    t.Parallel()
```

consensus/polybft/polybft_test.go
```diff
@@ -264,3 +266,29 @@ func Test_Factory(t *testing.T) {
...
+	expected := "127cfb8e2512b447056f33b91fca6cb2a7039e8b330edc4e5e5287f1c58bba5206373a97c9f09db144c8db5681c39e013ee6039ebbe36e0448e9f704f2d326c0"
+	bytes, _ := hex.DecodeString("3139343634393730313533353434353137333331343333303931343932303731313035313730303336303738373134363131303435323837383335373237343933383834303135343336383231")
+
+	pk, err := bls.UnmarshalPrivateKey(bytes)
+	require.NoError(t, err)
+
+	address := types.BytesToAddress((pk.PublicKey().Marshal())[:20])
```
Stefan-Ethernal: Is this suggestion correct?

```suggestion
	address := types.BytesToAddress((pk.PublicKey().Marshal())[:types.AddressLength])
```

Stefan-Ethernal (approved):
Generally looks good

server/server.go
igorcrevar: They are calling `(bool callSuccess, bytes memory returnData) = VALIDATOR_PKCHECK_PRECOMPILE.staticcall{` and there we use  `DomainCheckpointManager`

consensus/polybft/polybft_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/d282b8e0012ae417736023cb79f7f916ad08a38c

consensus/polybft/polybft_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/d282b8e0012ae417736023cb79f7f916ad08a38c

consensus/polybft/polybft_config.go
```diff
@@ -197,3 +255,17 @@ func (m *Manifest) Save(manifestPath string) error {
 
 	return nil
 }
+
+// MakeKoskSignature creates KOSK signature which prevents rogue attack
+func MakeKoskSignature(
```
vcastellm: nitpick: Should be `MakeKOSKSignature`

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1184/commits/96cf126ab95563ba348ca6ca1ffd515c1db49fc6

vcastellm (approved):

ferranbt (approved):

PR #1186, opened 2023-02-01
Fix  validation of state transactions

goran-ethernal:
# Description

This PR fixes the issue where a malicious node can include an `uptime` (`commitEpoch`) transaction in the middle of the epoch, causing the epoch to end abruptly.

Fix includes the check for `uptime` transaction no matter if it is not an epoch ending block.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -28,8 +28,9 @@ type blockBuilder interface {
 }
 
 var (
-	errUptimeTxDoesNotExist = errors.New("uptime transaction is not found in the epoch ending block")
-	errUptimeTxNotExpected  = errors.New("didn't expect uptime transaction in a non ending epoch block")
+	errUptimeTxDoesNotExist          = errors.New("uptime transaction is not found in the epoch ending block")
+	errUptimeTxNotExpected           = errors.New("didn't expect uptime transaction in a non epoch ending block")
+	errUptimeTxOnlyOneUptimeExpected = errors.New("only one uptime transaction is allowed in epoch ending block")
 )
 
 type fsm struct {
@@ -365,11 +366,11 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 		switch stateTxData := decodedStateTx.(type) {
 		case *CommitmentMessageSigned:
 			if !f.isEndOfSprint {
-				return fmt.Errorf("found commitment in block which should not contain it: tx = %v", tx.Hash)
+				return fmt.Errorf("found commitment tx in block which should not contain it: tx = %v", tx.Hash)
 			}
 
 			if commitmentMessageSignedExists {
-				return fmt.Errorf("only one commitment is allowed per block: %v", tx.Hash)
+				return fmt.Errorf("only one commitment tx is allowed per block: %v", tx.Hash)
 			}
 
 			commitmentMessageSignedExists = true
@@ -398,6 +399,13 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 				return fmt.Errorf("invalid signature for tx = %v", tx.Hash)
 			}
 		case *contractsapi.CommitEpochFunction:
+			if uptimeTransactionExists {
+				// if we already validated uptime transaction,
+				//that means someone added two or more uptime tx to block,
+				// which is not allowed
+				return errUptimeTxOnlyOneUptimeExpected
+			}
+
 			uptimeTransactionExists = true
 
 			if err := f.verifyValidatorsUptimeTx(tx); err != nil {
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -27,6 +27,10 @@ type blockBuilder interface {
 	Receipts() []*types.Receipt
 }
 
+var (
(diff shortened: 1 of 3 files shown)


consensus/polybft/fsm.go
```diff
@@ -517,13 +514,11 @@ func (f *fsm) verifyValidatorsUptimeTx(transactions []*types.Transaction) error
...
-	} else {
-		if blockUptimeTx != nil && blockUptimeTx.Hash == createdUptimeTx.Hash {
-			return errors.New("didn't expect uptime transaction in the middle of an epoch")
-		}
+
+		return nil
 	}
 
-	return nil
+	return errors.New("didn't expect uptime transaction in a non ending epoch block")
```
igorcrevar: This could go to `var` at the beginning where we do errors declaration

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1186/commits/3c3920420847ec1cb1b83e8ff589b424dee653d8

consensus/polybft/fsm.go
```diff
@@ -517,13 +517,11 @@ func (f *fsm) verifyValidatorsUptimeTx(transactions []*types.Transaction) error
 				createdUptimeTx.Hash,
 			)
 		}
-	} else {
-		if blockUptimeTx != nil && blockUptimeTx.Hash == createdUptimeTx.Hash {
-			return errors.New("didn't expect uptime transaction in the middle of an epoch")
```
vcastellm, line 522: This error message indicates that this was already checked, wasn't this working as expected?

vcastellm (approved):
LGTM aside from question

consensus/polybft/fsm.go
goran-ethernal: We didn't need this condition `blockUptimeTx != nil && blockUptimeTx.Hash == createdUptimeTx.Hash`. If we managed to decode state transaction, and that transaction was uptime, if it wasn't an end of epoch, we should always return this error.

consensus/polybft/fsm.go
```diff
@@ -27,6 +27,11 @@ type blockBuilder interface {
 	Receipts() []*types.Receipt
 }
 
+var (
+	errUptimeTxDoesNotExist = errors.New("uptime transaction is not found in the epoch ending block")
+	errUptimeTxNotExpected  = errors.New("didn't expect uptime transaction in a non ending epoch block")
```
```suggestion
errUptimeTxNotExpected  = errors.New("didn't expect uptime transaction in a non epoch ending block")
```

consensus/polybft/fsm.go
```diff
@@ -400,9 +397,21 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 			if !verified {
 				return fmt.Errorf("invalid signature for tx = %v", tx.Hash)
 			}
+		case *contractsapi.CommitEpochFunction:
+			uptimeTransactionExists = true
```
Stefan-Ethernal, line 409: Should we validate that uptime transaction is the first transaction in the block?

consensus/polybft/fsm.go
```diff
@@ -342,35 +347,27 @@ func (f *fsm) ValidateSender(msg *proto.Message) error {
...
 		decodedStateTx, err := decodeStateTransaction(tx.Input) // used to be Data
 		if err != nil {
-			return fmt.Errorf("state transaction error while decoding: tx = %v, err = %w", tx.Hash, err)
+			return fmt.Errorf("unknown state transaction: tx = %v, err = %w", tx.Hash, err)
 		}
 
 		switch stateTxData := decodedStateTx.(type) {
 		case *CommitmentMessageSigned:
+			if !f.isEndOfSprint {
+				return fmt.Errorf("found commitment in block which should not contain it: tx = %v", tx.Hash)
```
```suggestion
return fmt.Errorf("found commitment tx in block which should not contain it: tx = %v", tx.Hash)
```

consensus/polybft/fsm.go
goran-ethernal: I don't think we need that, since there is a discussion that `commitEpoch` is not necessarily the first transaction in block.
I would add a check that I figured out right now, is that there can only be one `commitEpoch` transaction. That is something we are missing.

consensus/polybft/fsm.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1186/commits/78df39a2c6c62985ced07d121da713eda5ec1245

consensus/polybft/fsm.go
goran-ethernal: Fix for two `commitEpoch` txs in block: https://github.com/0xPolygon/polygon-edge/pull/1186/commits/78df39a2c6c62985ced07d121da713eda5ec1245

consensus/polybft/fsm.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1186/commits/78df39a2c6c62985ced07d121da713eda5ec1245

igorcrevar (approved):

Nemanja0x (approved):

Stefan-Ethernal (approved):
🥇

PR #1191, opened 2023-02-02
EVM-453 Build proposal header's extra validators bug

igorcrevar:
# Description

Build proposal should calculate next validators hash and populate it to header's extra data, only after all other transactions have been applied to the block.

There is a problem if block contains stake transaction (which changes stake of some of existing validators and essentially change validator set).
In that case, validator stake change will not take effect on proposer node itself, but it is going to take effect on validators (since they would calculate state differently, by taking into account stake transaction as well)

# Changes include

# Checklist

## Testing


Stefan-Ethernal (approved):

ferranbt:
Can we have a test for this?

goran-ethernal (approved):

vcastellm (approved):
LGTM

b00ris (approved):

begmaroman (approved):

Nemanja0x (approved):

PR #1192, opened 2023-02-02
Introduce burn contract address 

begmaroman:
# Description

Add the burn contract addresses map

# Changes include

# Checklist

## Testing

### DIFF
--- a/chain/params.go
+++ b/chain/params.go
@@ -10,8 +10,8 @@ import (
 )
 
 var (
-	// ErrBurntContractAddressMissing is the error when a contract address is not provided
-	ErrBurntContractAddressMissing = errors.New("burnt contract address missing")
+	// ErrBurnContractAddressMissing is the error when a contract address is not provided
+	ErrBurnContractAddressMissing = errors.New("burn contract address missing")
 )
 
 // Params are all the set of params for the chain
@@ -22,19 +22,19 @@ type Params struct {
 	Whitelists     *Whitelists            `json:"whitelists,omitempty"`
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
 
-	// Governance contract where the token will be sent to and burnt in london fork
-	BurntContract map[string]string `json:"burntContract"`
+	// Governance contract where the token will be sent to and burn in london fork
+	BurnContract map[string]string `json:"burnContract"`
 }
 
-// CalculateBurntContract calculates burn contract address for the given block number
-func (p *Params) CalculateBurntContract(block uint64) (types.Address, error) {
-	keys := make([]string, 0, len(p.BurntContract))
-	for k := range p.BurntContract {
+// CalculateBurnContract calculates burn contract address for the given block number
+func (p *Params) CalculateBurnContract(block uint64) (types.Address, error) {
+	keys := make([]string, 0, len(p.BurnContract))
+	for k := range p.BurnContract {
 		keys = append(keys, k)
 	}
 
 	if len(keys) == 0 {
-		return types.ZeroAddress, ErrBurntContractAddressMissing
+		return types.ZeroAddress, ErrBurnContractAddressMissing
 	}
 
 	sort.Strings(keys)
@@ -44,11 +44,11 @@ func (p *Params) CalculateBurntContract(block uint64) (types.Address, error) {
 		valUintNext, _ := strconv.ParseUint(keys[i+1], 10, 64)
 
 		if block > valUint && block < valUintNext {
-			return types.StringToAddress(p.BurntContract[keys[i]]), nil
+			return types.StringToAddress(p.BurnContract[keys[i]]), nil
 		}
 	}
 
-	return types.StringToAddress(p.BurntContract[keys[len(keys)-1]]), nil
+	return types.StringToAddress(p.BurnContract[keys[len(keys)-1]]), nil
 }
 
 func (p *Params) GetEngine() string {
--- a/chain/params.go
+++ b/chain/params.go
@@ -1,9 +1,17 @@
 package chain
 
 import (
+	"errors"
 	"math/big"
+	"sort"
+	"strconv"
 
 	"github.com/0xPolygon/polygon-edge/types"
+)
+
+var (
+	// ErrBurntContractAddressMissing is the error when a contract address is not provided
+	ErrBurntContractAddressMissing = errors.New("burnt contract address missing")
 )
 
 // Params are all the set of params for the chain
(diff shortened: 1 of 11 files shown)


begmaroman:
Tests are failing because they don't support burn contracts yet. It will be updated in one of the next PRs.

chain/params.go
```diff
@@ -1,18 +1,54 @@
...
 	Engine         map[string]interface{} `json:"engine"`
 	Whitelists     *Whitelists            `json:"whitelists,omitempty"`
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
+
+	// Governance contract where the token will be sent to and burnt in london fork
+	BurntContract map[string]string `json:"burntContract"`
+}
+
+// CalculateBurntContract calculates burn contract address for the given block number
+func (p *Params) CalculateBurntContract(number uint64) (types.Address, error) {
```
vcastellm: If this is block number, shouldn't it be `block` or `height`?

vcastellm (commented):
Is it `burn contract` or `burnt contract`? Aside from that, would be great to include some tests for the new functions.

chain/params.go
begmaroman: Renamed

begmaroman:
@vcastellm fixed the name and added tests

chain/params.go
```diff
@@ -1,18 +1,54 @@
...
 // Params are all the set of params for the chain
 type Params struct {
 	Forks          *Forks                 `json:"forks"`
 	ChainID        int                    `json:"chainID"`
 	Engine         map[string]interface{} `json:"engine"`
 	Whitelists     *Whitelists            `json:"whitelists,omitempty"`
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
+
+	// Governance contract where the token will be sent to and burnt in london fork
+	BurntContract map[string]string `json:"burntContract"`
```
Stefan-Ethernal: As @vcastellm mentioned in his review, should this be renamed to `BurnContract`, instead of `BurntContract`? Here and all the other places across the codebase (refer to it as burn, instead of burnt, because it is more semantically correct). Agree?

chain/params_test.go
```diff
@@ -54,3 +56,59 @@ func TestParamsForksInTime(t *testing.T) {
 	expect("constantinople", ff.Constantinople, false)
 	expect("eip150", ff.EIP150, false)
 }
+
+func TestParams_CalculateBurntContract(t *testing.T) {
+	tests := []struct {
```
Stefan-Ethernal, line 65: Introduce parallel execution?
```suggestion
        t.Parallel()
	tests := []struct {
```

chain/params_test.go
```diff
@@ -54,3 +56,59 @@ func TestParamsForksInTime(t *testing.T) {
...
+				"15": "0x8888f1f195afa192cfee860698584c030f4c9db1",
+			},
+			block:   10,
+			want:    types.StringToAddress("0x8888f1f195afa192cfee860698584c030f4c9db2"),
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
Stefan-Ethernal: Parallel execution?

```suggestion
	for _, tt := range tests {
	        tt := tt
		t.Run(tt.name, func(t *testing.T) {
		    t.Parallel()
```

chain/params.go
begmaroman: Right, fixed

chain/params_test.go
begmaroman: Done

chain/params_test.go
begmaroman: Done

chain/params.go
```diff
@@ -1,18 +1,54 @@
...
+		return types.ZeroAddress, ErrBurnContractAddressMissing
+	}
+
+	sort.Strings(keys)
+
+	for i := 0; i < len(keys)-1; i++ {
+		valUint, _ := strconv.ParseUint(keys[i], 10, 64)
+		valUintNext, _ := strconv.ParseUint(keys[i+1], 10, 64)
+
+		if block > valUint && block < valUintNext {
```
igorcrevar: Just a question:
Shouldn't the condition be inclusive on one side?
If, for example,  keys are `[1, 10, 20, 30]` and block number is 10, than the burn contract for 30 will be returned
It seams to me that first part of condition should be `block >= valUint`

begmaroman: Good point. There was a bug because there was a string sorting but not int which is wrong. Fixed.

vcastellm (approved):
Are we going to provide an example burn contract?

goran-ethernal (approved):

begmaroman:
> Are we going to provide an example burn contract?

@vcastellm smart contracts team is working on it

command/genesis/polybft_params.go
```diff
@@ -122,6 +123,15 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 		}
 	}
 
+	for _, premine := range p.burnContracts {
+		block, addr, err := parseBurnContractInfo(premine)
```
Stefan-Ethernal:
```suggestion
	for _, burnContractCfg := range p.burnContracts {
		block, addr, err := parseBurnContractInfo(burnContractCfg)
```

chain/params_test.go
```diff
@@ -54,3 +56,76 @@ func TestParamsForksInTime(t *testing.T) {
...
+			got, err := p.CalculateBurnContract(tt.block)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("CalculateBurnContract() error = %v, wantErr %v", err, tt.wantErr)
+
+				return
+			}
+
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("CalculateBurnContract() got = %v, want %v", got, tt.want)
+			}
```
Stefan-Ethernal: WDYT?
```suggestion
                        got, err := p.CalculateBurnContract(tt.block)
			if tt.wantErr {
				require.Error(t, err, "CalculateBurnContract() error = %v, wantErr %v", err, tt.wantErr)
			} else {
				require.NoError(t, err)
				require.Equal(t, tt.want, got, "CalculateBurnContract() got = %v, want %v", got, tt.want)
			}
```

Stefan-Ethernal (approved):
LGTM, apart from minor suggestions

chain/params_test.go
begmaroman: Done

command/genesis/polybft_params.go
begmaroman: Done

igorcrevar (approved):

PR #1195, opened 2023-02-03
EVM-387(TOB-EDGE-11) & EVM-409 (TOB-EDGE-10) Fix ValidateSignatureValues

Kourin1996:
# Description

This PR fixes `ValidateSignatureValues` function that might pass wrong signature

TOB-EDGE-10: `ValidateSignatureValues` used absolute values of `r` and `s` and accepted negative `r` and `s` as inputs but `ValidateSignatureValues` shouldn't accept negative `r` and `s`.

TOB-EDGE-11: `bytes.Compare` compares two byte slices lexicographically, so it could pass some value which is greater than upper limit (secp256k1N).

TOB-EDGE-8: the `ValidateSignatureValues` function lacks one of the three criteria that the Ethereum Yellow Paper
defines to prevent ECDSA signature malleability attacks. 

That is `s` value must be in this range of valid values
0 < 𝑠 < 𝑠𝑒𝑐𝑝256𝑘1𝑛 ÷ 2 + 1

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


crypto/crypto.go
```diff
@@ -66,31 +62,29 @@ func trimLeftZeros(b []byte) []byte {
...
+	// v must be 0 or 1
+	if v > 1 {
 		return false
 	}
 
-	ss := s.Bytes()
-	ss = trimLeftZeros(ss)
-
-	if bytes.Compare(ss, secp256k1N) >= 0 || bytes.Compare(ss, one) < 0 {
+	// From Homestead, s must be less than secp256k1n/2
```
Stefan-Ethernal: Please double check me, but if we negate invalid condition for `s` validation, we are getting `s <= secp256k1n/2`?

```suggestion
        // From Homestead, s must be less or equal than secp256k1n/2
```

Stefan-Ethernal (approved):
Nice. 🚀 

Just to confirm, this addresses issues 8, 10 and 11 from the report? If so, just please amend issue 8 to the PR description as well.

crypto/crypto.go
```diff
@@ -66,31 +62,29 @@ func trimLeftZeros(b []byte) []byte {
...
+	if isHomestead && s.Cmp(secp256k1NHalf) > 0 {
```
igorcrevar: Maybe for Homestead it would be better if code looks like:
```
if isHomestead {
    // In Homestead, r must be less than secp256k1n and s must be less or equal than secp256k1NHalf
    return s.Cmp(secp256k1NHalf) <= 0 && r.Cmp(secp256k1N) < 0
}
```

Because if that check is false, we gonna check `r.Cmp(secp256k1N) < 0 && s.Cmp(secp256k1N) < 0` and that is unnecessary (if  `s.Cmp(secp256k1NHalf) < 0` is true then `s.Cmp(secp256k1N) < 0` is also true)

trimixlover:
@Stefan-Ethernal Yes it solves TOB-8 as well.

trimixlover (approved):

Stefan-Ethernal:
> @Stefan-Ethernal Yes it solves TOB-8 as well.

Great, then just amend that information to PR description if possible :slightly_smiling_face:

Kourin1996:
@Stefan-Ethernal This PR is created by the intension of fixing TOB-10 and TOB-11. But I'm sure this PR also deals with TOB-8. I checked both of Yellow Paper and Geth. Yellow paper always requires `s` to be less than half of `𝑠𝑒𝑐𝑝256𝑘1𝑛` but Geth enables such a rule from Homestead fork.

crypto/crypto.go
Kourin1996: Fixed 5c0da0216cb78fe695c7019347e5be3594afb623 @igorcrevar

crypto/crypto.go
Kourin1996: Fixed 5c0da0216cb78fe695c7019347e5be3594afb623 @Stefan-Ethernal

igorcrevar (approved):

yaziciahmet (approved):

PR #1204, opened 2023-02-06
Update state transition logic according to EIP-1559

begmaroman:
# Description

Update state transition logic in order to burn base fee and transfer the rest to the miner.

# Changes include

# Checklist

## Testing

### DIFF
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -41,6 +41,24 @@ func Max(a, b uint64) uint64 {
 	return b
 }
 
+// BigMax returns the largest of x or y.
+func BigMax(x, y *big.Int) *big.Int {
+	if x.Cmp(y) < 0 {
+		return y
+	}
+
+	return x
+}
+
+// BigMin returns the smallest of x or y.
+func BigMin(x, y *big.Int) *big.Int {
+	if x.Cmp(y) > 0 {
+		return y
+	}
+
+	return x
+}
+
 func ConvertUnmarshalledInt(x interface{}) (int64, error) {
 	switch tx := x.(type) {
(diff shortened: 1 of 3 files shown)


helper/common/common.go
```diff
@@ -61,12 +79,6 @@ func roundFloat(num float64) int64 {
 	return int64(num + math.Copysign(0.5, num))
 }
 
-func ToFixedFloat(num float64, precision int) float64 {
-	output := math.Pow(10, float64(precision))
-
-	return float64(roundFloat(num*output)) / output
-}
-
```
begmaroman, line 69: Not used

state/executor.go
```diff
@@ -510,10 +531,25 @@ func (t *Transition) apply(msg *types.Transaction) (*runtime.ExecutionResult, er
 	remaining := new(big.Int).Mul(new(big.Int).SetUint64(result.GasLeft), msg.GasPrice)
 	t.state.AddBalance(msg.From, remaining)
 
-	// pay the coinbase
-	coinbaseFee := new(big.Int).Mul(new(big.Int).SetUint64(result.GasUsed), msg.GasPrice)
+	// define effective tip based on tx type
+	effectiveTip := msg.GasPrice
```
igorcrevar, line 538: maybe it would be good to explain calculations of miner fee and burn amount with comment and link(s)?

begmaroman: The calculation is not complex at all: we just use the gas fee cap and gas tip cap depending on the london hard fork enablement. Added more comments.

igorcrevar (approved):

PR #1205, opened 2023-02-06
Message is not hashed before signature verification

Stefan-Ethernal:
# Description

IBFT messages were not hashed prior to signing it and we were lacking hashing in opposite direction as well (namely when recovering public key from the signature and checking the message sender identity).
Before signing a message it is important to hash it because only the first 32 bytes are being verified and some tampered messages, larger than 32 bytes, could be recognized as valid.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/wallet/key.go
+++ b/consensus/polybft/wallet/key.go
@@ -44,7 +44,7 @@ func (k *Key) SignEcdsaMessage(msg *proto.Message) (*proto.Message, error) {
 		return nil, fmt.Errorf("cannot marshal message: %w", err)
 	}
 
-	if msg.Signature, err = k.raw.Ecdsa.Sign(raw); err != nil {
+	if msg.Signature, err = k.raw.Ecdsa.Sign(crypto.Keccak256(raw)); err != nil {
 		return nil, fmt.Errorf("cannot create message signature: %w", err)
 	}
 
@@ -52,8 +52,8 @@ func (k *Key) SignEcdsaMessage(msg *proto.Message) (*proto.Message, error) {
 }
 
 // RecoverAddressFromSignature recovers signer address from the given digest and signature
-func RecoverAddressFromSignature(sig, msg []byte) (types.Address, error) {
-	pub, err := crypto.RecoverPubkey(sig, msg)
+func RecoverAddressFromSignature(sig, digest []byte) (types.Address, error) {
+	pub, err := crypto.RecoverPubkey(sig, digest)
 	if err != nil {
 		return types.Address{}, fmt.Errorf("cannot recover address from signature: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


goran-ethernal (approved):
LGTM :100:

consensus/polybft/wallet/key.go
```diff
@@ -20,40 +20,43 @@ func NewKey(raw *Account) *Key {
...
 	}
 
 	return msg, nil
 }
 
 // RecoverAddressFromSignature recovers signer address from the given digest and signature
-func RecoverAddressFromSignature(sig, msg []byte) (types.Address, error) {
-	pub, err := crypto.RecoverPubkey(sig, msg)
+func RecoverAddressFromSignature(sig, digest []byte) (types.Address, error) {
+	pub, err := crypto.RecoverPubkey(sig, digest)
```
goran-ethernal: Maybe consider hashing the message here, like we do with sign.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1205/commits/7cbd83dda9ce34d0fb3fe12facd9925c821ef12d

consensus/polybft/wallet/key.go
```diff
@@ -20,40 +20,44 @@ func NewKey(raw *Account) *Key {
...
-	return s.Marshal()
+	return signature.Marshal()
 }
 
-// SignEcdsaMessage signs the proto message with ecdsa
-func (k *Key) SignEcdsaMessage(msg *proto.Message) (*proto.Message, error) {
-	raw, err := protobuf.Marshal(msg)
+// SignIBFTMessage signs the IBFT consensus message with ECDSA key
+func (k *Key) SignIBFTMessage(msg *proto.Message) (*proto.Message, error) {
+	msgRaw, err := protobuf.Marshal(msg)
```
ferranbt, line 45: So the `Message` contains the `Signature` too? I know that technically it should not be a problem since `Signature` should be empty. However, this is traditionally done with two structures. A `Container` and a `SignedContainer`:
```
type Container struct { // Message
   Field1 uint64
   Field2 uint64
}

type SignedContainer struct {
    Container []byte
    Signature [32]byte
}
```
It is also easier to know this way the lifecycle of the objects (i.e. you get a `SignedContainer` you have to decode it, from that point onwards you can use the `Container`).

Stefan-Ethernal: Ok, agree. However, it is not strictly related to this PR, since I believe `go-ibft` needs to be changed.

In that case, let's not block delivering of this PR until we change it on `go-ibft` and adapt Edge accordingly.

ferranbt: Yup, also, it would mean breaking backward compatibility.

vcastellm (approved):
LGTM

begmaroman (approved):

b00ris (approved):

PR #1207, opened 2023-02-07
`Signer.Sender` does not validate whole EDCSA signature parameters

goran-ethernal:
# Description

This PR fixes the issue where Edge is verifying only a part of ECDSA signature, specifically the `V` part, because it takes its `big.Int` value and downcast it to byte without checking the `V` value range. This means that the downcasting eliminates any data that was stored in the high bits of the `big.Int` without raising an error.

This poses an issue for transaction hash calculation; signature verification is being performed on signature parameters where the big.Int value has been downcasted, but Edge calculates transaction hashes using transaction signature parameters where the `big.Int` has not been downcasted.

In EIP-155 specification, `V` can be larger than 8 bits, because it is calculated as follows: `{0,1} + CHAIN_ID * 2 + 35`, so we will not downcast it, until we successfully validate it.

### ECDSA

ECDSA (Elliptic Curve Digital Signature Algorithm) is a digital signature algorithm that uses elliptic curve cryptography.
It uses the private key  to generate a digital signature for the message hash. This is done by applying a mathematical operation to the message hash and the private key to produce two numbers, `R` and `S`. These numbers together form the digital signature.
ECDSA provides a way to recover the public key of the signer from provided signature and message hash, and that process requires additional information to be provided - `V` value, which represents a recovery ID or recovery parameter. It is usally a single byte, that tells the algorithm about the sign of `y` coordinate of the public key on the elliptic curve.
Alongside the `R` and `S` values, it enables the public key recovery of the message signer.

# Changes include

# Checklist

## Testing

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -51,7 +51,7 @@ type KeccakState interface {
 }
 
 // ValidateSignatureValues checks if the signature values are correct
-func ValidateSignatureValues(v byte, r, s *big.Int, isHomestead bool) bool {
+func ValidateSignatureValues(v, r, s *big.Int, isHomestead bool) bool {
 	// r & s must not be nil
 	if r == nil || s == nil {
 		return false
@@ -63,7 +63,7 @@ func ValidateSignatureValues(v byte, r, s *big.Int, isHomestead bool) bool {
 	}
 
 	// v must be 0 or 1
-	if v > 1 {
+	if v.Cmp(big.NewInt(1)) == 1 || v.Cmp(big.NewInt(0)) == -1 {
 		return false
 	}
 
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -26,6 +26,7 @@ var S256 = btcec.S256()
 var (
 	secp256k1N, _  = new(big.Int).SetString("fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141", 16)
 	secp256k1NHalf = new(big.Int).Div(secp256k1N, big.NewInt(2))
+	zero           = big.NewInt(0)
 	one            = big.NewInt(1)
 
 	ErrInvalidBLSSignature = errors.New("invalid BLS Signature")
@@ -63,7 +64,7 @@ func ValidateSignatureValues(v, r, s *big.Int, isHomestead bool) bool {
 	}
 
 	// v must be 0 or 1
-	if v.Cmp(big.NewInt(1)) == 1 || v.Cmp(big.NewInt(0)) == -1 {
+	if v.Cmp(zero) == -1 || v.Cmp(one) == 1 {
 		return false
 	}
 
--- a/crypto/crypto_test.go
+++ b/crypto/crypto_test.go
@@ -153,7 +153,7 @@ func TestValidateSignatureValues(t *testing.T) {
 	cases := []struct {
 		homestead bool
 		name      string
-		v         byte
+		v         *big.Int
 		r         *big.Int
 		s         *big.Int
 		res       bool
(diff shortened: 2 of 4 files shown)


crypto/crypto.go
```diff
@@ -66,13 +66,13 @@ func trimLeftZeros(b []byte) []byte {
...
 // ValidateSignatureValues checks if the signature values are correct
-func ValidateSignatureValues(v byte, r, s *big.Int) bool {
+func ValidateSignatureValues(v, r, s *big.Int) bool {
 	// TODO: ECDSA malleability
 	if r == nil || s == nil {
 		return false
 	}
 
-	if v > 1 {
+	if v.Cmp(big.NewInt(1)) == 1 || v.Cmp(big.NewInt(0)) == -1 {
```
Stefan-Ethernal: You can rely on already present variable `big1`. Note that you would need to introduce `big0` variable, the same way we have `big1` (and also this would clash a bit with #1195, but hopefully it isn't a major problem)
Also makes more sense to first check lower and then upper bound :)

```suggestion
	if v.Cmp(big0) == -1 || v.Cmp(big1) == 1 {
```

crypto/txsigner.go
```diff
@@ -228,15 +228,15 @@ func (e *EIP155Signer) CalculateV(parity byte) []byte {
...
+func encodeSignature(R, S, V *big.Int) ([]byte, error) {
 	if !ValidateSignatureValues(V, R, S) {
 		return nil, fmt.Errorf("invalid txn signature")
 	}
 
 	sig := make([]byte, 65)
 	copy(sig[32-len(R.Bytes()):32], R.Bytes())
 	copy(sig[64-len(S.Bytes()):64], S.Bytes())
-	sig[64] = V
+	sig[64] = byte(V.Int64()) // here is safe to convert it since ValidateSignatureValues will validate the v value
```
Stefan-Ethernal, line 241: Not sure myself about this proposal, but think it is more consistent and offering an alternative. However you wish.

```suggestion
sig[64] = V.Bytes()[0] // here is safe to convert it since ValidateSignatureValues will validate the v value
```

state/runtime/precompiled/base.go
```diff
@@ -35,7 +35,7 @@ func (e *ecrecover) run(input []byte, caller types.Address, _ runtime.Host) ([]b
 	r := big.NewInt(0).SetBytes(input[64:96])
 	s := big.NewInt(0).SetBytes(input[96:128])
 
-	if !crypto.ValidateSignatureValues(v, r, s) {
+	if !crypto.ValidateSignatureValues(big.NewInt(int64(v)), r, s) {
```
Stefan-Ethernal:
```suggestion
if !crypto.ValidateSignatureValues(new(big.Int).SetBytes([]byte{v}), r, s) {
```

crypto/crypto_test.go
```diff
@@ -116,44 +116,46 @@ func TestCreate2(t *testing.T) {
 }
 
 func TestValidateSignatureValues(t *testing.T) {
```
Stefan-Ethernal:
```suggestion
func TestValidateSignatureValues(t *testing.T) {
    t.Parallel()
```

Stefan-Ethernal (approved):

b00ris (approved):

begmaroman (approved):

crypto/crypto_test.go
ferranbt: Is there any other extra test to check the downcasting?

goran-ethernal: No, I'll try to add some new tests.

crypto/txsigner.go
goran-ethernal: I would leave as it is for now, since it is safe to convert it, because of the validation we perform before this.

crypto/crypto.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1207/commits/18fc03393b8272b15d3b90ae0dc9eb6a8bee607c

crypto/crypto_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1207/commits/18fc03393b8272b15d3b90ae0dc9eb6a8bee607c

vcastellm (approved):
LGTM

crypto/crypto_test.go
goran-ethernal: Whatever test I write, it is very similar to this, so I would say, this test should be enough.

PR #1208, opened 2023-02-08
Encapsulate signatures validation logic in `Extra`

Stefan-Ethernal:
# Description

This PR encapsulates signatures validation logic in a single place. Prior to this, we were having almost the same logic for `ParentSignature` validation spread to `fsm` and `polybft`. This way `Extra` is more encapsulated and it contains validation logic of its own state.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/extra.go
+++ b/consensus/polybft/extra.go
@@ -148,7 +148,7 @@ func (i *Extra) ValidateFinalizedHeader(header *types.Header, parent *types.Head
 	// validate committed signatures
 	blockNumber := header.Number
 	if i.Committed == nil {
-		return fmt.Errorf("failed to verify signatures for block %d because signatures are not present", blockNumber)
+		return fmt.Errorf("failed to verify signatures for block %d, because signatures are not present", blockNumber)
 	}
 
 	if i.Checkpoint == nil {
@@ -172,7 +172,7 @@ func (i *Extra) ValidateFinalizedHeader(header *types.Header, parent *types.Head
 
 	parentExtra, err := GetIbftExtra(parent.ExtraData)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to verify signatures for block %d: %w", blockNumber, err)
 	}
 
 	// validate parent signatures
--- a/consensus/polybft/extra.go
+++ b/consensus/polybft/extra.go
@@ -142,17 +142,13 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
 	return nil
 }
 
-// ValidateBasic contains extra data basic set of validations
-func (i *Extra) ValidateBasic(header *types.Header, parent *types.Header, parents []*types.Header,
+// ValidateFinalizedHeader contains extra data validations for finalized headers
+func (i *Extra) ValidateFinalizedHeader(header *types.Header, parent *types.Header, parents []*types.Header,
 	chainID uint64, consensusBackend polybftBackend, logger hclog.Logger) error {
+	// validate committed signatures
 	blockNumber := header.Number
 	if i.Committed == nil {
 		return fmt.Errorf("failed to verify signatures for block %d because signatures are not present", blockNumber)
-	}
-
-	parentExtra, err := GetIbftExtra(parent.ExtraData)
-	if err != nil {
-		return err
 	}
 
 	checkpointHash, err := i.Checkpoint.Hash(chainID, header.Number, header.Hash)
@@ -170,7 +166,12 @@ func (i *Extra) ValidateBasic(header *types.Header, parent *types.Header, parent
 			blockNumber, checkpointHash, err)
 	}
 
-	// validate the signatures for parent (skip block 1 because genesis does not have committed)
+	parentExtra, err := GetIbftExtra(parent.ExtraData)
+	if err != nil {
+		return err
+	}
+
+	// validate parent signatures
 	if err := i.ValidateParentSignatures(blockNumber, consensusBackend, parents,
 		parent, parentExtra, chainID, logger); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


vcastellm (approved):
LGTM

consensus/polybft/extra.go
```diff
@@ -142,14 +142,77 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
...
+// ValidateFinalizedHeader contains extra data validations for finalized headers
+func (i *Extra) ValidateFinalizedHeader(header *types.Header, parent *types.Header, parents []*types.Header,
+	chainID uint64, consensusBackend polybftBackend, logger hclog.Logger) error {
+	// validate committed signatures
+	blockNumber := header.Number
+	if i.Committed == nil {
+		return fmt.Errorf("failed to verify signatures for block %d because signatures are not present", blockNumber)
+	}
+
+	checkpointHash, err := i.Checkpoint.Hash(chainID, header.Number, header.Hash)
```
goran-ethernal, line 159: Since you are checking if checkpoint is `nil` in some functions, I would say, check it here as well.

consensus/polybft/fsm.go
```diff
@@ -262,6 +262,11 @@ func (f *fsm) Validate(proposal []byte) error {
 		return err
 	}
 
+	if err := extra.ValidateParentSignatures(block.Number(), f.polybftBackend, nil, f.parent, parentExtra,
```
goran-ethernal, line 273: I this `ValidateParentSignatures` you are using the `Checkpoint` before validating it. Then, in `validateExtraData` which is called after this function, you are validating checkpoint data based on parent extra. I would say, move this nil checks for parent and current extra Checkpoint to that `ValidateParentSignatures` and remove it from the `validateExtraData`.

goran-ethernal:
I am not sure, did we have unit tests for these functions? If we did not, it would be nice to add them, since these are now pretty isolated functions.

consensus/polybft/fsm.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1208/commits/e1b8a65f095fb8167da7c8befe45153449b3b997.

However, I didn't put it to `ValidateParentSignatures`, as it is not using `Checkpoint` instance.
Btw I had doubts about whether to remove entirely check about `CheckpointData` being `nil` since we don't expect such use case ever to happen.

consensus/polybft/extra.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1208/commits/e1b8a65f095fb8167da7c8befe45153449b3b997

Stefan-Ethernal:
> I am not sure, did we have unit tests for these functions? If we did not, it would be nice to add them, since these are now pretty isolated functions.

They were covered as part of `fsm` unit test, however not all unhappy paths are covered. Let me see if I can implement some more unit tests.

goran-ethernal (approved):

begmaroman (approved):

consensus/polybft/extra.go
```diff
@@ -142,14 +142,81 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
...
 }
 
-// Validate contains extra data validation logic
-func (i *Extra) Validate(parentExtra *Extra, currentValidators AccountSet, nextValidators AccountSet) error {
-	return i.Checkpoint.Validate(parentExtra.Checkpoint, currentValidators, nextValidators)
+// ValidateParentSignatures validates signatures for parent block
+func (i *Extra) ValidateParentSignatures(blockNumber uint64, consensusBackend polybftBackend, parents []*types.Header,
+	parent *types.Header, parentExtra *Extra, chainID uint64, logger hclog.Logger) error {
+	// skip block 1 because genesis does not have committed signatures
+	if blockNumber <= 1 {
```
b00ris, line 192: Does 1 genesis have signatures?

consensus/polybft/extra.go
```diff
@@ -142,14 +142,81 @@ func (i *Extra) UnmarshalRLPWith(v *fastrlp.Value) error {
...
+		return nil
+	}
+
+	if i.Parent == nil {
+		return fmt.Errorf("failed to verify signatures for parent of block %d because signatures are not present",
+			blockNumber)
+	}
+
+	parentValidators, err := consensusBackend.GetValidators(blockNumber-2, parents)
```
b00ris, line 201: Why do you want to check validators for  blockNumber - 2?

b00ris (approved):

consensus/polybft/extra.go
Stefan-Ethernal: No, the genesis block contains no signatures. Since we are validating finalized/parent signatures (namely validating signatures for block N-1 when validating block N), we skip genesis and block#1 validation. We are going to validate signatures for block#1 when validating block#2.

consensus/polybft/extra.go
Stefan-Ethernal: ## Long story short...
because we want to skip epoch-ending blocks, in order to retrieve the correct validator set.

## Example
Epoch size=10 blocks
We are currently validating block#11, which means that we need to check its parent signatures (signatures of validators that have sealed block#10). Since block#10 is an epoch-ending block, it contains the `commitEpoch` transaction (i.e. validator set transition). If we were about to retrieve the validator set on block#10, we would get a validator set for the next epoch. However, we don't expect validators for the next epoch, but the ones which sealed the epoch-ending block itself (which are the validators saved for block#9 or any other block within the epoch). So that's why we are substracting by 2 here, to make sure that we have skipped an epoch-ending block and query validator set from the block within the (previous) epoch.

PR #1210, opened 2023-02-09
Added a function to calculate gas price if not provided

begmaroman:
# Description

Calculate and return gas price of the transaction if the value is not provided.

TESTS COULD BE FAILED. WILL BE FIXED WITHIN NEXT PRs.

# Changes include

# Checklist

## Testing


consensus/polybft/blockchain_wrapper.go
```diff
@@ -152,20 +152,33 @@ func (p *blockchainWrapper) GetHeaderByHash(hash types.Hash) (*types.Header, boo
...
 	}
 
+	// Calculate next base fee based on the parent head
+	baseFee := p.blockchain.CalculateBaseFee(parent)
+
+	// Update base fee in tx pool so txs could be sorted properly.
+	// This is the right place to update a base fee because this function
+	// gets executed when creating a new block, i.e. in the beginning of
+	// a new block so there are no transactions yet.
+	txPool.SetBaseFee(baseFee)
```
igorcrevar: I would move `baseFee := p.blockchain.CalculateBaseFee(parent)` + `txPool.SetBaseFee(baseFee)` to `func (c *consensusRuntime) OnBlockInserted` right bellow `c.config.txPool.ResetWithHeaders(fullBlock.Block.Header)`. I realize that current solution will work but maybe its good to have everything in one place and to  update `baseFee` for all validators/nodes and not just proposer

begmaroman: It went over the code and it seems like it would require some structural changes there. This is a bit out of the scope of this PR so I would preffer to address this comment within the next PR. Anyway, this is the good point.

types/transaction.go
```diff
@@ -20,7 +20,7 @@ type TxType byte
 const (
 	LegacyTx     TxType = 0x0
 	StateTx      TxType = 0x7f
-	DynamicFeeTx TxType = 0x8f
+	DynamicFeeTx TxType = 0x02
```
begmaroman: As per ethereum spec, the dynamic gas fee Tx has "2" as a type identifier.

types/rlp_unmarshal.go
```diff
@@ -396,8 +396,10 @@ func (t *Transaction) unmarshalRLPFrom(p *fastrlp.Parser, v *fastrlp.Value) erro
 	switch t.Type {
 	case LegacyTx:
 		num = 9
-	case DynamicFeeTx, StateTx:
+	case StateTx:
 		num = 10
+	case DynamicFeeTx:
+		num = 12
```
begmaroman: Because it has chain ID and access list fields. We don't use them but should be compatible.

command/genesis/polybft_params.go
```diff
@@ -174,11 +174,11 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 	chainConfig.Genesis = &chain.Genesis{
 		GasLimit:   p.blockGasLimit,
 		Difficulty: 0,
+		BaseFee:    chain.GenesisBaseFee,
```
vcastellm: Why moving this around? 😄

types/rlp_marshal.go
```diff
@@ -191,13 +193,19 @@ func (t *Transaction) MarshalRLPTo(dst []byte) []byte {
 func (t *Transaction) MarshalRLPWith(arena *fastrlp.Arena) *fastrlp.Value {
 	vv := arena.NewArray()
 
+	// Specify zero chain ID as per spec.
+	// This is needed to be compatible with other EVM chains and have the same format.
+	if t.Type == DynamicFeeTx {
```
ferranbt: What is this new Type of transaction?

state/executor.go
```diff
@@ -918,67 +905,17 @@ func TransactionGasCost(msg *types.Transaction, isHomestead, isIstanbul bool) (u
 	return cost, nil
 }
 
-// checkDynamicFees checks correctness of the EIP-1559 feature-related fields.
-// Basically, makes sure gas tip cap and gas fee cap are good.
```
ferranbt: Why have been these checks moved?

jsonrpc/types.go
```diff
@@ -17,6 +17,8 @@ type transactionOrHash interface {
 type transaction struct {
```
ferranbt: Can you add a test to make sure the encoding is correct? Shouldn't the `gasTipCap` and `gasFeeCap` be optional?

jsonrpc/eth_endpoint.go
```diff
@@ -167,6 +170,8 @@ func (e *Eth) SendRawTransaction(buf argBytes) (interface{}, error) {
 		return nil, err
 	}
 
+	tx = fillTxFees(tx, e.store.GetBaseFee())
```
ferranbt: You cannot change the fields from a signed transaction sent by `JsonRPC`. The user signed this transactions, the fields cannot be mutated.

ferranbt (changes requested):

state/executor.go
begmaroman: These checks have been moved to the tx validation function in the tx pool. That is the most suitable part for this logic since this is validation so it should be done with the rest of the validation which is placed in the tx pool. 
Here: https://github.com/0xPolygon/polygon-edge/blob/develop/txpool/txpool.go#L576

jsonrpc/types.go
begmaroman: Done

types/rlp_marshal.go
begmaroman: Basically, the dynamic fee tx type has a [modified RLP structure](https://github.com/umbracle/ethgo/blob/main/structs_marshal_rlp.go#L33): 
- Chain ID. We don't have it but it still has to be populated with 0 in order to be compatible with other chains. 
- Access List. Same point as chain ID. 

This has been tested via metamask on my local env.

jsonrpc/eth_endpoint.go
begmaroman: We don't really mutate it, just populate either gas price OR gas fee cap + gas tip cap. This is used within the node only and gets broadcasted properly. Gas price is used in many places in the code so we should define it if not provided.

begmaroman: Also, this part is modified in the next PR in the chain: https://github.com/0xPolygon/polygon-edge/pull/1230

types/transaction.go
```diff
@@ -118,7 +118,14 @@ func (t *Transaction) Copy() *Transaction {
...
 // Cost returns gas * gasPrice + value
 func (t *Transaction) Cost() *big.Int {
-	total := new(big.Int).Mul(t.GasPrice, new(big.Int).SetUint64(t.Gas))
+	total := new(big.Int)
+
+	if t.GasFeeCap != nil || t.GasFeeCap.BitLen() > 0 {
+		total = total.Mul(t.GasFeeCap, new(big.Int).SetUint64(t.Gas))
+	} else {
+		total = total.Mul(t.GasPrice, new(big.Int).SetUint64(t.Gas))
+	}
```
vcastellm: Probably good to simplify to something like

```suggestion
        var factor uint64
	if t.GasFeeCap != nil || t.GasFeeCap.BitLen() > 0 {
                 factor = t.GasFeeCap
	} else {
		factor = t.GasPrice
	}
	total = total.Mul(factor, new(big.Int).SetUint64(t.Gas))
```

vcastellm (approved):
Aside from the comment LGTM

types/transaction.go
begmaroman: Done

command/genesis/polybft_params.go
begmaroman: No reason, put it back :)

jsonrpc/types.go
ferranbt: Are `gasTipCap` and `gasFeeCap` optional?

state/executor.go
ferranbt: This check should be done also at the executor level. Otherwise, if by some reason a transaction slips the checks and ends up in the transition, it should fail as well.

jsonrpc/eth_endpoint.go
ferranbt: Well, you are mutating the struct if you change the value of the fields. I think you are not mutating it enough to make it change the hash since the `rlp` function will discard the fields depending on the type. However, this changes an imporant invariant of the `Transaction`, this is, the idea that `Transaction` is an immutable object that has a lifecycle that does not change. Instead, now the `Transaction` is a wrapper around a user action.

jsonrpc/types.go
begmaroman: Yes

jsonrpc/eth_endpoint.go
begmaroman: Well, it would require a lot of refactoring which is overkill imo. Another option is to calculate gas price every time when we use it but this solution is completely inefficient. Also, We could delegate this part to the user but it might be really confusing. 
From my side, I don't see potential issues with this. We mutate the tx object when calculating hash as well. 
Do you have any suggestions?

state/executor.go
begmaroman: There is a similar check on the executor level - `subGasLimitPrice`. 
Another point is how a transaction could come to the executor level without passing a tx pool (theoretically)? And if it is possible, we have big problems :)

jsonrpc/eth_endpoint.go
ferranbt: Well, if we were to populate fields from the `Transaction`, we cannot do it here only. We should also do it over the p2p and the storage layer.
I do not have the full scope of where the gas price is used. The solution would depend on that. My mental model tells me that the "gas price" (or any other type of sorting function) is required over the txpool mainly. In that case, a possible solution is to create a `txpool.Transaction` wrapper over the `Transaction` which decides how is the sorting factor calculated.

jsonrpc/types.go
ferranbt: They are not optional right now.

state/executor.go
ferranbt: Well, a malicious validator could propose a block with the wrong values. In that case, this is the only way to check if those values are valid or not.

begmaroman: Got it. Reverted it back and made it more accurate.

jsonrpc/eth_endpoint.go
begmaroman: It is used in the following places: 
1. Tx pool
2. State Executor
3. Blockchain
4. JSON RPC
5. Tx signer
6. Types
7. Several tests

> Well, if we were to populate fields from the Transaction, we cannot do it here only. We should also do it over the p2p and the storage layer.

I could update RLP marshaling for storage so it would store all these 3 fields regardless of the tx type. The next PR contains some changes related to this part so it does this in the following places: 
1. JSON RPC when receiving transactions.
2. Entrypoint of tx pool for all incoming transactions including ones coming from the network (gossip). 
3. When decoding a transaction while reading it via RPC request.

begmaroman: Took changes from the next PR into this one since it would be more reasonable.

state/executor.go
begmaroman: Btw, in this case, the transaction won't pass even without this validation. Malicious validator won't achieve its goal.

state/executor.go
```diff
@@ -393,6 +393,12 @@ func (t *Transition) subGasLimitPrice(msg *types.Transaction) error {
 	upfrontGasCost := new(big.Int).Set(msg.GasPrice)
 	upfrontGasCost.Mul(upfrontGasCost, new(big.Int).SetUint64(msg.Gas))
 
+	// Apply EIP-1559 tx cost calculation logic
+	if msg.GasFeeCap.BitLen() > 0 {
+		upfrontGasCost = new(big.Int).SetUint64(msg.Gas)
+		upfrontGasCost = upfrontGasCost.Mul(upfrontGasCost, msg.GasFeeCap)
+	}
```
goran-ethernal: Since we are rewriting the `uprfrontGasCost` here, you can just put in `else` the calculation of the same variable when it is not `EIP-1559` transaction.

state/executor.go
```diff
@@ -414,6 +420,38 @@ func (t *Transition) nonceCheck(msg *types.Transaction) error {
 	return nil
 }
 
+// checkDynamicFees checks correctness of the EIP-1559 feature-related fields.
+// Basically, makes sure gas tip cap and gas fee cap are good.
+func (t *Transition) checkDynamicFees(msg *types.Transaction) error {
+	if msg.GasFeeCap.BitLen() == 0 && msg.GasTipCap.BitLen() == 0 {
```
goran-ethernal: Is it correct to return nil here if it is a dynamic fee transaction? It's ok if it's a legacy transaction, but correct me if I'm wrong, if a EIP-1559 transaction has 0 as GasFeeCap, doesn't that mean that the user that sent the transaction is not willing to pay anything for it, which is not a valid use case?

state/executor.go
begmaroman: Agree, done

state/executor.go
begmaroman: I'm not sure about that tho. In case when eip-1559 is not activated yet, users must be able to send transactions with empty dynamic fees. I can add a check to apply this validation only for dynamic tx when eip-1559 is activated already.

jsonrpc/eth_endpoint.go
begmaroman: Actually, I agree with you that we should not modify tx object, but there would be a lot of changes around this. I would suggest keeping it like that and doing research and updating it in one of the next PRs. It would be easier to understand and review.

ferranbt: I would prefer not to merge something that we are going to change soon after in a follow-up PR, many things on context can be lost along the way.

txpool/txpool.go
```diff
@@ -347,6 +354,11 @@ func (p *TxPool) getSealing() bool {
 	return atomic.LoadUint32(&p.sealing) == 1
 }
 
+// SetBaseFee sets the given base fee
+func (p *TxPool) SetBaseFee(baseFee uint64) {
```
ferranbt: Could we reuse `Prepare` function (i.e. Prepare(baseFee uint)? `baseFee` is a parameter of the block and `Prepare` is called before building a block.

state/executor.go
goran-ethernal: Ok, I've analyzed the code a bit further. So if someone does set transaction type to `DynamicFeeTx`, but doesn't set the `GasFeeCap` and `GasTipCap`, or if they are zero, we will consider the given transaction as a legacy transaction? Am I correct? If that is the case, then this part of code is ok.

jsonrpc/eth_endpoint.go
begmaroman: Anyway, we haven't decided on the solution yet. 
- Gas price is used by several packages.
- This function requires base fee value in order to pre-fill fields.
- This function is just a helper for a tx sender which does not really affect the logic and behavior.  

The option that we have atm is to introduce a new object within each of these packages and do calculations there. It would create several code duplications and affects the architecture since we would have to sync the base fee across all of them and make sure it is in sync every time. I see this option as overengineering (just my opinion).

state/executor.go
begmaroman: I think so. Do you think the behaviour should be different?

goran-ethernal: To my knowledge, on `ethereum`, if you do not specify the the maximum fee you are willing to pay per unit of gas, your transaction gets rejected, but maybe we can put this in the RFC for further discussion.

txpool/txpool.go
begmaroman: Moved the passing base fee to tx pool in a separate PR to review it easily - https://github.com/0xPolygon/polygon-edge/pull/1243. And applied your suggestion.

types/rlp_marshal.go
begmaroman: Moved to https://github.com/0xPolygon/polygon-edge/pull/1237. Added description and link to EIP to the code and to the RFC.

ferranbt (approved):

goran-ethernal (approved):

PR #1211, opened 2023-02-09
Remove codecov

vcastellm:
As we already have SonaQube covering same funtionality.

Inline codecov comments are annoying.


igorcrevar (approved):
Thank god (and in this case Victor)

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #1226, opened 2023-02-14
Execute block in `fsm.Insert` if cached block is undefined or mismatch detected

Stefan-Ethernal:
# Description

This PR fixes a situation when a `FullBlock` cached in the fsm, during `Validate` function execution remains undefined. It can happen in case any of the validations fail and `Validate` function doesn't get fully executed, due to some data invalidity.

The solution proposal consists of detecting two scenarios:
1. a cached proposal that we verified is undefined
2. a hash of the cached proposal that was verified doesn't match the hash of the received proposal via `Insert` function invocation.
In any of these two scenarios, an error is returned, and we let the syncer insert the block, just to make sure we don't have any unwanted errors and inconsistencies in our data. 

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -435,14 +435,9 @@ func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) 
 	}
 
 	if newBlock == nil || newBlock.Block.Hash() != proposedBlock.Hash() {
-		f.logger.Warn("insert block: cached block is either undefined or hashes don't match. Executing retrieved proposal...")
-		// execute block
-		finalBlock, err := f.backend.ProcessBlock(f.parent, &proposedBlock, nil)
-		if err != nil {
-			return nil, fmt.Errorf("failed to execute block %d: %w", proposedBlock.Number(), err)
-		}
-
-		newBlock = finalBlock
+		// if this is the case, we will let syncer insert the block
+		return nil, errors.New(`failed to insert proposal, because the validated proposal is either nil 
+		or it does not match the received one`)
 	}
 
 	// In this function we should try to return little to no errors since
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -429,10 +429,29 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) (*types.FullBlock, error) {
 	newBlock := f.target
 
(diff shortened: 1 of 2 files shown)


consensus/polybft/fsm.go
```diff
@@ -429,10 +429,29 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
 func (f *fsm) Insert(proposal []byte, committedSeals []*messages.CommittedSeal) (*types.FullBlock, error) {
 	newBlock := f.target
 
+	var proposedBlock types.Block
+	if err := proposedBlock.UnmarshalRLP(proposal); err != nil {
+		return nil, fmt.Errorf("failed to insert proposal, block unmarshaling failed: %w", err)
+	}
+
+	if newBlock == nil || newBlock.Block.Hash() != proposedBlock.Hash() {
```
Stefan-Ethernal, line 439: Should we return an error immediately in case `cachedBlock` and `proposedBlock` mismatch?

```suggestion
        if newBlock.Block.Hash() != proposedBlock.Hash() {
             return errors.New("failed to insert proposal: cached block and proposed blocks mismatch")
        }
	if newBlock == nil  {
```

goran-ethernal: In my opinion - Yes. If the proposal you validated is not equal to the proposal you got to insert, then some malicious behavior is possibly at work, or a bug on client side. Better let syncer insert the block, and make sure we didn't mess something up, just to be sure.

igorcrevar: I would do same as Goran suggested

igorcrevar: if `f.target` is `nil` i would return from function immediately with error

goran-ethernal: We agreed to return an error here.

goran-ethernal (approved):

Nemanja0x (approved):

consensus/polybft/fsm.go
```diff
@@ -429,10 +429,24 @@ func (f *fsm) VerifyStateTransactions(transactions []*types.Transaction) error {
...
+		// if this is the case, we will let syncer insert the block
+		return nil, errors.New(`failed to insert proposal, because the validated proposal is either nil 
+		or it does not match the received one`)
```
Stefan-Ethernal: Nitpick, consistency wise:
```suggestion
		return nil, errors.New("failed to insert proposal, because the validated proposal is either nil 
		or it does not match the received one")
```

igorcrevar (approved):

PR #1228, opened 2023-02-14
EVM-352 Event tracker block finalization

igorcrevar:
# Description

Block should not be considered final until at least `NumBlockConfirmations` number of children blocks appear.
Only logs of finalized blocks should be delegated to the subscriber.

The changes should only affect `stateSyncManager`, as it relies on the event tracker to retrieve logs from the `eth` node.
`BlockFinalityDepth` will be `0` for the `StateSyncRelayer` because we have instant finality on our chain.

Server command has been updated to include an optional `--num-block-confirmations` flag, which controls the minimum number of child blocks that must appear after a block before we consider that block finalized. By default, this value is set to 64.
That parameter can be set from configuration also with `num_block_confirmations`. 

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/server/config/config.go
+++ b/command/server/config/config.go
@@ -79,7 +79,7 @@ const (
 	// requests with fromBlock/toBlock values (e.g. eth_getLogs)
 	DefaultJSONRPCBlockRangeLimit uint64 = 1000
 
-	// after how many blocks we consider block is finalized, default value
+	// DefaultBlockFinalityDepth minimal number of child blocks required for the parent block to be considered final
 	DefaultBlockFinalityDepth uint64 = 100
 )
 
--- a/command/server/config/config.go
+++ b/command/server/config/config.go
@@ -80,7 +80,8 @@ const (
 	DefaultJSONRPCBlockRangeLimit uint64 = 1000
 
 	// DefaultBlockFinalityDepth minimal number of child blocks required for the parent block to be considered final
-	DefaultBlockFinalityDepth uint64 = 100
+	// on ethereum epoch lasts for 32 blocks. more details: https://www.alchemy.com/overviews/ethereum-commitment-levels
+	DefaultBlockFinalityDepth uint64 = 64
 )
 
 // DefaultConfig returns the default server configuration
--- a/command/server/config/config.go
+++ b/command/server/config/config.go
@@ -31,7 +31,9 @@ type Config struct {
 	JSONRPCBatchRequestLimit uint64     `json:"json_rpc_batch_request_limit" yaml:"json_rpc_batch_request_limit"`
 	JSONRPCBlockRangeLimit   uint64     `json:"json_rpc_block_range_limit" yaml:"json_rpc_block_range_limit"`
 	JSONLogFormat            bool       `json:"json_log_format" yaml:"json_log_format"`
-	Relayer                  bool       `json:"relayer" yaml:"relayer"`
+
+	Relayer                 bool   `json:"relayer" yaml:"relayer"`
+	BlockFinalizedThreshold uint64 `json:"block_finalized_threshold" yaml:"block_finalized_threshold"`
 }
 
 // Telemetry holds the config details for metric services.
@@ -76,6 +78,9 @@ const (
 	// DefaultJSONRPCBlockRangeLimit maximum block range allowed for json_rpc
 	// requests with fromBlock/toBlock values (e.g. eth_getLogs)
 	DefaultJSONRPCBlockRangeLimit uint64 = 1000
+
+	// after how many blocks we consider block is finalized, default value
+	DefaultBlockFinalizedThreshold uint64 = 100
 )
 
 // DefaultConfig returns the default server configuration
@@ -113,6 +118,7 @@ func DefaultConfig() *Config {
 		JSONRPCBatchRequestLimit: DefaultJSONRPCBatchRequestLimit,
 		JSONRPCBlockRangeLimit:   DefaultJSONRPCBlockRangeLimit,
 		Relayer:                  false,
+		BlockFinalizedThreshold:  DefaultBlockFinalizedThreshold,
 	}
 }
 
--- a/command/server/config/config.go
+++ b/command/server/config/config.go
@@ -32,8 +32,8 @@ type Config struct {
 	JSONRPCBlockRangeLimit   uint64     `json:"json_rpc_block_range_limit" yaml:"json_rpc_block_range_limit"`
 	JSONLogFormat            bool       `json:"json_log_format" yaml:"json_log_format"`
 
-	Relayer            bool   `json:"relayer" yaml:"relayer"`
-	BlockFinalityDepth uint64 `json:"block_finality_depth" yaml:"block_finality_depth"`
+	Relayer               bool   `json:"relayer" yaml:"relayer"`
+	NumBlockConfirmations uint64 `json:"num_block_confirmations" yaml:"num_block_confirmations"`
 }
 
 // Telemetry holds the config details for metric services.
@@ -79,9 +79,9 @@ const (
 	// requests with fromBlock/toBlock values (e.g. eth_getLogs)
 	DefaultJSONRPCBlockRangeLimit uint64 = 1000
 
-	// DefaultBlockFinalityDepth minimal number of child blocks required for the parent block to be considered final
+	// DefaultNumBlockConfirmations minimal number of child blocks required for the parent block to be considered final
 	// on ethereum epoch lasts for 32 blocks. more details: https://www.alchemy.com/overviews/ethereum-commitment-levels
-	DefaultBlockFinalityDepth uint64 = 64
+	DefaultNumBlockConfirmations uint64 = 64
 )
 
 // DefaultConfig returns the default server configuration
@@ -119,7 +119,7 @@ func DefaultConfig() *Config {
 		JSONRPCBatchRequestLimit: DefaultJSONRPCBatchRequestLimit,
 		JSONRPCBlockRangeLimit:   DefaultJSONRPCBlockRangeLimit,
 		Relayer:                  false,
-		BlockFinalityDepth:       DefaultBlockFinalityDepth,
+		NumBlockConfirmations:    DefaultNumBlockConfirmations,
 	}
 }
 
--- a/command/server/config/config.go
+++ b/command/server/config/config.go
@@ -32,8 +32,8 @@ type Config struct {
 	JSONRPCBlockRangeLimit   uint64     `json:"json_rpc_block_range_limit" yaml:"json_rpc_block_range_limit"`
 	JSONLogFormat            bool       `json:"json_log_format" yaml:"json_log_format"`
 
-	Relayer                 bool   `json:"relayer" yaml:"relayer"`
-	BlockFinalizedThreshold uint64 `json:"block_finalized_threshold" yaml:"block_finalized_threshold"`
+	Relayer            bool   `json:"relayer" yaml:"relayer"`
+	BlockFinalityDepth uint64 `json:"block_finality_depth" yaml:"block_finality_depth"`
 }
 
 // Telemetry holds the config details for metric services.
@@ -80,7 +80,7 @@ const (
 	DefaultJSONRPCBlockRangeLimit uint64 = 1000
 
 	// after how many blocks we consider block is finalized, default value
-	DefaultBlockFinalizedThreshold uint64 = 100
+	DefaultBlockFinalityDepth uint64 = 100
 )
 
 // DefaultConfig returns the default server configuration
@@ -118,7 +118,7 @@ func DefaultConfig() *Config {
 		JSONRPCBatchRequestLimit: DefaultJSONRPCBatchRequestLimit,
 		JSONRPCBlockRangeLimit:   DefaultJSONRPCBlockRangeLimit,
 		Relayer:                  false,
-		BlockFinalizedThreshold:  DefaultBlockFinalizedThreshold,
+		BlockFinalityDepth:       DefaultBlockFinalityDepth,
 	}
 }
 
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -232,7 +232,7 @@ func setFlags(cmd *cobra.Command) {
 		&params.rawConfig.BlockFinalityDepth,
 		blockFinalityDepthFlag,
 		defaultConfig.BlockFinalityDepth,
-		"after how many blocks we consider block is finalized",
+		"minimal number of child blocks required for the parent block to be considered final",
 	)
 
 	setLegacyFlags(cmd)
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -229,9 +229,9 @@ func setFlags(cmd *cobra.Command) {
 	)
 
 	cmd.Flags().Uint64Var(
-		&params.rawConfig.BlockFinalizedThreshold,
-		blockFinalizedThresholdFlag,
-		defaultConfig.BlockFinalizedThreshold,
+		&params.rawConfig.BlockFinalityDepth,
+		blockFinalityDepthFlag,
+		defaultConfig.BlockFinalityDepth,
 		"after how many blocks we consider block is finalized",
 	)
 
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -229,9 +229,9 @@ func setFlags(cmd *cobra.Command) {
 	)
 
 	cmd.Flags().Uint64Var(
-		&params.rawConfig.BlockFinalityDepth,
-		blockFinalityDepthFlag,
-		defaultConfig.BlockFinalityDepth,
+		&params.rawConfig.NumBlockConfirmations,
+		numBlockConfirmationsFlag,
+		defaultConfig.NumBlockConfirmations,
 		"minimal number of child blocks required for the parent block to be considered final",
 	)
 
--- a/command/server/server.go
+++ b/command/server/server.go
@@ -228,6 +228,13 @@ func setFlags(cmd *cobra.Command) {
 		"start the state sync relayer service (PolyBFT only)",
 	)
 
+	cmd.Flags().Uint64Var(
+		&params.rawConfig.BlockFinalizedThreshold,
+		blockFinalizedThresholdFlag,
+		defaultConfig.BlockFinalizedThreshold,
+		"after how many blocks we consider block is finalized",
+	)
+
 	setLegacyFlags(cmd)
 
 	setDevFlags(cmd)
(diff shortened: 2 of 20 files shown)


command/server/config/config.go
```diff
@@ -31,7 +31,9 @@ type Config struct {
 	JSONRPCBatchRequestLimit uint64     `json:"json_rpc_batch_request_limit" yaml:"json_rpc_batch_request_limit"`
 	JSONRPCBlockRangeLimit   uint64     `json:"json_rpc_block_range_limit" yaml:"json_rpc_block_range_limit"`
 	JSONLogFormat            bool       `json:"json_log_format" yaml:"json_log_format"`
-	Relayer                  bool       `json:"relayer" yaml:"relayer"`
+
+	Relayer                 bool   `json:"relayer" yaml:"relayer"`
+	BlockFinalizedThreshold uint64 `json:"block_finalized_threshold" yaml:"block_finalized_threshold"`
```
begmaroman: Maybe `block_finality_depth` ?

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/4e62780847c16aa37d8639b24e4e84b0e9027061

e2e-polybft/bridge_test.go
```diff
@@ -56,7 +56,10 @@ func checkLogs(
 }
 
 func TestE2E_Bridge_MainWorkflow(t *testing.T) {
-	const num = 10
+	const (
+		num                = 10
+		blockFinalityDepth = 4
```
goran-ethernal: We should update the `cluster` script, to have a similar `blockFinalityDepth` since we also set the `epochSize` to 10 in it. Since the script is used for testing/debugging, I think it should have a `blockFinalityDepth` similar to Bridge e2e test.

e2e-polybft/framework/test-cluster.go
```diff
@@ -47,6 +47,9 @@ const (
 
 	// prefix for validator directory
 	defaultValidatorPrefix = "test-chain-"
+
+	// after how many blocks we consider block is finalized
+	defaultBlockFinalityDepth = 0
```
goran-ethernal: Do we need this `constant`?

igorcrevar: This has been removed in the latest commit https://github.com/0xPolygon/polygon-edge/pull/1228/commits/eecb93dddc6af118fab9d94be99140b748755a51

e2e-polybft/bridge_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/8c2218bb0b63abbc49def4f4ab58fd3be471c300
Set `--block-finality-depth` to 2

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+func btoi(b []byte) uint64 {
+	return binary.BigEndian.Uint64(b)
+}
+
+func itob(u uint64) []byte {
+	buf := [8]byte{}
+	binary.BigEndian.PutUint64(buf[:], u)
+
+	return buf[:]
+}
```
Stefan-Ethernal: Think that you can use `common.EncodeUint64ToBytes`, which was introduced in #1055.

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+	if last, _ := bucket.Cursor().Last(); last != nil {
+		return btoi(last) + 1
+	}
+
+	return 0
+}
+
+func btoi(b []byte) uint64 {
+	return binary.BigEndian.Uint64(b)
+}
```
Stefan-Ethernal: Maybe we could move this function to `helper/common/common.go`?

command/server/server.go
```diff
@@ -228,6 +228,13 @@ func setFlags(cmd *cobra.Command) {
 		"start the state sync relayer service (PolyBFT only)",
 	)
 
+	cmd.Flags().Uint64Var(
+		&params.rawConfig.BlockFinalityDepth,
+		blockFinalityDepthFlag,
+		defaultConfig.BlockFinalityDepth,
```
Stefan-Ethernal: Is this setting related to the rootchain/Ethereum?

Maybe we could couple `blockFinalityDepth` with epoch size by default. It would mean that we consider a block is final if at least a single epoch is elapsed (so we could consider that blocks from the previous epoch are final).
If we assume that most of the time rootchain is going to be Ethereum, its current epoch size is 32 blocks (currently epoch lasts 6.4 minutes, where the target block time is 12 seconds).
This is the article, to which I've reached out and got the impression that block finalization should be related to epoch size: https://www.alchemy.com/overviews/ethereum-commitment-levels

consensus/polybft/statesyncrelayer/state_sync_relayer.go
```diff
@@ -87,6 +87,7 @@ func (r *StateSyncRelayer) Start() error {
 		r.rpcEndpoint,
 		r.stateReceiverAddr,
 		r,
+		0, // our (side-chain) is instant finality POS, so no need to wait
```
Stefan-Ethernal:
```suggestion
		0, // sidechain (Polygon POS) is instant finality, so no need to wait
```

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+	notifierCh chan<- []*ethgo.Log,
+	logger hcf.Logger) (*EventTrackerStore, error) {
+	db, err := bolt.Open(path, 0600, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	store := &EventTrackerStore{
+		conn:          db,
+		finalityDepth: finalizedTrashhold,
```
Stefan-Ethernal: You had a typo in `finalizedTrashhold`, so I propose to rename it to `finalityDepth` or `finalizedThreshold` (if you find it more suitable).
```suggestion
	finalityDepth uint64,
	notifierCh chan<- []*ethgo.Log,
	logger hcf.Logger) (*EventTrackerStore, error) {
	db, err := bolt.Open(path, 0600, nil)
	if err != nil {
		return nil, err
	}

	store := &EventTrackerStore{
		conn:          db,
		finalityDepth: finalityDepth,
```

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+			"from", lastLogIndx, "to", lastLogIndx+uint64(len(logs))-1, "block", lastBlockNumber)
+
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	lastBlockNumber = logs[len(logs)-1].BlockNumber
+
+	notifyLogs, lastProcessedIdx, err := e.getFinalizedLogs(lastBlockNumber - e.finalityDepth)
```
goran-ethernal: If `lastBlockNumber` is lower than `e.finalityDepth`, I would say, we don't need to do the `getFinalizedLogs` and `notifyFinalizedLogs`, and you don't need to open up a new `db` transaction.

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+
+func (e *Entry) notifyFinalizedLogs(logs []*ethgo.Log, lastProcessedIdx uint64) error {
+	if len(logs) == 0 {
+		return nil
+	}
+
+	e.logger.Info("notify event logs", "len", len(logs), "last processed", lastProcessedIdx)
+
+	if err := e.conn.Update(func(tx *bolt.Tx) error {
+		return tx.Bucket(e.bucketNextToProcess).Put([]byte("0"), itob(lastProcessedIdx+1))
```
Stefan-Ethernal: Could you extract this singular key (`[]byte("0")`) to a variable?

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,331 @@
...
+			if err := cursorLogs.Delete(); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	})
+}
+
+// GetLog implements the store interface
```
Stefan-Ethernal: Is this the correct comment? Also all the other functions maybe should have this comment (or possibly it is more meaningful to copy comment from the interface itself)?
```suggestion
// GetLog implements the Entry interface
```

tracker/event_tracker.go
```diff
@@ -44,19 +46,20 @@ func (e *EventTracker) Start(ctx context.Context) error {
...
 
-	e.logger.Info("Start tracking events", "contract address", e.contractAddr)
+	e.logger.Info("Start tracking events", "block finality depth", e.finalityDepth, "contract address", e.contractAddr)
 
 	tt, err := tracker.NewTracker(provider.Eth(),
 		tracker.WithBatchSize(10),
 		tracker.WithStore(store),
 		tracker.WithFilter(&tracker.FilterConfig{
-			Async: false,
+			Async: true,
```
Stefan-Ethernal, line 61: Heads up:
this was changed a couple of months ago because it was noticed that putting the tracker in async mode, resulted in gaps in captured logs (bridge e2e tests were failing occasionally).
If you were able to overcome such a problem with custom-made storage, then we are ok.

command/server/server.go
goran-ethernal: I agree with Stefan. The default should be 32 blocks (length of an Ethereum epoch).

tracker/event_tracker.go
igorcrevar: Old channel from tracker is not used anymore. By setting this configuration to true, we are preventing blocking of tracker. Otherwise we need to (dummy)read that old channel. It is easier this way.

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

command/server/server.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

consensus/polybft/statesyncrelayer/state_sync_relayer.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

tracker/event_tracker_store.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/7c00602ddd77823dd9574710897a8307d0399e5f

Stefan-Ethernal (approved):

goran-ethernal (approved):

begmaroman (approved):

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,324 @@
...
+				return err
+			}
+
+			if err := bucketLogs.Put(common.EncodeUint64ToBytes(logIdx), val); err != nil {
+				return err
+			}
+		}
+
+		e.logger.Info("write event logs",
+			"from", lastLogIndx, "to", lastLogIndx+uint64(len(logs))-1, "block", lastBlockNumber)
```
stana-miric: Here we are logging lastBlockNumber which is not set at this point

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/c2175b54214c888a06fba8980dbb78a59d7f7037

stana-miric (approved):

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,326 @@
...
+	if lastBlockNumber <= e.finalityDepth {
+		return nil
+	}
+
+	notifyLogs, lastProcessedIdx, err := e.getFinalizedLogs(lastBlockNumber - e.finalityDepth)
+	if err != nil {
+		return err
+	}
+
+	return e.notifyFinalizedLogs(notifyLogs, lastProcessedIdx)
```
ferranbt: This is not going to work in this scenario:
- 1 subscribed event is included in block 10.
- 1k blocks more are created without new subscribed events.

The first event will only be notified when a new event after the 1k blocks is inserted. The tracker should notify the event at block 10 + numBlockConfirmations.

ferranbt: There is an entry on the `Set` function that is used to notify when a new block is inserted, I would do this check there instead.

tracker/event_tracker_store.go
```diff
@@ -0,0 +1,326 @@
...
+		return tx.Bucket(e.bucketNextToProcess).Put(nextToProcessKey, common.EncodeUint64ToBytes(lastProcessedIdx+1))
+	}); err != nil {
+		return err
+	}
+
+	e.logger.Info("event logs have been notified to a subscriber", "len", len(logs), "last processed", lastProcessedIdx)
+
+	// notify logs - for testing purpose chan can be nil
+	if e.notifierCh != nil {
+		e.notifierCh <- logs
```
ferranbt: I would prefer to use a method from the `Subscription` interface here instead of a channel. It is too easy to create deadlocks with a channel.

tracker/event_tracker_store_test.go
```diff
@@ -0,0 +1,35 @@
...
+	require.NoError(t, err)
+
+	closeFn := func() {
+		require.NoError(t, os.RemoveAll(dir))
+	}
+
+	return store, closeFn
+}
+
+func TestBoltDBStore(t *testing.T) {
```
ferranbt: Could we have more unit tests on the store to cover the new uses cases?

ferranbt (changes requested):

tracker/event_tracker_store_test.go
igorcrevar: Fixed  https://github.com/0xPolygon/polygon-edge/pull/1228/commits/1e6f354eea1ee27452409969234d19382c79f894

tracker/event_tracker_store.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/1e6f354eea1ee27452409969234d19382c79f894

tracker/event_tracker_store.go
igorcrevar: Refactored https://github.com/0xPolygon/polygon-edge/pull/1228/commits/1e6f354eea1ee27452409969234d19382c79f894

tracker/event_tracker_store.go
```diff
@@ -25,15 +29,15 @@ var (
...
 	finalityDepth uint64
-	notifierCh    chan<- []*ethgo.Log
+	subscriber    eventSubscription
 	logger        hcf.Logger
 }
 
 // NewEventTrackerStore creates a new EventTrackerStore
 func NewEventTrackerStore(
 	path string,
 	finalityDepth uint64,
```
ferranbt: Can we call this `numBlockConfirmations`? That is how it is usually referred to this value.

tracker/event_tracker_store_test.go
```diff
@@ -1,35 +1,177 @@
...
+	entry.StoreLogs([]*ethgo.Log{
+		{BlockNumber: 1}, {BlockNumber: 5}, {BlockNumber: 8}, {BlockNumber: 11}, {BlockNumber: 12}, {BlockNumber: 15},
+	})
+
+	for i := 0; i < 10; i++ {
+		require.NoError(t, entry.saveNextToProcessIndx(uint64(i)))
 	}
+}
+
+func Test_EventTrackerStore_SetNotLastBlock(t *testing.T) {
```
ferranbt:
```suggestion
func TestEventTrackerStore_SetNotLastBlock(t *testing.T) {
```

tracker/event_tracker_store_test.go
```diff
@@ -1,35 +1,177 @@
...
+
+	store.TestStore(t, createSetupDB(nil, 2))
+}
+
+func Test_Entry_getFinalizedLogs(t *testing.T) {
+	t.Parallel()
+
+	const someFilterHash = "test"
+
+	tstore, fn := createSetupDB(nil, 3)(t)
```
ferranbt:
```suggestion
	tstore, closeFn := createSetupDB(nil, 3)(t)
```

tracker/event_tracker_store_test.go
```diff
@@ -1,35 +1,177 @@
...
+
+	entry, err := tstore.GetEntry(hash)
+	require.NoError(t, err)
+
+	require.NoError(t, entry.StoreLogs([]*ethgo.Log{
+		{BlockNumber: 1}, {BlockNumber: 2}, {BlockNumber: 3},
+	}))
+
+	for i := 0; i < 3; i++ {
+		block := ethgo.Block{Number: uint64(i + 2)}
```
ferranbt, line 168: Could you add a bit more description on how this is testing the finality?

ferranbt (approved):
Minor stuff. Otherwise LGTM.

tracker/event_tracker_store_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/58898271b21757ec25be593a5fcc7a983c032392

tracker/event_tracker_store_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/58898271b21757ec25be593a5fcc7a983c032392

tracker/event_tracker_store_test.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/58898271b21757ec25be593a5fcc7a983c032392

tracker/event_tracker_store.go
igorcrevar: renamed https://github.com/0xPolygon/polygon-edge/pull/1228/commits/2fe1d5c1fba88ff426e9a55dbbccf1e044921ad2

PR #1230, opened 2023-02-16
Implement EIP-1559 tx signer

begmaroman:
# Description

Implement tx signer according to EIP-1559 spec. Use the following k6 code to test a transaction:

```js
import eth from 'k6/x/ethereum';

var rpc_url = __ENV.RCP_URL
if (rpc_url == undefined) {
  rpc_url = "http://localhost:10002"
}

const client = new eth.Client({
  url: rpc_url,
  // You can also specify a private key here
  // privateKey: 'private key of your account',
  // or a mnemonic
  // mnemonic: 'my mnemonic'
});

// You can use an existing premined account
const root_address = "0x85da99c8a7c2c95964c8efd687e95e632fc533d6"

export function setup() {
  return { nonce: client.getNonce(root_address) };
}

export default function (data) {
  console.log(`nonce => ${data.nonce}`);
  const gas = client.gasPrice();
  console.log(`gas price => ${gas}`);

  const bal = client.getBalance(root_address, client.blockNumber());
  console.log(`bal => ${bal}`);

  const tx = {
    to: "0xDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF",
    value: Number(0.0001 * 1e18),
    gas_price: 1000000000,
    gas_fee_cap: 10000000000, 
    gas_tip_cap: 1000000000,
    nonce: data.nonce,
  };

  const txh = client.sendRawTransaction(tx)
  console.log("tx hash => " + txh);

  // Optional: wait for the transaction to be mined
  client.waitForTransactionReceipt(txh).then((receipt) => {
    console.log("tx block hash => " + receipt.block_hash);
    console.log(JSON.stringify(receipt))

    const balance = client.getBalance(tx.to, client.blockNumber())
    console.log("recipient balance:", balance)
  });

  data.nonce = data.nonce + 1;

}
```

# Changes include

# Checklist

## Testing


crypto/txsigner_london.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// LondonSigner implements signer for EIP-1559
+type LondonSigner struct {
+	chainID        uint64
+	fallbackSigner TxSigner
+}
+
+// NewLondonSigner returns a new LondonSigner object
+func NewLondonSigner(chainID uint64, fallbackSigner TxSigner) *LondonSigner {
```
ferranbt: Why do we need a fallback signer?

crypto/txsigner_eip155.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"crypto/ecdsa"
+	"math/big"
+	"math/bits"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type EIP155Signer struct {
+	chainID uint64
```
igorcrevar: Maybe `chainID` should be `int64` because that is the type used in configuration? Also, there a lot conversions later on when doing math operations

crypto/txsigner_london_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+func TestLondonSigner_Sender(t *testing.T) {
```
igorcrevar: Both tests have same name, the only difference is `_` character between Test prefix. Can you name at least one of them differently? Are they testing all different possibilities for the transaction?

crypto/txsigner_london.go
begmaroman: We use london signer if a london hardfork is enabled. Even with enabled london hardfork, users could still send legacy transactions that should be handled by fallback signer.

begmaroman: Also, I didn't want to duplicate the code by moving the legacy tx encoding to the london signer.

crypto/txsigner_eip155.go
begmaroman: I just moved an existing code to a separate file. I would love to change it but within a separate PR because this one is only about adding a new signer.

crypto/txsigner_london_test.go
begmaroman: Done

vcastellm (approved):
LGTM

PR #1231, opened 2023-02-16
Add output flag to polybft-secrets

vcastellm:
# Description

The output flag indicates the secrets command that it should output the existing secrets in case they are found in local disk or in the vault configured in the secrets manager conf.

Example usages:

`$ polygon-edge polybft-secrets --data-dir test-chain- --insecure --output --num 4`
`$ polygon-edge polybft-secrets --data-dir test-chain-1 --insecure --output `

# Changes include

# Checklist

## Testing

### Manual tests

CLI

# Documentation update

### DIFF
--- a/command/polybftsecrets/params.go
+++ b/command/polybftsecrets/params.go
@@ -215,6 +215,13 @@ func (ip *initParams) getResult(secretsManager secrets.SecretsManager) (command.
 			}
 
 			res.PrivateKey = hex.EncodeToString(pk)
+
+			blspk, err := account.Bls.Marshal()
+			if err != nil {
+				return nil, err
+			}
+
+			res.BLSPrivateKey = hex.EncodeToString(blspk)
 		}
 	}
 
--- a/command/polybftsecrets/params.go
+++ b/command/polybftsecrets/params.go
@@ -20,6 +20,7 @@ const (
 	privateKeyFlag = "private"
 	networkFlag    = "network"
 	numFlag        = "num"
+	outputFlag     = "output"
 
 	// maxInitNum is the maximum value for "num" flag
 	maxInitNum = 30
@@ -50,6 +51,8 @@ type initParams struct {
 	numberOfSecrets int
 
 	insecureLocalStore bool
+
+	output bool
 }
 
 func (ip *initParams) validateFlags() error {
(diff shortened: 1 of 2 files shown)


begmaroman (approved):

Stefan-Ethernal:
Leaving a couple of comments:
- maybe we should consider putting `--account` and `--network` mutually exclusive to `--output`, because if we don't already have keys generated, command execution will fail.
- when having secrets already generated in local, I'm not able to output all the secrets at once. Namely this command is failing:
`go run . polybft-secrets --data-dir test-chain- --insecure --output`
with following error message
`failed to read account data: unable to read secret from disk (test-chain-/consensus/validator.key), open test-chain-/consensus/validator.key: no such file or directory`.

vcastellm:
> * when having secrets already generated in local, I'm not able to output all the secrets at once. Namely this command is failing:
>   `go run . polybft-secrets --data-dir test-chain- --insecure --output`
>   with following error message
>   `failed to read account data: unable to read secret from disk (test-chain-/consensus/validator.key), open test-chain-/consensus/validator.key: no such file or directory`.

Take a look to the path, you need to specify the `--num` param if you are relying in our local "automagic" folders 😄

vcastellm:
> * maybe we should consider putting `--account` and `--network` mutually exclusive to `--output`, because if we don't already have keys generated, command execution will fail.

It works if you specify the same command used to generate the secrets, eg:
`$polygon-edge polybft-secrets --data-dir test-chain- --insecure --num 4 --output --account=false`

🤔

Stefan-Ethernal:
> It works if you specify the same command used to generate the secrets, eg: `$polygon-edge polybft-secrets --data-dir test-chain- --insecure --num 4 --output --account=false`

Ok seems like you are right, it is working properly when `--num` is submitted. 😉 
If secrets are generated already, we don't need to provide neither `--account` nor `--network` flags.

As a side note: it is anyways going to output public identities (ECDSA address, BLS public key and p2p node id) with or without `--output` provided when generating secrets. Not sure do we want to change it? 🤔 Probably doesn't matter... WDYT?

Stefan-Ethernal (approved):
LGTM

goran-ethernal (approved):

vcastellm:
> As a side note: it is anyways going to output public identities (ECDSA address, BLS public key and p2p node id) with or without `--output` provided when generating secrets. Not sure do we want to change it? 🤔 Probably doesn't matter... WDYT?

It's convenient to have the output when generating the secrets, but we need also a way to print them, because for example, in cloud secrets, are not stored in FS

PR #1233, opened 2023-02-16
Introduce signature file in validator secrets to handle validator key signature

vcastellm:
# Description

This PR introduces a new file in validator secrets to handle validator key signature, this is a necessary change to allow secure deployments.

Also implement generating secrets that weren't previously generated with the init command, and letting the use know:

```
[SECRETS GENERATED]
network-key, validator-key, validator-bls-key, validator-bls-signature
```

# Changes include

# Breaking changes

The secrets init command now accepts the `chain-id` param

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/genesis/utils.go
+++ b/command/genesis/utils.go
@@ -137,7 +137,7 @@ func ReadValidatorsByPrefix(dir, prefix string) ([]*polybft.Validator, error) {
 	for i, file := range validatorKeyFiles {
 		path := filepath.Join(dir, file)
 
-		account, nodeID, err := getSecrets(path)
+		account, nodeID, blsSignature, err := getSecrets(path)
 		if err != nil {
 			return nil, err
 		}
@@ -146,6 +146,7 @@ func ReadValidatorsByPrefix(dir, prefix string) ([]*polybft.Validator, error) {
 			Address:       types.Address(account.Ecdsa.Address()),
 			BlsPrivateKey: account.Bls,
 			BlsKey:        hex.EncodeToString(account.Bls.PublicKey().Marshal()),
+			BlsSignature:  blsSignature,
 			NodeID:        nodeID,
 		}
 	}
@@ -153,7 +154,7 @@ func ReadValidatorsByPrefix(dir, prefix string) ([]*polybft.Validator, error) {
 	return validators, nil
 }
 
-func getSecrets(directory string) (*wallet.Account, string, error) {
+func getSecrets(directory string) (*wallet.Account, string, string, error) {
 	baseConfig := &secrets.SecretsManagerParams{
 		Logger: hclog.NewNullLogger(),
 		Extra: map[string]interface{}{
@@ -163,18 +164,20 @@ func getSecrets(directory string) (*wallet.Account, string, error) {
 
 	localManager, err := local.SecretsManagerFactory(nil, baseConfig)
 	if err != nil {
-		return nil, "", fmt.Errorf("unable to instantiate local secrets manager, %w", err)
+		return nil, "", "", fmt.Errorf("unable to instantiate local secrets manager, %w", err)
 	}
 
 	nodeID, err := helper.LoadNodeID(localManager)
 	if err != nil {
-		return nil, "", err
+		return nil, "", "", err
 	}
 
 	account, err := wallet.NewAccountFromSecret(localManager)
 	if err != nil {
-		return nil, "", err
+		return nil, "", "", err
 	}
 
-	return account, nodeID, nil
+	blsSignature, err := helper.LoadBLSSignature(localManager)
+
+	return account, nodeID, blsSignature, err
 }
--- a/command/polybftmanifest/manifest_init.go
+++ b/command/polybftmanifest/manifest_init.go
@@ -21,6 +21,7 @@ const (
 	validatorsFlag        = "validators"
 	validatorsPathFlag    = "validators-path"
 	validatorsPrefixFlag  = "validators-prefix"
+	chainIDFlag           = "chain-id"
 
 	defaultValidatorPrefixPath = "test-chain-"
 	defaultManifestPath        = "./manifest.json"
@@ -88,6 +89,13 @@ func setFlags(cmd *cobra.Command) {
 		"the amount which will be pre-mined to all the validators",
 	)
 
+	cmd.Flags().Int64Var(
+		&params.chainID,
+		chainIDFlag,
+		command.DefaultChainID,
+		"the ID of the chain",
+	)
+
 	cmd.MarkFlagsMutuallyExclusive(validatorsFlag, validatorsPathFlag)
 	cmd.MarkFlagsMutuallyExclusive(validatorsFlag, validatorsPrefixFlag)
 }
@@ -103,7 +111,7 @@ func runCommand(cmd *cobra.Command, _ []string) {
 		return
 	}
 
-	manifest := &polybft.Manifest{GenesisValidators: validators}
+	manifest := &polybft.Manifest{GenesisValidators: validators, ChainID: params.chainID}
 	if err = manifest.Save(params.manifestPath); err != nil {
 		outputter.SetError(fmt.Errorf("failed to save manifest file '%s': %w", params.manifestPath, err))
 
@@ -119,6 +127,7 @@ type manifestInitParams struct {
 	validatorsPrefixPath string
 	premineValidators    string
 	validators           []string
+	chainID              int64
 }
 
 func (p *manifestInitParams) validateFlags() error {
--- a/command/polybftmanifest/manifest_init.go
+++ b/command/polybftmanifest/manifest_init.go
@@ -86,7 +86,7 @@ func setFlags(cmd *cobra.Command) {
 		&params.validators,
 		validatorsFlag,
 		[]string{},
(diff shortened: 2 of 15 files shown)


command/polybftsecrets/params.go
```diff
@@ -196,13 +210,32 @@ func (ip *initParams) getResult(secretsManager secrets.SecretsManager) (command.
 		res.Address = types.Address(account.Ecdsa.Address())
 		res.BLSPubkey = hex.EncodeToString(account.Bls.PublicKey().Marshal())
 
+		s, err := polybft.MakeKOSKSignature(account.Bls, types.Address(account.Ecdsa.Address()), 10, bls.DomainValidatorSet)
```
vcastellm: Chain id?

secrets/helper/helper.go
```diff
@@ -139,6 +143,39 @@ func InitNetworkingPrivateKey(secretsManager secrets.SecretsManager) (libp2pCryp
...
+
+	// Write the signature to the secrets manager storage
+	if setErr := secretsManager.SetSecret(
+		secrets.ValidatorBLSSignature,
+		[]byte(hex.EncodeToString(sb)),
+	); setErr != nil {
+		return nil, setErr
+	}
+
+	return sb, err
```
Stefan-Ethernal:
```suggestion
	return sb, nil
```

command/genesis/utils.go
```diff
@@ -163,18 +164,20 @@ func getSecrets(directory string) (*wallet.Account, string, error) {
...
 	account, err := wallet.NewAccountFromSecret(localManager)
 	if err != nil {
-		return nil, "", err
+		return nil, "", "", err
 	}
 
-	return account, nodeID, nil
+	blsSignature, err := helper.LoadBLSSignature(localManager)
+
+	return account, nodeID, blsSignature, err
```
Stefan-Ethernal, line 182: Nitpick:
it is worth considering introducing an abstraction for `account+nodeID+blsSignature` (if not even putting it all together into `Account` object itself). WDYT?

secrets/helper/helper.go
```diff
@@ -139,6 +143,39 @@ func InitNetworkingPrivateKey(secretsManager secrets.SecretsManager) (libp2pCryp
...
+		return nil, err
+	}
+
+	// Write the signature to the secrets manager storage
+	if setErr := secretsManager.SetSecret(
+		secrets.ValidatorBLSSignature,
+		[]byte(hex.EncodeToString(sb)),
+	); setErr != nil {
+		return nil, setErr
+	}
```
Stefan-Ethernal: Optional

```suggestion
	if err := secretsManager.SetSecret(
		secrets.ValidatorBLSSignature,
		[]byte(hex.EncodeToString(sb)),
	); err != nil {
		return nil, err
	}
```

Stefan-Ethernal:
Leaving some minor suggestions.

However, I'm still bothered with following:
- tighting BLS signature creation to `polybft-secrets` command limits us to use it only in testing environments as BLS signature is populated only when creating account(s). 
Imagine we have already a private key provided, which needs to sign agreed format and create KOSK BLS signature. Still don't see we are able to accomplish it? Therefore, although I'm all for it to have as minimal commands as possible, I think we would need additional command here, which would acquire private key from secrets manager and just call `InitValidatorBLSSignature`. WDYT?
- for some reason bridge related e2e tests are failing. Not sure why. I've randomly checked a couple of logs of failing tests, but haven't found anything suspicious there.

secrets/helper/helper.go
```diff
@@ -139,6 +143,39 @@ func InitNetworkingPrivateKey(secretsManager secrets.SecretsManager) (libp2pCryp
...
+
+	sb, err := s.Marshal()
+	if err != nil {
+		return nil, err
+	}
+
+	// Write the signature to the secrets manager storage
+	if setErr := secretsManager.SetSecret(
+		secrets.ValidatorBLSSignature,
+		[]byte(hex.EncodeToString(sb)),
```
begmaroman, line 171: Can't we just same `sb` instead of hex encoding it to string and then converting back to bytes?

secrets/helper/helper.go
```diff
@@ -238,3 +286,17 @@ func InitCloudSecretsManager(secretsConfig *secrets.SecretsManagerConfig) (secre
 
 	return secretsManager, nil
 }
+
+// MakeKOSKSignature creates KOSK signature which prevents rogue attack
+func MakeKOSKSignature(
+	privateKey *bls.PrivateKey, address types.Address, chainID int64, domain []byte) (*bls.Signature, error) {
+	message, err := abi.Encode(
+		[]interface{}{address, big.NewInt(chainID)},
+		abi.MustNewType("tuple(address, uint256)"))
```
begmaroman, line 295: This type could be moved to per-package variables instead of creating it every time. It is ok if we call this function only once.

vcastellm:
> * tighting BLS signature creation to `polybft-secrets` command limits us to use it only in testing environments as BLS signature is populated only when creating account(s).
>   Imagine we have already a private key provided, which needs to sign agreed format and create KOSK BLS signature. Still don't see we are able to accomplish it? Therefore, although I'm all for it to have as minimal commands as possible, I think we would need additional command here, which would acquire private key from secrets manager and just call `InitValidatorBLSSignature`. WDYT?

This is why I commented this in the RFC:

My recommendation here is that we adapt the secrets command in order to generate missing parts when others already exits. Currently it simply fails but we can change that.

Stefan-Ethernal:
> My recommendation here is that we adapt the secrets command in order to generate missing parts when others already exits. Currently it simply fails but we can change that.

Ok, but I feel like this is the PR that should change it as well or in the very near future. Do you agree?

vcastellm:
> Ok, but I feel like this is the PR that should change it as well or in the very near future. Do you agree?

Yes, going to do it

Stefan-Ethernal (approved):
Generally looks good. Leaving some minor thoughts to consider is it worth changing...

command/genesis/utils.go
vcastellm: I was thinking on it, and actually it can provide a some clarity to some parts of the code, but I wanted to keep it separated because not knowing what side effects we can find, do you think is worth it?

secrets/helper/helper.go
igorcrevar: I think secret manager can not handle any byte slice, i think it should be human readable string (converted to bytes)

secrets/helper/helper.go
igorcrevar: It is called only when someone wants to generate KOSK. Later on it will be kept in secrets (which kind of bothers me because KOSK is not secret at all :) )

vcastellm: I've my doubts here, because:
1. we're using it only here
2. Secrets init is just used once at setup time.

I prefer not to pollute the package vars with it considering 1 and 2

secrets/helper/helper.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1233/commits/042fb94759287b200e2a68211f691410f64b191d

secrets/helper/helper.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1233/commits/c0ad2b574fab129797bb7a09d8dc8d63eadae77b

secrets/helper/helper.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1233/commits/c0ad2b574fab129797bb7a09d8dc8d63eadae77b

command/genesis/utils.go
vcastellm: In any case, let's do this in a follow-up PR in case we want

secrets/helper/helper.go
vcastellm: Reverted, I forgot that I tried it and the marshaling is binary so it ends up garbled chars in json, we need to encode to string then pass bytes.

command/genesis/utils.go
Stefan-Ethernal: Yes, agreed. This was just a nice-to-have proposal.

command/polybftmanifest/manifest_init.go
```diff
@@ -75,18 +74,11 @@ func setFlags(cmd *cobra.Command) {
 		"folder prefix names for polybft validator keys",
 	)
 
-	cmd.Flags().Int64Var(
-		&params.chainID,
-		chainIDFlag,
-		command.DefaultChainID,
-		"the ID of the chain",
-	)
```
Stefan-Ethernal, line 83: This is the problem why e2e test is failing:
we are providing different chain id when initializing CheckpointManager smart contract (https://github.com/0xPolygon/core-contracts/blob/dev/contracts/root/CheckpointManager.sol#L42) and when calculating checkpoint hash (which takes chain id as well into account: https://github.com/0xPolygon/polygon-edge/blob/develop/consensus/polybft/blockchain_wrapper.go#L183). Later one is read from Genesis configuration and it defaults to 100, whereas this one in manifest file is left to be 0 (since it is not populated and it is defaulted to 0). Therefore we have mismatch between checkpoint hash calculated on the smart contract vs the one calculated and signed by the client.

igorcrevar (approved):

PR #1234, opened 2023-02-17
TOB-EDGE-13 Merkle Tree bugs

goran-ethernal:
# Description

Addressed the comments from the TOB audit regarding `merkle_tree.go` that we use as a Merkle Tree implementation for both `state syncs` and `exit events`.

What is done through this PR:
1. Introduced concepts of left and right node.
2. Removed padding of the tree with zero hash values. Now we duplicate the last node if the tree has uneven number of leaf nodes.
3. Introduced additional validations of leaf index and proof length when verifying proof.
4. Added additional Unit Tests.
5. Moved the merkle tree implementation and tests to a separate package.

### Merkle Tree
A Merkle tree, also known as a hash tree, is a data structure used in cryptography and computer science to efficiently verify the integrity and authenticity of large datasets.

A Merkle tree is constructed by recursively hashing pairs of data until a single hash value, known as the root hash or Merkle root, is produced. Each leaf of the tree represents a piece of data, such as a transaction in a blockchain, while the internal nodes represent the hashes of their children nodes. The root hash of the tree is then used to represent the entire dataset.

A Merkle tree example:

image

Each leaf node (`L0 - L7`) contains a data value, and each intermediate node (`h1 - h7`) contains the hash of its two child nodes. The root node (`ROOT`) contains the hash of the entire tree. To prove the inclusion of a leaf node in the tree, a Merkle proof would consist of the leaf node, along with the hash values of its sibling nodes on the path from the leaf to the root.

So the proof for leaf node `L0` will look like this:
`H(Hash(L1), Hash(H2), Hash(H7))`

# Changes include

# Checklist

## Testing

### DIFF
--- a/merkle-tree/merkle_tree.go
+++ b/merkle-tree/merkle_tree.go
@@ -1,4 +1,4 @@
-package polybft
+package merkle
 
 import (
 	"bytes"
@@ -14,12 +14,12 @@ import (
 
 // A Merkle tree example:
 //
-//             ROOT
-//            /    \
-//          h3      h7
-//         /  \    /  \
-//       h1   h2  h5   h6
-//      / \   / \ / \   / \
+//              ROOT
+//            /       \
+//          h3         h7
+//         /  \      /    \
+//       h1   h2    h5    h6
+//      / \   / \   / \   / \
 //     L0 L1 L2 L3 L4 L5 L6 L7
 //
 // Each leaf node (L0 - L7) contains a data value, and each intermediate node
@@ -33,8 +33,6 @@ import (
 
 var (
 	errLeafNotFound = errors.New("leaf not found")
-	leftNodeDomain  = []byte("leftNodeDomain")
-	rightNodeDomain = []byte("rightNodeDomain")
 )
 
 // MerkleNode represents a single node in merkle tree
@@ -56,9 +54,7 @@ func newMerkleNode(left, right *MerkleNode, data []byte, hasher hash.Hash) *Merk
 		dataToHash = data
 	} else {
 		// it's an inner node
-		dataToHash = append(leftNodeDomain, left.hash...)
-		dataToHash = append(dataToHash, rightNodeDomain...)
-		dataToHash = append(dataToHash, right.hash...)
+		dataToHash = append(left.hash, right.hash...)
 	}
 
 	hasher.Reset()
@@ -214,14 +210,10 @@ func getProofHash(index uint64, leaf []byte, proof []types.Hash, hasher hash.Has
 		hasher.Reset()
 
 		if index%2 == 0 {
-			hasher.Write(leftNodeDomain)
 			hasher.Write(computedHash)
-			hasher.Write(rightNodeDomain)
 			hasher.Write(proof[i].Bytes())
 		} else {
-			hasher.Write(leftNodeDomain)
 			hasher.Write(proof[i].Bytes())
-			hasher.Write(rightNodeDomain)
 			hasher.Write(computedHash)
 		}
 
--- a/merkle-tree/merkle_tree.go
+++ b/merkle-tree/merkle_tree.go
@@ -41,12 +41,12 @@ type MerkleNode struct {
 	right  *MerkleNode
 	parent *MerkleNode
 	data   []byte
-	hash   []byte
+	hash   types.Hash
 }
 
 // newMerkleNode creates a new merkle node
 func newMerkleNode(left, right *MerkleNode, data []byte, hasher hash.Hash) *MerkleNode {
-	node := MerkleNode{left: left, right: right, data: data}
+	node := &MerkleNode{left: left, right: right, data: data}
 
 	var dataToHash []byte
 	if left == nil && right == nil {
@@ -54,14 +54,14 @@ func newMerkleNode(left, right *MerkleNode, data []byte, hasher hash.Hash) *Merk
 		dataToHash = data
 	} else {
 		// it's an inner node
-		dataToHash = append(left.hash, right.hash...)
+		dataToHash = append(left.hash.Bytes(), right.hash.Bytes()...)
 	}
 
 	hasher.Reset()
 	hasher.Write(dataToHash)
-	node.hash = hasher.Sum(nil)
+	node.hash = types.BytesToHash(hasher.Sum(nil))
 
-	return &node
+	return node
 }
 
 // MerkleTree is the structure for the Merkle tree.
@@ -132,7 +132,7 @@ func (t *MerkleTree) LeafIndex(leaf []byte) (uint64, error) {
 
 // Hash is the Merkle Tree root hash
 func (t *MerkleTree) Hash() types.Hash {
-	return types.BytesToHash(t.rootNode.hash)
+	return t.rootNode.hash
 }
 
 // String implements the stringer interface
@@ -155,11 +155,11 @@ func (t *MerkleTree) GenerateProof(leaf []byte) ([]types.Hash, error) {
 	}
 
 	node := leafNode
-	for !bytes.Equal(node.hash, t.rootNode.hash) {
-		if bytes.Equal(node.parent.left.hash, node.hash) {
-			proof = append(proof, types.BytesToHash(node.parent.right.hash))
+	for !bytes.Equal(node.hash.Bytes(), t.rootNode.hash.Bytes()) {
+		if bytes.Equal(node.parent.left.hash.Bytes(), node.hash.Bytes()) {
+			proof = append(proof, node.parent.right.hash)
 		} else {
-			proof = append(proof, types.BytesToHash(node.parent.left.hash))
+			proof = append(proof, node.parent.left.hash)
 		}
 
 		node = node.parent
@@ -202,6 +202,7 @@ func VerifyProofUsing(index uint64, leaf []byte, proof []types.Hash, root types.
 	return nil
 }
 
+// getProofHash uses the leaf and proof to recalculate the root hash of a tree that contains given leaf
 func getProofHash(index uint64, leaf []byte, proof []types.Hash, hasher hash.Hash) []byte {
 	hasher.Write(leaf)
 	computedHash := hasher.Sum(nil)
--- a/merkle-tree/merkle_tree_benchmark_test.go
+++ b/merkle-tree/merkle_tree_benchmark_test.go
@@ -7,43 +7,83 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Benchmark_MerkleTreeCreation(b *testing.B) {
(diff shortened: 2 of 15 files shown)


Nemanja0x (approved):

vcastellm (commented):
Can we separate this to a package and maintain it separately? Can be useful for other people

igorcrevar (approved):

Nemanja0x (approved):

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	} else {
+		// it's an inner node
+		dataToHash = append(left.hash, right.hash...)
+	}
+
+	hasher.Reset()
+	hasher.Write(dataToHash)
+	node.hash = hasher.Sum(nil)
+
+	return &node
```
Stefan-Ethernal:
```suggestion
	node := &MerkleNode{left: left, right: right, data: data}

	var dataToHash []byte
	if left == nil && right == nil {
		// it's a leaf node
		dataToHash = data
	} else {
		// it's an inner node
		dataToHash = append(left.hash, right.hash...)
	}

	hasher.Reset()
	hasher.Write(dataToHash)
	node.hash = hasher.Sum(nil)

	return node
```

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	// rootNode is the root node of the tree
+	rootNode *MerkleNode
+	// leaf nodes is the list of leaf nodes of the tree (the lowest level of the tree)
+	leafNodes []*MerkleNode
+}
+
+// NewMerkleTree creates a new Merkle tree from the provided data and using the default hashing (Keccak256).
+func NewMerkleTree(data [][]byte) (*MerkleTree, error) {
+	return NewMerkleTreeWithHashing(data, crypto.NewKeccakState())
+}
```
Stefan-Ethernal, line 80: Did we say we would get rid off this function in favor of `NewMerkleTreeWithHashing` so that we don't depend on crypto package anymore? Is it the time to do it now or we are going to consider it when transferring logic into a separate library?

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,225 @@
...
+
+	computedHash := getProofHash(index, leaf, proof, hasher)
+	if !bytes.Equal(root.Bytes(), computedHash) {
+		return fmt.Errorf("leaf with index %v, not a member of merkle tree. Merkle root hash: %v", index, root)
+	}
+
+	return nil
+}
+
+func getProofHash(index uint64, leaf []byte, proof []types.Hash, hasher hash.Hash) []byte {
```
Stefan-Ethernal, line 206: Consider adding a comment.

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,225 @@
...
+
+		node = node.parent
+	}
+
+	return proof, nil
+}
+
+// findLeafNode finds the given leaf node that corresponds to given leaf data
+func (t *MerkleTree) findLeafNode(leaf []byte) *MerkleNode {
+	for _, leafNode := range t.leafNodes {
```
Stefan-Ethernal, line 173: Does it make sense introducing a `map[types.Hash]*MerkleNode` and rely on it when searching for `MerkleNode` instead of looping through all leaf nodes. WDYT?

So when adding `MerkleNode` to the map, we would just hash data using provided hasher and put it into the map. And here, when looking up for a node, we would again hash provided leaf data and just access to the map and return a `MerkleNode` if found.

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	errLeafNotFound = errors.New("leaf not found")
+)
+
+// MerkleNode represents a single node in merkle tree
+type MerkleNode struct {
+	left   *MerkleNode
+	right  *MerkleNode
+	parent *MerkleNode
+	data   []byte
+	hash   []byte
```
Stefan-Ethernal: Represent it as `types.Hash` perhaps? Eventually if and when this becomes a library, we would have some Hash type alias defined there and use that one instead. WDYT?

merkle-tree/merkle_tree.go
goran-ethernal: I prefer to not complicate this. If this is going to be a library, this can be used for anything, theoretically, you can have leaf nodes that contain the same data (meaning their hashes will be the same), so the map won't work in that case.

merkle-tree/merkle_tree.go
goran-ethernal: Let's do it when we transfer the logic in the separate library. Since it is in the polygon edge client codebase, this function makes sense.

merkle-tree/merkle_tree.go
Stefan-Ethernal: > theoretically, you can have leaf nodes that contain the same data (meaning their hashes will be the same), so the map won't work in that case.

If leaf data are going to be the same, then the outcome is the same. But it is faster to look up in the map for large amount of leaf nodes.

goran-ethernal: Ok, I see your point, My reasoning is that it if we use a map, then `GetLeafIndex` function will not work. If we are to keep leafs in both slice and map, then there should be a reason for it (generating proof for large trees is to slow, for example). I'll write some benchmark tests, and we will see if generating proof is slow in large trees. If it is, using a map makes sense, if it's reasonable, then I would say lets leave it as it is.

merkle-tree/merkle_tree.go
goran-ethernal: Why not

merkle-tree/merkle_tree.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1234/commits/9ce1f4656d13a5c90ec4a241a980c7b0fd433658

merkle-tree/merkle_tree.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1234/commits/9ce1f4656d13a5c90ec4a241a980c7b0fd433658

merkle-tree/merkle_tree.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1234/commits/9ce1f4656d13a5c90ec4a241a980c7b0fd433658

merkle-tree/merkle_tree.go
goran-ethernal: I've pushed the Benchmark tests, since they are a nice thing to have. I've added the map as you suggested and didn't see any improvements, so let's leave it as is.

merkle-tree/merkle_tree.go
```diff
@@ -155,11 +155,11 @@ func (t *MerkleTree) GenerateProof(leaf []byte) ([]types.Hash, error) {
 	}
 
 	node := leafNode
-	for !bytes.Equal(node.hash, t.rootNode.hash) {
-		if bytes.Equal(node.parent.left.hash, node.hash) {
-			proof = append(proof, types.BytesToHash(node.parent.right.hash))
+	for !bytes.Equal(node.hash.Bytes(), t.rootNode.hash.Bytes()) {
```
Stefan-Ethernal, line 158: You could compare hashes with == as well.

merkle-tree/merkle_tree_benchmark_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"testing"
+
+	"github.com/0xPolygon/polygon-edge/helper/common"
+	"github.com/stretchr/testify/require"
+)
+
+func Benchmark_MerkleTreeCreation(b *testing.B) {
+	const numOfLeaves = 10_000
```
Stefan-Ethernal: You could have created different test cases and variate number of leaves. That way we could have better understanding how algorithm performs, since we would have more results to compare against and therefore we could conclude what is are the time and space complexity.

You can take a look to this as a reference:
https://github.com/maticnetwork/bor/pull/262/files#diff-a424e26a1e19cffe181aa20565feb4f3dc617fc9e6e54192351555517f5c92f0R36-R73

merkle-tree/merkle_tree_benchmark_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for i := 0; i < b.N; i++ {
+		tree.GenerateProof(data[0])
+	}
+}
+
+func Benchmark_VerifyProof(b *testing.B) {
+	const (
+		numOfLeaves = 10_000
+		leafIndex   = 0
```
Stefan-Ethernal: You could vary `leafIndex`? For example 0, somewhere in the middle and very end of `numOfLeaves`. Does it make sense?

merkle-tree/merkle_tree.go
```diff
@@ -0,0 +1,226 @@
...
+func (t *MerkleTree) GenerateProof(leaf []byte) ([]types.Hash, error) {
+	proof := []types.Hash{}
+
+	leafNode := t.findLeafNode(leaf)
+	if leafNode == nil {
+		return nil, fmt.Errorf("given data not in merkle tree")
+	}
+
+	node := leafNode
+	for !bytes.Equal(node.hash.Bytes(), t.rootNode.hash.Bytes()) {
```
Stefan-Ethernal, line 158: If I'm correct, you could compare it this way as well (but double check before applying suggestion) :slightly_smiling_face: 
```suggestion
	for node.hash != t.rootNode.hash {
```

merkle-tree/merkle_tree_benchmark_test.go
goran-ethernal: Isn't that fuzz testing?

merkle-tree/merkle_tree_benchmark_test.go
goran-ethernal: There is no need to randomize leaf index. Verifying and Generating proof should produce the same performance results since the tree is balanced, and you always start generating and verifying proof from the lowest level of the tree.

merkle-tree/merkle_tree_benchmark_test.go
Stefan-Ethernal: No, it isn't. I'm not saying that you should randomize tree size, but just have more than a single test case so that you are able to see how the algorithm behaves for different data set sizes.

goran-ethernal: Ok. I'll add more cases for different tree sizes.

Stefan-Ethernal (approved):

vcastellm (approved):
LGTM

PR #1238, opened 2023-02-20
Fix eth_getCode for non-empty accounts without code

tkporter:
# Description

Ran into an issue where `eth_getCode` would give an error for addresses that are "non-empty" (meaning their nonce, balance, or storage hash is non-zero), but they don't have any code.

Example of the error I'd see for an account that had a balance but no code:
```
$ cast code 0xBBA5a06C75b62dd43E7ab94a6C13696b6010B546 --rpc-url $RPC_URL
Error:
(code: -32600, message: unable to fetch code, data: None)
```

This is my understanding of the bug:

When an `eth_getCode` RPC comes in, it runs [this bit of code](https://github.com/0xPolygon/polygon-edge/blob/develop/jsonrpc/eth_endpoint.go#L684-L702) in `jsonrpc/eth_endpoint.go`:
```go
func (e *Eth) GetCode(address types.Address, filter BlockNumberOrHash) (interface{}, error) {
	header, err := GetHeaderFromBlockNumberOrHash(filter, e.store)
	if err != nil {
		return nil, err
	}

	emptySlice := []byte{}
	code, err := e.store.GetCode(header.StateRoot, address)

		// If the account doesn't exist / is not initialized yet,
		// return the default value
		return "0x", nil
	} else if err != nil {
		return argBytesPtr(emptySlice), err
	}

	return argBytesPtr(code), nil
}
```

When an account is empty, meaning it has all zero/default values for the nonce, balance, account storage root (as the `EmptyRootHash`), and code hash, `ErrStateNotFound` is returned by `e.store.GetCode`. So this results in these accounts correctly returning the code of `0x`.

However in a case where there's an account that is not empty but has no code, the db is called to get the code for the given code hash. It seems that the db doesn't have an entry for the empty code hash, so it returns an error.

I decided to add a check before actually making a db read, where if the empty code hash is being requested, just an empty byte slice is returned. I'm very open to this change being more appropriately made somewhere else!

# Changes include

# Breaking changes

I'm unsure if the change to the `State`'s `GetCode` function is totally backward compatible

# Checklist

## Testing

### Manual tests

Just ran it locally and made eth_getCode RPCs to the node


tkporter:

vcastellm (commented):
LGTM but can you add a test for this case?

Stefan-Ethernal:
recheck

Stefan-Ethernal (approved):

Stefan-Ethernal:
Continuation of this PR is done through #1474 (namely implementing necessary test(s)).

PR #1239, opened 2023-02-20
Add dynamic tx fees to jsonrpc tx object

begmaroman:
# Description

Added dynamic tx fees fields to the jsonrpc tx object 

# Changes include

# Checklist

## Testing


goran-ethernal (approved):

jsonrpc/types.go
```diff
@@ -17,6 +17,8 @@ type transactionOrHash interface {
 type transaction struct {
 	Nonce       argUint64      `json:"nonce"`
 	GasPrice    argBig         `json:"gasPrice"`
+	GasTipCap   *argBig        `json:"gasTipCap,omitempty"`
+	GasFeeCap   *argBig        `json:"gasFeeCap,omitempty"`
```
Stefan-Ethernal, line 21: Are these fields pointers so that they can accept `nil` values (which is the case for Legacy transactions)?

Stefan-Ethernal (approved):

jsonrpc/types.go
begmaroman: Yea, there is a test to check it because they are optional. Addressed there: https://github.com/0xPolygon/polygon-edge/pull/1210#discussion_r1109764684

PR #1241, opened 2023-02-20
Added dynamic fees validation to tx pool and state executor 

begmaroman:
# Description

Added dynamic fees validation to tx pool and state executor. 
The logic of updating base fee in tx pool will be implemented in a separate PR. 

# Changes include

# Checklist

## Testing


state/executor.go
```diff
@@ -414,6 +423,38 @@ func (t *Transition) nonceCheck(msg *types.Transaction) error {
 	return nil
 }
 
+// checkDynamicFees checks correctness of the EIP-1559 feature-related fields.
+// Basically, makes sure gas tip cap and gas fee cap are good.
+func (t *Transition) checkDynamicFees(msg *types.Transaction) error {
```
goran-ethernal, line 428: We could use some Unit Tests for this function if there aren't any.

goran-ethernal (approved):
Except for the minor comment.

state/executor.go
begmaroman: Done

PR #1244, opened 2023-02-22
TOB-EDGE-22: Unmarshalling Transaction in RLP format can panic

stana-miric:
# Description

In function unmarshalRLPFrom, in the beginning of the fn there is a check if elements length is less than 9 which is considered invalid. Afterwards, in the case of state transaction, the code accessing 10th element (calls v.Get(9) function) without checking the length first. The check if this element exists is added before accessing it. The Get function of the fastrlp library itself doesn't check index in a best way since it returns if i > len(v.a)  instead of if i >= len(v.a) but this added protection solves the originally reported problem  in the unmarshalRLPFrom function and prevent Get function accessing the non existent element. 

# Changes include

# Checklist

## Testing

### DIFF
--- a/types/rlp_encoding_test.go
+++ b/types/rlp_encoding_test.go
@@ -167,15 +167,15 @@ func TestRLPMarshall_And_Unmarshall_TypedTransaction(t *testing.T) {
 }
 
 func TestRLPMarshall_Unmarshall_Missing_Data(t *testing.T) {
+	t.Parallel()
+
 	txTypes := []TxType{
 		StateTx,
 		LegacyTx,
 	}
 
-	arena := fastrlp.DefaultArenaPool.Get()
-	defer fastrlp.DefaultArenaPool.Put(arena)
-
 	for _, txType := range txTypes {
+		txType := txType
 		testTable := []struct {
 			name          string
 			expectedErr   bool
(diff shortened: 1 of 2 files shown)


goran-ethernal (approved):

igorcrevar (approved):

types/rlp_encoding_test.go
```diff
@@ -162,6 +162,67 @@ func TestRLPMarshall_And_Unmarshall_TypedTransaction(t *testing.T) {
 
 		unmarshalledTx.ComputeHash()
 		assert.Equal(t, originalTx.Type, unmarshalledTx.Type)
+		assert.Equal(t, originalTx.Hash, unmarshalledTx.Hash)
+	}
+}
+
+func TestRLPMarshall_Unmarshall_Missing_Data(t *testing.T) {
+	txTypes := []TxType{
```
Stefan-Ethernal: Consider adding parallel execution?

```suggestion
func TestRLPMarshall_Unmarshall_Missing_Data(t *testing.T) {
        t.Parallel()
	txTypes := []TxType{
```

types/rlp_encoding_test.go
```diff
@@ -162,6 +162,67 @@ func TestRLPMarshall_And_Unmarshall_TypedTransaction(t *testing.T) {
...
+				},
+				fromAddrSet: false,
+			},
+			{
+				name:          "Address set for state tx only",
+				expectedErr:   false,
+				ommitedValues: map[string]bool{},
+				fromAddrSet:   txType == StateTx,
+			},
+		}
```
Stefan-Ethernal: Seems like you could have moved the definition of `testTable` above the for loop?

types/rlp_encoding_test.go
```diff
@@ -162,6 +162,67 @@ func TestRLPMarshall_And_Unmarshall_TypedTransaction(t *testing.T) {
...
+
+		for _, tt := range testTable {
```
Stefan-Ethernal, line 209: Consider using `t.Run` to run subtests and include parallel execution? In that case, don't forget to capture `tt` to a new local variable and analyze whether it is better for each test to use its own instance of `arena` (not sure if it is thread-safe).

Stefan-Ethernal (approved):
LGTM, leaving minor suggestions for test.

types/rlp_encoding_test.go
stana-miric: txType is used in "Address set for state tx only" test case when setting fromAddrSet. Other option is to include txType in each test case but that means that we would have duplicated number of cases.

types/rlp_encoding_test.go
stana-miric: fixed 673cce8

types/rlp_encoding_test.go
stana-miric: fixed 673cce8

PR #1249, opened 2023-02-23
Update tx pool ordering logic according to EIP-1559

begmaroman:
# Description

Update tx pool ordering logic according to EIP-1559 spec. 

# Changes include

# Checklist

## Testing


types/encoding.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/0xPolygon/polygon-edge/helper/hex"
 )
 
+// ParseUint64orHex parses the given string as uint64 in hex
+// It should go to the common package from the logical perspective
+// as well as avoiding cycle imports.
+// DEPRECATED. Use common.ParseUint64orHex.
 func ParseUint64orHex(val *string) (uint64, error) {
```
Stefan-Ethernal, line 16: I hope we are about to remove this function through some of the follow-up PRs.

helper/common/common.go
```diff
@@ -79,6 +79,22 @@ func ConvertUnmarshalledUint(x interface{}) (uint64, error) {
 	}
 }
 
+func ParseUint64orHex(val *string) (uint64, error) {
```
Stefan-Ethernal, line 84: Add a comment?

Stefan-Ethernal (approved):
Looks good.

Slightly off topic question, however EIP-1559 related one:
should we check if London hard-fork is enabled e.g. in `txpool.validateTx` and if it is not and we found `DynamicFeeTx` to return an error?
Also worth double checking is there any other place where we should explicitly check is London hard-fork enabled.

types/encoding.go
begmaroman: Yeah, we will remove it

begmaroman:
@Stefan-Ethernal I think it still should be possible to send legacy txs even if the hardfork is enabled.

Stefan-Ethernal:
@begmaroman it is not a question about sending legacy transactions, but dynamic fee transactions if EIP-1559 hardfork is not enabled. My point is that we should have a validation in tx pool if EIP-1559 hardfork is not enabled. Makes sense?

begmaroman:
@Stefan-Ethernal good point. Would you be able to add this comment to the parent PR since this is a little bit out of the current scope?

helper/common/common.go
begmaroman: Done

types/transaction.go
```diff
@@ -150,7 +151,13 @@ func (t *Transaction) Cost() *big.Int {
 //   - or calculate a value with formula: min(gasFeeCap, gasTipCap * baseFee);
 func (t *Transaction) GetGasPrice(baseFee uint64) *big.Int {
 	if (t.GasPrice != nil && t.GasPrice.BitLen() > 0) || baseFee == 0 {
-		return new(big.Int).Set(t.GasPrice)
```
igorcrevar, line 153: Is this same as
```
} else if baseFee == 0  {
    return new(big.Int)
}
```
? somehow it is easier for me if there is one less if inside first if block

begmaroman: Done

helper/common/common.go
```diff
@@ -79,6 +79,24 @@ func ConvertUnmarshalledUint(x interface{}) (uint64, error) {
 	}
 }
 
+// ParseUint64orHex parses the given uint64 hex string into the number.
+// It can parse the string with 0x prefix as well.
+func ParseUint64orHex(val *string) (uint64, error) {
```
goran-ethernal, line 84: Don't we have the same function in `types` package?

goran-ethernal (approved):
Except for the minor comment LGTM

helper/common/common.go
Stefan-Ethernal: Yes, we do have it, but as Roman explained, the aim is to remove that one in some of the following PRs and use this one instead.
https://github.com/0xPolygon/polygon-edge/pull/1249/files#diff-c99d168ac05d1f4057606db9fa5ac73a6c7f5d00ef27273a912cb6cb174a8a45R12-R15

goran-ethernal: Ah, OK, didn't see that he already explained that somewhere.

PR #1250, opened 2023-02-24
Commitment vote signature verification

stana-miric:
# Description

The signature of the vote for a commitment was not verified against the sender address, this PR includes this fix.

In the audit report, where this issue is originally reported, there is a concern if EpochNumber (included in a TransportMessage and used in a key when saving votes) can be used for replay attack. This is not possible because:
- if the malicious user replays the signed message with the same EpochNumber, the existing vote will be overwritten with the same data
- if the malicious user replays the signed message setting different EpochNumber, the vote would not be taken into consideration. Before this fix, votes with equal or greater EpochNumber than the current epoch are taken into consideration, but irrelevant votes with the higher epoch number would not be saved anyway since the bucket for that epoch is not yet created. The fix now allows only votes for the current epoch, since there is no point of processing votes for higher epoch
- even without EpochNumber check mentioned above, if the replayed vote is saved for the higher epoch it would never participate in a tally since the commitment with the hash from the vote is deleted from the pending commitments upon is submitted for the current epoch or the epoch change happens

This PR also includes a minor change of renaming 'NodeID' of TransportMessage to 'From', since it represents the address of the message signer and it's not the node identifier. 

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -39,8 +39,8 @@ type TransportMessage struct {
 	Hash []byte
 	// Message signature
 	Signature []byte
-	// Node identifier
-	NodeID string
+	// Address of the message signer
+	From string
 	// Number of epoch
 	EpochNumber uint64
 }
--- a/consensus/polybft/state.go
+++ b/consensus/polybft/state.go
@@ -39,7 +39,7 @@ type TransportMessage struct {
 	Hash []byte
 	// Message signature
 	Signature []byte
-	// Address of the message signer
+	// From is the address of the message signer
 	From string
 	// Number of epoch
 	EpochNumber uint64
--- a/consensus/polybft/state_sync_manager.go
+++ b/consensus/polybft/state_sync_manager.go
@@ -210,7 +210,7 @@ func (s *stateSyncManager) verifyVoteSignature(valSet ValidatorSet, signer types
 
 	unmarshaledSignature, err := bls.UnmarshalSignature(signature)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal signature: %w", err)
+		return fmt.Errorf("failed to unmarshal signature from signer %s, %w", signer.String(), err)
 	}
 
 	if !unmarshaledSignature.Verify(validator.BlsKey, hash, bls.DomainCheckpointManager) {
--- a/consensus/polybft/state_sync_manager.go
+++ b/consensus/polybft/state_sync_manager.go
@@ -172,7 +172,7 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
 	s.lock.RUnlock()
 
 	if valSet == nil || msg.EpochNumber != epoch {
-		// Epoch metadata is undefined or received message for the unrelvant epoch
+		// Epoch metadata is undefined or received a message for the irrelevant epoch
 		return nil
 	}
 
--- a/consensus/polybft/state_sync_manager.go
+++ b/consensus/polybft/state_sync_manager.go
@@ -176,8 +176,7 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
 		return nil
 	}
 
-	err := s.verifyVoteSignature(valSet, types.StringToAddress(msg.From), msg.Signature, msg.Hash)
-	if err != nil {
+	if err := s.verifyVoteSignature(valSet, types.StringToAddress(msg.From), msg.Signature, msg.Hash); err != nil {
 		return fmt.Errorf("error verifying vote signature: %w", err)
 	}
 
@@ -201,6 +200,7 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
 	return nil
 }
 
+// Verifies signature of the message against the public key of the signer and checks if the signer is a validator
 func (s *stateSyncManager) verifyVoteSignature(valSet ValidatorSet, signer types.Address, signature []byte,
 	hash []byte) error {
 	validator := valSet.Accounts().GetValidatorMetadata(signer)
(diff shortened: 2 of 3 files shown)


consensus/polybft/state_sync_manager.go
```diff
@@ -171,17 +171,18 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
 	valSet := s.validatorSet
 	s.lock.RUnlock()
 
-	if valSet == nil || msg.EpochNumber < epoch {
-		// Epoch metadata is undefined or received message for some of the older epochs
+	if valSet == nil || msg.EpochNumber != epoch {
+		// Epoch metadata is undefined or received message for the unrelvant epoch
```
```suggestion
		// Epoch metadata is undefined or received a message for the irrelevant epoch
```

consensus/polybft/state_sync_manager.go
```diff
@@ -193,13 +194,32 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
...
+func (s *stateSyncManager) verifyVoteSignature(valSet ValidatorSet, signer types.Address, signature []byte,
+	hash []byte) error {
+	validator := valSet.Accounts().GetValidatorMetadata(signer)
+	if validator == nil {
+		return fmt.Errorf("unable to resolve validator %s", signer)
+	}
+
+	unmarshaledSignature, err := bls.UnmarshalSignature(signature)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshall signature: %w", err)
```
Stefan-Ethernal: Nitpick:
```suggestion
		return fmt.Errorf("failed to unmarshal signature: %w", err)
```

consensus/polybft/state_sync_manager_test.go
```diff
@@ -142,6 +142,50 @@ func TestStateSyncManager_MessagePool_SenderIsNoValidator(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestStateSyncManager_MessagePool_InvalidEpoch(t *testing.T) {
```
Stefan-Ethernal, line 145: You can probably add to other tests as well if you want.
```suggestion
func TestStateSyncManager_MessagePool_InvalidEpoch(t *testing.T) {
     t.Parallel()
```

consensus/polybft/state_sync_manager.go
```diff
@@ -171,17 +171,18 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
...
-		// Epoch metadata is undefined or received message for some of the older epochs
+	if valSet == nil || msg.EpochNumber != epoch {
+		// Epoch metadata is undefined or received message for the unrelvant epoch
 		return nil
 	}
 
-	if !valSet.Includes(types.StringToAddress(msg.NodeID)) {
-		return fmt.Errorf("validator is not among the active validator set")
+	err := s.verifyVoteSignature(valSet, types.StringToAddress(msg.From), msg.Signature, msg.Hash)
+	if err != nil {
```
Stefan-Ethernal: We could inline this?
```suggestion
	if err := s.verifyVoteSignature(valSet, types.StringToAddress(msg.From), msg.Signature, msg.Hash); err != nil {
```

consensus/polybft/state_sync_manager.go
```diff
@@ -193,13 +194,32 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
...
 		"hash", hex.EncodeToString(msg.Hash),
-		"sender", msg.NodeID,
+		"sender", msg.From,
 		"signatures", numSignatures,
 	)
 
 	return nil
 }
 
+func (s *stateSyncManager) verifyVoteSignature(valSet ValidatorSet, signer types.Address, signature []byte,
```
Stefan-Ethernal, line 204: Consider adding a comment. WDYT?

Stefan-Ethernal (commented):
Generally looks good. Leaving some minor suggestions.

consensus/polybft/state_sync_manager.go
stana-miric: fixed 91d52fe

consensus/polybft/state_sync_manager_test.go
stana-miric: fixed 91d52fe

consensus/polybft/state_sync_manager.go
stana-miric: fixed 91d52fe

consensus/polybft/state_sync_manager.go
stana-miric: fixed 91d52fe

consensus/polybft/state.go
```diff
@@ -39,8 +39,8 @@ type TransportMessage struct {
 	Hash []byte
 	// Message signature
 	Signature []byte
-	// Node identifier
-	NodeID string
+	// Address of the message signer
```
begmaroman: Minor one, but the comment should start from the type name. E.g.: "From is the address of the message signer"

consensus/polybft/state_sync_manager.go
```diff
@@ -193,13 +193,33 @@ func (s *stateSyncManager) saveVote(msg *TransportMessage) error {
...
+	hash []byte) error {
+	validator := valSet.Accounts().GetValidatorMetadata(signer)
+	if validator == nil {
+		return fmt.Errorf("unable to resolve validator %s", signer)
+	}
+
+	unmarshaledSignature, err := bls.UnmarshalSignature(signature)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal signature: %w", err)
```
igorcrevar: Maybe we should include signer address in this error message too

igorcrevar (approved):
LGTM, just minor comment

consensus/polybft/state_sync_manager.go
stana-miric: fixed 23e1551

consensus/polybft/state.go
stana-miric: fixed f3fe533

begmaroman (approved):

Stefan-Ethernal (approved):

PR #1260, opened 2023-03-01
TOB-EDGE-27: RecoverPubk should validate provided hashed message

goran-ethernal:
# Description

`crypto` package provides a way to recover the public from provided signature and hashed message, but before this PR, it did not validate the provided hashed message, meaning the function will return an arbitrary public key and then a malicious signer could impersonate other parties even without knowledge of their secret keys.

For example:
An attacker Eve finds a component inside of Polygon Edge that was changed recently and does not properly validate messages and submits a signature on an empty message that results in the `RecoverPubkey` function returning the public key for another user Bob. Eve is then able to impersonate Bob.

# Changes include

# Checklist

## Testing

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -30,7 +30,8 @@ var (
 	one            = big.NewInt(1)
 
 	ErrInvalidBLSSignature = errors.New("invalid BLS Signature")
-	errEmptyOrZeroHash     = errors.New("can not recover public key from zero or empty message hash")
+	errZeroHash            = errors.New("can not recover public key from zero or empty message hash")
+	errHashOfInvalidLength = errors.New("message hash of invalid length")
 	errInvalidSignature    = errors.New("invalid signature")
 )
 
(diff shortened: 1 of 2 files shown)


crypto/crypto.go
```diff
@@ -144,6 +142,10 @@ func Ecrecover(hash, sig []byte) ([]byte, error) {
 // RecoverPubkey verifies the compact signature "signature" of "hash" for the
 // secp256k1 curve.
 func RecoverPubkey(signature, hash []byte) (*ecdsa.PublicKey, error) {
+	if len(hash) == 0 || types.BytesToHash(hash) == types.ZeroHash {
```
igorcrevar: Maybe we should compare `len` to `types.HashLength`? Or we should allow at least `> 0 && < types.HashLength` hash sizes?
I am not sure if we should allow arbitrary number of bytes?

goran-ethernal: It makes sense, since we are using the `secp256k1` curve. I'll add the check.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1260/commits/73bb389efc7dec12005c8486505518c121910a04

igorcrevar (approved):
LGTM

Nemanja0x (approved):

Stefan-Ethernal (approved):
LGTM

vcastellm (commented):
Late LGTM

PR #1263, opened 2023-03-02
ToB code quality issues

stana-miric:
This PR includes most of the fixes from the ToB audit section 'Code Quality'. Some of the issues are left for future iteration since those require deeper analyzes and are more time and tests. 
Fixed issues(**relevant for this PR**):
TOB-Code Quality: 1. Local variable shadows global name
TOB-Code Quality: 3. Wrong order of arguments in tests
TOB-Code Quality: 5. Default branch in switch statement should return an error
TOB-Code Quality: 6. No explicit txType validation in validateTx
TOB-Code Quality: 8. Unchecked type assertions
TOB-Code Quality:  9. DropHexPrefix function may panic for 1 byte inputs
TOB-Code Quality: 10. Use constants instead of hardcoded values
TOB-Code Quality: 13. Possible integer overﬂows
TOB-Code Quality: 14.  Safety checks for receipts and logs should be implemented

Previously solved issues:
TOB-Code Quality:  11. Unused variable

Not applicable issues:
TOB-Code Quality: 12. Unstable Sort function
-not applicable since it is called only from ibft, but also is irrelevant since there are no two snapshots for the same height (2 equal elements) and thus use of unstable sort would not have any impact anyway 

Issues to solve through future task:
TOB-Code Quality:  2. hash.Hash interface is used instead of hash.Hash32.
-not fixed at this moment because if decided to be changed, it will require more time to be fixed and testes to be added
TOB-Code Quality: 7. LibP2Addr is not used, Network.Addr is used instead
-not fixed at this moment because it needs further investigation
TOB-Code Quality: 4. Redundant checks might introduce vulnerabilities in the future
-not fixed at this moment because it needs further investigation if it is safe to change it

# Changes include

# Checklist

## Testing


helper/hex/hex.go
```diff
@@ -95,5 +95,9 @@ func DecodeHexToBig(hexNum string) *big.Int {
 }
 
 func DropHexPrefix(input []byte) []byte {
-	return input[2:]
+	if len(input) > 2 {
```
igorcrevar: Maybe we should check 
`if len(input) >= 2 && strings.ToLower(input[:2]) == "0x"` 
No need to skip first two bytes if there are no hex prefix.

igorcrevar: Do we need `input[:]`(this will create a new slice but on same underlying array) in else?

consensus/polybft/checkpoint_manager.go
```diff
@@ -369,22 +369,24 @@ func getExitEventsFromReceipts(epoch, block uint64, receipts []*types.Receipt) (
 	events := make([]*ExitEvent, 0)
 
 	for i := 0; i < len(receipts); i++ {
-		for _, log := range receipts[i].Logs {
-			if log.Address != contracts.L2StateSenderContract {
-				continue
+		if *receipts[i].Status == types.ReceiptSuccess {
```
igorcrevar: Because Status is pointer (`Status *ReceiptStatus`), maybe we should check for `nil` value also? 

I am not sure about this and this does not make a difference, but maybe it would be good to revert condition and `continue` if status is not `ReceiptSuccess`? All the code bellow uses `continue` logic

consensus/polybft/state_transaction.go
```diff
@@ -80,6 +80,10 @@ func (cm *CommitmentMessageSigned) VerifyStateSyncProof(proof []types.Hash,
 		return errors.New("no state sync event")
 	}
 
+	if stateSync.ID.Uint64() < cm.Message.StartID.Uint64() {
```
goran-ethernal: `VerifyStateSyncProof` is only used for unit tests in edge client, since proof verification is done in smart contracts, but nevertheless, this is not needed, since if it does not belong to the commitment, `VerifyProof` on merkle tree will return an error (meaning the state sync does not belong to given commitment).
I mean we can leave it as an optimization, but I would say then add the check for the `stateSyncID > commitment.EndID`.

goran-ethernal (approved):
Apart from a small comment LGTM

helper/hex/hex.go
stana-miric: fix b952ec8

consensus/polybft/checkpoint_manager.go
stana-miric: fix b952ec8

consensus/polybft/state_transaction.go
stana-miric: fix b952ec8

igorcrevar (approved):
LGTM

Nemanja0x (approved):

helper/hex/hex.go
```diff
@@ -95,5 +95,9 @@ func DecodeHexToBig(hexNum string) *big.Int {
 }
 
 func DropHexPrefix(input []byte) []byte {
-	return input[2:]
+	if len(input) >= 2 && strings.ToLower(string(input[:2])) == "0x" {
+		return input[2:]
+	}
+
+	return input
```
Stefan-Ethernal: Just dropping as an idea, not sure what is better:
```suggestion
	return []byte(strings.TrimPrefix(strings.ToLower(string(input)), "0x"))
```
What I don't like with my proposal is casting back and forth to string and back to a byte slice. Although proposed it anyways because of sparing some lines of code and checks :slightly_smiling_face: 

You decide what makes more sense to you.

txpool/txpool_test.go
```diff
@@ -243,6 +256,27 @@ func TestAddTxErrors(t *testing.T) {
...
+		pool := setupPool()
+
+		// fill the pool leaving only 1 slot
+		pool.gauge.increase(defaultMaxSlots - 1)
+
+		// create tx requiring 1 slot
+		tx := newTx(defaultAddr, 0, 1)
+		tx = signTx(tx)
+
+		//	enqueue tx
```
Stefan-Ethernal, line 270: Nitpick:
```suggestion
		// enqueue tx
```

txpool/txpool_test.go
```diff
@@ -243,6 +256,27 @@ func TestAddTxErrors(t *testing.T) {
 		)
 	})
 
+	t.Run("FullTxPoolToTheLimit", func(t *testing.T) {
```
Stefan-Ethernal: WDYT?
```suggestion
	t.Run("FillTxPoolToTheLimit", func(t *testing.T) {
```

Stefan-Ethernal (approved):
Nice :raised_hands: 

Leaving a few minor comments to consider.

helper/hex/hex.go
vcastellm: I prefer this one-liner, and dropping `DropHexPrefix`

helper/hex/hex_test.go
```diff
@@ -38,3 +38,45 @@ func TestDecodeUint64(t *testing.T) {
 		assert.Equal(t, uint64Array[index], decodedValue)
 	}
 }
+
+func TestDropHexPrefix(t *testing.T) {
```
vcastellm: No need for this if we use `TrimPrefix`

helper/hex/hex.go
stana-miric: Fixed as Victor suggested 031bb22

txpool/txpool_test.go
stana-miric: fixed 031bb22

txpool/txpool_test.go
stana-miric: fixed 031bb22

helper/hex/hex_test.go
stana-miric: fixed 031bb22

vcastellm (approved):
LGTM
G
T
M

PR #1269, opened 2023-03-06
EVM-495 AA - Relay transactions from the pool

igorcrevar:
# Description

In this task there will be implemented two things:
- pool of sorted transactions
- service which will relay tx from the pool to the actual node 

Account abstraction command will be executed like:
`./polygon-edge aarelayer --addr 127.0.0.1:8198 --data-dir ./test-chain-1 --jsonrpc 127.0.01:10002`

# Changes include

# Checklist

## Testing

### Manual tests

### DIFF
--- a/command/aarelayer/service/aa_relayer_service.go
+++ b/command/aarelayer/service/aa_relayer_service.go
@@ -17,8 +17,9 @@ type AARelayerService struct {
 	state        AATxState
 	txSender     AATxSender
 	key          ethgo.Key
-	pullTime     time.Duration
+	pullTime     time.Duration // pull from txpool every `pullTime` second/millisecond
 	receiptDelay time.Duration
+	numRetries   int
 }
 
 func NewAARelayerService(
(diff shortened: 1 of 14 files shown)


command/aarelayer/service/aa_relayer_service.go
```diff
@@ -0,0 +1,161 @@
...
+		fmt.Printf("error while updating state tx = %s after sending it, err = %v", stateTx.ID, err)
+	}
+
+	select {
+	case <-ctx.Done():
+		return nil
+	default:
+	}
+
+	recipt, err := rs.txRelayer.WaitForReceipt(ctx, hash)
```
ferranbt: At this point, the relayer is only an abstraction to send a transaction (`SendTransactionWithoutReceipt`) and wait for the receipt (`WaitForReceipt`). You are not leveraging the utilities of the relayer (i.e. fire and forget the execution of a transaction). Thus, you can use directly the `jsonrpc.Client` package to send and wait for the receipt without complicating the `txrelayer`.

ferranbt (changes requested):

command/aarelayer/service/aa_relayer_service.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1269/commits/c3a0bc274da59b886131b859103a8e06f894ce7d

ferranbt (approved):

Nemanja0x (approved):

PR #1272, opened 2023-03-07
Provide multi addresses instead of p2p peer ids to the manifest file

Stefan-Ethernal:
# Description

This PR changes the way we are expecting validators' network parameters to be specified when creating the manifest file. Prior to this, when specifying validators via CLI flags, only P2P peer id was expected, and then when genesis configuration was generated multi-address was assuming the local host IP address.

Now manifest generation expects entire multi-address to be provided if case genesis validators are being specified via CLI flag. Otherwise, if secrets are provided on local file storage, multi addresses are being constructed with localhost IP addresses.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/polybftmanifest/manifest_init.go
+++ b/command/polybftmanifest/manifest_init.go
@@ -12,6 +12,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/command/genesis"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft"
 	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/multiformats/go-multiaddr"
 	"github.com/spf13/cobra"
 )
 
@@ -26,7 +27,6 @@ const (
 	defaultValidatorPrefixPath = "test-chain-"
 	defaultManifestPath        = "./manifest.json"
 
-	nodeIDLength       = 53
 	ecdsaAddressLength = 40
 	blsKeyLength       = 256
 	blsSignatureLength = 128
@@ -79,7 +79,7 @@ func setFlags(cmd *cobra.Command) {
 		&params.validators,
 		validatorsFlag,
 		[]string{},
-		"validators defined by user (format: <node id>:<ECDSA address>:<public BLS key>:<BLS signature>)",
+		"validators defined by user (format: <P2P multi address>:<ECDSA address>:<public BLS key>:<BLS signature>)",
 	)
 
 	cmd.Flags().StringVar(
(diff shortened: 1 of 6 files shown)


vcastellm (commented):
Will the multiaddr be output by the `polybft-secrets` command?

Stefan-Ethernal:
> Will the multiaddr be output by the `polybft-secrets` command?

Nope, it just prints out P2P node id. The thing is that AFAIK, we don't keep IP addresses and ports inside secrets, but only P2P node IDs, so we are not able to reconstruct the entire multiaddress.

vcastellm:
> Nope, it just prints out P2P node id. The thing is that AFAIK, we don't keep IP addresses and ports inside secrets, but only P2P node IDs, so we are not able to reconstruct the entire multiaddress.

Ok, so it's the user that should build the multiaddr format for feeding the `manifest` command, right? We should specify that in docs

vcastellm (approved):
LGTM

Stefan-Ethernal:
> Ok, so it's the user that should build the multiaddr format for feeding the `manifest` command, right? We should specify that in docs

Yes, that is correct. :+1:

Stefan-Ethernal:
> We should specify that in docs

Updated README on my end: https://github.com/0xPolygon/polygon-edge/pull/1272/commits/fa743f1def053a6e081de0ec1151732d0577d1cb

command/polybftmanifest/manifest_init.go
```diff
@@ -153,31 +153,30 @@ func (p *manifestInitParams) getValidatorAccounts() ([]*polybft.Validator, error
...
 			}
 
 			if len(parts[3]) != blsSignatureLength {
-				return nil, fmt.Errorf("invalid bls signature: %s", parts[3])
+				return nil, fmt.Errorf("invalid BLS signature: %s", parts[3])
 			}
 
 			validators[i] = &polybft.Validator{
-				NodeID:       parts[0],
+				MultiAddr:    parts[0],
```
igorcrevar, line 181: Use `strings.TrimPrefix(parts[1], "0x")` and `strings.TrimPrefix(parts[2], "0x")` here also

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1272/commits/13cda702f09e49d9b7533bb7d11a7907c003faa6

igorcrevar (approved):

PR #1278, opened 2023-03-08
Fix `bridge-json-rpc` flag in genesis command

Stefan-Ethernal:
# Description

This PR fixes `bridge-json-rpc` flag behavior. Prior to this, there was obscure behavior meant to allow specifying this flag, without providing a value and it should then assign the default value. However, this resulted in not being possible to provide some different value than the default one.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -181,7 +181,7 @@ func setFlags(cmd *cobra.Command) {
 			&params.bridgeJSONRPCAddr,
 			bridgeFlag,
 			"",
-			"the rootchain JSON RPC IP address. If present, node is running in bridge mode.",
+			"the rootchain JSON RPC endpoint",
 		)
 
 		cmd.Flags().Uint64Var(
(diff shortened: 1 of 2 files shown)


vcastellm:
This is  working as expected, actually it's possible to change the default value, only that the format changes, it should be specified as `bridge-json-rpc=value` it doesn't work using space, that's the only caveat.

I agree if the intention is to make it explicit.

Stefan-Ethernal:
> This is working as expected, actually it's possible to change the default value, only that the format changes, it should be specified as `bridge-json-rpc=value` it doesn't work using space, that's the only caveat.

Oh ok didn't know that. :+1: 

> I agree if the intention is to make it explicit.

Yes, it is easy to make a mistake there and then the question is about consistency and why we aren't treating `--bridge-json-rpc` the same way as other flags. On the other hand, benefit is pretty small if any (allows us to specify localhost json rpc address easier, but it wouldn't be the case outside of dev environments).

vcastellm (approved):

goran-ethernal (approved):

PR #1281, opened 2023-03-09
Add burn contract to tests

begmaroman:
# Description

Update tests according to EIP-1559 changes made before. New tests will be added in the next PR.

# Changes include

# Checklist

## Testing


state/executor.go
```diff
@@ -574,7 +578,7 @@ func (t *Transition) apply(msg *types.Transaction) (*runtime.ExecutionResult, er
 	// We use EIP-1559 fields of the tx if the london hardfork is enabled.
 	// Effective tip be came to be either gas tip cap or (gas fee cap - current base fee)
 	effectiveTip := new(big.Int).Set(gasPrice)
-	if t.config.London && msg.Type != types.StateTx {
+	if t.config.London && msg.Type == types.DynamicFeeTx {
```
goran-ethernal, line 581: There is a condition in this `apply` function:

```
// Burn some amount if the london hardfork is applied.
// Basically, burn amount is just transferred to the current burn contract.
if t.config.London && msg.Type != types.StateTx {
```

Should we change it to check if `msg.Type == types.DynamicFeeTx` like we did here? Maybe we can move it in this if block.

types/rlp_encoding_test.go
```diff
@@ -281,53 +296,59 @@ func TestRLPMarshall_And_Unmarshall_TxType(t *testing.T) {
 func testRLPData(arena *fastrlp.Arena, ommitValues map[string]bool) []byte {
 	vv := arena.NewArray()
 
-	_, ommit := ommitValues["Nonce"]
-	if !ommit {
+	if ommit, _ := ommitValues["ChainID"]; !ommit {
```
vcastellm: I hate to be the nitpicker here but it's `omit` and `omitted` 😄

vcastellm (commented):
Minor comment

types/rlp_encoding_test.go
begmaroman: Fixed

state/executor.go
begmaroman: The idea here is that there are not gas fee cap and gas tip cap in non-dynamic-gas-fee tx so it jus does not make any sense to apply this check for legacy or state txs.

vcastellm (approved):

PR #1282, opened 2023-03-09
Print output of cluster bash script to both terminal and file

Stefan-Ethernal:
# Description

This PR changes the helper cluster script to print the output of both `stdout` and `stderr` to both terminal and file for each validator node.

**Note:** would be nice to maybe group it somehow to a folder because currently files are written to the root folder of the project and it gets overwritten if cluster script is run multiple times.

# Changes include

# Checklist

## Testing

### DIFF
--- a/scripts/cluster
+++ b/scripts/cluster
@@ -27,10 +27,18 @@ function createGenesis() {
 }
 
 function startServerFromBinary() {
-  ./polygon-edge server --data-dir ./test-chain-1 --chain genesis.json --grpc-address :10000 --libp2p :30301 --jsonrpc :10002 --num-block-confirmations 2 --seal --log-level DEBUG &
-  ./polygon-edge server --data-dir ./test-chain-2 --chain genesis.json --grpc-address :20000 --libp2p :30302 --jsonrpc :20002 --num-block-confirmations 2 --seal --log-level DEBUG &
-  ./polygon-edge server --data-dir ./test-chain-3 --chain genesis.json --grpc-address :30000 --libp2p :30303 --jsonrpc :30002 --num-block-confirmations 2 --seal --log-level DEBUG &
-  ./polygon-edge server --data-dir ./test-chain-4 --chain genesis.json --grpc-address :40000 --libp2p :30304 --jsonrpc :40002 --num-block-confirmations 2 --seal --log-level DEBUG &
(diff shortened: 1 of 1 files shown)


vcastellm (commented):
Why this change? I'm not against it but also not much in favor, search in terminal works well

Stefan-Ethernal:
> Why this change? I'm not against it but also not much in favor, search in terminal works well

I can think of at least two reasons at a glance:
1. Terminal output has limited capacity (at least on VS Code Ubuntu) and if we are running cluster, it gets clipped away pretty fast, which makes searching inconvenient,
2. Logs from different nodes are interleaved, which makes it hard to follow if we want to track execution on a specific node.

So this way the best of both approaches so depending on use case, one could opt to search through the files or the terminal itself.

vcastellm:
> > Why this change? I'm not against it but also not much in favor, search in terminal works well
> 
> I can think of at least two reasons at a glance:
> 
> 1. Terminal output has limited capacity (at least on VS Code Ubuntu) and if we are running cluster, it gets clipped away pretty fast, which makes searching inconvenient,

You can change that in settings in majority of terminals
image

> 2. Logs from different nodes are interleaved, which makes it hard to follow if we want to track execution on a specific node.
> 

For this I prefer to identify each node with a prefix + color similar to docker compose, if this is complex I'm ok with this.

> So this way the best of both approaches so depending on use case, one could opt to search through the files or the terminal itself.

Stefan-Ethernal:
> You can change that in settings in majority of terminals

Yes, but still I've experienced that during running a cluster if I was about to search Terminal content at the same time, it would start highlighting wrong words (believe this is because of terminal content constantly being changed) and experienced it at least on Ubuntu (haven't tried on Mac OS though) which makes debugging harder. 

So I'd still would like to keep option to output logs to the files and make it optional (https://github.com/0xPolygon/polygon-edge/pull/1282/commits/eb57954ed46334d31ffa215baf8cc094d1d70dd7; not sure if we can make this if/else to have less repetitions, by for example extracting common parameters to local variable(s)).

> For this I prefer to identify each node with a prefix + color similar to docker compose, if this is complex I'm ok with this.

Not sure how to accomplish this, to be honest with you.

vcastellm (approved):

b00ris (approved):

ferranbt (approved):

PR #1283, opened 2023-03-09
EVM-441: Add gRPC data validation

stana-miric:
# Description

This PR adds validation of grpc incoming messages (requests). It does not include validation for proto messages which are not used as grps service request, but since the infrastructure for validation is integrated, if needed those validations can be optionally added in the future. The changes contain:

1. Modification of protoc command in makefile, in order for validation files to be generated. The files are generated using protoc-gen-validate plugin, which is chosen over go-proto-validators because it is more flexible, extensible, and it has wide range of built-in validators. It must be installed before running 'sudo make protoc' command and installation is simply (as for protoc-gen-go and protoc-gen-go-grpc which are already in use by the current protoc command)
2. Validation of a following grpc requests:
 2.1 polygon-edge/consensus/ibft/proto/ibft_operator.proto This is skipped since it's irelevant (ibft)
 2.2 polygon-edge/network/proto/discovery.proto
    message FindPeersReq {
        string key = 1;   
        int64 count = 2;
    }
key - empty value or the value matching the regex for libp2p PeerId are acceptable. Empty value is acceptable, because when   processing the request in the case of empty key, code sets it to grpcContext.PeerID
count - valid values are from 1 to maxDiscoveryPeerReqCount
 2.3 polygon/polygon-edge/network/proto/testing.proto This is skipped since it's irrelevant (tests)
 2.4 polygon/polygon-edge/syncer/proto/syncer.proto
    message GetBlocksRequest {
      uint64 from = 1; 
    }
    from: no validation is added, any block number can be selected as starting block
    **Q: path for syncer proto file is not included in proto generation when calling protoc in makefile. Is this for a reason, or it should be included?**
 2.5 /polygon-edge/txpool/proto/operator.proto
     message AddTxnReq {
       google.protobuf.Any raw = 1; 
       string from = 2; 
     }
 raw - value must be set
 from - should match pattern for eth address, also empty value is allowed (**Q: should we allow empty?**)

     message SubscribeRequest {
       repeated EventType types = 1; 
     }
 types - all values must be unique, it has to be minimum 1 value and all are defined in EventType enum
 2.6 polygon-edge/server/proto/system.proto
     message PeersAddRequest {
       string id = 1;
     }
 id - matches pattern for multiaddress
     message PeersStatusRequest {
       string id = 1;
     }
 id - matches pattern for libp2p PeerId
     message BlockByNumberRequest {
       uint64 number = 1;
     }
 number - no limitation
      message ExportRequest {
       uint64 from = 1;
       uint64 to = 2;
     }
 from - no limitation, string block can be any
 to - no limitation, ending block can be any
 2.7 polygon-edge/network/proto/identity.proto This is skipped since Hello method is used for testing only
3. Unit test are added to test each request for invalid data, and e2e test is added to check triggering of interceptor 

 *Note: Currently, protoc-gen-validate does not support skipping unannotated files for golang. So, for few proto files *.validate.go is generated even though its validate method does not contain any rules. Those files are committed in order to avoid noise when running make protoc command, and maybe in future some validation will be added to those messages as well. 

# Changes include

# Checklist

## Testing

- [x] I have tested this code with the unit tests


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/btcsuite/btcd v0.22.1
 	github.com/containerd/cgroups v1.0.4 // indirect
 	github.com/elastic/gosigar v0.14.2 // indirect
+	github.com/envoyproxy/protoc-gen-validate v0.9.1
```
vcastellm, line 9: It looks like the updated dep is this https://github.com/bufbuild/protoc-gen-validate

server/proto/system_test.go
```diff
@@ -0,0 +1,88 @@
...
+			name: "PeersAddRequest: valid id format",
+			req: &PeersAddRequest{
+				Id: "/ip4/127.0.0.1/tcp/30301/p2p/16Uiu2HAm698AJQJJjZj4ef8PPmE1YDnPAuZnf3wGCp7ZyM3uhcAu",
+			},
+			valid: true,
+		},
+		{
+			name: "PeersAddRequest: valid id format relay circuit",
+			req: &PeersAddRequest{
+				Id: "/ip4/198.51.100.0/tcp/4242/p2p/QmRelay/p2p-circuit/p2p/QmRelayedPeer",
```
vcastellm, line 45: Check also with `/dns4/domain.name/....`

go.mod
stana-miric: module path is declared as github.com/envoyproxy/protoc-gen-validate, [here](https://github.com/bufbuild/protoc-gen-validate#-yes-our-go-module-path-is-githubcomenvoyproxyprotoc-gen-validate-not-bufbuild-this-is-intentional) is a short statement about it. github.com/envoyproxy/protoc-gen-validate redirects to bufbuild but go modul is the same

server/proto/system_test.go
stana-miric: fixed b7f6b44

vcastellm (approved):
LGTM

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #1289, opened 2023-03-13
TOB-EDGE-15: Some errors are not handled

goran-ethernal:
# Description

Added error checking to missing places as per TOB audit. Also, included the `errcheck` lint rule to our configuration so that it is enabled for future development.

# Changes include

# Checklist

## Testing

### DIFF
--- a/helper/hex/hex.go
+++ b/helper/hex/hex.go
@@ -89,8 +89,9 @@ func EncodeBig(bigint *big.Int) string {
 // DecodeHexToBig converts a hex number to a big.Int value
 func DecodeHexToBig(hexNum string) (*big.Int, error) {
 	createdNum := new(big.Int)
+	cleaned := strings.TrimPrefix(hexNum, "0x")
 
-	_, ok := createdNum.SetString(hexNum, 16)
+	_, ok := createdNum.SetString(cleaned, 16)
 	if !ok {
 		return nil, fmt.Errorf("failed to convert string: %s to big.Int with base: 16", hexNum)
 	}
--- a/helper/hex/hex.go
+++ b/helper/hex/hex.go
@@ -89,6 +89,7 @@ func EncodeBig(bigint *big.Int) string {
 // DecodeHexToBig converts a hex number to a big.Int value
 func DecodeHexToBig(hexNum string) (*big.Int, error) {
 	createdNum := new(big.Int)
+
 	_, ok := createdNum.SetString(hexNum, 16)
 	if !ok {
 		return nil, fmt.Errorf("failed to convert string: %s to big.Int with base: 16", hexNum)
--- a/helper/hex/hex.go
+++ b/helper/hex/hex.go
@@ -88,13 +88,12 @@ func EncodeBig(bigint *big.Int) string {
 
 // DecodeHexToBig converts a hex number to a big.Int value
 func DecodeHexToBig(hexNum string) (*big.Int, error) {
-	createdNum := new(big.Int)
 	cleaned := strings.TrimPrefix(hexNum, "0x")
 
-	_, ok := createdNum.SetString(cleaned, 16)
+	value, ok := new(big.Int).SetString(cleaned, 16)
 	if !ok {
 		return nil, fmt.Errorf("failed to convert string: %s to big.Int with base: 16", hexNum)
 	}
 
-	return createdNum, nil
+	return value, nil
 }
--- a/helper/hex/hex.go
+++ b/helper/hex/hex.go
@@ -87,9 +87,12 @@ func EncodeBig(bigint *big.Int) string {
 }
 
 // DecodeHexToBig converts a hex number to a big.Int value
-func DecodeHexToBig(hexNum string) *big.Int {
+func DecodeHexToBig(hexNum string) (*big.Int, error) {
 	createdNum := new(big.Int)
-	createdNum.SetString(hexNum, 16)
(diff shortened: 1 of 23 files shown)


b00ris (approved):

stana-miric (approved):

helper/hex/hex.go
```diff
@@ -87,9 +87,14 @@ func EncodeBig(bigint *big.Int) string {
...
-	createdNum.SetString(hexNum, 16)
+	cleaned := strings.TrimPrefix(hexNum, "0x")
 
-	return createdNum
+	_, ok := createdNum.SetString(cleaned, 16)
+	if !ok {
+		return nil, fmt.Errorf("failed to convert string: %s to big.Int with base: 16", hexNum)
+	}
+
+	return createdNum, nil
```
Stefan-Ethernal: Spare 1 line :smile:

```suggestion
	cleaned := strings.TrimPrefix(hexNum, "0x")

	value, ok := new(big.Int).SetString(cleaned, 16)
	if !ok {
		return nil, fmt.Errorf("failed to convert string: %s to big.Int with base: 16", hexNum)
	}

	return value, nil
```

helper/hex/hex_test.go
```diff
@@ -38,3 +39,26 @@ func TestDecodeUint64(t *testing.T) {
...
 	}
 }
+
+func TestDecodeHexToBig(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Error on decoding", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := DecodeHexToBig("012345q")
```
Stefan-Ethernal: Maybe you can assert that returned value is nil as well?

helper/hex/hex_test.go
```diff
@@ -38,3 +39,26 @@ func TestDecodeUint64(t *testing.T) {
...
+		require.ErrorContains(t, err, "failed to convert string")
+	})
+
+	t.Run("Happy path", func(t *testing.T) {
+		t.Parallel()
+
+		big, err := DecodeHexToBig("0123456")
```
Stefan-Ethernal: I'd add prefix, just to stress that it is hexadecimal number. WDYT?
```suggestion
		big, err := DecodeHexToBig("0x0123456")
```

helper/keccak/keccak.go
```diff
@@ -22,7 +22,7 @@ type Keccak struct {
 // WriteRlp writes an RLP value
 func (k *Keccak) WriteRlp(dst []byte, v *fastrlp.Value) []byte {
 	k.buf = v.MarshalTo(k.buf[:0])
-	k.Write(k.buf)
+	k.Write(k.buf) //nolint:errcheck
```
Stefan-Ethernal, line 25: I'm aware that you didn't want to change signature of these functions to propagate an error, however maybe to consider propagating errors, not sure...?

Stefan-Ethernal (approved):

helper/hex/hex.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1289/commits/8513e9d2713d95a0bbb1df4eed52c8d0422cc6a0

helper/hex/hex_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1289/commits/8513e9d2713d95a0bbb1df4eed52c8d0422cc6a0

helper/hex/hex_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1289/commits/8513e9d2713d95a0bbb1df4eed52c8d0422cc6a0

helper/keccak/keccak.go
goran-ethernal: Write does not return any error in the Keccak256 implementation we use. I checked the code. Error is there because it implements that `Writer` interface. Take a look at this:

```
// Only use this function if you require compatibility with an existing cryptosystem
// that uses non-standard padding. All other users should use New256 instead.
func NewLegacyKeccak256() hash.Hash { return &state{rate: 136, outputLen: 32, dsbyte: 0x01} }
```

```
// Write absorbs more data into the hash's state. It produces an error
// if more data is written to the ShakeHash after writing
func (d *state) Write(p []byte) (written int, err error) {
	if d.state != spongeAbsorbing {
		panic("sha3: write to sponge after read")
	}
	if d.buf == nil {
		d.buf = d.storage.asBytes()[:0]
	}
	written = len(p)

	for len(p) > 0 {
		if len(d.buf) == 0 && len(p) >= d.rate {
			// The fast path; absorb a full "rate" bytes of input and apply the permutation.
			xorIn(d, p[:d.rate])
			p = p[d.rate:]
			keccakF1600(&d.a)
		} else {
			// The slow path; buffer the input until we can fill the sponge, and then xor it in.
			todo := d.rate - len(d.buf)
			if todo > len(p) {
				todo = len(p)
			}
			d.buf = append(d.buf, p[:todo]...)
			p = p[todo:]

			// If the sponge is full, apply the permutation.
			if len(d.buf) == d.rate {
				d.permute()
			}
		}
	}

	return
}
```

Stefan-Ethernal: Yes and that's fine, but if the underlying implementation changes at some point... Anyhow it is good for me (for now).

PR #1291, opened 2023-03-13
EVM-516 AA Update EVM with opcodes

igorcrevar:
# Description

`EIP-3074`: AUTH and AUTHCALL opcodes

# Changes include

# Checklist

## Testing

### DIFF
--- a/crypto/txsigner.go
+++ b/crypto/txsigner.go
@@ -9,6 +9,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/chain"
 	"github.com/0xPolygon/polygon-edge/helper/keccak"
 	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/umbracle/ethgo"
 	"github.com/umbracle/fastrlp"
 )
 
@@ -98,7 +99,7 @@ func (f *FrontierSigner) Sender(tx *types.Transaction) (types.Address, error) {
 
 	refV.Sub(refV, big27)
 
-	sig, err := encodeSignature(tx.R, tx.S, refV, f.isHomestead)
+	sig, err := EncodeSignature(tx.R, tx.S, refV, f.isHomestead)
 	if err != nil {
 		return types.Address{}, err
 	}
@@ -181,7 +182,7 @@ func (e *EIP155Signer) Sender(tx *types.Transaction) (types.Address, error) {
 	bigV.Sub(bigV, mulOperand)
 	bigV.Sub(bigV, big35)
 
-	sig, err := encodeSignature(tx.R, tx.S, bigV, e.isHomestead)
+	sig, err := EncodeSignature(tx.R, tx.S, bigV, e.isHomestead)
 	if err != nil {
 		return types.Address{}, err
 	}
@@ -229,8 +230,8 @@ func (e *EIP155Signer) CalculateV(parity byte) []byte {
 	return reference.Bytes()
 }
 
-// encodeSignature generates a signature value based on the R, S and V value
-func encodeSignature(R, S, V *big.Int, isHomestead bool) ([]byte, error) {
+// EncodeSignature generates a signature value based on the R, S and V value
+func EncodeSignature(R, S, V *big.Int, isHomestead bool) ([]byte, error) {
 	if !ValidateSignatureValues(V, R, S, isHomestead) {
 		return nil, fmt.Errorf("invalid txn signature")
 	}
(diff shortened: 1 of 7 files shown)


paulgoleary:
Hi all. Just wanted to make sure that it's known that this impl of 3074 is based on an older spec and needs to be updated.
Comparable work has already been done on V1: https://github.com/maticnetwork/bor/commits/poleary/wip-eip-3074

state/runtime/evm/instructions.go
```diff
@@ -1191,10 +1191,16 @@ func (c *state) buildCallContract(op OpCode) (*runtime.Contract, uint64, uint64,
...
 	var value *big.Int
-	if op == CALL || op == CALLCODE {
+	if op == CALL || op == CALLCODE || op == AUTHCALL {
 		value = c.pop()
 	}
 
+	var extValue *big.Int
+	if op == AUTHCALL {
+		extValue = c.pop() // todo: propagate the extValue
+		_ = extValue       // ignore for now
```
ferranbt: The EIP specifies a behaviour for this value:
```
If valueExt is not zero, the instruction immediately returns 0. In this case the gas that would have been passed into the call is refunded, but not the gas consumed by the AUTHCALL opcode itself. In the future, this restriction may be relaxed to externally transfer value out of the authorized account.
```

crypto/txsigner.go
```diff
@@ -242,3 +243,15 @@ func encodeSignature(R, S, V *big.Int, isHomestead bool) ([]byte, error) {
...
+// Eip3074Magic serialize EIP-3074 messages in form
+// keccak256(type ++ invoker ++ commit)
+func Eip3074Magic(commit []byte, invokerAddr types.Address) []byte {
+	var msg [64]byte
+
+	msg[0] = 0x03
+	copy(msg[13:33], invokerAddr.Bytes())
+	copy(msg[33:], commit)
+
+	return ethgo.Keccak256(msg[:])
```
ferranbt: `edge` has already a function to calculate the `keccak` [here](https://github.com/0xPolygon/polygon-edge/blob/develop/crypto/crypto.go#L244).

state/runtime/evm/instructions.go
```diff
@@ -1403,6 +1414,48 @@ func opHalt(op OpCode) instruction {
...
+		c.exit(fmt.Errorf("invalid signature values: %w", err))
+
+		return
+	}
+
+	signAddr := crypto.PubKeyToAddress(pubKey)
+	c.authorized = &signAddr
+
+	c.ret = make([]byte, 12)
+	c.ret = append(c.ret, signAddr.Bytes()...)
```
ferranbt: I do not see on the spec that the return value has to be modified.

state/runtime/evm/instructions.go
```diff
@@ -1403,6 +1414,48 @@ func opHalt(op OpCode) instruction {
 	}
 }
 
+func opAuth(c *state) {
+	c.authorized = nil
+
+	commit := c.pop()
```
ferranbt, line 1444: Is this aligned with the spec [here](https://eips.ethereum.org/EIPS/eip-3074#authcall-0xf7)? `AUTH` opcode has three values on the stack and they are memory pointers not the signature values.

ferranbt (changes requested):

crypto/txsigner.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1291/commits/0cc913505670adaf9460093e843f7d99109774b1

state/runtime/evm/instructions.go
igorcrevar: Is this correct fix https://github.com/0xPolygon/polygon-edge/pull/1291/commits/669eddc7cb87bb8478602ed5a6ab8b0cbee4c133

state/runtime/evm/instructions.go
igorcrevar: The current solidity compiler used by our smart contract team contains a patch that is only compatible with the previous revision of the specification. While it would be relatively easy for us to convert the code from the `bor` repository (https://github.com/maticnetwork/bor/commit/54df42f550d57b27423ed178f7469b011ac609b7) into `edge` we are facing issues with the smart contract side of the conversion

state/runtime/evm/instructions.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1291/commits/fed0808cda37cb7e40af514931fa0dc9dc27bcc5

state/runtime/evm/instructions.go
igorcrevar: This will be solved on base branch later on https://github.com/0xPolygon/polygon-edge/pull/1298

Nemanja0x (approved):

Stefan-Ethernal (approved):

goran-ethernal (approved):
After going through changes with @igorcrevar, I am approving the PR :rocket:

PR #1297, opened 2023-03-15
Add EIP-1559 tests to e2e framework 

begmaroman:
# Description

Update e2e framework by adding dynamic gas fee txs

# Changes include

# Checklist

## Testing


e2e/broadcast_test.go
```diff
@@ -8,43 +8,125 @@ import (
...
 	"github.com/0xPolygon/polygon-edge/e2e/framework"
 	"github.com/0xPolygon/polygon-edge/helper/tests"
 	"github.com/0xPolygon/polygon-edge/types"
-	"github.com/stretchr/testify/assert"
 )
 
 func TestBroadcast(t *testing.T) {
 	// This test is not stable
 	// Opened the ticket to check + fix it
 	t.Skip()
```
Stefan-Ethernal: Two points that I want to make:

- This test is anyways skipped. Do you think investing some time right now to figure out which part introduces instability? Otherwise don't see the full benefit of adding new test cases.
- This is part of the IBFT e2e framework, which is probably be deprecated at some point. Maybe it would make more sense to implement the e2e test in e2e-polybft, using its polybft e2e testing framework.

WDYT?

e2e/pos_test.go
```diff
@@ -499,28 +510,36 @@ func TestPoS_StakeUnstakeExploit(t *testing.T) {
...
-			Gas:      framework.DefaultGasLimit,
-			Value:    value,
-			V:        big.NewInt(1), // it is necessary to encode in rlp
-			Input:    framework.MethodSig(methodName),
-		}, senderKey)
-
-		if signErr != nil {
-			t.Fatalf("Unable to sign transaction, %v", signErr)
+	generateTx := func(i int, value *big.Int, methodName string) *types.Transaction {
+		unsignedTx := &types.Transaction{
```
Stefan-Ethernal, line 520: Again, not sure if it is justified to adapt E2E tests that are using the IBFT codebase. I'd prefer implementing a new one from scratch but using polybft e2e framework.

e2e/broadcast_test.go
begmaroman: Agree. Since I've already updated them, do you think it makes sense to revert it back?

Stefan-Ethernal: Just remember to reuse this code when revamping this test on e2e polybft.

Stefan-Ethernal (approved):
LGTM with a side note that EIP-1559 E2E tests should be implemented using polybft framework as well in some of the following PRs.

begmaroman:
Going to add polybft e2e tests in the next PR

PR #1304, opened 2023-03-16
EVM-440-Using-UTC-time

goran-ethernal:
# Description

Using local validators’ time zones may result in chain fork. So through this PR, we are using the `UTC` time throughout the code.

PR also introduces a lint rule that checks if `time.Now` variations are being used in code without calling the `UTC` function as well.

# Changes include

# Checklist

## Testing


vcastellm (changes requested):
I don't find it necessary to create a package out of the stdlib for wrapping the same calls to the stdlib, it creates unnecessary roundtrip for the reader, the value provided is very limited and it's very prone to be skipped in new code.

I vote for just using stdlib

goran-ethernal:
@vcastellm Those were my initial thoughts as well, but I've gone with Evgeny's proposal on that task, to have it wrapped for easier usage. I am down whatever we find more comfortable.

vcastellm (approved):
LGTM

Stefan-Ethernal (approved):
🚀

igorcrevar (approved):

PR #1306, opened 2023-03-17
Feat allow list

ferranbt:
# Description

This PR introduces two concepts:
- An `AllowList` framework as a Golang built-in precompile that can handle authorization tasks (i.e. this address is enabled to perform the task, and this address can update who is enabled to perform the task). The framework does not contain any business-level logic on where the allow list framework is used and serves as a generic building block.
- The application of the `AllowList` framework to control who can deploy smart contracts.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
 	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true E2E_TESTS_TYPE=integration \
-	go test -v -timeout=45m ./e2e-polybft/e2e/...
+	go test -v -timeout=45m ./e2e-polybft/e2e/... -run TestAllowList_ContractDeployment
 
 test-property-polybft:
     # We can not build with race because of a bug in boltdb dependency
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
 	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true E2E_TESTS_TYPE=integration \
-	go test -v -timeout=45m ./e2e-polybft/e2e/... -run TestAllowList_ContractDeployment
+	go test -v -timeout=45m ./e2e-polybft/e2e/...
 
 test-property-polybft:
     # We can not build with race because of a bug in boltdb dependency
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -203,8 +203,15 @@ func setFlags(cmd *cobra.Command) {
 	// Allow list
 	{
 		cmd.Flags().StringArrayVar(
-			&params.adminAllowList,
-			"allow-list-admin",
+			&params.contractDeployerAllowListAdmin,
+			contractDeployedAllowListAdminFlag,
+			[]string{},
+			"",
+		)
+
+		cmd.Flags().StringArrayVar(
+			&params.contractDeployerAllowListEnabled,
+			contractDeployedAllowListEnabledFlag,
 			[]string{},
 			"",
 		)
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -206,14 +206,14 @@ func setFlags(cmd *cobra.Command) {
 			&params.contractDeployerAllowListAdmin,
 			contractDeployedAllowListAdminFlag,
 			[]string{},
-			"",
+			"list of addresses to use as admin accounts in the contract deployer allow list",
 		)
 
 		cmd.Flags().StringArrayVar(
 			&params.contractDeployerAllowListEnabled,
 			contractDeployedAllowListEnabledFlag,
 			[]string{},
-			"",
+			"list of addresses to enable by default in the contract deployer allow list",
 		)
 	}
 }
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -199,6 +199,16 @@ func setFlags(cmd *cobra.Command) {
 				"(format: <contract address>:<start block>)",
 		)
 	}
+
+	// Allow list
+	{
+		cmd.Flags().StringArrayVar(
+			&params.adminAllowList,
+			"allow-list-admin",
+			[]string{},
+			"",
+		)
+	}
 }
 
 // setLegacyFlags sets the legacy flags to preserve backwards compatibility
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -38,6 +38,9 @@ const (
 	defaultBridge           = false
 	defaultEpochReward      = 1
 
+	contractDeployedAllowListAdminFlag   = "contract-deployer-allow-list-admin"
+	contractDeployedAllowListEnabledFlag = "contract-deployer-allow-list-enabled"
+
 	bootnodePortStart = 30301
 )
 
@@ -178,15 +181,11 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 		Mixhash:    polybft.PolyBFTMixDigest,
 	}
 
-	// allowlist
-	if len(p.adminAllowList) != 0 {
-		addrs := []types.Address{}
-		for _, addr := range p.adminAllowList {
-			addrs = append(addrs, types.StringToAddress(addr))
-		}
-
-		chainConfig.Params.ContractsAllowList = &chain.AllowListConfig{
-			AdminAddresses: addrs,
+	if len(p.contractDeployerAllowListAdmin) != 0 || len(p.contractDeployerAllowListEnabled) != 0 {
+		// allow list for contract deployment is enabled
+		chainConfig.Params.ContractDeployerAllowList = &chain.AllowListConfig{
+			AdminAddresses:   stringSliceToAddressSlice(p.contractDeployerAllowListAdmin),
+			EnabledAddresses: stringSliceToAddressSlice(p.contractDeployerAllowListEnabled),
 		}
 	}
 
@@ -270,4 +269,12 @@ func generateExtraDataPolyBft(validators []*polybft.ValidatorMetadata) ([]byte, 
 	extra := polybft.Extra{Validators: delta, Checkpoint: &polybft.CheckpointData{}}
 
 	return append(make([]byte, polybft.ExtraVanity), extra.MarshalRLPTo(nil)...), nil
+}
+
+func stringSliceToAddressSlice(addrs []string) []types.Address {
+	res := []types.Address{}
+	for _, addr := range addrs {
+		res = append(res, types.StringToAddress(addr))
+	}
+	return res
 }
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -273,9 +273,9 @@ func generateExtraDataPolyBft(validators []*polybft.ValidatorMetadata) ([]byte, 
 }
 
 func stringSliceToAddressSlice(addrs []string) []types.Address {
-	res := []types.Address{}
-	for _, addr := range addrs {
-		res = append(res, types.StringToAddress(addr))
+	res := make([]types.Address, len(addrs))
+	for indx, addr := range addrs {
+		res[indx] = types.StringToAddress(addr)
 	}
 
 	return res
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -181,8 +181,9 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 		Mixhash:    polybft.PolyBFTMixDigest,
 	}
 
-	if len(p.contractDeployerAllowListAdmin) != 0 || len(p.contractDeployerAllowListEnabled) != 0 {
-		// allow list for contract deployment is enabled
+	if len(p.contractDeployerAllowListAdmin) != 0 {
+		// only enable allow list if there is at least one address as **admin**, otherwise
+		// the allow list could never be updated
 		chainConfig.Params.ContractDeployerAllowList = &chain.AllowListConfig{
 			AdminAddresses:   stringSliceToAddressSlice(p.contractDeployerAllowListAdmin),
 			EnabledAddresses: stringSliceToAddressSlice(p.contractDeployerAllowListEnabled),
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -277,5 +277,6 @@ func stringSliceToAddressSlice(addrs []string) []types.Address {
 	for _, addr := range addrs {
 		res = append(res, types.StringToAddress(addr))
 	}
+
 	return res
 }
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -178,6 +178,18 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 		Mixhash:    polybft.PolyBFTMixDigest,
 	}
 
+	// allowlist
+	if len(p.adminAllowList) != 0 {
+		addrs := []types.Address{}
+		for _, addr := range p.adminAllowList {
+			addrs = append(addrs, types.StringToAddress(addr))
+		}
+
+		chainConfig.Params.ContractsAllowList = &chain.AllowListConfig{
+			AdminAddresses: addrs,
+		}
+	}
+
 	return helper.WriteGenesisConfigToDisk(chainConfig, params.genesisPath)
 }
 
--- a/e2e-polybft/e2e/allowlist_test.go
+++ b/e2e-polybft/e2e/allowlist_test.go
@@ -1,0 +1,152 @@
+package e2e
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/0xPolygon/polygon-edge/e2e-polybft/framework"
+	"github.com/0xPolygon/polygon-edge/helper/hex"
+	"github.com/0xPolygon/polygon-edge/state/runtime/allowlist"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/stretchr/testify/require"
+	"github.com/umbracle/ethgo/compiler"
+	"github.com/umbracle/ethgo/wallet"
+)
+
+const (
+	statusFailed  = uint64(0)
+	statusSucceed = uint64(1)
+)
+
+var allowListE2EContractCode = `// SPDX-License-Identifier: MIT
+pragma solidity ^0.5.5;
+
+interface AllowListInterface {
+    function readAllowList() external view;
+}
+
+contract AllowListProxy {
+	AllowListInterface allowlist = AllowListInterface(0x0200000000000000000000000000000000000000);
+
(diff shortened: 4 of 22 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ test-e2e-polybft:
     # We can not build with race because of a bug in boltdb dependency
 	go build -o artifacts/polygon-edge .
 	env EDGE_BINARY=${PWD}/artifacts/polygon-edge E2E_TESTS=true E2E_LOGS=true E2E_TESTS_TYPE=integration \
-	go test -v -timeout=45m ./e2e-polybft/e2e/...
+	go test -v -timeout=45m ./e2e-polybft/e2e/... -run TestAllowList_ContractDeployment
```
ferranbt: This will get removed once all the testing is done.

command/genesis/polybft_params.go
```diff
@@ -178,6 +181,15 @@ func (p *genesisParams) generatePolyBftChainConfig() error {
 		Mixhash:    polybft.PolyBFTMixDigest,
 	}
 
+	if len(p.contractDeployerAllowListAdmin) != 0 {
+		// only enable allow list if there is at least one address as **admin**, otherwise
```
ferranbt, line 185: It does not make sense to enable the allow list if there is no admin key to manage it.

e2e-polybft/e2e/allowlist_test.go
```diff
@@ -0,0 +1,142 @@
...
+		input, _ := allowlist.SetEnabledSignatureFunc.Encode([]interface{}{types.ZeroAddress})
+
+		adminSetFailTxn := cluster.MethodTxn(t, target, allowlist.AllowListContractsAddr, input)
+		require.NoError(t, adminSetFailTxn.Wait())
+		require.True(t, adminSetFailTxn.Failed())
+		expectRole(types.ZeroAddress, allowlist.NoRole)
+	}
+
+	/*
+		fmt.Println("_ CONTRACT _", deployedContractAddr)
```
ferranbt: This is still pending.

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +591,185 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+	return t.SendTxn(tt, sender, &ethgo.Transaction{To: &targetAddr, Value: value})
+}
+
+func (t *TestCluster) MethodTxn(tt *testing.T, sender ethgo.Key, target types.Address, input []byte) *TestTxn {
+	targetAddr := ethgo.Address(target)
+	return t.SendTxn(tt, sender, &ethgo.Transaction{To: &targetAddr, Input: input})
+}
+
+// SendTxn sends a transaction
+func (t *TestCluster) SendTxn(tt *testing.T, sender ethgo.Key, txn *ethgo.Transaction) *TestTxn {
```
ferranbt: This is a wrapper around the execution of transactions in the `e2e` framework. I tried to use the relayer but I found it more helpful to have a custom thing on `TestCluster` to which we can attach specific methods and wrappers for testing.

state/executor.go
```diff
@@ -531,6 +543,11 @@ func (t *Transition) Call2(
 }
 
 func (t *Transition) run(contract *runtime.Contract, host runtime.Host) *runtime.ExecutionResult {
+	// check allow list (if any)
+	if t.deploymentAllowlist != nil && t.deploymentAllowlist.Addr() == contract.CodeAddress {
```
ferranbt, line 545: This enables the allow list to be reachable from other smart contracts and from transactions made by the user.

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +591,185 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
 
 	return result, nil
 }
+
+var (
+	defaultGasPrice = uint64(1879048192) // 0x70000000
```
b00ris: Is it make sense to reuse this variables from txrelayer package?

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +591,185 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+	data, err := hex.DecodeString(resp[2:])
+	require.NoError(tt, err)
+
+	output, err := method.Decode(data)
+	require.NoError(tt, err)
+
+	return output
+}
+
+type TestCall struct {
```
b00ris: Do you need this type?

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +591,185 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+}
+
+// Wait waits for the transaction to be executed
+func (t *TestTxn) Wait() error {
+	return t.WaitWithDuration(1 * time.Minute)
+}
+
+// WaitWithDuration waits for the transaction to be executed with a given timeout
+func (t *TestTxn) WaitWithDuration(timeout time.Duration) error {
+	for {
```
b00ris, line 751: Why do you need for there?

ferranbt: This was used before because I thought one of the transactions would never be executed.

server/server.go
```diff
@@ -201,6 +202,11 @@ func NewServer(config *Config) (*Server, error) {
 		m.executor.GenesisPostHook = factory(m.config.Chain, engineName)
 	}
 
+	// apply allow list genesis data
+	if m.config.Chain.Params.ContractDeployerAllowList != nil {
```
b00ris, line 206: Is it make sense to make a part of genesis post hook?

command/genesis/polybft_params.go
```diff
@@ -259,3 +271,12 @@ func generateExtraDataPolyBft(validators []*polybft.ValidatorMetadata) ([]byte,
 
 	return append(make([]byte, polybft.ExtraVanity), extra.MarshalRLPTo(nil)...), nil
 }
+
+func stringSliceToAddressSlice(addrs []string) []types.Address {
+	res := []types.Address{}
```
goran-ethernal: You can initialize it with `make([]types.Address, len(addrs))` and use index in the for loop.

e2e-polybft/framework/test-cluster.go
```diff
@@ -453,6 +486,12 @@ func (c *TestCluster) WaitUntil(dur time.Duration, handler func() bool) error {
 	}
 }
 
+func (c *TestCluster) WaitForReady(t *testing.T) {
```
goran-ethernal, line 490: Do you need this function? Seems like you can just call `WaitForBlock` in test.

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +593,207 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+	code, err := client.Eth().GetCode(addr, ethgo.Latest)
+	if err != nil {
+		return false
+	}
+
+	if code == "0x" {
+		return false
+	}
+
+	return true
```
goran-ethernal:
```suggestion
	return code != "0x"
```

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +593,207 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+}
+
+func (c *TestCluster) Call(t *testing.T, to types.Address, method *abi.Method,
```
goran-ethernal, line 612: I think you can use the transaction relayer for the most of the code here.

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +593,207 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+func (c *TestCluster) MethodTxn(t *testing.T, sender ethgo.Key, target types.Address, input []byte) *TestTxn {
+	t.Helper()
+
+	targetAddr := ethgo.Address(target)
+
+	return c.SendTxn(t, sender, &ethgo.Transaction{To: &targetAddr, Input: input})
+}
+
+// SendTxn sends a transaction
+func (c *TestCluster) SendTxn(t *testing.T, sender ethgo.Key, txn *ethgo.Transaction) *TestTxn {
```
goran-ethernal, line 663: You can use the transaction relayer as well in this function, and remove bunch of code.

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,179 @@
...
+type Role types.Hash
+
+var (
+	NoRole      Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000000"))
+	EnabledRole Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000001"))
+	AdminRole   Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000002"))
+)
+
+func (r Role) Uint64() uint64 {
+	if r == EnabledRole {
```
goran-ethernal:
```suggestion
	 switch (r) {
               case EnabledRole: return 1
               case AdminRole: return 2
               default: return 0
         }
```

server/server.go
ferranbt: Genesis post hook was meant for specific consensus changes. Not for blockchain-wide configuration changes like this. Technically, allow list would apply to all consensus.

e2e-polybft/framework/test-cluster.go
ferranbt: I commented out on this [here](https://github.com/0xPolygon/polygon-edge/pull/1306#discussion_r1139882067)

e2e-polybft/framework/test-cluster.go
ferranbt: So we always do an initial wait for block to make sure everything is ready, instead of writing `WaitForBlock(n)` this abstract that into its own concept.

e2e-polybft/framework/test-cluster.go
goran-ethernal: Yeah, but the code for creating transaction, getting nonce, signing and sending is practically the same. I'm maybe missing the point, but I am not sure what else could we need there. If you wanted to make some custom processing of a transaction execution result, then I would say we can just let the `relayer` return the result, and then have a custom function to process the result. For waiting of the transaction receipt, we have that `WithReceiptTimeout` config function that allows us to configure a larger duration to wait for the receipt on the `relayer`.

ferranbt: Yep, so initially, I was doing a bit more magic on the `TestTxn` type, for example:
```
.SendTxn().
  OnFinished(func() error { // custom on finish logic }).
  Wait()
```
but you are right that now it does not have any of those methods. Lemme see how many things I can move to the relayer.

ferranbt: I had to remember a little bit about why I did this. So the `txrelayer` was not flexible enough for some of the patterns I had in mind (i.e send a specific transaction that you expect to fail) so this is how it evolved.

I think some (if not all) of the patterns can be moved to the `txrelayer` but I wanted to avoid changing code in other places out of the scope of this PR. I propose to have a followup PR to extend the usage of this framework to other unit tests (i.e. txpool e2e test would be sliced in half) to gather a bit of requirements and then move everything to `txrelayer`.

goran-ethernal: Agreed.

e2e-polybft/framework/test-cluster.go
b00ris: Just notice, that timeout will be recreated for every loop step(it's not a timeout for the whole function).

b00ris:
Good job! Generally - good for me.

b00ris (approved):

goran-ethernal (approved):
LGTM. Just don't forget to edit the PR description :muscle:

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return res
+}
+
+var (
+	errNoFunctionSignature = fmt.Errorf("input is too short for a function call")
+	errInputTooShort       = fmt.Errorf("wrong input size, expected 32")
+	errFunctionNotFound    = fmt.Errorf("function not found")
+	errWriteProtection     = fmt.Errorf("write protection")
+	errNotAuth             = fmt.Errorf("not auth")
```
Stefan-Ethernal: WDYT about using `ErrNotAuth` from runtime.go?

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+func (a *AllowList) runInputCall(caller types.Address, input []byte,
+	gas uint64, isStatic bool) ([]byte, uint64, error) {
+	// decode the function signature from the input
+	if len(input) < 4 {
```
Stefan-Ethernal: Generally speaking, would ge great to have this as a constant somewhere in `types` or `common` package and use it from there. WDYT?

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, 0, errNoFunctionSignature
+	}
+
+	sig, inputBytes := input[:4], input[4:]
+
+	// all the functions have the same input (i.e. tuple(address)) which
+	// in abi gets codified as a 32 bytes array with the first 20 bytes
+	// encoding the address
+	if len(inputBytes) != 32 {
```
Stefan-Ethernal, line 72: Generally speaking, would ge great to have these constants somewhere in `types` or `common` package and use it from there. WDYT?

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return Role(res)
+}
+
+type Role types.Hash
+
+var (
+	NoRole      Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000000"))
+	EnabledRole Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000001"))
+	AdminRole   Role = Role(types.StringToHash("0x0000000000000000000000000000000000000000000000000000000000000002"))
```
Stefan-Ethernal, line 148: Just use a shorthand, it should be enough. WDYT?

```suggestion
	NoRole      Role = Role(types.StringToHash("0x0"))
	EnabledRole Role = Role(types.StringToHash("0x1"))
	AdminRole   Role = Role(types.StringToHash("0x2"))
```

state/runtime/allowlist/allowlist.go
```diff
@@ -0,0 +1,178 @@
...
+	"fmt"
+
+	"github.com/0xPolygon/polygon-edge/chain"
+	"github.com/0xPolygon/polygon-edge/state/runtime"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/umbracle/ethgo/abi"
+)
+
+var (
+	AllowListContractsAddr = types.StringToAddress("0x0200000000000000000000000000000000000000")
```
Stefan-Ethernal: WDYT about moving this to `system_addresses.go?` The intention is to have those system-predefined addresses in one place.

command/genesis/genesis.go
```diff
@@ -199,6 +199,23 @@ func setFlags(cmd *cobra.Command) {
...
+			contractDeployedAllowListAdminFlag,
+			[]string{},
+			"",
+		)
+
+		cmd.Flags().StringArrayVar(
+			&params.contractDeployerAllowListEnabled,
+			contractDeployedAllowListEnabledFlag,
+			[]string{},
+			"",
```
Stefan-Ethernal: Add a description to the flag?

command/genesis/genesis.go
```diff
@@ -199,6 +199,23 @@ func setFlags(cmd *cobra.Command) {
...
 		)
 	}
+
+	// Allow list
+	{
+		cmd.Flags().StringArrayVar(
+			&params.contractDeployerAllowListAdmin,
+			contractDeployedAllowListAdminFlag,
+			[]string{},
+			"",
```
Stefan-Ethernal: Add a descriptionto the flag?

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +594,195 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+
+// Receipt returns the receipt of the transaction
+func (t *TestTxn) Receipt() *ethgo.Receipt {
+	return t.receipt
+}
+
+const (
+	statusFailed  = uint64(0)
+	statusSucceed = uint64(1)
+)
```
Stefan-Ethernal: Didn't want to use constants from the `types` package?

```go
type ReceiptStatus uint64

const (
	ReceiptFailed ReceiptStatus = iota
	ReceiptSuccess
)
```

e2e-polybft/framework/test-cluster.go
```diff
@@ -554,3 +594,195 @@ func (c *TestCluster) InitSecrets(prefix string, count int) ([]types.Address, er
...
+				return nil
+			}
+
+		case <-time.After(timeout):
+			return fmt.Errorf("timeout")
+		}
+	}
+}
+
+func sliceArrayToSliceString(addrs []types.Address) []string {
```
Stefan-Ethernal: Rename?
```suggestion
func sliceAddressToSliceString(addrs []types.Address) []string {
```

state/runtime/allowlist/genesis.go
```diff
@@ -0,0 +1,51 @@
...
+	if alloc.Storage == nil {
+		alloc.Storage = map[types.Hash]types.Hash{}
+	}
+
+	alloc.Storage[key] = value
+}
+
+func (g *genesisState) GetStorage(addr types.Address, key types.Hash) types.Hash {
+	panic("not used")
+}
```
Stefan-Ethernal: Maybe to consider this to be no-op implementation, just returning `types.ZeroHash` all the time. Is it a better idea? I know that intention 
here is to prevent somebody to call this function, but again, not sure is the panic right way.

Would it harm if we were returning `Storage` from the `alloc` (if I'm correct there)?
So it would go like this:

```go
if alloc, exists := g.chain.Alloc[addr];exists {
    return alloc.Storage[key]
}

return types.ZeroHash
```

Anyways if nothing of afore-mentioned is not appropriate, then rebase PR on `develop`, since Goran introduced custom linting rule, which prevents explicit panicing, so suppress it here.

Stefan-Ethernal (commented):
Generally looks good, but leaving some thoughts to consider.

state/runtime/allowlist/allowlist.go
ferranbt: I would agree with the signature size, but this is specific for this input and only used once.

state/runtime/allowlist/allowlist.go
ferranbt: Well, this is a hack that `StringToHash` does. I prefer to be explicit here so that we are fully aware that `0x1` is `0x0...1` and not `0x01...0`.

state/runtime/allowlist/genesis.go
ferranbt: It would not really affect it because it cannot ever be called at all. But given Goran work on the linting rules, I will return an empty `types.Hash` and comment out that the function will never be called.

Stefan-Ethernal (approved):

PR #1310, opened 2023-03-17
Fix proposer logging for HasQuorum function

stana-miric:
In HasQuorum fn, if the view(height or round) from the message does not match the view from the proposer calculator, that will be logged as debug instead of warning and false should be returned. This is valid case because node can run the sequence for the height lower than the height from calculator since the calculator is updated on block insertion (which can come from the syncer as well and thus the proposer calculator can have different height that message). 

# Changes include

# Checklist

## Testing


Stefan-Ethernal (approved):

Nemanja0x (approved):

vcastellm (approved):
LGTM

b00ris (approved):

PR #1312, opened 2023-03-19
Simplify state transactions function in FSM

Stefan-Ethernal:
# Description

As `stateTransactions` function returns only single transaction, which is the one for bridge commitment registration, it is renamed and it's signature is changed, so it returns only single instance instead of slice.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -195,7 +195,7 @@ func (f *fsm) applyBridgeCommitmentTx() error {
 	if f.proposerCommitmentToRegister != nil {
 		bridgeCommitmentTx, err := f.createBridgeCommitmentTx()
 		if err != nil {
-			return err
+			return fmt.Errorf("creation of bridge commitment transaction failed: %w", err)
 		}
 
 		if err := f.blockBuilder.WriteTx(bridgeCommitmentTx); err != nil {
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -114,10 +114,8 @@ func (f *fsm) BuildProposal(currentRound uint64) ([]byte, error) {
 	}
 
 	if f.config.IsBridgeEnabled() {
-		if bridgeCommitmentTx := f.getBridgeCommitmentTx(); bridgeCommitmentTx != nil {
-			if err := f.blockBuilder.WriteTx(bridgeCommitmentTx); err != nil {
-				return nil, fmt.Errorf("failed to apply commitment state sync state transaction. Error: %w", err)
-			}
+		if err := f.applyBridgeCommitmentTx(); err != nil {
+			return nil, err
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


consensus/polybft/fsm.go
```diff
@@ -184,25 +184,20 @@ func (f *fsm) BuildProposal(currentRound uint64) ([]byte, error) {
 	return stateBlock.Block.MarshalRLP(), nil
 }
 
-func (f *fsm) stateTransactions() []*types.Transaction {
-	var txns []*types.Transaction
-
+// getBridgeCommitmentTx builds state transaction which contains data for bridge commitment registration
+func (f *fsm) getBridgeCommitmentTx() *types.Transaction {
```
igorcrevar: Maybe we can make this function to be `writeBridgeTx` and do actual `f.blockBuilder.WriteTx(` here, so we dont need two ifs outside?

Stefan-Ethernal: Is it better like this?

https://github.com/0xPolygon/polygon-edge/pull/1312/commits/3c04e6ae93fa6903e6f83138094723f3abb4b7ee

consensus/polybft/fsm.go
```diff
@@ -192,25 +190,34 @@ func (f *fsm) BuildProposal(currentRound uint64) ([]byte, error) {
...
-
+// applyBridgeCommitmentTx builds state transaction which contains data for bridge commitment registration
+func (f *fsm) applyBridgeCommitmentTx() error {
 	if f.proposerCommitmentToRegister != nil {
-		// add register commitment transaction
-		inputData, err := f.proposerCommitmentToRegister.EncodeAbi()
+		bridgeCommitmentTx, err := f.createBridgeCommitmentTx()
 		if err != nil {
-			f.logger.Error("StateTransactions failed to encode input data for state sync commitment registration", "Error", err)
+			return err
```
b00ris: Could you wrap the error?

b00ris (approved):

igorcrevar (approved):

consensus/polybft/fsm.go
```diff
@@ -192,25 +190,34 @@ func (f *fsm) BuildProposal(currentRound uint64) ([]byte, error) {
...
+	}
 
-		txns = append(txns,
-			createStateTransactionWithData(f.config.StateReceiverAddr, inputData))
+	return nil
+}
+
+// createBridgeCommitmentTx builds bridge commitment registration transaction
+func (f *fsm) createBridgeCommitmentTx() (*types.Transaction, error) {
+	if f.proposerCommitmentToRegister == nil {
```
goran-ethernal: I think this check is redundant, since in the `applyBridgeCommitmentTx` function we check if this is nil.

goran-ethernal (approved):
Except for a small comment, LGTM :100:

consensus/polybft/fsm.go
Stefan-Ethernal: I have added it because of unit tests (just in case somebody calls it without setting `proposerCommitmentToRegister` field) and to prevent any unwanted panics in future (better to return an error, just in case).

consensus/polybft/fsm.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1312/commits/f5b15a7d1471a073e50fb208ceb26d644930e583

consensus/polybft/fsm.go
goran-ethernal: I agree with you what you said about panics, but I would not conform the client code itself to tests, especially since we don't  expect this to be called if `proposerCommitmentToRegister` is not set.

Stefan-Ethernal: Fixed [68c80c0](https://github.com/0xPolygon/polygon-edge/pull/1312/commits/68c80c07fcd84bd0d22f00f8bf6000b45befb148)

PR #1313, opened 2023-03-20
Disable console precompile for prod networks

ferranbt:
# Description

Only local development networks have a use case to deploy the `console.log` utilities, but, in the production scenarios in which `Edge` is meant to run, this precompile is not necessary and could introduce attack vectors (i.e. Ddos the logs). This PR disables the precompile altogether. I would propose discussing two options to follow this up:
- To introduce a `--console` flag that enables the precompile. This could only run in local networks and not distributed networks since it would break consensus (some nodes logging and others not and consuming gas).
- To leave it like this and uncomment the line in the cases where we need to debug something from the local smart contracts.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/state/runtime/precompiled/precompiled.go
+++ b/state/runtime/precompiled/precompiled.go
@@ -77,7 +77,7 @@ func (p *Precompiled) setupContracts() {
 	p.register(contracts.BLSAggSigsVerificationPrecompile.String(), &blsAggSignsVerification{})
 
 	// Console precompile
-	p.register(contracts.ConsolePrecompile.String(), &console{})
+	// p.register(contracts.ConsolePrecompile.String(), &console{})
 }
 
 func (p *Precompiled) register(addrStr string, b contract) {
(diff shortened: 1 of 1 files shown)


state/runtime/precompiled/precompiled.go
```diff
@@ -77,7 +77,7 @@ func (p *Precompiled) setupContracts() {
 	p.register(contracts.BLSAggSigsVerificationPrecompile.String(), &blsAggSignsVerification{})
 
 	// Console precompile
-	p.register(contracts.ConsolePrecompile.String(), &console{})
+	// p.register(contracts.ConsolePrecompile.String(), &console{})
```
Stefan-Ethernal, line 77: Linter complains, because the block is ending with a comment. We need to trick it out somehow...

Stefan-Ethernal (approved):
Linter nitpick, otherwise LGTM.

I'm for creating a simple flag that enables console precompile via CLI flag, I think it is more convenient than (un)commenting code itself.

Nemanja0x (approved):

state/runtime/precompiled/precompiled.go
ferranbt: Solved!

PR #1320, opened 2023-03-21
Some SyncPeerStatus messages were not published

stana-miric:
The problem fixed in this PR is related to publishing SyncPeerStatus. Some messages were not published because of the bug in startNewBlockProcess fn of syncPeerClient where each time after the notification is dispatched on block insertions the new channel is created for the same syncPeerClient subscription. This could be noticed on non validator nodes since they are not producing block but instead receiving them upon SyncPeerStatus is published (when syncer's newStatusCh is set). On those nodes as reproduced on the test system as well, blocks were not arriving one by one.
-The main fix is in syncer/client.go
-Unit test is written and for that purpose,  mocked subscriber is adjusted to look more like real subscriber in order to be able to test this scenario (it does not affect other test, the previous version was more simplified because it was sufficient for existing tests). 
-blockchain.go is changed because source of the block is added to the log since we have that data and it can be useful 

# Changes include

# Checklist

## Testing

### DIFF
--- a/blockchain/subscription.go
+++ b/blockchain/subscription.go
@@ -19,28 +19,51 @@ type Subscription interface {
 // FOR TESTING PURPOSES //
 
 type MockSubscription struct {
-	eventCh chan *Event
+	updateCh chan *Event // Channel for update information
+	closeCh  chan void   // Channel for close signals
 }
 
 func NewMockSubscription() *MockSubscription {
-	return &MockSubscription{eventCh: make(chan *Event)}
+	return &MockSubscription{
+		updateCh: make(chan *Event),
+		closeCh:  make(chan void),
+	}
 }
 
 func (m *MockSubscription) Push(e *Event) {
-	m.eventCh <- e
+	m.updateCh <- e
 }
 
 func (m *MockSubscription) GetEventCh() chan *Event {
-	return m.eventCh
+	eventCh := make(chan *Event)
+
+	go func() {
+		for {
+			evnt := m.GetEvent()
+			if evnt == nil {
+				return
+			}
+			eventCh <- evnt
+		}
+	}()
+
+	return eventCh
 }
 
 func (m *MockSubscription) GetEvent() *Event {
-	evnt := <-m.eventCh
-
-	return evnt
+	for {
+		// Wait for an update
+		select {
+		case ev := <-m.updateCh:
+			return ev
+		case <-m.closeCh:
+			return nil
+		}
+	}
 }
 
 func (m *MockSubscription) Close() {
+	close(m.closeCh)
 }
 
 // subscription is the Blockchain event subscription object
--- a/blockchain/subscription.go
+++ b/blockchain/subscription.go
@@ -19,49 +19,19 @@ type Subscription interface {
 // FOR TESTING PURPOSES //
 
 type MockSubscription struct {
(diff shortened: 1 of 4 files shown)


blockchain/subscription.go
```diff
@@ -19,28 +19,51 @@ type Subscription interface {
 // FOR TESTING PURPOSES //
 
 type MockSubscription struct {
```
goran-ethernal, line 21: Isn't this `MockSubscription` now the same implementation as the `subscription` struct that is actually used. Could we just expose that `subscription` struct and use it in tests instead of having this mock?

stana-miric: Mocked implementation has Push fn added, but also I'm not sure if we want to expose the real implementation(was thinking about that)

goran-ethernal: Then I would say, make `MockSubscription` an extension of `subscription`, and remove all functions except `Push`.

```suggestion
type MockSubscription struct {
	*subscription
}

func NewMockSubscription() *MockSubscription {
	return &MockSubscription{
		subscription: &subscription{
			updateCh: make(chan *Event),
			closeCh:  make(chan void),
		},
	}
}

func (m *MockSubscription) Push(e *Event) {
	m.updateCh <- e
}
```

igorcrevar (approved):

Nemanja0x (approved):

blockchain/subscription.go
stana-miric: fixed f973b3b

goran-ethernal (approved):

vcastellm (approved):
🎩 LGTM

syncer/client_test.go
```diff
@@ -524,3 +524,102 @@ func Test_syncPeerClient_GetBlocks(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	testGossip := func(t *testing.T, blocksNum int) {
+		t.Helper()
+
+		var wgForGossip sync.WaitGroup
+
+		wgForGossip.Add(blocksNum)
+
+		assert.NoError(t, topic.Subscribe(func(_ interface{}, id peer.ID) {
```
Stefan-Ethernal:
```suggestion
		assert.NoError(t, topic.Subscribe(func(_ interface{}, _ peer.ID) {
```

syncer/client_test.go
```diff
@@ -524,3 +524,102 @@ func Test_syncPeerClient_GetBlocks(t *testing.T) {
...
+		client.Close()
+	})
+
+	err := network.JoinAndWaitMultiple(
+		network.DefaultJoinTimeout,
+		clientSrv,
+		peerSrv,
+	)
+
+	assert.NoError(t, err)
```
Stefan-Ethernal: WDYT about using requires instead of assert in the entire test? I believe test should fail as soon as any of the assertions fails?

```suggestion
	require.NoError(t, err)
```

Stefan-Ethernal (approved):
Aside from nitpicks, LGTM. Nice one

syncer/client_test.go
stana-miric: fixed 0c17957

syncer/client_test.go
stana-miric: fixed 0c17957

PR #1321, opened 2023-03-21
EVM-535: Use contractsapi bindings throught code

goran-ethernal:
# Description

This PR uses the `contractsapi` package and generated `go` stubs from contract functions and events, on places where it was missing.

PR expanded the `ParseLog` function on generated `go` stubs for contract events, so that the function returns two values:
- `boolean` - indicating if the given log matches the given event.
- `error` - if a log matches the event, but its parsing fails for some reason.

This simplifies the code on client that first matched the events with log, and then parsed it, shortening the code, and removing the manually added references to an event abi.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/bridge/exit/exit.go
+++ b/command/bridge/exit/exit.go
@@ -233,7 +233,8 @@ func createExitTxn(sender ethgo.Address, proof types.Proof) (*ethgo.Transaction,
 		return nil, nil, fmt.Errorf("failed to unmarshal exit event from JSON. Error: %w", err)
 	}
 
-	exitEventEncoded, err := polybft.ExitEventInputsABIType.Encode(exitEvent)
+	var exitEventApi contractsapi.L2StateSyncedEvent
+	exitEventEncoded, err := exitEventApi.Encode(exitEvent)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to encode exit event: %w", err)
 	}
--- a/command/bridge/exit/exit.go
+++ b/command/bridge/exit/exit.go
@@ -233,8 +233,9 @@ func createExitTxn(sender ethgo.Address, proof types.Proof) (*ethgo.Transaction,
 		return nil, nil, fmt.Errorf("failed to unmarshal exit event from JSON. Error: %w", err)
 	}
 
-	var exitEventApi contractsapi.L2StateSyncedEvent
-	exitEventEncoded, err := exitEventApi.Encode(exitEvent)
+	var exitEventAPI contractsapi.L2StateSyncedEvent
+
+	exitEventEncoded, err := exitEventAPI.Encode(exitEvent)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to encode exit event: %w", err)
 	}
--- a/command/sidechain/registration/register_validator.go
+++ b/command/sidechain/registration/register_validator.go
@@ -4,7 +4,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math/big"
 
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/0xPolygon/polygon-edge/command/helper"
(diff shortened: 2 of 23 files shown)


command/sidechain/registration/register_validator.go
```diff
@@ -21,10 +20,8 @@ import (
 )
 
 var (
-	stakeManager         = contracts.ValidatorSetContract
-	stakeFn              = contractsapi.ChildValidatorSet.Abi.Methods["stake"]
-	newValidatorEventABI = contractsapi.ChildValidatorSet.Abi.Events["NewValidator"]
-	stakeEventABI        = contractsapi.ChildValidatorSet.Abi.Events["Staked"]
+	stakeManager = contracts.ValidatorSetContract
+	stakeFn      = contractsapi.ChildValidatorSet.Abi.Methods["stake"]
```
Stefan-Ethernal, line 24: Didn't want to create a binding for it?

consensus/polybft/checkpoint_manager_test.go
```diff
@@ -520,7 +520,7 @@ func createTestLogForExitEvent(t *testing.T, exitEventID uint64) *types.Log {
 	t.Helper()
 
 	topics := make([]types.Hash, 4)
-	topics[0] = types.Hash(ExitEventABI.ID())
+	topics[0] = types.Hash(contractsapi.L2StateSender.Abi.Events["L2StateSynced"].ID())
```
Stefan-Ethernal: Think it would be beneficial to wrap ethgo.Event to each of the generated event bindings. Then you could use `ID()` function, instead of retrieveing event from the map. WDYT?

Ideally, outside of `contractsapi.go` we don't need to access the `Events` map from the `ABI` instance. Agree?

Stefan-Ethernal (approved):
Nice.

Leaving some thoughts to consider.

command/sidechain/registration/register_validator.go
goran-ethernal: Didn't want to create it because `stake` function doesn't have arguments on the contract, so the generated struct would be empty :confused:

consensus/polybft/checkpoint_manager_test.go
goran-ethernal: Fixed: https://github.com/0xPolygon/polygon-edge/pull/1321/commits/24b7b7904d4d31a6604b86f0a56c5fab848cca8f

command/bridge/exit/exit.go
```diff
@@ -233,7 +233,9 @@ func createExitTxn(sender ethgo.Address, proof types.Proof) (*ethgo.Transaction,
 		return nil, nil, fmt.Errorf("failed to unmarshal exit event from JSON. Error: %w", err)
 	}
 
-	exitEventEncoded, err := polybft.ExitEventInputsABIType.Encode(exitEvent)
+	var exitEventAPI contractsapi.L2StateSyncedEvent
+
+	exitEventEncoded, err := exitEventAPI.Encode(exitEvent)
```
vcastellm, line 238: This API looks a bit strange, should `Encode` and other methods be moved to a non-member method?

goran-ethernal: We could move `Encode` function on events to non-member method, since it doesn't change the state of the object (in this case `L2StateSyncedEvent`). I just wanted to keep it tight to the corresponding stubs. But yeah, I can move it to non-member. Just keep in mind we would have `n` number of `Encode{NameOfEvent}` functions, where `n` is the number of event stubs we generate.

vcastellm: this generated API doesn't looks neat at all, but it's ok for now but we can remove the object in the member method. `(o *Object) Method() {}` -> `(*Object) Method() {}` if `o` is not used in the member method

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1321/commits/ba6096ec656b5b8b1a27892a9f6f61a33f493f95

vcastellm (approved):

PR #1322, opened 2023-03-22
Deploy NFT contracts

Stefan-Ethernal:
# Description

This PR introduces NFT predicates and mock NFT tokens to the Edge.

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [ ] I have assigned this PR to myself
- [ ] I have added at least 1 reviewer
- [ ] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [ ] I have tested this code with the official test suite
- [ ] I have tested this code manually

PR #1323, opened 2023-03-22
Add EIP1559Burn contract to genesis rootchain command 

begmaroman:
# Description

Add EIP1559Burn contract to genesis rootchain command 

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -272,6 +272,11 @@ func (p *genesisParams) deployContracts(totalStake *big.Int) (map[types.Address]
 			artifact: contractsapi.L2StateSender,
 			address:  contracts.L2StateSenderContract,
 		},
+		{
+			// EIP1559Burn contract
+			artifact: contractsapi.EIP1559Burn,
+			address:  contracts.EIP1559BurnContract,
+		},
 	}
 
 	allocations := make(map[types.Address]*chain.GenesisAccount, len(genesisContracts))
--- a/contracts/system_addresses.go
+++ b/contracts/system_addresses.go
@@ -19,6 +19,8 @@ var (
 	ChildERC20Contract = types.StringToAddress("0x1003")
 	// ChildERC20PredicateContract is an address of child ERC20 predicate contract on the child chain
 	ChildERC20PredicateContract = types.StringToAddress("0x1004")
+	// EIP1559BurnContract is an address of EIP-1559 burn contract on the child chain
+	EIP1559BurnContract = types.StringToAddress("0x1005")
 	// SystemCaller is address of account, used for system calls to smart contracts
 	SystemCaller = types.StringToAddress("0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE")
 
--- a/contracts/system_addresses.go
+++ b/contracts/system_addresses.go
@@ -20,7 +20,7 @@ var (
 	// ChildERC20PredicateContract is an address of child ERC20 predicate contract on the child chain
 	ChildERC20PredicateContract = types.StringToAddress("0x1004")
 	// EIP1559BurnContract is an address of EIP-1559 burn contract on the child chain
-	EIP1559BurnContract = types.StringToAddress("0x1005")
+	EIP1559BurnContract = types.StringToAddress("0x104")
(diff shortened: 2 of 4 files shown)


command/genesis/polybft_params.go
```diff
@@ -272,6 +272,11 @@ func (p *genesisParams) deployContracts(totalStake *big.Int) (map[types.Address]
 			artifact: contractsapi.L2StateSender,
 			address:  contracts.L2StateSenderContract,
 		},
+		{
```
Stefan-Ethernal, line 275: Don't you need to invoke initialize function?
https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/EIP1559Burn.sol#L29-L34

contracts/system_addresses.go
```diff
@@ -19,6 +19,8 @@ var (
 	ChildERC20Contract = types.StringToAddress("0x1003")
 	// ChildERC20PredicateContract is an address of child ERC20 predicate contract on the child chain
 	ChildERC20PredicateContract = types.StringToAddress("0x1004")
+	// EIP1559BurnContract is an address of EIP-1559 burn contract on the child chain
+	EIP1559BurnContract = types.StringToAddress("0x1005")
```
Stefan-Ethernal: WDYT about deploying it to the `0x104` address? Asking this, because I'm going to deploy some other bridge-related smart contracts to this address range ([0x1005...0x1008]): https://github.com/0xPolygon/polygon-edge/pull/1322/files#diff-cb0d9c0c8a217bb409554571a2a979efd58f1dbf13e059824db4b1df98abed3cR24-R30. 

I think it is a good idea to have bridge contracts clustered in the same range. Not sure who is deciding about addresses, but maybe to raise that topic within the team.

Stefan-Ethernal:
Not sure who is accountable for calling `withdraw` function (https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/EIP1559Burn.sol#L40). Is there going to be some parallel process, which is going to call withdraw periodically or...?

begmaroman:
> Not sure who is accountable for calling withdraw function (https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/EIP1559Burn.sol#L40). Is there going to be some parallel process, which is going to call withdraw periodically or...?

@Stefan-Ethernal something like that. WIll be done in the next PR.

contracts/system_addresses.go
begmaroman: Done

command/genesis/polybft_params.go
begmaroman: We do need it, but I don't think it should be done within the contracts deployment command. It will be done in a different place tho.

contracts/system_addresses.go
```diff
@@ -19,6 +19,8 @@ var (
 	ChildERC20Contract = types.StringToAddress("0x1003")
 	// ChildERC20PredicateContract is an address of child ERC20 predicate contract on the child chain
 	ChildERC20PredicateContract = types.StringToAddress("0x1004")
+	// EIP1559BurnContract is an address of EIP-1559 burn contract on the child chain
+	EIP1559BurnContract = types.StringToAddress("0x104")
```
Stefan-Ethernal, line 23: Nitpick:
move it below `MerkleContract` (https://github.com/0xPolygon/polygon-edge/pull/1323/files#diff-cb0d9c0c8a217bb409554571a2a979efd58f1dbf13e059824db4b1df98abed3cR11), so that the addresses are ordered by values ascending.

Stefan-Ethernal (approved):

command/genesis/polybft_params.go
```diff
@@ -272,6 +272,11 @@ func (p *genesisParams) deployContracts(totalStake *big.Int) (map[types.Address]
...
+			// EIP1559Burn contract
+			artifact: contractsapi.EIP1559Burn,
```
ferranbt, line 277: From what I understand, this is a requirement for EIP-1559. But, what would happen if a network starts without it (in one our of the current versions) and then enables it, and the contract is not there? It would break forever I guess.

begmaroman: So it would make more sense to don't pre-deploy it here at all. Also, this address could be changed as well. Good point.

PR #1330, opened 2023-03-23
EVM-475 Create a Relayer Service logging + cors

igorcrevar:
# Description

Some additional this for AA I did not implement:
- logger
- rest server CORS

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/aarelayer/params.go
+++ b/command/aarelayer/params.go
@@ -10,6 +10,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/command/helper"
 	"github.com/0xPolygon/polygon-edge/command/polybftsecrets"
 	sidechainHelper "github.com/0xPolygon/polygon-edge/command/sidechain"
+	"github.com/hashicorp/go-hclog"
 	"github.com/spf13/cobra"
 )
 
@@ -18,8 +19,11 @@ const (
 	dbPathFlag      = "db-path"
 	chainIDFlag     = "chain-id"
 	invokerAddrFlag = "invoker-addr"
+	logPathFlag     = "log-path"
+	logLevelFlag    = "log-level"
 
-	defaultPort = 8198
+	defaultLogLevel = "info"
+	defaultPort     = 8198
 )
 
 type aarelayerParams struct {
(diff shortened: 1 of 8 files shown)


command/aarelayer/service/aa_relayer_service.go
```diff
@@ -9,9 +9,12 @@ import (
 	"time"
 
 	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/hashicorp/go-hclog"
 	"github.com/umbracle/ethgo"
 )
 
+const receiptSuccess = 1
```
Stefan-Ethernal, line 16: Use `types.ReceiptSuccess` instead.
```suggestion
```

Stefan-Ethernal (approved):

command/aarelayer/service/aa_relayer_service.go
igorcrevar: I know, but this will be moved to separate repository, so I am trying to depend on edge fw as less as I can

Stefan-Ethernal: Ok, though you are already relying on `types` package, so it makes sense to start decoupling it.
https://github.com/0xPolygon/polygon-edge/pull/1330/files#diff-fdffb0664e91ae7d97d676b8b7f901903b54a0fae76111e0069bea688c96349bR11

command/aarelayer/params.go
```diff
@@ -54,6 +60,28 @@ func (rp *aarelayerParams) validateFlags() error {
...
+
+		return hclog.New(&hclog.LoggerOptions{
+			Name:       "aarelayer",
+			Output:     logFileWriter,
+			Level:      hclog.LevelFromString(rp.logLevel),
+			JSONFormat: false,
+		}), nil
+	}
+
+	return hclog.New(&hclog.LoggerOptions{
```
goran-ethernal, line 78: You could remove the nested return and just set a local variable `var outputter io.Writer` and set it in the lower return. In case the log path was not provided, it will be `nil`, and `hclog` will use its `DefaultOutput`.

goran-ethernal (approved):
Except for a minor comment, LGTM

PR #1332, opened 2023-03-24
ERC 1155 tokens bridge integration

Stefan-Ethernal:
# Description

This PR integrates predicate contracts for bridging ERC 1155 tokens from a root to a child chain (and vice versa).

# Changes include

# Breaking changes

PR deploys a set of  ERC 1155 bridge contracts to the genesis.

# Checklist

## Testing

# Additional comments

PR is dependent upon https://github.com/0xPolygon/core-contracts/pull/204.


command/bridge/withdraw/erc1155/withdraw_erc1155.go
```diff
@@ -0,0 +1,261 @@
...
+// extractExitEventID tries to extract exit event id from provided receipt
+func extractExitEventID(receipt *ethgo.Receipt) (*big.Int, error) {
+	var exitEvent contractsapi.L2StateSyncedEvent
+	for _, log := range receipt.Logs {
+		doesMatch, err := exitEvent.ParseLog(log)
+		if !doesMatch {
+			continue
+		}
+
+		if err != nil {
```
igorcrevar, line 218: Maybe handle error first (if error occurs I suppose `doesMatch` will be false in most cases) and then match case later:
```
		doesMatch, err := exitEvent.ParseLog(log)		
		if err != nil {
			return nil, err
		}

                if doesMatch {
			return exitEvent.ID, nil
		}
```

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1332/commits/8bb5ab6d1c470d7037b91d446da3857e673268b5

command/bridge/withdraw/erc1155/withdraw_erc1155.go
```diff
@@ -0,0 +1,261 @@
...
+
+	_ = withdrawCmd.MarkFlagRequired(common.ReceiversFlag)
+	_ = withdrawCmd.MarkFlagRequired(common.AmountsFlag)
+	_ = withdrawCmd.MarkFlagRequired(common.TokenIDsFlag)
+
+	return withdrawCmd
+}
+
+func preRun(_ *cobra.Command, _ []string) error {
+	if err := wp.Validate(); err != nil {
```
igorcrevar: `return wp.Validate()`

command/bridge/deposit/erc1155/deposit_erc1155.go
```diff
@@ -0,0 +1,328 @@
...
+	cmdHelper "github.com/0xPolygon/polygon-edge/command/helper"
+	"github.com/0xPolygon/polygon-edge/command/rootchain/helper"
+	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
+	"github.com/0xPolygon/polygon-edge/txrelayer"
+	"github.com/0xPolygon/polygon-edge/types"
+)
+
+type depositERC1155Params struct {
+	*common.ERC1155BridgeParams
+	testMode bool
```
goran-ethernal, line 23: I think `testMode` can also go to that `BridgeParams` struct.

command/bridge/common/params.go
```diff
@@ -0,0 +1,78 @@
...
+	SenderKey     string
+	Receivers     []string
+	TokenAddr     string
+	PredicateAddr string
+	JSONRPCAddr   string
+}
+
+type ERC20BridgeParams struct {
+	*BridgeParams
+	Amounts []string
```
goran-ethernal, line 43: Could we move this `Amounts` slice to `BridgeParams` struct, since you use it in both `ERC20BridgeParams` and in `ERC1155BridgeParams`?

command/bridge/deposit/erc1155/deposit_erc1155.go
Stefan-Ethernal: I haven't put it there, because it is applicable only for deposit workflow and not withdrawal. So I still think it is good to leave it here.

command/bridge/common/params.go
Stefan-Ethernal: It would be misleading for ERC721 tokens, because they don't have it. So I'd still prefer to leave it as is.

command/bridge/withdraw/erc1155/withdraw_erc1155.go
Stefan-Ethernal: Tnx, forgot to change it.

Fixed https://github.com/0xPolygon/polygon-edge/pull/1332/commits/bfd33f9fd6247c291fd8f675b98ee002c9964ccb

igorcrevar (approved):
approved, left a comment for a trifle.

e2e-polybft/e2e/bridge_test.go
```diff
@@ -192,6 +199,208 @@ func TestE2E_Bridge_DepositAndWithdrawERC20(t *testing.T) {
...
+	logs, err := childEthEndpoint.GetLogs(filter)
+	require.NoError(t, err)
+
+	// assert that all deposits are executed successfully.
+	// All deposits are sent using a single transaction, so arbitrary message bridge emits two state sync events:
+	// MAP_TOKEN_SIG and DEPOSIT_BATCH_SIG state sync events
+	checkStateSyncResultLogs(t, logs, 2)
+
+	// retrieve child token address
+	rootToChildTokenFn := contractsapi.ChildERC1155Predicate.Abi.Methods["rootTokenToChildToken"]
```
goran-ethernal, line 364: You don't want to use `contractsapi` here?

goran-ethernal (approved):
Except for small comments, LGTM :100:

vcastellm:
I'm wondering if it'd be possible to merge all commands for ERC-721 and ERC-1155?

consensus/polybft/contracts_initializer.go
```diff
@@ -21,6 +21,8 @@ var (
 	nativeTokenName     = "Polygon"
 	nativeTokenSymbol   = "MATIC"
 	nativeTokenDecimals = uint8(18)
+
+	disabledBridgeRootPredicateAddr = "0xDEAD"
```
vcastellm, line 19: const?

e2e-polybft/e2e/bridge_test.go
Stefan-Ethernal: Yes, I wanted to do so, but it wasn't feasible, unfortunately. Do you remember that we tried it out together and it didn't work (the bindings generator created a strange binding struct with an empty `abi` tag, for the `Address` field)?

consensus/polybft/contracts_initializer.go
Stefan-Ethernal: Sure https://github.com/0xPolygon/polygon-edge/pull/1332/commits/2efb94ee7f0c7ad3ef386d909fbaeb74ab7f47d7

Stefan-Ethernal:
> I'm wondering if it'd be possible to merge all commands for ERC-721 and ERC-1155?

We could probably do something, although the most problematic part would be exposing extra flags to the commands (e.g. exposing amounts flag for ERC-721 or token ids for ERC-20, though we could validate it). Not sure if is it possible to register a flag "dynamically/conditionally", namely only after we tested the value of some other flag.

However, maybe do it as a separate PR, when all these are in place? WDYT?

e2e-polybft/e2e/bridge_test.go
goran-ethernal: Ah yes. I'll resolve the comment.

Nemanja0x (approved):

vcastellm (approved):

PR #1333, opened 2023-03-24
Withdraw burnt fees once an epoch

begmaroman:
# Description

Implement existing mechanism for burnt fees: send withdrawal transaction to EIP1559Burnt contract once an epoch. 

# Changes include

# Checklist

## Testing


consensus/polybft/burnt_fees_manager.go
```diff
@@ -0,0 +1,140 @@
...
+	txRelayer txrelayer.TxRelayer
+
+	// calculateBurnContract returns burnt contract address based on the given block
+	calculateBurnContract func(block uint64) (types.Address, error)
+
+	// forks is needed to check if london hardfork is enabled for the given block
+	forks *chain.Forks
+
+	// checkpointsOffset represents offset between withdrawal blocks (applicable only for non-epoch ending blocks)
+	withdrawalOffset uint64
```
goran-ethernal, line 51: Do we need the offset if we are going to do this once per epoch?

consensus/polybft/burnt_fees_manager.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if receipt.Status == uint64(types.ReceiptFailed) {
+		return fmt.Errorf("burnt fees withdrawal transaction failed for block %d", latestHeader.Number)
+	}
+
+	// Update last withdrawal block
+	m.lastWithdrawalBlock = req.FullBlock.Block.Number()
+
+	// update burnt fees withdrawal block number metrics
+	metrics.SetGauge([]string{"bridge", "checkpoint_block_number"}, float32(latestHeader.Number))
```
goran-ethernal: Did you want to name the metric something else? You probably forgot to rename it.

consensus/polybft/consensus_runtime.go
```diff
@@ -157,7 +171,7 @@ func (c *consensusRuntime) close() {
 func (c *consensusRuntime) initStateSyncManager(logger hcf.Logger) error {
```
vcastellm, line 171: Why not reuse here the `logger` instance in `consensusRuntime`?

consensus/polybft/consensus_runtime.go
```diff
@@ -210,6 +218,39 @@ func (c *consensusRuntime) initCheckpointManager(logger hcf.Logger) error {
 	return nil
 }
 
+// initBurntFeesManager initializes burnts fees manager
+// if bridge is not enabled, then a dummy burnt fees manager will be used
+func (c *consensusRuntime) initBurntFeesManager(logger hcf.Logger) error {
```
vcastellm, line 223: Same here with the logger.

consensus/polybft/polybft.go
```diff
@@ -295,6 +295,9 @@ func (p *Polybft) initRuntime() error {
 		txPool:                p.txPool,
 		bridgeTopic:           p.bridgeTopic,
 		numBlockConfirmations: p.config.NumBlockConfirmations,
+		jsonrpcAddr:           p.config.Config.JSONRPCAddr,
```
vcastellm, line 298: Why this was not present before? are you sure this can be done safely here without bridge config?

consensus/polybft/burnt_fees_manager.go
begmaroman: Just in case if something went wrong during the epoch block run so it could be started again based on the offset.

consensus/polybft/burnt_fees_manager.go
begmaroman: Good point, done

consensus/polybft/consensus_runtime.go
begmaroman: No idea, just following the current pattern. We could create a ticket to clean up this part of the code.

consensus/polybft/consensus_runtime.go
begmaroman: No idea, just following the current pattern. We could create a ticket to clean up this part of the code.

consensus/polybft/polybft.go
begmaroman: We don't need bridge here since the childchain smart contract should be triggered, not the rootchain. 
Something similar implemented in consensus/polybft/statesyncrelayer/state_sync_relayer.go

vcastellm: Where is this `JSONRPCAddr` config value coming from? I mean, without checking I imagine it's coming from the bridge config, if that's true, it can be empty.

consensus/consensus.go
```diff
@@ -60,6 +61,9 @@ type Config struct {
 
 	// Path is the directory path for the consensus protocol to store information
 	Path string
+
+	// JSONRPCAddr is the address of JSONRPC server running within the node
+	JSONRPCAddr *net.TCPAddr
```
goran-ethernal, line 66: I would say just pass a string instead of `net.TCPAddr` object, since we only use its string representation.

consensus/polybft/burnt_fees_manager.go
goran-ethernal: I am still trying to figure out, do we need to execute this transaction at the end of epoch, or do we just need to create it at the end of epoch and add it to the pool? As it is now, we just add it to the pool after the epoch ending block gets executed.

begmaroman: Correct, we just need to add it to the pool basically. But the question is what should be the tx type and who is going to pay for it?

consensus/polybft/polybft.go
begmaroman: It uses the localhost if it is empty.

consensus/polybft/burnt_fees_manager.go
goran-ethernal: Yeah. Maybe it might make sense to even have it as a state transaction, since it is a form of `automatic` burning of child tokens.

begmaroman:
Decided to do not automate the withdrawal process for now.

PR #1349, opened 2023-03-31
Transactions allow list

vcastellm:
# Description

In this PR we introduce the allow list for transactions, allowing the network to setup an allow list for transaction execution. The transactions won't fail but will be reverted with an error message.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

Sending manual transactions

### DIFF
--- a/chain/params.go
+++ b/chain/params.go
@@ -15,8 +15,8 @@ type Params struct {
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
 
 	// AllowList configuration
-	ContractDeployerAllowList *AllowListConfig `json:"contractDeployerAllowListConfig,omitempty"`
-	TransactionsAllowList     *AllowListConfig `json:"transactionsAllowListConfig,omitempty"`
+	ContractDeployerAllowList *AllowListConfig `json:"contractDeployerAllowList,omitempty"`
+	TransactionsAllowList     *AllowListConfig `json:"transactionsAllowList,omitempty"`
 }
 
 type AllowListConfig struct {
--- a/chain/params.go
+++ b/chain/params.go
@@ -16,6 +16,7 @@ type Params struct {
 
 	// AllowList configuration
 	ContractDeployerAllowList *AllowListConfig `json:"contractDeployerAllowListConfig,omitempty"`
+	TransactionsAllowList     *AllowListConfig `json:"transactionsAllowListConfig,omitempty"`
 }
 
 type AllowListConfig struct {
--- a/contracts/system_addresses.go
+++ b/contracts/system_addresses.go
@@ -30,4 +30,6 @@ var (
 	ConsolePrecompile = types.StringToAddress("0x000000000000000000636F6e736F6c652e6c6f67")
 	// AllowListContractsAddr is the address of the contract deployer allow list
 	AllowListContractsAddr = types.StringToAddress("0x0200000000000000000000000000000000000000")
+	// AllowListTransactionsAddr is the address of the transactions allow list
+	AllowListTransactionsAddr = types.StringToAddress("0x0200000000000000000000000000000000000002")
 )
--- a/e2e-polybft/e2e/allowlist_test.go
+++ b/e2e-polybft/e2e/allowlist_test.go
@@ -155,8 +155,9 @@ func TestAllowList_Transactions(t *testing.T) {
 
 	{
 		// Step 2. 'targetAddr' **cannot** send a normal transaction because it is not whitelisted.
-		err := cluster.Transfer(t, target, types.ZeroAddress, big.NewInt(1)).Wait()
-		require.Error(t, err)
+		targetTxn := cluster.Transfer(t, target, types.ZeroAddress, big.NewInt(1))
+		require.NoError(t, targetTxn.Wait())
+		require.True(t, targetTxn.Reverted())
 	}
 
 	{
--- a/e2e-polybft/e2e/allowlist_test.go
+++ b/e2e-polybft/e2e/allowlist_test.go
@@ -114,14 +114,14 @@ func TestAllowList_Transactions(t *testing.T) {
 	// both enabled and non-enabled roles.
 	admin, _ := wallet.GenerateKey()
 	target, _ := wallet.GenerateKey()
+	other, _ := wallet.GenerateKey()
 
 	adminAddr := types.Address(admin.Address())
 	targetAddr := types.Address(target.Address())
-
-	otherAddr := types.Address{0x1}
+	otherAddr := types.Address(other.Address())
 
 	cluster := framework.NewTestCluster(t, 3,
-		framework.WithPremine(adminAddr, targetAddr),
+		framework.WithPremine(adminAddr, targetAddr, otherAddr),
 		framework.WithTransactionsAllowListAdmin(adminAddr),
 		framework.WithTransactionsAllowListEnabled(otherAddr),
 	)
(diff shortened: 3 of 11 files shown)


state/runtime/allowlist/allowlist.go
```diff
@@ -131,6 +131,7 @@ func (a *AllowList) runInputCall(caller types.Address, input []byte,
 }
 
 func (a *AllowList) SetRole(addr types.Address, role Role) {
+	fmt.Println("SetRole", addr, role)
```
goran-ethernal: Is this some leftover from debugging?

vcastellm: Yes, removed

state/executor.go
```diff
@@ -562,6 +567,21 @@ func (t *Transition) run(contract *runtime.Contract, host runtime.Host) *runtime
 		return t.deploymentAllowlist.Run(contract, host, &t.config)
 	}
 
+	if t.txnAllowList != nil {
```
goran-ethernal, line 570: What happens if we have both `deployment` and `txn` allow lists defined, but an address is not in the `txn` allow list? Will this condition block him to deploy a contract? Correct me if I'm wrong, but if you are deploying a new contract, you will check the `txnAllowList` as well, and if you are not in that list, you'll be rejected.

vcastellm: > What happens if we have both deployment and txn allow lists defined, but an address is not in the txn allow list? Will this condition block him to deploy a contract? 

Yes, it will block the user from deploying a contract (create transaction), all transactions follow this code path.

> Correct me if I'm wrong, but if you are deploying a new contract, you will check the txnAllowList as well, and if you are not in that list, you'll be rejected.

This is what's happening, I added a new test case to ensure it's like that.

goran-ethernal: Yep, but is that something we want and expect? To me it's a little bit confusing. Maybe it's not a valid case to be in deployment allow list, but not in transaction allow list, but it can happen in theory right?

e2e-polybft/e2e/allowlist_test.go
```diff
@@ -107,3 +107,97 @@ func TestAllowList_ContractDeployment(t *testing.T) {
...
+	cluster.WaitForReady(t)
+
+	// bytecode for an empty smart contract
+	bytecode, _ := hex.DecodeString("6080604052348015600f57600080fd5b50603e80601d6000396000f3fe6080604052600080fdfea265627a7a7231582027748e4afe5ee282a786005d286f4427f13dac1b62e03f9aed311c2db7e8245364736f6c63430005110032")
+
+	expectRole := func(addr types.Address, role allowlist.Role) {
+		out := cluster.Call(t, contracts.AllowListTransactionsAddr, allowlist.ReadAllowListFunc, addr)
+
+		num, ok := out["0"].(*big.Int)
+		if !ok {
```
igorcrevar: Nitpick: `require.True(ok)` or `assert.True(ok)`

state/executor.go
vcastellm: Yes it can happen, but in security, the norm is that the most restrictive permission apply, so even if you are allowed to deploy but not allowed to send transactions, you shouldn't be able to send transactions.

e2e-polybft/e2e/allowlist_test.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1349/commits/3692badaa4e9300c8691a88752fe68c72114db65

goran-ethernal (approved):
LGTM :100:

server/server.go
```diff
@@ -212,6 +212,12 @@ func NewServer(config *Config) (*Server, error) {
 			m.config.Chain.Params.ContractDeployerAllowList)
 	}
 
+	// apply allow list genesis data
```
Stefan-Ethernal:
```suggestion
	// apply transactions execution allow list genesis data
```

contracts/system_addresses.go
```diff
@@ -30,4 +30,6 @@ var (
 	ConsolePrecompile = types.StringToAddress("0x000000000000000000636F6e736F6c652e6c6f67")
 	// AllowListContractsAddr is the address of the contract deployer allow list
 	AllowListContractsAddr = types.StringToAddress("0x0200000000000000000000000000000000000000")
+	// AllowListTransactionsAddr is the address of the transactions allow list
+	AllowListTransactionsAddr = types.StringToAddress("0x0200000000000000000000000000000000000002")
```
Stefan-Ethernal, line 34: Just curious, why we haven't taken `0x0200000000000000000000000000000000000001` for address? This way we have a gap in addresses.
```suggestion
	AllowListTransactionsAddr = types.StringToAddress("0x0200000000000000000000000000000000000001")
```

igorcrevar (approved):

server/server.go
Stefan-Ethernal: Similar would go for `ContractDeployerAllowList`, modify comment like this for example:
```go
// apply allow list contracts deployer genesis data
```

Stefan-Ethernal (approved):
Generally looks good and leaving some comments to consider.

chain/params.go
```diff
@@ -16,6 +16,7 @@ type Params struct {
 
 	// AllowList configuration
 	ContractDeployerAllowList *AllowListConfig `json:"contractDeployerAllowListConfig,omitempty"`
+	TransactionsAllowList     *AllowListConfig `json:"transactionsAllowListConfig,omitempty"`
```
Stefan-Ethernal: Minor nitpick, I would probably consider shortening the JSON element names to just `transactionsAllowList` and `contractDeployerAllowList`.

state/executor.go
```diff
@@ -562,6 +567,21 @@ func (t *Transition) run(contract *runtime.Contract, host runtime.Host) *runtime
...
+		if t.txnAllowList.Addr() == contract.CodeAddress {
+			return t.txnAllowList.Run(contract, host, &t.config)
+		}
+
+		if contract.Caller != contracts.SystemCaller {
+			txnAllowListRole := t.txnAllowList.GetRole(contract.Caller)
+			if !txnAllowListRole.Enabled() {
+				return &runtime.ExecutionResult{
+					Err: fmt.Errorf("caller not authorized"),
+				}
```
Stefan-Ethernal, line 580: Extract error to a variable?

contracts/system_addresses.go
Stefan-Ethernal: Ok now I see, we are following the Avalanche specification, it is described in the JIRA task itself.

chain/params.go
vcastellm: Agree

e2e-polybft/e2e/allowlist_test.go
```diff
@@ -107,3 +107,94 @@ func TestAllowList_ContractDeployment(t *testing.T) {
...
+		framework.WithPremine(adminAddr, targetAddr, otherAddr),
+		framework.WithTransactionsAllowListAdmin(adminAddr),
+		framework.WithTransactionsAllowListEnabled(otherAddr),
+	)
+	defer cluster.Stop()
+
+	cluster.WaitForReady(t)
+
+	// bytecode for an empty smart contract
+	bytecode, _ := hex.DecodeString("6080604052348015600f57600080fd5b50603e80601d6000396000f3fe6080604052600080fdfea265627a7a7231582027748e4afe5ee282a786005d286f4427f13dac1b62e03f9aed311c2db7e8245364736f6c63430005110032")
```
Stefan-Ethernal, line 133: Add an assertion for error?

chain/params.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1349/commits/a9ee37f3e386d6b41b8bb11e956076c54e0c725a

server/server.go
vcastellm: Done in https://github.com/0xPolygon/polygon-edge/pull/1349/commits/a9ee37f3e386d6b41b8bb11e956076c54e0c725a

e2e-polybft/e2e/allowlist_test.go
vcastellm: I think it's not worth it.

PR #1363, opened 2023-04-05
refactor(server): remove block time configuration from config and server params

flavor-town:
# Description

The block time configuration was removed from the raw config and server params. 
The initBlockTime function was also removed from the server params.
`polygon-edge server` command was not doing anything with those values. 

# Changes include

# Checklist

## Testing


flavor-town:

Stefan-Ethernal:
Generally the problem we are currently facing is that server config (https://github.com/flavor-town/polygon-edge/blob/EVM-571/server/config.go#L27), that is used for providing syncer block time remains 0 with this change. Therefore we need to figure out a way to fetch block time from consensus parameters.

`PolyBFTConfig` already has such field, however not sure for IBFT.

PR #1364, opened 2023-04-06
EVM-584 Panic -> send on closed channel inside syncer

igorcrevar:
# Description

Bug fix for panic that occurs frequently on CI unit tests
```
panic: send on closed channel

goroutine 1185 [running]:

[github.com/0xPolygon/polygon-edge/syncer.(*syncPeerClient](http://github.com/0xPolygon/polygon-edge/syncer.(*syncPeerClient)).handleStatusUpdate(0xc0026cc500, {0x1086380?, 0xc002a38690}, {0xc0028e9ec0, 0x27})

 /home/runner/work/polygon-edge/polygon-edge/syncer/client.go:219 +0x216

[Network Graph · 0xPolygon/polygon-edge](http://github.com/0xPolygon/polygon-edge/network.(*Topic) ).readLoop.func2()

 /home/runner/work/polygon-edge/polygon-edge/network/gossip.go:106 +0xcb

created by [Network Graph · 0xPolygon/polygon-edge](http://github.com/0xPolygon/polygon-edge/network.(*Topic) ).readLoop

 /home/runner/work/polygon-edge/polygon-edge/network/gossip.go:98 +0x14a
```

# Changes include

# Checklist

## Testing

### DIFF
--- a/network/gossip.go
+++ b/network/gossip.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"reflect"
+	"sync"
 	"sync/atomic"
 
 	"github.com/hashicorp/go-hclog"
@@ -22,10 +23,13 @@ const (
 type Topic struct {
 	logger hclog.Logger
 
-	topic   *pubsub.Topic
-	typ     reflect.Type
-	closeCh chan struct{}
-	closed  *uint64
+	topic  *pubsub.Topic
+	typ    reflect.Type
+	closed *uint64
+
+	ctx       context.Context
+	cancelFn  context.CancelFunc
+	waitGroup sync.WaitGroup
 }
 
 func (t *Topic) createObj() proto.Message {
@@ -43,12 +47,14 @@ func (t *Topic) Close() {
 		return
 	}
 
+	t.cancelFn()       // close all subscribers
+	t.waitGroup.Wait() // wait for all the subscribers to finish
+
+	// if all subscribers are finished, close the topic
 	if t.topic != nil {
 		t.topic.Close()
 		t.topic = nil
 	}
-
-	close(t.closeCh)
 }
 
 func (t *Topic) Publish(obj proto.Message) error {
@@ -75,18 +81,14 @@ func (t *Topic) Subscribe(handler func(obj interface{}, from peer.ID)) error {
 }
 
 func (t *Topic) readLoop(sub *pubsub.Subscription, handler func(obj interface{}, from peer.ID)) {
-	ctx, cancelFn := context.WithCancel(context.Background())
-
-	go func() {
-		<-t.closeCh
-		cancelFn()
-	}()
+	t.waitGroup.Add(1)
+	defer t.waitGroup.Done()
 
 	for {
-		msg, err := sub.Next(ctx)
+		msg, err := sub.Next(t.ctx)
 		if err != nil {
 			// Above cancelFn() called.
-			if errors.Is(err, ctx.Err()) {
+			if errors.Is(err, t.ctx.Err()) {
 				break
 			}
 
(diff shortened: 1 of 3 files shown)


syncer/client.go
```diff
@@ -210,16 +219,15 @@ func (m *syncPeerClient) handleStatusUpdate(obj interface{}, from peer.ID) {
...
+	defer m.peerStatusUpdateChLock.Unlock()
 
-		return
-	}
-
-	m.peerStatusUpdateCh <- &NoForkPeer{
-		ID:       from,
-		Number:   status.Number,
-		Distance: m.network.GetPeerDistance(from),
+	if !m.peerStatusUpdateChClosed {
```
Stefan-Ethernal, line 225: Sorry for my ignorance in advance, but could we use query `closeCh` state to determine if `Close` function has been called?

Something like this:

```suggestion
    _, isCloseChAlive := <-m.closeCh
	if !isCloseChAlive {
```

If we could do it, then I believe we can get rid of `peerStatusUpdateChLock` and `peerStatusUpdateChClosed`, which would be awesome. WDYT?

network/gossip.go
```diff
@@ -22,10 +23,13 @@ const (
...
-	topic   *pubsub.Topic
-	typ     reflect.Type
-	closeCh chan struct{}
-	closed  *uint64
+	topic  *pubsub.Topic
+	typ    reflect.Type
+	closed *uint64
+
+	ctx       context.Context
+	cancelFn  context.CancelFunc
```
Stefan-Ethernal: It would be better to preserve `closeCh` to signal cancellation since having a `context.Context` captured within the state is considered a bad practice (didn't know myself either, until Evgeny pointed that out on one occasion in one of the mine PRs). 

 Generally, `context` is meant to be created when needed locally, and propagated in the call chain via parameter. So in this case, it would be perfectly fine to have `closeCh` captured within the state, and then listen to it to cancel the locally created context instance. Though I must agree that it is a tempting to put `context` to the state :slightly_smiling_face: 

Here are the reasons why it is considered a bad practice.
> There are a few reasons why it's generally considered bad practice to add a `context.Context `reference to a state object in Golang:
> 
> - It violates the principle of separation of concerns: Adding a `context.Context` reference to a state object means that your state object is now responsible for both managing its own internal state and handling request-scoped values. This can make your code harder to understand and reason about.
> 
> - It can lead to coupling: Storing a `context.Context` reference in your state object means that your state object is now tightly coupled to the request context. This can make it harder to reuse your state object in other contexts, since it now depends on a specific `context.Context` implementation.
> 
> - It can make testing more difficult: Including a `context.Context `reference in your state object can make it harder to test your code, since you now need to provide a valid `context.Context` for your tests to run. This can make your tests more complex and harder to maintain.
> 
> - It can introduce subtle bugs: Storing a `context.Context` reference in your state object can lead to subtle bugs if you're not careful. For example, if you pass a reference to a `context.Context` object to a goroutine, that goroutine will continue to run even if the original request context is canceled. This can lead to resource leaks and other issues.
> 
> Overall, while there may be some cases where it makes sense to include a `context.Context` reference in a state object, it's generally considered bad practice to do so. Instead, you should look for alternative ways to manage your state and handle request-scoped values, such as passing values explicitly through function arguments or using other data structures to manage your state.

stana-miric (approved):

network/gossip.go
igorcrevar: reverted https://github.com/0xPolygon/polygon-edge/pull/1364/commits/e4fa5d62fac2498a5ee0740964dc1aa6f5a277c1

syncer/client.go
igorcrevar: It wont work because in line 225 it will be alive and in line 226 it wont. I think this is only approach that will work. Its complicated because we are closing channel from totally different routine

Stefan-Ethernal (approved):
LGTM. Really glad this is finally resolved :rocket:

goran-ethernal (approved):

vcastellm (commented):
🎩

PR #1365, opened 2023-04-06
Obsolete block time flag from server command

Stefan-Ethernal:
# Description

This PR removes `block-time` flag from the server command and server config. Instead, it resolves it from a consensus engine parameters map. `PolyBFT` and `IBFT` consensus protocols have defined block time. `Dev` and `Dummy` consensuses are initialized with 0 block time, as those are used solely for testing.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/server/init.go
+++ b/command/server/init.go
@@ -19,7 +19,6 @@ import (
 )
 
 var (
-	errInvalidBlockTime       = errors.New("invalid block time specified")
 	errDataDirectoryUndefined = errors.New("data directory not defined")
 )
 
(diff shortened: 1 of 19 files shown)


command/server/init.go
```diff
@@ -66,14 +61,6 @@ func (p *serverParams) initRawParams() error {
 	return p.initAddresses()
 }
 
-func (p *serverParams) initBlockTime() error {
-	if p.rawConfig.BlockTime < 1 {
```
igorcrevar, line 70: Do we need to keep this condition somewhere? maybe in `extractBlockTime`?

Stefan-Ethernal: You are right, fixed: https://github.com/0xPolygon/polygon-edge/pull/1365/commits/2b0c446f29fb07fe86e6c2ea100ff409cbcac570

igorcrevar (approved):

goran-ethernal (approved):

PR #1371, opened 2023-04-10
Change folders hierarchy for property tests logs

goran-ethernal:
# Description

Since property tests run `cluster` multiple times (and logs of all runs are in one folder), for better separation of logs, this PR (in the case of a property test), creates this folder hierarchy for logs:

image

# Changes include

# Checklist

## Testing

### DIFF
--- a/e2e-polybft/property/property_test.go
+++ b/e2e-polybft/property/property_test.go
@@ -3,6 +3,7 @@ package property
 import (
 	"fmt"
 	"math"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


e2e-polybft/property/property_test.go
```diff
@@ -35,6 +35,7 @@ func TestProperty_DifferentVotingPower(t *testing.T) {
 		}
 
 		cluster := framework.NewTestCluster(t, int(numNodes),
```
Stefan-Ethernal: Thinking out loud:
maybe it is better to create a separate function called `NewPropertyTestCluster`, which would set `IsPropertyTest` to `true` and wrap `NewTestCluster`. That way we would be more explicit that it is a cluster created for the property-based test. The way it is currently, it is easy to forget that `WithPropertyTestLogging` needs to be called.

Stefan-Ethernal (approved):

e2e-polybft/property/property_test.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1371/commits/4819acf67a6fd96c1e52460fee14cc0453f70368

igorcrevar (approved):

PR #1373, opened 2023-04-10
Handle block tracker init/start failures and treat as fatal

sergerad:
# Description

Change Event Tracker creation logic to treat BlockTracker and Tracker errors as fatal.

Removes the possibility of running a validator without any block tracking against the rootchain due to init/start error of event tracker.

# Changes include

# Checklist

## Testing

### Manual tests

Found that the EventTracker Start() func took <20ms on my laptop. So these changes should not affect startup perf at all

### DIFF
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -443,6 +443,11 @@ func Test_NewConsensusRuntime(t *testing.T) {
 	require.NoError(t, err)
 
 	polyBftConfig := &PolyBFTConfig{
+		Bridge: &BridgeConfig{
+			StateSenderAddr:       types.Address{0x13},
+			CheckpointManagerAddr: types.Address{0x10},
+			JSONRPCEndpoint:       "testEndpoint",
+		},
 		EpochSize:  10,
 		SprintSize: 10,
 		BlockTime:  common.Duration{Duration: 2 * time.Second},
@@ -452,6 +457,7 @@ func Test_NewConsensusRuntime(t *testing.T) {
 
 	systemStateMock := new(systemStateMock)
 	systemStateMock.On("GetEpoch").Return(uint64(1)).Once()
+	systemStateMock.On("GetNextCommittedIndex").Return(uint64(1)).Once()
 
 	blockchainMock := &blockchainMock{}
 	blockchainMock.On("CurrentHeader").Return(&types.Header{Number: 1, ExtraData: createTestExtraForAccounts(t, 1, validators, nil)})
@@ -480,7 +486,9 @@ func Test_NewConsensusRuntime(t *testing.T) {
 	assert.Equal(t, uint64(10), runtime.config.PolyBFTConfig.SprintSize)
 	assert.Equal(t, uint64(10), runtime.config.PolyBFTConfig.EpochSize)
 	assert.Equal(t, "0x0000000000000000000000000000000000000101", contracts.ValidatorSetContract.String())
-	assert.False(t, runtime.IsBridgeEnabled())
+	assert.Equal(t, "0x1300000000000000000000000000000000000000", runtime.config.PolyBFTConfig.Bridge.StateSenderAddr.String())
+	assert.Equal(t, "0x1000000000000000000000000000000000000000", runtime.config.PolyBFTConfig.Bridge.CheckpointManagerAddr.String())
+	assert.True(t, runtime.IsBridgeEnabled())
 	systemStateMock.AssertExpectations(t)
 	blockchainMock.AssertExpectations(t)
 	polybftBackendMock.AssertExpectations(t)
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -437,6 +437,7 @@ func TestConsensusRuntime_FSM_EndOfEpoch_BuildCommitEpoch(t *testing.T) {
 }
 
 func Test_NewConsensusRuntime(t *testing.T) {
+	t.Skip("fails on no such host for JSON RPC endpoint in event tracker init/start")
 	t.Parallel()
 
 	_, err := os.Create("/tmp/consensusState.db")
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -437,18 +437,12 @@ func TestConsensusRuntime_FSM_EndOfEpoch_BuildCommitEpoch(t *testing.T) {
 }
 
 func Test_NewConsensusRuntime(t *testing.T) {
-	t.Skip("fails on no such host for JSON RPC endpoint in event tracker init/start")
 	t.Parallel()
 
 	_, err := os.Create("/tmp/consensusState.db")
 	require.NoError(t, err)
 
 	polyBftConfig := &PolyBFTConfig{
-		Bridge: &BridgeConfig{
-			StateSenderAddr:       types.Address{0x13},
-			CheckpointManagerAddr: types.Address{0x10},
-			JSONRPCEndpoint:       "testEndpoint",
-		},
 		EpochSize:  10,
 		SprintSize: 10,
 		BlockTime:  common.Duration{Duration: 2 * time.Second},
@@ -458,7 +452,6 @@ func Test_NewConsensusRuntime(t *testing.T) {
 
 	systemStateMock := new(systemStateMock)
 	systemStateMock.On("GetEpoch").Return(uint64(1)).Once()
-	systemStateMock.On("GetNextCommittedIndex").Return(uint64(1)).Once()
 
 	blockchainMock := &blockchainMock{}
 	blockchainMock.On("CurrentHeader").Return(&types.Header{Number: 1, ExtraData: createTestExtraForAccounts(t, 1, validators, nil)})
@@ -487,9 +480,7 @@ func Test_NewConsensusRuntime(t *testing.T) {
 	assert.Equal(t, uint64(10), runtime.config.PolyBFTConfig.SprintSize)
 	assert.Equal(t, uint64(10), runtime.config.PolyBFTConfig.EpochSize)
 	assert.Equal(t, "0x0000000000000000000000000000000000000101", contracts.ValidatorSetContract.String())
-	assert.Equal(t, "0x1300000000000000000000000000000000000000", runtime.config.PolyBFTConfig.Bridge.StateSenderAddr.String())
-	assert.Equal(t, "0x1000000000000000000000000000000000000000", runtime.config.PolyBFTConfig.Bridge.CheckpointManagerAddr.String())
-	assert.True(t, runtime.IsBridgeEnabled())
+	assert.False(t, runtime.IsBridgeEnabled())
 	systemStateMock.AssertExpectations(t)
 	blockchainMock.AssertExpectations(t)
 	polybftBackendMock.AssertExpectations(t)
--- a/tracker/event_tracker.go
+++ b/tracker/event_tracker.go
@@ -45,7 +45,10 @@ func NewEventTracker(
 	}
 }
 
-func (e *EventTracker) Start(ctx context.Context) error {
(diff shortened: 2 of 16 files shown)


sergerad:
@Stefan-Ethernal could you please look at my changes and explain why the test error occurs
```
--- FAIL: Test_NewConsensusRuntime (0.02s)
    consensus_runtime_test.go:480: 
        	Error Trace:	/home/runner/work/polygon-edge/polygon-edge/consensus/polybft/consensus_runtime_test.go:480
        	Error:      	Received unexpected error:
        	            	failed to init event tracker. Error: failed to sync: lookup : no such host
        	Test:       	Test_NewConsensusRuntime
```

I'm assuming that something is initialized after those goroutines are started, which they depend on. Hence my wait approach does not work.

If this is the case, would we consider an `os.Exit()` or similar solution here so that event monitor failure causes application exit? Or is it intentional that a validator continues without a block tracker when it is required? Thanks

tracker/event_tracker.go
```diff
@@ -99,10 +114,23 @@ func (e *EventTracker) Start(ctx context.Context) error {
...
+
 		if err := tt.Sync(ctx); err != nil {
-			e.logger.Error("failed to sync", "error", err)
+			errCh <- fmt.Errorf("failed to sync: %w", err)
+
+			return
 		}
 	}()
 
+	wg.Wait()
```
sergerad: I think there is a downstream dependancy which this wait precludes

igorcrevar: Can we use `errgroup` instead of `waitgroup` + `channel` https://pkg.go.dev/golang.org/x/sync/errgroup ?

igorcrevar: Also, I am not sure... will this block server if there is no error at all?

sergerad: Will do but I think we can't do the wait approach unless we refactor the usage here because it seems that something lower in the stack needs to execute for those goroutines to work - and that thing is being executed after the goroutines are run

tracker/event_tracker.go
```diff
@@ -101,6 +101,7 @@ func (e *EventTracker) Start(ctx context.Context) error {
 	go func() {
 		if err := tt.Sync(ctx); err != nil {
 			e.logger.Error("failed to sync", "error", err)
+			panic(err)
```
sergerad: @igorcrevar Added this to show it occurs in the tests as they currently exist. My earlier changes with the wait also uncovered this. Have tested this locally, expecting it to happen in PR also.

Is this intentional / expected behaviour? If so, I will just add wait for the other goroutine which is what I require

sergerad: Same thing for
```
		if err := blockTracker.Init(); err != nil {
			e.logger.Error("failed to init blocktracker", "error", err)

			panic(err)
		}
```

sergerad: Potentially these errors are unhandled for the sake of these tests that are failing, which will cause issues for people like myself. This config causes the error
```
			JSONRPCEndpoint: "testEndpoint",
```

igorcrevar: The more I think about this problem, the more I would remove those two go routines all-together and keep the code like this

```
	tt, err := tracker.NewTracker(provider.Eth(),
		tracker.WithBatchSize(10),
		tracker.WithBlockTracker(blockTracker),
		tracker.WithStore(store),
		tracker.WithFilter(&tracker.FilterConfig{
			Async: true,
			Address: []ethgo.Address{
				e.contractAddr,
			},
			Start: e.startBlock,
		}),
	)
	if err != nil {
		return err
	}

	if err := blockTracker.Init(); err != nil {
		return err
	}

	if err := blockTracker.Start(); err != nil {
		return err
	}

		e.logger.Error("failed to sync", "error", err)
	}

	go func() { // this should stay in go routine
		<-ctx.Done()
		blockTracker.Close()
		store.Close()
	}()

	return nil
```
With synchronization we are just blocking current routine until initialization finishes...

Did you measure how much time takes for block tracker initialization/start and for initial batch sync?

consensus/polybft/consensus_runtime_test.go
```diff
@@ -435,6 +435,7 @@ func TestConsensusRuntime_FSM_EndOfEpoch_BuildCommitEpoch(t *testing.T) {
 }
 
 func Test_NewConsensusRuntime(t *testing.T) {
+	t.Skip("fails on no such host for JSON RPC endpoint in event tracker init/start")
```
igorcrevar: I think we should fix this by just disabling bridge for this test

```
func (c *consensusRuntime) IsBridgeEnabled() bool {
	return c.config.PolyBFTConfig.IsBridgeEnabled()
}
```

tracker/event_tracker.go
sergerad: > Did you measure how much time takes for block tracker initialization/start and for initial batch sync?

No I haven't and I don't know the significance of those timings

igorcrevar: Maybe it would be good to know for how much time the main server routine gets blocked before initialization of event tracker finishes.

tracker/event_tracker.go
```diff
@@ -79,30 +86,15 @@ func (e *EventTracker) Start(ctx context.Context) error {
...
-
-			return
-		}
-
-		if err := blockTracker.Start(); err != nil {
-			e.logger.Error("failed to start blocktracker", "error", err)
-		}
-	}()
+	if err := tt.Sync(ctx); err != nil {
+		e.logger.Error("failed to sync", "error", err)
```
igorcrevar: Maybe we should return error here also?

tracker/event_tracker.go
sergerad: I'm seeing `func (e *EventTracker) Start(ctx context.Context) error` take <20ms on my laptop. Those goroutines seemed unnecessary in the first place b/c the tracker and block tracker funcs were mainly just launching other goroutines themselves

tracker/event_tracker.go
sergerad: mb will fix

igorcrevar (approved):
LGTM!

vcastellm (approved):
LGTM

tracker/event_tracker.go
```diff
@@ -79,30 +87,15 @@ func (e *EventTracker) Start(ctx context.Context) error {
...
-			e.logger.Error("failed to init blocktracker", "error", err)
-
-			return
-		}
-
-		if err := blockTracker.Start(); err != nil {
-			e.logger.Error("failed to start blocktracker", "error", err)
-		}
-	}()
+	if err := tt.Sync(ctx); err != nil {
```
Stefan-Ethernal: I'm just a bit puzzled is it correct to invoke `Sync` synchronously and could there be some unwanted consequences? There is some logic in the `syncImpl` function in the `ethgo.Tracker`, which, at least to me, could run for some time...

Stefan-Ethernal (approved):

tracker/event_tracker.go
sergerad: Yes this is a good point.

I am not able to test a situation where `func (t *Tracker) findAncestor(block, pivot *ethgo.Block) (uint64, error)` might involve a large range of blocks and many RPC requests.

My main question is - should the runtime be synchronous here notwithstanding the potential in delay given the importance of the event tracker for a running validator?

If we are not happy being synchronous here in the worst case, should the client exit on failure of `tt.Sync()` despite being async?

Stefan-Ethernal: > My main question is - should the runtime be synchronous here notwithstanding the potential in delay given the importance of the event tracker for a running validator?

IMO we could live with potential delays due to the async nature of the tracker synchronization.

> If we are not happy being synchronous here in the worst case, should the client exit on failure of tt.Sync() despite being async?

Not sure about this either. For example, for the `StateSyncRelayer` component, which executes bridge events (transfers assets from the root chain to the child chain) I don't feel it is critical enough to shut down the validator in case of `Sync` failure. On the other hand for the `stateSyncManager,` it might make sense.

I've talked a bit with @goran-ethernal and his idea was maybe to consider propagating errors using some error channel. In that case, we could once again closely analyze whether it makes sense to revert everything to async behavior or do it only partially (e.g. make only `Sync` invocation async).

sergerad: > propagating errors using some error channel. In that case, we could once again closely analyze whether it makes sense to revert everything to async behavior or do it only partially (e.g. make only Sync invocation async).

Makes sense. I will update the PR

sergerad: @igorcrevar @vcastellm @Stefan-Ethernal I have updated the PR based on the above. The relayer and state sync manager treat the event tracker sync error differently. I left a note on the os.Exit line which is worth considering for future

sergerad:
@Stefan-Ethernal this is still an issue in our deployments b/c it silently leaves our system in a broken state sometimes. We can create alerting to capture the error log but that is coupling our alerts to this impl in a very undesirable way.

Would you like to have me look at doing a wider refactor or is the os.Exit() approach OK?

Thanks

Stefan-Ethernal:
Sorry for waiting this long.

> Would you like to have me look at doing a wider refactor?

I have brought merging this PR to the team and conclusion was to try out not use `os.Exit()`, although PR looks good. Could you try propagating error message there (or if you have some other approach on mind...)?

Tnx 🙏

sergerad:
Yep sounds good I will aim to have something ready by Monday

sergerad:
@Stefan-Ethernal I had a quick crack at it before signing off for the weekend.
https://github.com/0xPolygon/polygon-edge/pull/1373/commits/f241329d41d76d05263c0965b294f51371aa19ac

Its not tested. Just wanted to give you an idea of what would be involved in:
1. Keeping the event tracker sync functionality concurrent with the rest of the server; and
2. Ensuring that an error in the event tracker sync is handled by the server.

Note I found another concurrent operation that was being logged if error so I merged that into these changes.

Please have a look and let me know if you would consider these kind of diffs appropriate. The existing code is not designed to propagate concurrent errors back back down the stack but I have tried to be as consistent with the current impl as much as possible.

Side note: the `server.Close()` approach (used in `HandleSignals()`) is not what I would call the canonical approach for closing resources in Go. Ideally a root context should exist somewhere in the start of the command stack and be used to create child contexts for resources that can be cleaned up in case of `ctx.Done()`. This way, a signal capture logic could use the root `cancel()` to shutdown all resources, rather than doing the `server.Close()` approach. This is a purely code maintenance concern.

Thanks!

Stefan-Ethernal:
The latest changes resulted in a segmentation fault (seems that's the reason why all the e2e tests failed):

```go
[signal SIGSEGV: segmentation violation code=0x1 addr=0x80 pc=0x169a976]

goroutine 136 [running]:
github.com/0xPolygon/polygon-edge/consensus/polybft/statesyncrelayer.(*StateSyncRelayer).SyncError(...)
	/home/runner/work/polygon-edge/polygon-edge/consensus/polybft/statesyncrelayer/state_sync_relayer.go:227
github.com/0xPolygon/polygon-edge/server.(*Server).Error(0xc000c59970)
	/home/runner/work/polygon-edge/polygon-edge/server/server.go:942 +0x36
github.com/0xPolygon/polygon-edge/command/server.runServerLoop.func2()
	/home/runner/work/polygon-edge/polygon-edge/command/server/server.go:329 +0x56
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.2.0/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.2.0/errgroup/errgroup.go:72 +0xa5
```

sergerad:
@Stefan-Ethernal I am closing this PR in favour of https://github.com/0xPolygon/polygon-edge/pull/1503

I realized that I was introducing a paradigm with this PR that is inconsistent with the error handling approach used in the Edge application in general (the approach is to keep running despite error). Please see the PR for more info 🙇

PR #1388, opened 2023-04-14
Allowlist package refactor and blocklist implementation

vcastellm:
# Description

This is a refactor of the allowlist package to make it more generic as it's now used for both allowlist and blocklist. It also adds the blocklist implementation.

# Changes include

# Checklist

## Testing


goran-ethernal (approved):
:100:

command/genesis/polybft_params.go
```diff
@@ -40,8 +40,12 @@ const (
 
 	contractDeployerAllowListAdminFlag   = "contract-deployer-allow-list-admin"
 	contractDeployerAllowListEnabledFlag = "contract-deployer-allow-list-enabled"
+	contractDeployerBlockListAdminFlag   = "contract-deployer-block-list-admin"
+	contractDeployerBlockListEnabledFlag = "contract-deployer-block-list-enabled"
 	transactionsAllowListAdminFlag       = "transactions-allow-list-admin"
 	transactionsAllowListEnabledFlag     = "transactions-allow-list-enabled"
+	transactionsBlockListAdminFlag       = "transactions-block-list-admin"
+	transactionsBlockListEnabledFlag     = "transactions-block-list-enabled"
```
Stefan-Ethernal, line 52: WDYT should we consider introducing some kind of JSON config which would hold all these access lists? Then we would provide only `AccessListConfig` path to the genesis command.
Such `AccessListConfig` could probably populated manually and we can have empty JSON template as part of codebase. Not sure does it make sense, however these flag names are pretty long one and there are quite a few of them, so just thinking how to simplify this?

chain/params.go
```diff
@@ -14,12 +14,14 @@ type Params struct {
 	Whitelists     *Whitelists            `json:"whitelists,omitempty"`
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
 
-	// AllowList configuration
-	ContractDeployerAllowList *AllowListConfig `json:"contractDeployerAllowList,omitempty"`
-	TransactionsAllowList     *AllowListConfig `json:"transactionsAllowList,omitempty"`
+	// ACList configuration
```
Stefan-Ethernal: Nitpick:
```suggestion
	// Access control configuration
```

state/executor.go
```diff
@@ -593,21 +603,42 @@ func (t *Transition) Call2(
...
+			}
+		}
+	} else if t.txnBlockList != nil {
+		if t.txnBlockList.Addr() == contract.CodeAddress {
+			return t.txnBlockList.Run(contract, host, &t.config)
+		}
+
+		if contract.Caller != contracts.SystemCaller {
+			role := t.txnBlockList.GetRole(contract.Caller)
+			if role == addresslist.EnabledRole {
```
Stefan-Ethernal, line 638: Should we check `AdminRole` here as well (it takes precedence over `EnabledRole` to my understanding)?

```suggestion
			if role.Enabled() {
```

e2e-polybft/e2e/acls_test.go
```diff
@@ -0,0 +1,365 @@
+package e2e
```
Stefan-Ethernal, line 1: Consider naming entire feature either `AccessList` or `AddressList` consistently? E.g. most of the time we are referring as `AddressList`, but in the genesis there was `AccessList` and also this file is referred as `AccessList` (`acls_test.go`). 

Think that following single naming convention would remove any potential ambiguities in the future.

state/executor.go
```diff
@@ -749,15 +780,24 @@ func (t *Transition) applyCreate(c *runtime.Contract, host runtime.Host) *runtim
...
 		if !role.Enabled() {
 			return &runtime.ExecutionResult{
 				GasLeft: 0,
 				Err:     runtime.ErrNotAuth,
 			}
 		}
+	} else if t.deploymentBlockList != nil {
+		role := t.deploymentBlockList.GetRole(c.Caller)
+
+		if role == addresslist.EnabledRole {
```
Stefan-Ethernal, line 795: Check `AdminRole` as well?

```suggestion
		if role.Enabled() {
```

e2e-polybft/e2e/acls_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+	function readAllowList(address addr) public returns (uint256) {
+		(bool success, bytes memory returnData) = allowList.call(abi.encodeWithSignature("readAllowList(address)", addr));
+		(uint256 val) = abi.decode(returnData, (uint256));
+		return val;
+	}
+}
+*/
+
+func TestE2E_AllowList_ContractDeployment(t *testing.T) {
```
Stefan-Ethernal, line 33: Are we able to unify these tests and run them against the same cluster? Would it be hard to accomplish, because of state needs to be resetted at the end of the each test etc.?

e2e-polybft/e2e/acls_test.go
vcastellm: It's not the same, AddressList is a generalization, `addresslist` package doesn't do anything controlling access, it's just a mapping in the state. The implementation that uses `addresslist` is an `AccessList`, and it can be of two types, `AllowList` or `BlockList`.

state/executor.go
vcastellm: With your suggestion we'll block admins or the ones in the list, we don't want that, `AdminRole` will simply pass the check forward, that's what we want.

command/genesis/polybft_params.go
vcastellm: I was thinking hard on how to simplify this, we can introduce a wrapper config to hold all values but I don't clearly see how're we going to handle that from the CLI, the user needs to be explicit with all lists, currently the config stays in commandline, explicit, I don't feel good about introducing a separate config file only for this feature, I find it not operator friendly and more complex to explain in docs.

I don't see a clear advantage of providing a new config file that will generate another config file (genesis) in another or very similar format. However, it's possible that we need a kind of genesis template generation "assistant" to simplify genesis creation.

state/executor.go
vcastellm: https://github.com/0xPolygon/polygon-edge/pull/1388#discussion_r1171011573

e2e-polybft/e2e/acls_test.go
vcastellm: Exactly

state/executor.go
Stefan-Ethernal: I suggested this since the same way it is checked for allowed list:
if an account has either `AdminRole` or `EnabledRole` it is considered to be whitelisted. Although not sure if that makes sense either. To me `AdminRole` is allowed to set roles, but probably should not be implicitly considered as enabled. Not sure, what are your thoughts here? :thinking:

command/genesis/polybft_params.go
Stefan-Ethernal: My feeling is that having 8 flags to configure allow and black lists becomes cumbersome and complex enough to consider introducing a separate config (JSON format for example).
 
You are right with regards to how such a config should be populated (and that was my doubt as well), but I was thinking of populating it manually and then just providing a path to it in the genesis command, validating it, and eventually importing it if it is valid.

Stefan-Ethernal (approved):

state/executor.go
vcastellm: I think the expectation is that Admin is enabled by default. That said, I think semantically `Enabled` doesn't makes sense in BlockLists.

Do you feel it's worth to add also `DisabledRole` and use that for BlockLists?

Stefan-Ethernal: No need to add another role, it is ok like this: if an account is enabled in block list => it is disabled.

PR #1389, opened 2023-04-14
Remove manifest init step from network bootstrap workflow

Stefan-Ethernal:
# Description

Changelog:
- `manifest` command has been removed in favour of genesis command.
- `premine-validators` flag became redundant and is removed. Validators are premined using `--premine` as it is done with all the other accounts.
- `rootchain init-contracts` command is renamed into `rootchain deploy`, as it is shorter and enough self-explanatory.
- `rootchain deploy` instead of manifest, expects genesis file. It's absolute path is provided using `--genesis` flag.
- `--tracker-start-blocks` is removed as the event tracker configuration is now populated automatically into the `genesis.json` configuration as part of rootchain deployment.
- sequence of commands is changed: now the `genesis` command must be run before the `rootchain deploy`. Exact sequence is documented [here](https://github.com/0xPolygon/polygon-edge/blob/EVM-573-remove-manifest-initialization-from-the-bootstrap-workflow/consensus/polybft/README.md).
- updated README files.

# Changes include

# Breaking changes

Genesis configuration has changed.

# Checklist

## Testing


Stefan-Ethernal:
cc @gatsbyz, @praetoriansentry, @DannyS03 to follow this PR, as it is going to affect both documentation and deployment procedure.

command/rootchain/README.md
```diff
@@ -4,33 +4,35 @@ Top level command for manipulating rootchain server.
...
-$ polygon-edge rootchain server
+polygon-edge rootchain server
 ```
 
 ## Fund initialized accounts
 
 This command funds the initialized accounts via `polygon-edge polybft-secrets` command.
 
 ```bash
-$ polygon-edge rootchain fund --data-dir data-dir- --num 2
```
igorcrevar, line 18: Why removing `$`? Its usual way to describe that the following command should be executed in the system shell

Stefan-Ethernal: The markdown warning analyzer was complaining for some reason.

command/genesis/params.go
```diff
@@ -142,12 +144,12 @@ func (p *genesisParams) validateFlags() error {
 		return errInvalidEpochSize
 	}
 
-	// Validate min and max validators number
-	if err := command.ValidateMinMaxValidatorsNumber(p.minNumValidators, p.maxNumValidators); err != nil {
-		return err
+	if _, err := os.Stat(p.validatorsPath); errors.Is(err, os.ErrNotExist) {
```
vcastellm: Will this work with passing validators as param?

Stefan-Ethernal: Yes, it would work, due to the fact that the default value for `validatorsPath` is set to the current directory ("./").
The exact same code was present in the manifest command: https://github.com/0xPolygon/polygon-edge/pull/1389/files#diff-81dfae315a2b068d32b49fb15f77a51a35a3825447063099b021f800156b45f7L150.

However, I'm going to add an explicit check and validate this value only if `validators` are not provided via flag.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1389/commits/63f163fbc16b21636aa10541ccb7ab99e8131fca

command/rootchain/README.md
Stefan-Ethernal: Returned $ sign eventually:
https://github.com/0xPolygon/polygon-edge/pull/1389/commits/63f163fbc16b21636aa10541ccb7ab99e8131fca

scripts/cluster
```diff
@@ -14,9 +14,8 @@ function initIbftConsensus() {
 
 function initPolybftConsensus() {
     echo "Running with polybft consensus"
-    genesis_params="--consensus polybft --bridge-json-rpc http://127.0.0.1:8545"
+    genesis_params="--consensus polybft"
```
vcastellm, line 17: Don't we need the bridge RPC address?

vcastellm (approved):

scripts/cluster
Stefan-Ethernal: Nope, we don't need it eventually. :slightly_smiling_face:

The mindset was like this:
let's configure everything related to the rootchain in `rootchain deploy` command:

- we are populating JSON RPC endpoint address in that particular command, as part of `RootchainConfig` (https://github.com/0xPolygon/polygon-edge/pull/1389/files#diff-db0b676aabd14c80bf4a9bfa6c9353fc37b51d3a75763cf23ebb59ea8ddf462aR301-R303),
- it is then converted into `BridgeConfig` (https://github.com/0xPolygon/polygon-edge/pull/1389/files#diff-db0b676aabd14c80bf4a9bfa6c9353fc37b51d3a75763cf23ebb59ea8ddf462aR207) and 
- written into the genesis file (https://github.com/0xPolygon/polygon-edge/pull/1389/files#diff-db0b676aabd14c80bf4a9bfa6c9353fc37b51d3a75763cf23ebb59ea8ddf462aR223).

command/genesis/params.go
```diff
@@ -142,12 +144,15 @@ func (p *genesisParams) validateFlags() error {
 		return errInvalidEpochSize
 	}
 
-	// Validate min and max validators number
-	if err := command.ValidateMinMaxValidatorsNumber(p.minNumValidators, p.maxNumValidators); err != nil {
-		return err
+	// Validate validatorsPath only if validators information were not provided via CLI flag
+	if len(p.validators) == 0 {
+		if _, err := os.Stat(p.validatorsPath); errors.Is(err, os.ErrNotExist) {
```
goran-ethernal: Shouldn't we handle all errors?

goran-ethernal (approved):
We say goodbye to a not so dear friend. He will not be missed :stuck_out_tongue: .

Good job! :100:

command/genesis/params.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1389/commits/b9d96c8adb9c3a856fc88b242dce5f3d0bdcdcad

igorcrevar (approved):

PR #1403, opened 2023-04-20
WriteFailedReceipt removed

Nemanja0x:
# Description

`WriteFailedReceipt` is a function of the `State Transition Function (STF) `that writes a “failed” transaction into the block, a “failed” transaction (given this logic) is a transaction that exceeds the block gas limit of the block (not the gas left on the block though). For example, if transaction A has a gas limit of 100 and the gas limit of the block is 99, the transaction gets included with _WriteFailedReceipt_ on the block. That means that the transaction is not executed but it is included in the block with a receipt. If the transaction is not executed, it means that there is no gas cost for the user to include the transaction, this is, there is no cost to fill the time of the block.
https://polygon.atlassian.net/browse/EVM-583

# Changes include

# Breaking changes
No more failed transactions included into a block. 

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -130,7 +130,7 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*types.FullBlock, e
 func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 	if tx.ExceedsBlockGasLimit(b.params.GasLimit) {
 		b.params.Logger.Info("Transaction gas limit exceedes block gas limit", "hash", tx.Hash,
-			"tx gas limit", tx.Gas, "bloc gas limt", b.params.GasLimit)
+			"tx gas limit", tx.Gas, "block gas limt", b.params.GasLimit)
 
 		return txpool.ErrBlockLimitExceeded
 	}
--- a/consensus/polybft/block_builder.go
+++ b/consensus/polybft/block_builder.go
@@ -131,6 +131,7 @@ func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 	if tx.ExceedsBlockGasLimit(b.params.GasLimit) {
 		b.params.Logger.Info("Transaction gas limit exceedes block gas limit", "hash", tx.Hash,
 			"tx gas limit", tx.Gas, "bloc gas limt", b.params.GasLimit)
+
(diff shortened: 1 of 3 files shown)


state/executor.go
```diff
@@ -137,10 +137,6 @@ func (e *Executor) ProcessBlock(
 
 	for _, t := range block.Transactions {
 		if t.ExceedsBlockGasLimit(block.Header.GasLimit) {
-			if err := txn.WriteFailedReceipt(t); err != nil {
```
Stefan-Ethernal, line 140: Worth considering do we even need this check in this place, since the block is already built if I'm correct?

```go
if t.ExceedsBlockGasLimit(block.Header.GasLimit)
```

igorcrevar: Maybe someone malicious built block with larger block gas limit then it should be?

Stefan-Ethernal: Ok, but it is already being checked when building a block itself, so not sure do we need a check here as well. If we are about to leave it, then maybe consider adding a log.

igorcrevar (approved):

state/executor.go
Nemanja0x: @Stefan-Ethernal fixed: https://github.com/0xPolygon/polygon-edge/pull/1403/commits/615b8f04cedb9c0ec24700600f5a6ac4aa829a49

consensus/polybft/block_builder.go
```diff
@@ -129,9 +129,8 @@ func (b *BlockBuilder) Build(handler func(h *types.Header)) (*types.FullBlock, e
 // WriteTx applies given transaction to the state. If transaction apply fails, it reverts the saved snapshot.
 func (b *BlockBuilder) WriteTx(tx *types.Transaction) error {
 	if tx.ExceedsBlockGasLimit(b.params.GasLimit) {
-		if err := b.state.WriteFailedReceipt(tx); err != nil {
-			return err
-		}
+		b.params.Logger.Info("Transaction gas limit exceedes block gas limit", "hash", tx.Hash,
+			"tx gas limit", tx.Gas, "bloc gas limt", b.params.GasLimit)
```
Stefan-Ethernal:
```suggestion
			"tx gas limit", tx.Gas, "block gas limt", b.params.GasLimit)
```

Stefan-Ethernal (approved):
LGTM :rocket:

PR #1406, opened 2023-04-20
[RFC-201/205]: EVM-603-Register supernet and child chain

goran-ethernal:
# Description

This PR registers child chain and supernet manager to root `StakeManager` contract. Now, `chainID` parameter (ID of child chain) is gotten when registering it to rootchain.

After `StakeManager` gets deployed, as well as `SupernetManager`, we register the child chain by sending the `registerChildChain` transaction to `StakeManager`, which emits a `ChildManagerRegistered` event with `ID` of the child chain. This is all done while deploying the root contracts in the `deploy` command.

After the transaction is executed successfully, we will save the `childID` to the `genesis` file.

**IMPORTANT NOTE**: This PR does not remove the `chainID` flag from the `genesis` command. It removes it's setting when running `genesis` for `polybft`. This will be revised later on. `E2E` are expected to fail.

# Changes include

# Breaking changes

Removes the setting of `chainID` from the `chai-id` flag in `genesis` command for `polybft`. For `polybft` this is updated in the `genesis` file, after the root contracts are deployed.

# Checklist

## Testing

### DIFF
--- a/command/rootchain/deploy/deploy.go
+++ b/command/rootchain/deploy/deploy.go
@@ -320,6 +320,10 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
 			artifact: contractsapi.StateSender,
 		},
 		{
+			name:     checkpointManagerName,
+			artifact: contractsapi.CheckpointManager,
+		},
+		{
 			name:     blsName,
 			artifact: contractsapi.BLS,
 		},
@@ -379,13 +383,6 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
 
 				return append(artifact.Bytecode, encoded...), nil
 			},
-		},
-		{
-			// since checkpoint manager needs chain Id,
-			// and since chain id is gotten when registering chain on stake manager
-			// we must deploy it the last
-			name:     checkpointManagerName,
-			artifact: contractsapi.CheckpointManager,
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


command/rootchain/deploy/deploy.go
```diff
@@ -478,6 +485,57 @@ func populateExistingTokenAddr(eth *jsonrpc.Eth, tokenAddr, tokenName string,
...
+
+	for _, log := range receipt.Logs {
+		doesMatch, err := childChainRegisteredEvent.ParseLog(log)
+		if !doesMatch {
+			continue
+		}
+
+		if err != nil {
+			return 0, err
+		}
```
Stefan-Ethernal: Nitpick. I'm aware that `ParseLog` first returns false if signature is not matched, but anyways it is better to first check if there are errors, just in case.
```suggestion
		if err != nil {
			return 0, err
		}

                if !doesMatch {
			continue
		}
```

Stefan-Ethernal (approved):

command/rootchain/deploy/deploy.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1406/commits/e747436ded2572c7cb896cb1deb8daaccc4b451a

Nemanja0x (approved):

igorcrevar (approved):

stana-miric (approved):

PR #1407, opened 2023-04-20
Utilize typed atomics

Stefan-Ethernal:
# Description

This PR utilizes typed atomics where appropriate, such as: `atomic.Bool` and `atomic.Pointer[T]`, which is possible now when we are running golang 1.19.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime_test.go
+++ b/consensus/polybft/consensus_runtime_test.go
@@ -301,15 +301,15 @@ func TestConsensusRuntime_FSM_NotInValidatorSet(t *testing.T) {
 		Key: createTestKey(t),
 	}
 	runtime := &consensusRuntime{
-		proposerCalculator:  NewProposerCalculatorFromSnapshot(snapshot, config, hclog.NewNullLogger()),
-		activeValidatorFlag: 1,
-		config:              config,
+		proposerCalculator: NewProposerCalculatorFromSnapshot(snapshot, config, hclog.NewNullLogger()),
+		config:             config,
 		epoch: &epochMetadata{
 			Number:     1,
 			Validators: validators.getPublicIdentities(),
 		},
 		lastBuiltBlock: &types.Header{},
 	}
+	runtime.setIsActiveValidator(true)
 
 	err := runtime.FSM()
 	assert.ErrorIs(t, err, errNotAValidator)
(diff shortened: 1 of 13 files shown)


consensus/polybft/consensus_runtime_test.go
```diff
@@ -301,15 +301,15 @@ func TestConsensusRuntime_FSM_NotInValidatorSet(t *testing.T) {
 		Key: createTestKey(t),
 	}
 	runtime := &consensusRuntime{
-		proposerCalculator:  NewProposerCalculatorFromSnapshot(snapshot, config, hclog.NewNullLogger()),
-		activeValidatorFlag: 1,
```
Nemanja0x, line 305: Why are we missing this here and in all the other tests? True by default?

network/gossip.go
```diff
@@ -126,7 +126,6 @@ func (s *Server) NewTopic(protoID string, obj proto.Message) (*Topic, error) {
 		topic:   topic,
 		typ:     reflect.TypeOf(obj).Elem(),
 		closeCh: make(chan struct{}),
-		closed:  new(uint64),
```
Nemanja0x, line 129: Default value?

network/gossip.go
```diff
@@ -71,7 +71,7 @@ func (t *Topic) Subscribe(handler func(obj interface{}, from peer.ID)) error {
 	}
 
 	// Mark topic active.
-	atomic.StoreUint64(t.closed, 0)
+	t.closed.Swap(false)
```
Nemanja0x: Why do not use `Store()`

Nemanja0x (approved):
Please look at the comments. They are not blockers.

consensus/polybft/consensus_runtime_test.go
Stefan-Ethernal: Nope, but I've set it to true here (seems that I cannot set it directly to true without need to call `setIsActiveValidator`, but maybe I'm missing out something):
`runtime.setIsActiveValidator(true)`

network/gossip.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1407/commits/09b94cecd1c6f00cc37961536058a8c6d0dbeed6. 

Although the documentation states that by default it is `false`.

network/gossip.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1407/commits/09b94cecd1c6f00cc37961536058a8c6d0dbeed6

igorcrevar (approved):

vcastellm (commented):
Late LGTM

PR #1409, opened 2023-04-21
[RFC-201/205]: EVM-608-Supernet command

goran-ethernal:
# Description

Through this PR we added a new command called `supernet`. The command will be in e2e tests to finalize initial validator set on rootchain, and to enable staking, before starting the child chain.

Command flags:
- `data-dir` - the directory for the Polygon Edge data if the local FS is used. (mutually exclusive with the `config` and `private-key` flags).
- `config` - the path to the SecretsManager config file, if omitted, the local FS secrets manager is used (mutually exclusive with the `data-dir` and `private-key` flags).
- `private-key` - hex encoded private key of the `SupernetManager` contract deployer. This flag is only a temporary solution, since Secret Manager (for now) does not support the absence of a bls key ((mutually exclusive with the `data-dir` and `config` flags).
- `finalize-genesis-set` - finalizes genesis validator set on root contract (`SupernetManager`).
- `enable-staking` - enables staking on root chain (`SupernetManager` and `StakeManager`).

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/rootchain/helper/metadata.go
+++ b/command/rootchain/helper/metadata.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/0xPolygon/polygon-edge/command/polybftsecrets"
+	polybftWallet "github.com/0xPolygon/polygon-edge/consensus/polybft/wallet"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"github.com/umbracle/ethgo"
@@ -88,4 +90,27 @@ func ReadRootchainIP() (string, error) {
 	}
 
 	return fmt.Sprintf("http://%s:%s", ports[0].HostIP, ports[0].HostPort), nil
+}
+
+func GetECDSAKey(privateKey, accountDir, accountConfig string) (ethgo.Key, error) {
+	if privateKey != "" {
+		key, err := GetRootchainPrivateKey(privateKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to initialize private key: %w", err)
+		}
+
+		return key, nil
+	}
+
+	secretsManager, err := polybftsecrets.GetSecretsManager(accountDir, accountConfig, true)
+	if err != nil {
+		return nil, err
+	}
+
+	key, err := polybftWallet.GetEcdsaFromSecret(secretsManager)
+	if err != nil {
+		return nil, err
+	}
+
+	return key, nil
 }
--- a/command/rootchain/helper/metadata.go
+++ b/command/rootchain/helper/metadata.go
@@ -107,10 +107,5 @@ func GetECDSAKey(privateKey, accountDir, accountConfig string) (ethgo.Key, error
 		return nil, err
 	}
 
-	key, err := polybftWallet.GetEcdsaFromSecret(secretsManager)
-	if err != nil {
-		return nil, err
-	}
-
-	return key, nil
+	return polybftWallet.GetEcdsaFromSecret(secretsManager)
 }
--- a/command/rootchain/supernet/supernet.go
+++ b/command/rootchain/supernet/supernet.go
@@ -99,7 +99,7 @@ func runCommand(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("enlist validator failed: %w", err)
 	}
 
-	gasPrice, err := rootHelper.GetGasPriceOnRoot(params.jsonRPC)
+	gasPrice, err := txRelayer.GetGasPrice()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


stana-miric (approved):

command/rootchain/helper/metadata.go
```diff
@@ -115,3 +117,26 @@ func GetGasPriceOnRoot(jsonrpcEndpoint string) (uint64, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	key, err := polybftWallet.GetEcdsaFromSecret(secretsManager)
+	if err != nil {
+		return nil, err
+	}
+
+	return key, nil
```
Stefan-Ethernal:
```suggestion
	return polybftWallet.GetEcdsaFromSecret(secretsManager)
```

command/rootchain/whitelist/whitelist_validators.go
```diff
@@ -84,27 +83,9 @@ func runCommand(cmd *cobra.Command, _ []string) error {
...
-			return err
-		}
-
-		key, err := wallet.GetEcdsaFromSecret(secretsManager)
-		if err != nil {
-			return err
-		}
-
-		ecdsaKey = key
+	ecdsaKey, err := rootHelper.GetECDSAKey(params.privateKey, params.accountDir, params.accountDir)
```
Stefan-Ethernal:
```suggestion
	ecdsaKey, err := rootHelper.GetECDSAKey(params.privateKey, params.accountDir, params.accountConfig)
```

command/rootchain/supernet/supernet.go
```diff
@@ -0,0 +1,163 @@
...
+	defer outputter.WriteOutput()
+
+	ownerKey, err := rootHelper.GetECDSAKey(params.privateKey, params.accountDir, params.accountConfig)
+	if err != nil {
+		return err
+	}
+
+	txRelayer, err := txrelayer.NewTxRelayer(txrelayer.WithIPAddress(params.jsonRPC))
+	if err != nil {
+		return fmt.Errorf("enlist validator failed: %w", err)
```
Stefan-Ethernal, line 99: This error message doesn't seem correct.

command/rootchain/supernet/supernet.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/spf13/cobra"
+	"github.com/umbracle/ethgo"
+)
+
+var params supernetParams
+
+func GetCommand() *cobra.Command {
+	registerCmd := &cobra.Command{
+		Use:     "supernet",
+		Short:   "Supernet command",
```
Stefan-Ethernal: WDYT?
```suggestion
		Short:   "supernet initialization finalization command",
```

command/rootchain/supernet/supernet.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/spf13/cobra"
+	"github.com/umbracle/ethgo"
+)
+
+var params supernetParams
+
+func GetCommand() *cobra.Command {
+	registerCmd := &cobra.Command{
+		Use:     "supernet",
```
Stefan-Ethernal, line 21: Thinking out loud and it could be that I'm not right...

Does it make sense to split into two separate sub-commands:
1. `supernet finalize-init`
2. `supernet enable-staking`

goran-ethernal: I would prefer not to. We can use it with both flags, or with one, no need to separate it into two, if we want two steps. But also, when it's a single command, we can immediately finalize genesis set and enable staking, which we will need for e2e tests.

command/rootchain/helper/metadata.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1409/commits/fcd365aa753c53dd5bedfa7fad43094e7950aa00

command/rootchain/supernet/supernet.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1409/commits/fcd365aa753c53dd5bedfa7fad43094e7950aa00

command/rootchain/whitelist/whitelist_validators.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1409/commits/fcd365aa753c53dd5bedfa7fad43094e7950aa00

command/rootchain/supernet/supernet.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1409/commits/de82b3eb3e18a20f8970aeee84f000e18399c0e7

Nemanja0x (approved):

Stefan-Ethernal (approved):

PR #1410, opened 2023-04-21
Cleanup `build` and `fuzz-test` GH workflows

Stefan-Ethernal:
# Description

This PR cleans up GH workflows in a way that it doesn't checkout git submodules anymore for the workflows which don't need it. It also adds names to some steps which were missing it.

# Changes include

# Checklist

## Testing


.github/workflows/fuzz-test.yml
```diff
@@ -27,11 +27,6 @@ jobs:
 
       - name: Checkout Code
         uses: actions/checkout@v3
-        with:
-          submodules: recursive
-
-      - name: Install Dependencies
-        run: ./setup-ci.sh
```
Nemanja0x, line 34: Why is this removed? What if want to run fuzz test independently?

Stefan-Ethernal: Glad that you have asked, somehow I felt you were going to ask me these kinds of questions. :smile: 

> Why is this removed?

It is removed because it is not needed and it has a negative aspect on CI duration.

> What if want to run fuzz test independently?

You are able to do so because `fuzz-test.yml` is in fact independent GH workflow.

In all seriousness, as you can see I have manually triggered `fuzz-tests` here and it was successfully executed
(https://github.com/0xPolygon/polygon-edge/actions/runs/4765173994).

Now to explain what I did here:
- we don't need git submodules in place for fuzz testing and that is why I've removed `submodules: recursive`
- we don't need a Solidity compiler to be installed on the CI for the fuzz tests and that is why I have removed these two lines:

```yml
 - name: Install Dependencies
   run: ./setup-ci.sh
```

Nemanja0x (approved):

stana-miric (approved):

vcastellm (commented):
nice

PR #1440, opened 2023-04-26
TxnType is not returned in JSON-RPC response

flavor-town:
# Description

The transaction type is not returned in the json-rpc reponse leading to in correct client hash calculations. 
Default TxType is 0 = LegacyTx. 

# Changes include

# Breaking changes

The transaction response now includes a `type` field, which will tell the end user if the txn is a `LegacyTx`, `StateTx`, or `DynamicFeeTx`. I believe this is required for EIP1559 txns to be processed and displayed correctly. 

# Checklist

## Testing


jsonrpc/eth_endpoint_test.go
```diff
@@ -242,6 +242,61 @@ func TestEth_GetNextNonce(t *testing.T) {
 	}
 }
 
+func TestEth_TxnType(t *testing.T) {
```
flavor-town, line 245: Looking at this again, this doesn't actually make any requests and just validates decoding

Stefan-Ethernal (approved):
Pay attention to this failing unit test: `TestBlock_Encoding`. Otherwise looks good.

jsonrpc/testsuite/transaction-eip1559.json
```diff
@@ -14,5 +14,6 @@
     "from": "0x0300000000000000000000000000000000000000",
     "blockHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
     "blockNumber": "0x1",
-    "transactionIndex": "0x2"
+    "transactionIndex": "0x2",
+    "type": "0x0"
```
Stefan-Ethernal: Nitpick: 
since this is a dynamic fee tx, probably it would be more correct to put dynamic fee tx type

```suggestion
    "type": "0x02"
```

flavor-town: Doing this the marshaling is different. , any idea why this would happen: 
```        	            	
- (string) (len=4) "type": (string) (len=4) "0x02",
+ (string) (len=4) "type": (string) (len=3) "0x2",
 ```

goran-ethernal (approved):

PR #1445, opened 2023-04-27
[RFC-201/205]: EVM-622-Save full validator set in db

goran-ethernal:
# Description

In order to have correct validator set update, we need to store entire validator set to `boltdb` (not just active ones).
This is done on every block finalization, where we update the full validator set in our `boltdb` using the transfer events fired by `ValidatorSet`contract on child chain, when validator stake changes on child chain.

**IMPORTANT NOTE**: `e2e` tests will fail, until the last PR in the RFC-201/205 saga, which fixes them.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/stake_manager.go
+++ b/consensus/polybft/stake_manager.go
@@ -341,6 +341,7 @@ func (sc *validatorStakeMap) removeStake(address types.Address, amount *big.Int)
 // getActiveValidators returns all validators (*ValidatorMetadata) in sorted order
 func (sc validatorStakeMap) getActiveValidators(maxValidatorSetSize int) []*ValidatorMetadata {
 	activeValidators := make([]*ValidatorMetadata, 0, len(sc))
+
 	for _, v := range sc {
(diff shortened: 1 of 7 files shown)


consensus/polybft/stake_manager.go
```diff
@@ -95,33 +86,14 @@ func (s *stakeManager) PostBlock(req *PostBlockRequest) error {
...
-
-		return &ValidatorSetDelta{}, nil
+		return err
 	}
 
-	// stake counter holds sorted stakes by current and (possible) new validators
-	// he will add to map current stake (voting power) of the current validators
-	// on object instantiation
-	stakeCounter := newStakeCounter(currentValidatorSet.Copy())
+	stakeCounter := newStakeCounter(fullValidatorSet, fullValidatorSet, len(fullValidatorSet))
```
Nemanja0x: Both times `fullValidatorSet` is submitted

Nemanja0x (approved):

Stefan-Ethernal (approved):

PR #1456, opened 2023-04-30
Integrate bridge access list contracts

vcastellm:
# Description

This PR integrates the new AccessList enabled versions of the ERC20/721/1155 Predicates for the bridge In the same address as the original ones.

# Changes include

# Checklist

## Testing


consensus/polybft/polybft_config.go
```diff
@@ -44,6 +44,12 @@ type PolyBFTConfig struct {
 	// NativeTokenConfig defines name, symbol and decimal count of the native token
 	NativeTokenConfig *TokenConfig `json:"nativeTokenConfig"`
 
+	// BridgeAllowListActive indicates whether bridge allow list is active
+	BridgeAllowListAdmin types.Address `json:"bridgeAllowListAdmin"`
```
goran-ethernal, line 47: Should we add these to the `BridgeConfig` struct?

e2e-polybft/e2e/bridge_test.go
```diff
@@ -124,7 +126,7 @@ func TestE2E_Bridge_Transfers(t *testing.T) {
 		rawKey, err := senderAccount.Ecdsa.MarshallPrivateKey()
 		require.NoError(t, err)
 
-		// send withdraw transaction
+		// try to withdraw again
```
Stefan-Ethernal:
```suggestion
		// send withdraw transaction
```

consensus/polybft/polybft_config.go
vcastellm: I was thinking on this, yes it looks cleaner

vcastellm: I propose to do it in a follow-up PR

e2e-polybft/e2e/bridge_test.go
```diff
@@ -54,7 +55,8 @@ func TestE2E_Bridge_Transfers(t *testing.T) {
 
 	cluster := framework.NewTestCluster(t, 5,
 		framework.WithNumBlockConfirmations(numBlockConfirmations),
-		framework.WithEpochSize(epochSize))
+		framework.WithEpochSize(epochSize),
+	)
```
Stefan-Ethernal: Nit
```suggestion
		framework.WithEpochSize(epochSize))
```

e2e-polybft/e2e/bridge_test.go
```diff
@@ -787,3 +789,202 @@ func TestE2E_Bridge_ChangeVotingPower(t *testing.T) {
...
+
+	cluster.WaitForReady(t)
+
+	polybftCfg, err := polybft.LoadPolyBFTConfig(path.Join(cluster.Config.TmpDir, chainConfigFileName))
+	require.NoError(t, err)
+
+	validatorSrv := cluster.Servers[0]
+	childEthEndpoint := validatorSrv.JSONRPC().Eth()
+
+	t.Run("bridge ERC 20 tokens", func(t *testing.T) {
```
Stefan-Ethernal, line 835: Probably no need to create a sub-test (at least for now). WDYT?

e2e-polybft/e2e/bridge_test.go
```diff
@@ -787,3 +789,202 @@ func TestE2E_Bridge_ChangeVotingPower(t *testing.T) {
...
+
+		senderAccount, err := sidechain.GetAccountFromDir(validatorSrv.DataDir())
+		require.NoError(t, err)
+
+		t.Logf("Withdraw sender: %s\n", senderAccount.Ecdsa.Address())
+
+		rawKey, err := senderAccount.Ecdsa.MarshallPrivateKey()
+		require.NoError(t, err)
+
+		// send withdraw transaction
```
Stefan-Ethernal: WDYT?
```suggestion
		// send withdraw transaction. 
        // It should fail because sender is not white-listed.
```

e2e-polybft/e2e/bridge_test.go
```diff
@@ -787,3 +789,202 @@ func TestE2E_Bridge_ChangeVotingPower(t *testing.T) {
...
+		err = cluster.Bridge.Withdraw(
+			common.ERC20,
+			hex.EncodeToString(rawKey),
+			strings.Join(receivers[:], ","),
+			strings.Join(amounts[:], ","),
+			"",
+			validatorSrv.JSONRPCAddr(),
+			contracts.NativeERC20TokenContract)
+		require.Error(t, err)
```
Stefan-Ethernal, line 906: Does it make sense to check for specific error message?

e2e-polybft/e2e/bridge_test.go
```diff
@@ -787,3 +789,202 @@ func TestE2E_Bridge_ChangeVotingPower(t *testing.T) {
...
+
+		{
+			input, _ := addresslist.SetEnabledSignatureFunc.Encode([]interface{}{senderAccount.Ecdsa.Address()})
+			adminSetTxn := cluster.MethodTxn(t, admin, contracts.AllowListBridgeAddr, input)
```
Stefan-Ethernal: Reword to `setEnabledTxn` maybe?

e2e-polybft/e2e/bridge_test.go
```diff
@@ -787,3 +789,202 @@ func TestE2E_Bridge_ChangeVotingPower(t *testing.T) {
...
+			strings.Join(receivers[:], ","),
+			strings.Join(amounts[:], ","),
+			"",
+			validatorSrv.JSONRPCAddr(),
+			contracts.NativeERC20TokenContract)
+		require.NoError(t, err)
+
+		{
+			input, _ := addresslist.SetEnabledSignatureFunc.Encode([]interface{}{senderAccount.Ecdsa.Address()})
+			adminSetTxn := cluster.MethodTxn(t, admin, contracts.BlockListBridgeAddr, input)
```
Stefan-Ethernal: Reword to `setDisabledTxn` maybe?

consensus/polybft/polybft_config.go
```diff
@@ -43,6 +43,12 @@ type PolyBFTConfig struct {
 	// NativeTokenConfig defines name, symbol and decimal count of the native token
 	NativeTokenConfig *TokenConfig `json:"nativeTokenConfig"`
 
+	// BridgeAllowListActive indicates whether bridge allow list is active
```
Stefan-Ethernal: Comment is not up to date.

consensus/polybft/polybft_config.go
```diff
@@ -43,6 +43,12 @@ type PolyBFTConfig struct {
...
+	BridgeAllowListAdmin types.Address `json:"bridgeAllowListAdmin"`
+
+	// BridgeBlockListActive indicates whether bridge block list is active
```
Stefan-Ethernal: Comment is not up to date.

consensus/polybft/polybft_config.go
Stefan-Ethernal: Do we even need this data redundancy? We already should be able to have bridge access lists (together with other access lists) persisted in `*chain.Chain.Params`. From there we could extract all the information needed (owner of predicate, is given access list enabled or not etc.).
I'd like to hear your opinions as well.

e2e-polybft/e2e/bridge_test.go
vcastellm: It provides better naming and no drawbacks, I prefer to keep it.

consensus/polybft/contracts_initializer.go
```diff
@@ -76,6 +76,41 @@ func getInitChildERC20PredicateInput(config *BridgeConfig) ([]byte, error) {
...
 
+// getInitChildERC20PredicateAccessListInput builds input parameters for ChildERC20PredicateAccessList SC initialization
+func getInitChildERC20PredicateAccessListInput(config PolyBFTConfig) ([]byte, error) {
+	//nolint:godox
+	// to be fixed with EVM-541
+	// TODO: @Stefan-Ethernal Temporary workaround just to be able to run cluster in non-bridge mode, until SC is fixed
+	rootERC20PredicateAddr := types.StringToAddress(disabledBridgeRootPredicateAddr)
+	rootERC20Addr := types.ZeroAddress
+
+	if config.Bridge != nil {
```
goran-ethernal, line 89: I think you can remove this condition an TODO when you merge with `develop` since RFC-201/205 changes are merged. We will always have `Bridge` configured.

consensus/polybft/polybft_config.go
vcastellm: I find cleaner to separate the logic and send the appropriate value here to the GenesisPostHook but it's true that it's redundant. We'll also need to modify the contract init getters to pass more variables or the `chain.Params`

consensus/polybft/contracts_initializer.go
vcastellm: Yes, but it means we can remove the previous 2 lines as well and that's out of the scope of the PR, adding TODO

consensus/polybft/polybft_config.go
vcastellm: done

consensus/polybft/polybft_config.go
vcastellm: done

e2e-polybft/e2e/bridge_test.go
vcastellm: The error is `failed to execute withdrawal` is that specific enough?

e2e-polybft/e2e/bridge_test.go
vcastellm: done

e2e-polybft/e2e/bridge_test.go
vcastellm: done

Stefan-Ethernal (approved):

consensus/polybft/polybft_config.go
goran-ethernal: > I propose to do it in a follow-up PR

I am thinking out loud, but if we are to move this later to the Bridge config, doesn't that break the old `genesis` file? If those values were on one place in one release, and then in another place in the second release, then we break the genesis file? Do we even care at this point? Probably not.

goran-ethernal (approved):

consensus/polybft/polybft_config.go
vcastellm: Good question, but I think the important part that is the Allocation are not going to move, so extra values in genesis won't be a problem and won't break genesis, I think we are good in that regard.

PR #1463, opened 2023-05-01
Polybft adaptation for writing traces on all validators

stana-miric:
Before this PR, trace was written only for ibft protocol (on proposer node). Polybft protocol is added to feat-zero branch in previous commit when the 0.8.1 version of edge is merged into it but the trace functionality is not integrated for polybft at that moment.
This PR adds:
- support for writing traces for polybft protocol in all cases (does not meter if the block is added by consensus or by syncer, or the note is proposer or regular validator). This means that queering any validator node to get the trace for block n will return result.
This PR does not fix the issue on ibft related to the fact that the trace for the block n is written only on validator for block n (it is plan to use polybft for zero integration)

The changes include:
- extending FullBlock type with the Trace
- adding new config type for the blockchain, it wraps existing chain config and adds data dir which is needed when writing trace on block insertion
- trace data is added to the full block when: 
  - the proposer builds the block (and saves it in the fsm target for the later insertion in the blockchain),  
  - the validator process received block from the consensus (and saves it in the fsm target for the later insertion in the blockchain),
  - the syncer receives the block which will be inserted to the blockchain  

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1,9 +1,12 @@
 package blockchain
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/big"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 
@@ -46,8 +49,8 @@ type Blockchain struct {
 	executor  Executor
 	txSigner  TxSigner
 
-	config  *chain.Chain // Config containing chain information
-	genesis types.Hash   // The hash of the genesis block
+	config  *Config    // Config containing chain information
+	genesis types.Hash // The hash of the genesis block
 
 	headersCache    *lru.Cache // LRU cache for the headers
 	difficultyCache *lru.Cache // LRU cache for the difficulty
@@ -101,6 +104,7 @@ type BlockResult struct {
 	Root     types.Hash
 	Receipts []*types.Receipt
 	TotalGas uint64
+	Trace    *types.Trace
 }
 
 // updateGasPriceAvg updates the rolling average value of the gas price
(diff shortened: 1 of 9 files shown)


ferranbt (approved):

igorcrevar (approved):

Nemanja0x (approved):

blockchain/blockchain.go
```diff
@@ -1282,6 +1290,22 @@ func (b *Blockchain) writeFork(header *types.Header) error {
...
 }
 
+// writeTrace writes the block trace to the file
+func (b *Blockchain) writeTrace(trace *types.Trace, blockNumber uint64) error {
+	raw, err := json.Marshal(trace)
+	if err != nil {
+		return err
+	}
+
+	consensusDir := filepath.Join(b.config.DataDir, "consensus")
```
vcastellm, line 1300: Move this to a constant?

blockchain/blockchain.go
```diff
@@ -1282,6 +1290,22 @@ func (b *Blockchain) writeFork(header *types.Header) error {
...
+	if err := ioutil.WriteFile(
+		filepath.Join(consensusDir, fmt.Sprintf("trace_%d.json", blockNumber)), raw, 0600); err != nil {
```
vcastellm, line 1302: Should all traces be kept in disk?

stana-miric: This is a good point for discussion since I was wondering the same and already created a [task](https://polygon.atlassian.net/browse/EVM-634) to discuss that. I can schedule a meeting next week when everyone is back from the holiday.

blockchain/blockchain.go
stana-miric: Will address this after the meeting, once we decide if we're going with the db or sticking with the files on disk

PR #1466, opened 2023-05-02
Update Ethereum tests spec to v10.4

Stefan-Ethernal:
# Description

Updated test cases suite to v10.4. This version includes london hardfork test cases and does not inlude the merge one and all further updates. Currently, london hardfork test cases are ignored since there are some differences in gas consumption. 

Also, TestEVM is not needed anymore since those test cases were moved to the regular state tests. Eventually, if there is any difference in the tx execution, the block hash is going to be different and a test will fail. 

# Changes include

# Checklist

## Testing


vcastellm (approved):
LGTM

tests/state_test.go
```diff
@@ -15,40 +15,57 @@ import (
...
-	msg, err := c.Transaction.At(p.Indexes)
+	var baseFee *big.Int
+
+	if config.IsLondon(0) {
+		if c.Env.BaseFee != "" {
+			baseFee = stringToBigIntT(t, c.Env.BaseFee)
+		} else {
+			// Retesteth uses `0x10` for genesis baseFee. Therefore, it defaults to
+			// parent - 2 : 0xa as the basefee for 'this' context.
+			baseFee = big.NewInt(0x0a)
```
igorcrevar: maybe put this 0x0a value inside some const?

igorcrevar (approved):
Approved, just a small neat-pick

tests/state_test.go
begmaroman: Done

PR #1475, opened 2023-05-06
Allow min commitment of single state sync event

Stefan-Ethernal:
# Description

Through this PR we are supporting a single state sync event per commitment. There is a real case where one deposit can be put on hold until another one happens, because, before this PR, minimum commitment size was 2. 

This now helps users to quickly execute their deposit on `L2`, since if no other deposit happens in the given sprint, commitment with only that one state sync event will be committed to `StateReceiver` contract, and user can execute its deposit without the need to wait for another one.

# Changes include

# Checklist

## Testing


Stefan-Ethernal (commented):
:heavy_check_mark:  LGTM (cannot approve it because I'm a creator of the PR).

Nemanja0x (approved):

igorcrevar (approved):

PR #1493, opened 2023-05-10
Extracted validator related structures into consensus/polybft/validator package

dusan-maksimovic:
# Description

This PR includes validator related files into separate package.
- Files validator_metadata.go and validator_set.go, with related test files, extracted to consensus/polybft/validator/ package
- ValidatorSetDelta and associated methods moved from polybft/extra.go to polybft/validators/validator_set.go
- Validator, validatorRaw and associated methods moved from polybft/polybft_config.go to polybft/validators/validator_config.go
	- Renamed Validator to GenesisValidator
	- Renamed validatorRaw to genesisValidatorRaw
- PostEpochRequest struct modified to use ValidatorSet interface instead of unexported validatorSet structure
- testValidator, testValidators and accociated methods moved from consensus/polybft/mocks_test.go to consensus/polybft/validator/test_mocks.go
	- unexported structs, fields and functions converted to exported, since they are used by various tests from the polybft package

# Changes include

# Checklist

## Testing


dusan-maksimovic:

command/genesis/polybft_params.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/bitmap"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi/artifact"
+	val "github.com/0xPolygon/polygon-edge/consensus/polybft/validator"
```
vcastellm: Nitpick but IMO it's ok with the full package name `validator`

consensus/polybft/handlers.go
```diff
@@ -23,5 +26,5 @@ type PostEpochRequest struct {
 	SystemState SystemState
 
 	// ValidatorSet is the validator set for the new epoch
-	ValidatorSet *validatorSet
+	ValidatorSet validator.ValidatorSet
```
vcastellm, line 29: Shouldn't this be a pointer?

vcastellm (approved):
LGTM aside from the comments

consensus/polybft/validator/validator_set.go
```diff
@@ -0,0 +1,244 @@
...
+func (vs validatorSet) Len() int {
+	return vs.validators.Len()
+}
+
+func (vs validatorSet) TotalVotingPower() big.Int {
+	return *vs.totalVotingPower
+}
+
+// ValidatorSetDelta holds information about added and removed validators compared to the previous epoch
+type ValidatorSetDelta struct {
```
goran-ethernal: I would move the `ValidatorSetDelta` to its own file.

consensus/polybft/validator/validator_config.go
```diff
@@ -0,0 +1,97 @@
+package validator
```
goran-ethernal: Since it only contains `GenesisValidator` struct and its functions, I would rename this file to `genesis_validator.go`.

command/genesis/polybft_params.go
dusan-maksimovic: There was an iteration variable called validator in getValidatorAccounts() so i had to rename the package import. I will switch back to full package name and rename the iteration variable instead.

consensus/polybft/handlers.go
dusan-maksimovic: Here I switched from (now) unexported structure validatorSet (now in new validator package) to the ValidatorSet interface. This field of the PostEpochRequest struct is always used for read-only access, so it didn't need to be a pointer in the first place. Also, there are other places where this interface was already used (e.g. stateSyncManager, fsm) also without the pointer. The structure validatorSet itself implements interface as value-receiver, which also indicates it is intended to be used in read-only fashion.

goran-ethernal (approved):

consensus/polybft/validator/validator_metadata_test.go
```diff
@@ -11,10 +12,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// generateRandomBytes generates byte array with random data of 32 bytes length
+func generateRandomBytes(t *testing.T) (result []byte) {
```
Stefan-Ethernal, line 16: I'd consider removing this function altogether. It is used in a single place and is used to set an address of validator. Instead we can use some hardcoded address instead (because this function basically creates a hash, i.e. random byte array of 32 bytes length, whereas address is 20 bytes long, so it is a bit misleading).

consensus/polybft/extra_test.go
```diff
@@ -559,11 +560,11 @@ func TestExtra_CreateValidatorSetDelta_Cases(t *testing.T) {
 func TestExtra_CreateValidatorSetDelta_BlsDiffer(t *testing.T) {
```
Stefan-Ethernal: Does it make sense to extract all validator set delta unit tests to a separate file named `validator_set_delta_test.go` inside validator package? Not sure is it feasible though, because of potential import cycle, but think it is worth trying. WDYT?

Stefan-Ethernal (commented):
Generally looks good, but leaving some suggestions.

igorcrevar (approved):

Nemanja0x (approved):

consensus/polybft/extra_test.go
dusan-maksimovic: I extracted all of them except TestExtra_InitGenesisValidatorsDelta because this one would cause issues with circular references. I had to move CreateValidatorSetDelta() into validator package as well.

Stefan-Ethernal (approved):
:1st_place_medal:

PR #1461, opened 2023-05-01
Fix RLP marshal byte array and enhancements to stake manager

Nemanja0x:
# Description

A bug is discovered in the `fastrlp` library in the `NewBytes` method. We should use `NewCopyBytes` wherever we have a non-transient slice/array, since `NewBytes` was not resetting memory properly. In a meanwhile, `fastrlp` library was fixed by wrapping `NewCopyBytes` function within the `NewBytes` (https://github.com/umbracle/fastrlp/pull/14), meaning that these two functions are interchangeable, so this PR updates the version of `fastrlp` library as well.
The stake manager is reverted to the behavior before the workaround for the bug in `fastrlp` is applied.

# Changes include

- [X] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backward-compatible and/or changes current functionality)

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Checklist

- [X] I have assigned this PR to myself
- [ ] I have added at least 1 reviewer
- [ ] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [X] I have tested this code with the official test suite
- [ ] I have tested this code manually

PR #1500, opened 2023-05-11
Added Stake Manager fuzz tests

jelacamarko:
# Description

Added Stake Manager fuzz tests

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


jelacamarko:

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,281 @@
...
+func generateTestAccountF(f *testing.F) *wallet.Account {
+	f.Helper()
+
+	acc, err := wallet.GenerateAccount()
+	require.NoError(f, err)
+
+	return acc
+}
+
+func newTestStateF(f *testing.F) *State {
```
goran-ethernal: No need for this function. Just change the original `newTestState` function in `helpers_test.go` to receive `testing.TB`. It's an interface which is implemented by all testing structs.

`// newTestState creates new instance of state used by tests.
func newTestState(t testing.TB) *State {`

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,281 @@
...
+	f.Helper()
+
+	return &testValidator{
+		alias:       alias,
+		votingPower: votingPower,
+		account:     generateTestAccountF(f),
+	}
+}
+
+func generateTestAccountF(f *testing.F) *wallet.Account {
```
goran-ethernal: Same here. No need for this function. We already have it in the `helpers_test.go` just change it to use `testing.TB`:

`func generateTestAccount(t testing.TB) *wallet.Account {`

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	aliases := make([]string, validatorsCount)
+	for i := 0; i < validatorsCount; i++ {
+		aliases[i] = strconv.Itoa(i)
+	}
+
+	return newTestValidatorsWithAliasesF(f, aliases)
+}
+
+func newTestValidatorsWithAliasesF(f *testing.F, aliases []string, votingPowers ...[]uint64) *testValidators {
```
goran-ethernal: Same here.

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+		fullValidatorSet := validators.getPublicIdentities().Copy()
+		validatorToUpdate := fullValidatorSet[x]
+		validatorToUpdate.VotingPower = big.NewInt(votingPower)
+
+		_, _ = stakeManager.UpdateValidatorSet(epoch, validators.getPublicIdentities())
+	})
+}
+
+func newTestValidatorsF(f *testing.F, validatorsCount int) *testValidators {
```
goran-ethernal: Same here.

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,205 @@
...
+		if err != nil {
+			return
+		}
+		newEpochID, err := tp.GetUint64()
+		if err != nil {
+			return
+		}
+		var validators AccountSet
+		err = tp.Fill(validators)
+		if err != nil {
```
Stefan-Ethernal: Nit

```suggestion
		if err = tp.Fill(validators); err != nil {
```

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,205 @@
...
+			return
+		}
+
+		_ = stakeManager.PostEpoch(&PostEpochRequest{
```
Stefan-Ethernal: Consider adding error check here. WDYT?

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		if x > uint64(len(aliases)-1) {
+			t.Skip()
+		}
+		votingPower, err := tp.GetInt64()
+		if err != nil {
+			return
+		}
+
+		_, _ = stakeManager.UpdateValidatorSet(epoch, validators.getPublicIdentities(aliases[x:]...))
```
Stefan-Ethernal: Add assertion that there are no errors?

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+		fullValidatorSet := validators.getPublicIdentities().Copy()
+		validatorToUpdate := fullValidatorSet[x]
+		validatorToUpdate.VotingPower = big.NewInt(votingPower)
+
+		_, _ = stakeManager.UpdateValidatorSet(epoch, validators.getPublicIdentities())
```
Stefan-Ethernal: Error assertion?

consensus/polybft/stake_manager_fuzz_test.go
jelacamarko: We can not assert any function responses because we don't know what will fuzzer generate. We can only compare responses (if we had that case, e.g. two functions that should have same response)

.github/workflows/lint.yml
```diff
@@ -2,13 +2,13 @@
 name: Lint
```
Stefan-Ethernal: I think that we need to revert changes in this file, because it reverts behavior from #1512.

stana-miric (approved):

Nemanja0x (approved):

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,288 @@
...
+	fieldNum := structVal.NumField()
+
+	for i := 0; i < fieldNum; i++ {
+		field := structVal.Field(i)
+		fieldName := structType.Field(i).Name
+
+		isSet := field.IsValid() && !field.IsZero()
+
+		if !isSet {
+			err = fmt.Errorf("%w%s in not set; ", err, fieldName)
```
Stefan-Ethernal:
```suggestion
			err = fmt.Errorf("%w%s is not set; ", err, fieldName)
```

Btw you didn't want to jump out from function as soon as first error occurs?
It seems to me that it is not necessary to collect all the errors, but data are considered invalid as soon as first field is not set. WDYT?

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,288 @@
...
+type postBlockStructF struct {
+	EpochID     uint64
+	ValidatorID uint64
+	BlockID     uint64
+	StakeValue  uint64
+}
+
+type updateValidatorSetF struct {
+	EpochID     uint64
+	X           uint64
```
Stefan-Ethernal: What `X` stands for? If I'm correct, maybe it is better to name it `index` (or something like that). WDYT?

consensus/polybft/stake_manager_fuzz_test.go
```diff
@@ -0,0 +1,288 @@
...
+		if err := ValidateStruct(data); err != nil {
+			t.Skip(err)
+		}
+
+		if data.ValidatorID > uint64(len(initialSetAliases)-1) {
+			t.Skip()
+		}
+
+		systemStateMock := new(systemStateMock)
+		systemStateMock.On("GetStakeOnValidatorSet", mock.Anything).Return(big.NewInt(int64(data.StakeValue)), nil).Once()
```
Stefan-Ethernal: Not needed anymore, since #1461 got merged in a meantime

```suggestion
```

consensus/polybft/stake_manager_fuzz_test.go
Nemanja0x: I would leave all the errors.

consensus/polybft/stake_manager_fuzz_test.go
jelacamarko: Yes, x is an index for the array. I will change it.

consensus/polybft/stake_manager_fuzz_test.go
jelacamarko: It's not necessary. I left it that way to get the info on what fields we have problems with because the initial idea was to validate the entire struct, but I found issues when validating a struct that has an array field.
Should I then leave all errors, or return when the first one happens?

Stefan-Ethernal: My opinion is that first error is enough, Nemanja doesn't agree, so your call, do it however you like better (there are not too many fields, so looping through each one of it would not harm) :slightly_smiling_face: 

Just please fix this typo which I stated in the suggestion (`in` instead of `is` in the error msg) :pray:

Stefan-Ethernal (approved):

PR #1503, opened 2023-05-14
Add common.RetryForever() and use for concurrent sync operations

sergerad:
# Description

Add common.RetryForever() and use for concurrent block/event synchronisation operations that can take a long time and fail at any point.

# Background

The Edge application approaches some errors by logging and running the rest of the program without retrying the failed operation. This is undesirable when those operations are critical (such as relayer startup and sync, which can mean that deposits no longer work). Critical operations should be retried because:
1. It allows the application to self heal, rather than run without critical capabilities; and
2. It allows for more reliable alerting and monitoring based on repeating critical errors.

This change allows critical operations to be retried every N seconds. Some such errors are transient and will fix themselves on retry. Other errors will have to be manually fixed by Edge operators.

Note that in a previous [PR](https://github.com/0xPolygon/polygon-edge/pull/1373), we implemented an approach which propagated concurrent errors back to the server shutdown logic. However, this was much more complicated and introduced a paradigm to error handling that is not consistent with how Edge is implemented today.

# Changes include

# Breaking changes

None

# Checklist

## Testing

### DIFF
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -38,6 +38,7 @@ func RetryForever(ctx context.Context, interval time.Duration, fn func(context.C
 		if err := fn(ctx); err != nil {
 			return retry.RetryableError(err)
 		}
+
 		return nil
 	})
 }
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"context"
 	"encoding/binary"
 	"encoding/json"
 	"errors"
@@ -18,6 +19,7 @@ import (
 	"time"
 
 	"github.com/0xPolygon/polygon-edge/helper/hex"
+	"github.com/sethvargo/go-retry"
 )
 
 var (
@@ -29,6 +31,16 @@ var (
 
 	errInvalidDuration = errors.New("invalid duration")
 )
+
+// RetryForever will execute a function until it completes without error
+func RetryForever(ctx context.Context, interval time.Duration, fn func(context.Context) error) {
+	_ = retry.Do(ctx, retry.NewConstant(interval), func(context.Context) error {
+		if err := fn(ctx); err != nil {
+			return retry.RetryableError(err)
+		}
+		return nil
+	})
+}
 
 // Min returns the strictly lower number
 func Min(a, b uint64) uint64 {
--- a/network/gossip.go
+++ b/network/gossip.go
@@ -6,7 +6,9 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
+	"time"
 
+	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/hashicorp/go-hclog"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -102,16 +104,16 @@ func (t *Topic) readLoop(sub *pubsub.Subscription, handler func(obj interface{},
 			continue
 		}
 
-		go func() {
+		go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
 			obj := t.createObj()
 			if err := proto.Unmarshal(msg.Data, obj); err != nil {
 				t.logger.Error("failed to unmarshal topic", "err", err)
-
-				return
+				return err
 			}
 
 			handler(obj, msg.GetFrom())
-		}()
+			return nil
+		})
 	}
 }
 
--- a/network/gossip.go
+++ b/network/gossip.go
@@ -6,9 +6,7 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
-	"time"
 
-	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/hashicorp/go-hclog"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -104,16 +102,16 @@ func (t *Topic) readLoop(sub *pubsub.Subscription, handler func(obj interface{},
 			continue
 		}
 
-		go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
+		go func() {
 			obj := t.createObj()
 			if err := proto.Unmarshal(msg.Data, obj); err != nil {
 				t.logger.Error("failed to unmarshal topic", "err", err)
-				return err
+
+				return
 			}
 
 			handler(obj, msg.GetFrom())
-			return nil
-		})
+		}()
 	}
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -1003,6 +1003,7 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
 		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
 			s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
+			return err
 		}
 		return nil
 	})
--- a/server/server.go
+++ b/server/server.go
@@ -1001,7 +1001,10 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 
 	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
 	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
-		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+		if err := srv.ListenAndServe(); err != nil {
+			if errors.Is(err, http.ErrServerClosed) {
+				return nil
+			}
 			s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
 			return err
 		}
--- a/server/server.go
+++ b/server/server.go
@@ -998,6 +998,7 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 	}
 
 	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+
 	go func() {
 		if err := srv.ListenAndServe(); err != nil {
 			if !errors.Is(err, http.ErrServerClosed) {
--- a/server/server.go
+++ b/server/server.go
@@ -918,13 +918,11 @@ func (s *Server) setupGRPC() error {
 	}
 
 	// Start server with infinite retries
-	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
+	go func() {
 		if err := s.grpcServer.Serve(lis); err != nil {
 			s.logger.Error(err.Error())
-			return err
 		}
-		return nil
-	})
+	}()
 
 	s.logger.Info("GRPC server running", "addr", s.config.GRPCAddr.String())
 
@@ -1000,16 +998,13 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 	}
 
 	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
-	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
+	go func() {
 		if err := srv.ListenAndServe(); err != nil {
-			if errors.Is(err, http.ErrServerClosed) {
-				return nil
+			if !errors.Is(err, http.ErrServerClosed) {
+				s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
 			}
-			s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
-			return err
 		}
-		return nil
-	})
+	}()
 
 	return srv
 }
--- a/server/server.go
+++ b/server/server.go
@@ -917,11 +917,14 @@ func (s *Server) setupGRPC() error {
 		return err
 	}
 
-	go func() {
+	// Start server with infinite retries
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
 		if err := s.grpcServer.Serve(lis); err != nil {
 			s.logger.Error(err.Error())
+			return err
 		}
-	}()
+		return nil
+	})
 
 	s.logger.Info("GRPC server running", "addr", s.config.GRPCAddr.String())
 
@@ -996,13 +999,13 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 		ReadHeaderTimeout: 60 * time.Second,
 	}
 
-	go func() {
-		s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
-
+	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
 		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
 			s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
 		}
-	}()
+		return nil
+	})
 
 	return srv
 }
--- a/tracker/event_tracker.go
+++ b/tracker/event_tracker.go
@@ -2,6 +2,7 @@ package tracker
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/0xPolygon/polygon-edge/helper/common"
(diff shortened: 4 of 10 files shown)


sergerad:
Note that the failing `TestE2E_Bridge_DepositAndWithdrawERC1155` passes on my machine. I assume its flakey due to the heuristic it uses for waiting for blocks after the deposit txs

server/server.go
```diff
@@ -996,13 +999,13 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 		ReadHeaderTimeout: 60 * time.Second,
 	}
 
-	go func() {
-		s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
-
+	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
 		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
 			s.logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
```
goran-ethernal: Return error here?

sergerad: Fixed, thanks

helper/common/common.go
```diff
@@ -30,6 +32,16 @@ var (
 	errInvalidDuration = errors.New("invalid duration")
 )
 
+// RetryForever will execute a function until it completes without error
+func RetryForever(ctx context.Context, interval time.Duration, fn func(context.Context) error) {
```
igorcrevar, line 36: Do we need to have context as a parameter here?

network/gossip.go
```diff
@@ -102,16 +104,16 @@ func (t *Topic) readLoop(sub *pubsub.Subscription, handler func(obj interface{},
 			continue
 		}
 
-		go func() {
+		go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
```
igorcrevar: I don't think we need retry pattern here because some node can send some malicious  message that can not be unmarshalled

server/server.go
```diff
@@ -917,11 +917,14 @@ func (s *Server) setupGRPC() error {
 		return err
 	}
 
-	go func() {
+	// Start server with infinite retries
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
```
igorcrevar: What if parameters for staring server are wrong? like port is already used or similar

tracker/event_tracker.go
```diff
@@ -64,6 +66,26 @@ func (e *EventTracker) Start(ctx context.Context) error {
...
 	blockTracker := blocktracker.NewBlockTracker(provider.Eth(), blocktracker.WithBlockMaxBacklog(blockMaxBacklog))
 
+	go func() {
+		<-ctx.Done()
+		blockTracker.Close()
+		store.Close()
+	}()
+
+	// Init and start block tracker concurrently, with infinite retries
+	go common.RetryForever(ctx, time.Second, func(context.Context) error {
```
igorcrevar, line 76: What if block tracker configuration is not populated how it should be? then we will have endless retries

sergerad: > then we will have endless retries

This is desired behaviour because point 2 in description:
```
2. It allows for more reliable alerting and monitoring based on repeating critical errors.
```

server/server.go
sergerad: if a port is in use, the app will self heal on retry when the port is made free. this is desired behaviour, instead of continuing a process that is broken - in this case an admin grpc endpoint being down

network/gossip.go
sergerad: Thanks, removed

helper/common/common.go
sergerad: Yes a valid use case is to pass in a `fn` which will exit with err == nil on `ctx.Done()` of the ctx passed in here

server/server.go
sergerad: I edited the error handling logic - it didn't look right to me. E.G. it was logging error when nil

server/server.go
```diff
@@ -996,13 +999,17 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
...
 
-	go func() {
-		s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
-
-		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
+		if err := srv.ListenAndServe(); err != nil {
+			if errors.Is(err, http.ErrServerClosed) {
+				return nil
```
sergerad: @goran-ethernal I updated this - can you please double check this was the intention before my changes

helper/common/common.go
igorcrevar: Ok, just make it sure that we are aligned with these chatgpt suggestions

- `Error handling`: In your code, you're converting the error returned by the function fn into a retry.RetryableError if it fails. This implies that any error encountered during the execution of fn should be considered retryable. Make sure this behavior aligns with your requirements and that you're not inadvertently retrying non-retryable errors.
- `Context usage`: The provided ctx parameter is used for cancellation. Ensure that you propagate the cancellation appropriately within the fn function to avoid unnecessary retries when the context is canceled.

server/server.go
goran-ethernal: Looks good.

helper/common/common.go
sergerad: > Error handling: In your code, you're converting the error returned by the function fn into a retry.RetryableError if it fails. This implies that any error encountered during the execution of fn should be considered retryable. Make sure this behavior aligns with your requirements and that you're not inadvertently retrying non-retryable errors.

Yep, `RetryForever()` is a simple func. We are free to replace it with more sophisticated retry logic anywhere it is being used. The contents of `fn` would decide whether a retryable error is returned or not. I implemented `RetryForever()` to avoid using the new retry module in multiple pkgs if we didn't need the more sophisticated logic.

> Context usage: The provided ctx parameter is used for cancellation. Ensure that you propagate the cancellation appropriately within the fn function to avoid unnecessary retries when the context is canceled.

This is actually incorrect. The `fn` does not need to execute `cancel` to avoid unnecessary retries. Instead, it needs to `select { case := <-ctx.Done(): return nil }`. Either way, this is still up to the implementation of `fn`, which is up to the caller of `RetryForever()`.

server/server.go
vcastellm: I don't find this a good practice, infinitely waiting for the port to be free is not a desired behaviour. The port can be busy forever.

helper/common/common.go
igorcrevar: Thanks for explanation. I did not know that wrapping error with `retry.RetryableError(err)` allows `go-retry` library to differentiate between errors that should be retried  and errors that should not be retried. you can resolve this issue

server/server.go
```diff
@@ -996,13 +999,17 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 		ReadHeaderTimeout: 60 * time.Second,
 	}
 
-	go func() {
-		s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
-
-		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+	go common.RetryForever(context.Background(), time.Second, func(context.Context) error {
```
vcastellm: This shouldn't be retried forever, just return the error and stop execution notifying the operator

tracker/event_tracker.go
vcastellm: Critical errors on startup should not be retried unless they can be temporary in nature and doesn't need user interaction to be fixed.

vcastellm (changes requested):
In some cases retrying an error could make sense but in the majority of cases, the approach should be to error as soon as possible and stop the execution from progressing.

sergerad:
> the approach should be to error as soon as possible and stop the execution from progressing.

@vcastellm this notion is contrary to the existing Edge code base which does not handle critical concurrent errors. Edge is a monolithic application and exiting on failure of one of its components is potentially contrary to design.

I started a PR that implemented logic to stop the program based on concurrent errors here: https://github.com/0xPolygon/polygon-edge/pull/1373

I closed it because I felt like I was rewriting Edge's approach to those errors. I am happy to pick that PR up again if you guys prefer that approach.

Alternatively, we could update this PR to add more sophisticated logic so that not everything is retried.

sergerad:
> In some cases retrying an error could make sense but in the majority of cases, the approach should be to error as soon as possible and stop the execution from progressing.

I initially misinterpreted this to mean stop the program - but I understand that you mean stop the goroutine that is erroring.

This approach makes it very hard to operate and debug Edge instances. We could have an instance that failed to sync rootchain events many hours ago and we would have to search for that singular ERROR log in an unknown range of time. If the sync was retried, we would see the ERROR log regularly. If we alert on every ERROR log, then we are inundated with alerts - I believe because dispatcher can ERROR log if RPC requests from users fail.

If we keep the status quo, and don't retry (or exit the program) for failures like gRPC server startup, event tracker sync, and prometheus server, what is your advice for cutting through the logs and being able to alert on ERRORs? Am I incorrect about the dispatcher logging ERROR on user requests? Thanks

Example of user-initiated RPC requests that results in ERROR logs (using eth_sendRawTransaction):
```
message:"2023-05-15T20:58:25.180Z [ERROR] polygon.server.dispatcher: failed to dispatch: method=eth_sendRawTransaction err=\"cannot parse RLP: cannot parse short bytes: cannot parse array value: cannot parse short bytes: cannot parse array value: cannot parse short bytes: length is not enough\""
```

tracker/event_tracker.go
sergerad: Yes but these are concurrent, unhandled errors. The initialization we can make synchronous and error out trivially. However the sync could take a long time and needs to be concurrent. In that case, failure needs to be propagated to the server logic so that we don't have a deployed Edge relayer failing to handle events, blocking deposits.

This PR began to implement concurrent error propagation https://github.com/0xPolygon/polygon-edge/pull/1373 but in the end I thought it was diverging from the current design of Edge.

server/server.go
sergerad: This would be fine if we could alert on every ERROR log. However, as per my other comment, I think that the dispatcher logs ERROR for user request failures, so it is not practical to alert on ERROR

sergerad:
Looking at tt.Sync in ethgo https://github.com/umbracle/ethgo/blob/main/tracker/tracker.go#L578.
The RetryForever here only helps for batchsync, because the longrunning for loop after it is handled concurrently by the Sync func itself. That means the following change is useless apart from retrying the batchsync before the for loop
```go
	go common.RetryForever(ctx, time.Second, func(context.Context) error {
			e.logger.Error("failed to sync", "error", err)
			return err
		}
		return nil
	})
```
IDK why the sync func in ethgo completely ignores those errors in the for loop

server/server.go
sergerad: Can refactor this to return error from parent func

Stefan-Ethernal: I agree with Victor and Igor here. Retry makes sense only for some transient errors, but port could be busy forever.

Stefan-Ethernal (commented):
I feel that we should keep retrying logic only in `event_tracker.go` and in `polybft.go` (synchronization protocol).

As a side note, please set `golangci-lint` as a linting tool and run `make lint` command. Seems like there are a couple of linter errors which should be fixed.

server/server.go
sergerad: Happy to rm this code of course.

Agreed that a port could be busy forever. It would become freed by an operator. At that point, Edge would "self heal" with respect to this functionality, which is a handy property for a distributed app.

For context, we run on Kubernetes so a port clash will never happen.

sergerad:
@Stefan-Ethernal Yep I can see the golangci-lint warnings in my IDE about line cuddling etc. I was ignoring them as I wasn't sure whether you guys really wanted to enforce those spacing requirements given the github workflow does not enforce them.

I am happy to add a workflow step that enforces the linter if you would like. Or perhaps there is a reason you do not want this. Thanks

server/server.go
```diff
@@ -996,11 +997,13 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
 		ReadHeaderTimeout: 60 * time.Second,
 	}
 
-	go func() {
-		s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
+	s.logger.Info("Prometheus server started", "addr=", listenAddr.String())
 
-		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
```
sergerad, line 1002: This would enter into error log statement when err is nil. e.g. https://goplay.tools/snippet/2NTrqFJzhYB

Stefan-Ethernal:
> @Stefan-Ethernal Yep I can see the golangci-lint warnings in my IDE about line cuddilng etc. I was ignoring them as I wasn't sure whether you guys really wanted to enforce those spacing requirements given the github workflow does not enforce them.
> 
> I am happy to add a workflow step that enforces the linter if you would like. Or perhaps there is a reason you do not want this. Thanks

@sergerad In fact we do have GH workflow for linters as well, although for some reason it hasn't been triggering for external collaborators. I have created a #1512  which got merged into the `develop`, so if you can rebase to the latest commit on the `develop` branch and let's see if is it any better now (I expect it to trigger this time). 🙂 🤞  

EDIT: It didn't help, so now I'm suspecting that it may be related to repository settings.

sergerad:
Have removed the retry logic apart from polybft sync / event tracker. If we merge this PR, I will need to do a follow up PR to ethgo to address the fact that this error is not handled: https://github.com/umbracle/ethgo/blob/main/tracker/tracker.go#LL582C7-L582C22

@vcastellm are you happy doing the retries on the sync logic? Currently in this PR we are also concurrently retrying initialization logic for the trackers because the tests rely on the fact that those errors are not handled yet. If we want initialization logic out of the concurrent retries, I can update the tests in this PR.

If you are not happy doing retries on sync logic at all, I can close this PR

server/server.go
paulgoleary: @sergerad raises a good point here, which is: is there a clearly defined 'design' for log levels and what the semantics are?
For example, in my thinking anything logged as an ERROR should be a critical system failure - i.e. a highly irregular situation that should be noticed and diagnosed. It would be inappropriate - as @sergerad suggest - to log a simple API error that's caused by malformed use input; since the system is designed to correctly handle this case, a user error should not be considered a critical failure. In fact, I'm not sure I would even WARN on this.

server/server.go
paulgoleary: FWIW, in the past I've occasionally applied a 'warn + sleep N times and then fail' for some of these type of instances where failure could be transient but also catastrophic if not resolved. A port being free might be an example where it might be expected to be transiently occupied - i.e. while something else shuts down - so it may be useful to retry a few times. But then if after some time - say 10s of trying - it doesn't resolved then fail.
But I recognize that it's not a good idea to blankety apply this idea so it may need to be done on a case-by-case basis.

sergerad: Yep and if we do decide to employ retries, we have the flexibility to customize the retry logic based on errors returned from the retryable functionality. RetryForever() is a simple/dumb approach that can be used as stop-gap measure

Stefan-Ethernal (approved):
LGTM.

As a side note: please rebase again to the most recent commit on the develop branch, since we have finally fixed lint workflow and then just resolve pending linting issues.

sergerad:
> As a side note: please rebase again to the most recent commit on the develop branch, since we have finally fixed lint workflow and then just resolve pending linting issues.

I did rebase when you mentioned the linting. I can see this commit in my logs
```
commit c3490646a75df27a36831e1a56873a5c46d01d61
Author: Stefan Negovanović <93934272+Stefan-Ethernal@users.noreply.github.com>
Date:   Tue May 16 20:43:07 2023 +0200

    Fix Lint GH workflow (#1512)
    
    * Remove ignored branches and increase timeout
    
    * Broader triggers
    
    * Add verbose output
```

Will rebase again now in any case

Stefan-Ethernal:
> I did rebase when you mentioned the linting. I can see this commit in my logs

Yep, but that one didn't work, so we have opened and merged another PR today (#1526) and now it's working properly :slightly_smiling_face:

goran-ethernal (approved):
LGTM. Thanks for taking into consideration comments from the guys.

server/server.go
vcastellm: Totally, but that's another topic, we don't have a clear definition of criticalities doesn't mean we can hold the execution forever retrying an error, as it's not an expected behaviour.

vcastellm (approved):
> @vcastellm are you happy doing the retries on the sync logic?

LGTM now, thanks for the contribution and sorry for the late reply

PR #1509, opened 2023-05-15
Reject future blocks 

Nemanja0x:
# Description

The PR add validations for header in `validationHeaderFields`. Changes includes check for future block, which includes checking if the timestamp is within predefined time slot of `time.Now().UTC().Unix()`. If not, we assume that it is a future block and reject it.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


consensus/polybft/polybft_config.go
```diff
@@ -45,6 +45,8 @@ type PolyBFTConfig struct {
 
 	// RewardConfig defines rewards configuration
 	RewardConfig *RewardsConfig `json:"rewardConfig"`
+
+	BlockTimeDrift int64
```
stana-miric: I suppose that negative values are not allowed? If that is the case should be uint64, also it would be nice to add a comment explaining the purpose of this configuration. This parameter is not supposed to be exposed in genesis file?

goran-ethernal:
```suggestion
	BlockTimeDrift int64 `json:blockTimeDrift`
```

consensus/polybft/polybft_config.go
```diff
@@ -59,6 +61,9 @@ func LoadPolyBFTConfig(chainConfigFile string) (PolyBFTConfig, int64, error) {
 		return PolyBFTConfig{}, 0, err
 	}
 
+	// hardcoded for now
```
goran-ethernal, line 64:
```suggestion
	// hardcoded until we implement on-governance
```

goran-ethernal (changes requested):
I think we should add a flag to the `genesis` command as well. Even though we will have this parameter as a part of on-chain governance soon, for the first start of the chain, we need to specify it in the `genesis`.

stana-miric (approved):

Stefan-Ethernal (commented):
Agree with @goran-ethernal and @stana-miric:
- `BlockTimeDrift` should be `uint64`, and
- a flag should be exposed in the genesis command in order to configure it initially.

Also, 2 out of 3 tests in the `TestE2E_Consensus_Basic_WithNonValidators` test are failing (the ones which are restarting nodes and letting them sync). I haven't realized why exactly, but have seen this error message in basically all of the validator nodes:

```go
2023-05-17T09:29:38.694Z [DEBUG] polygon.server.polybft.consensus_runtime: HasQuorum has been called but proposer could not be retrieved: error="get latest proposer not found - height: 14, round: 3, pc height: 14, pc round: 0"
```

goran-ethernal (approved):

command/genesis/genesis.go
```diff
@@ -239,6 +239,13 @@ func setFlags(cmd *cobra.Command) {
 			"",
 			"configuration of reward wallet in format <address:amount>",
 		)
+
+		cmd.Flags().Uint64Var(
+			&params.blockTimeDrift,
+			blockTimeDriftFlag,
+			defaultBlockTimeDrift,
+			"configuration for block time drift value",
```
Stefan-Ethernal: It is in seconds right? Anyways, think that it is wise to specify it flag description.
```suggestion
			"configuration for block time drift value (in seconds)",
```

consensus/polybft/polybft_config.go
```diff
@@ -48,6 +48,8 @@ type PolyBFTConfig struct {
 
 	// RewardConfig defines rewards configuration
 	RewardConfig *RewardsConfig `json:"rewardConfig"`
+
+	BlockTimeDrift uint64 `json:"blockTimeDrift"`
```
Stefan-Ethernal, line 53: Please add a comment here, which explains what `BlockTimeDrift` is.

consensus/polybft/fsm.go
```diff
@@ -644,6 +645,10 @@ func validateHeaderFields(parent *types.Header, header *types.Header) error {
 	if header.Number != parent.Number+1 {
 		return fmt.Errorf("invalid number")
 	}
+	// verify time is from the future
+	if header.Timestamp > (uint64(time.Now().UTC().Unix()) + blockTimeDrift) {
+		return fmt.Errorf("block from the future. block time: %d, time drift %d", header.Timestamp, blockTimeDrift)
```
```suggestion
		return fmt.Errorf("block from the future. block timestamp: %d, configured block time drift %d seconds", header.Timestamp, blockTimeDrift)
```

Also thinking out loud: should we convert `header.Timestamp` into a proper time format and report it that way in an error message, instead of logging (raw) elapsed seconds? The mindset is that the log message would contain time stamp in date/time format, so it would be maybe easier figuring out what is going on.

Stefan-Ethernal (approved):

consensus/polybft/fsm.go
jelacamarko: I agree. I will change that also.

Stefan-Ethernal (approved):

PR #1510, opened 2023-05-15
Adding benchmark test - root/child chain measurements

stana-miric:
# Description

This PR includes adding benchmark test for testing tx execution on root and child chain. Also, benchmark-test command is added for the purpose of testing on a test environment were we have both root and child chain running. The changes includes:

1. Adding benchmark directory that includes tests and helpers for tests.
- helpers.go that has helper functions needed in tests like deploying contracts and similar,
- root_child_benchmark_test.go tests execution of tx by calling the node on both root and child chain
- transition_benchmark_test.go tests calling contracts on transition instance. This test uses SampleContract which is taken from go-ethereum test since we want to compare execution on both edge and geth

2. Adding benchmark-test command(which tests the same thing as root_child_benchmark_test): the command has rootJSONRPC, childJSONRPC, startCluster and privateKey flags. If startCluster is false, then the local cluster will not be started and the provided addresses will be used as the endpoints to the root and child chains. If startCluster is set to true, the local cluster will be started automatically. If the private key is specified, it will be used as the transaction sender. Otherwise, the local cluster will generate a sender key. If startCluster is set to false, then the sender must have enough funds for sending transactions.. Example how to run the command: 
polygon-edge benchmark-test --childJSONRPC="http://127.0.0.1:12001" --rootJSONRPC="http://127.0.0.1:8545" --privateKey="aa75e9a7d427efc732f8e4f1a5b7646adcc61fd5bae40f80d13c8419c9f43d6d"
or
polygon-edge benchmark-test --startCluster
3. There are changes in e2e testing framework which replaces testing.T with testing.TB interface in order to use some functionality in the benchmark tests as well
4. tx relayer exposes WaitForReceipt and SumbitTransaction

# Changes include

# Checklist

## Testing


benchmark/common/contract_codes.go
```diff
@@ -0,0 +1,81 @@
+//nolint:lll
+package common
```
goran-ethernal: A suggestion, maybe it will be easier, without the need to have byte code hardcoded. 
We already have some test contracts we use in `e2e` tests, UTs, and other forms of testing. They are in `consensus/polybft/contractsapi/test-contracts` folder. Basically what we do when we need a test contract:

1. Add a test sol contract file to `core-contracts` repo, and compile the contracts.
2. Hardhat generates a `json` file (which is basically the artifact of the given contract) in `core-contracts/artifacts` folder.
3. Take that `json` file and copy it to the `consensus/polybft/contractsapi/test-contracts` folder.
4. In `consensus/polybft/contractsapi/init.go` you can add you test contracts by looking an example of `TestRewardToken`.
5. Wherever you use these `ContByteCode` fields, just reference your `TestContract (*artifact.Artifact).ByteCode`.

I think its much easier to modify these test contracts in the future if we need to, plus the artifact `json` file can be quite useful in debugging, if something goes wrong in the benchmark tests.

benchmark/common/helper.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/0xPolygon/polygon-edge/txrelayer"
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/stretchr/testify/require"
+	"github.com/umbracle/ethgo"
+	"github.com/umbracle/ethgo/abi"
+	"github.com/umbracle/ethgo/wallet"
+)
+
+// DeployContractOnRootAndChild deploys contract code on both root and child chain
+func DeployContractOnRootAndChild(
```
goran-ethernal: Is there a plan to use these functions in any new files? If not I would say move this entire file content to `root_child_send_tx.go` or just move it to `benchmark` folder and do not expose these functions since they are pretty much only for benchmark tests.

benchmark/common/helper.go
```diff
@@ -0,0 +1,99 @@
...
+	require.NotEqual(b, ethgo.ZeroAddress, receipt.ContractAddress)
+
+	return receipt.ContractAddress
+}
+
+// GetTxInput returns input for sending tx, given the abi encoded method and call parameters
+func GetTxInput(b *testing.B, method *abi.Method, args interface{}) []byte {
+	b.Helper()
+
+	var input []byte
```
goran-ethernal:
```suggestion
	var (
	     input []byte
	     err error
	)
```

command/benchmark/benchmark.go
```diff
@@ -0,0 +1,121 @@
...
+		"JSONRPC address of the child node",
+	)
+	cmd.Flags().StringVar(
+		&params.privateKey,
+		privateKeyFlag,
+		"",
+		"private key that will be used to send tx (it is expected that this address has enough funds)",
+	)
+
+	_ = cmd.MarkFlagRequired(rootJSONRPCFlag)
```
goran-ethernal: In the `README` file you are mentioning that these flags are not required, and command has a default behavior for each case. Should these `MarkFlagRequired` be removed?

command/benchmark/benchmark.go
```diff
@@ -0,0 +1,121 @@
...
+	_ = cmd.MarkFlagRequired(childJSONRPCFlag)
+	_ = cmd.MarkFlagRequired(privateKeyFlag)
+}
+
+func (p *benchmarkParams) validateFlags() error {
```
goran-ethernal: Remove this since it doesn't do anything and is not used.

command/benchmark/benchmark.go
stana-miric: I've marked these flags as mandatory because the command will be mostly used in a test system environment and we don't wan to run the cluster if some flag is omitted by mistake. On the local environment the test will likely be run with the go test framework where we can choose if we want to start the cluster or not. That was motivation behind those specific settings but it is not a problem to change it to support both.

benchmark/common/helper.go
stana-miric: fix 726fbea
The file is moved to the root folder. It is not deleted, because we might add new benchmark tests. Also, other files from common are moved and the folder is deleted. The explanation for the specific test written in this PR is moved to the root_child_send_tx.go and README file is removed as well

benchmark/common/helper.go
stana-miric: fix 726fbea

command/benchmark/benchmark.go
stana-miric: fix 726fbea

command/benchmark/benchmark.go
stana-miric: fix 81ed307
I've introduced flag startCluster, which is mutual exclusive with root and child json rpc addresses.

benchmark/root_child_send_tx.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/umbracle/ethgo/wallet"
+)
+
+var (
+	singleContCalcFunc   = abi.MustNewMethod("function compute(uint256 x, uint256 y) public returns (uint256)")
+	singleContGetFunc    = abi.MustNewMethod("function getValue() public returns (uint256[] memory)")
+	singleContSetFunc    = abi.MustNewMethod("function addValue(uint256 value) public")
+	multiContSetAddrFunc = abi.MustNewMethod("function setContractAddr(address _contract) public")
+	multiContFnA         = abi.MustNewMethod("function fnA() public returns (uint256)")
+	RootJSONRPC          = flag.String("rootJSONRPC", "", "JSONRPC address of the root node")
```
goran-ethernal: Do we need these flags here? I am not a fan of setting a global variable value from a different file and package. Could we just pass some struct or arguments to the `RootChildSendTxSetUp` function.

benchmark/common/contract_codes.go
stana-miric: fix 99efd02

benchmark/root_child_send_tx.go
stana-miric: fix 38bc3ce
we needed those flags before the command was introduced. i've removed them since we can run both scenarios from the command now

goran-ethernal (approved):

vcastellm (changes requested):
Let's talk about this

PR #1521, opened 2023-05-17
Differentiate `chain-id` and `supernet-id`

goran-ethernal:
# Description

Through this PR we are differentiating `chain-id` from `supernet-id` in the `genesis` file.

Before, `chain-id` was equal to `supernet-id` provided by `StakeManager` when supernet was registered to it in the `rootchain deploy` command.

Now `genesis` file will have both `chain-id` and `supernet-id` values, where `supernet-id` will be in the `polybftconfig` struct in `genesis`.
`chain-id` flag in `genesis` command stays, so that the users can use a custom `chain-id` for their supernet,

**NOTE**: `supernet-id` is not a flag in `genesis` command, but a value in the `genesis` file which is populated by the `rootchain deploy` command when supernet is registered to `StakeManager` contract.

`supernet-id` is used for staking, since validators need to provide the `supernet id` when calling the `stakeFor` function on `StakeManager` so that contract knows on which supernet validator is trying to stake.

Commands changed in this PR:
- `stake` command - flag `chain-id` is renamed to `supernet-id`.

Tagging @DannyS03 to update the docs.

# Changes include

# Checklist

## Testing


igorcrevar (approved):

Nemanja0x (approved):

stana-miric (approved):

vcastellm (commented):
Late LGTM

Issue #1447, opened 2023-04-27
eth_getStorageAt returns 0


fedeaviles:
# eth_getStorageAt returns 0
## Description
`eth_getStorageAt` method always returns `0x0000000000000000000000000000000000000000000000000000000000000000` in versions `0.6.2`, `0.6.3`, `0.7.3-beta3`, `0.8.0` and `0.8.1`. But it works fine in version `0.6.1`

## Your environment
- OS and version. macOs Ventura 13.3.1
- The version of the Polygon Edge.  0.6.2, 0.6.3, 0.7.3-beta3, 0.8.0, 0.8.1
- The branch that causes this issue. v0.6.2, v0.6.3, v0.7.3-beta3, v0.8.0, v0.8.1
- Locally or Cloud hosted (which provider). Local and also Cloud (Linux with polygon-edge v0.6.3 on AWS)
- Please confirm if the validators are running under containerized environment (K8s, Docker, etc.). Docker

## Steps to reproduce
- Tell us how to reproduce this issue.
 1. deploy a smart contract
```
 // SPDX-License-Identifier: MIT
pragma solidity ^0.8.9;

contract Storage {
    uint256 public value;
    
    constructor() {
        value = 42;
    }
    
    function setValue(uint256 _newValue) public {
        value = _newValue;
    }
}
```
 2. Use the `eth_getStorageAt` method (with the correct contract address)
  ```
  curl --request POST \
     --url http://localhost:10002/ \
     --header 'accept: application/json' \
     --header 'content-type: application/json' \
     --data '
{
  "method": "eth_getStorageAt",
  "params": [
    "0xAE975071Be8F8eE67addBC1A82488F1C24858067",
    "0x0",
    "latest"
  ],
  "id": 1,
  "jsonrpc": "2.0"
}
'
  ```

## Expected behavior
- Tell us what should happen.
It should return:
```
 {
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x000000000000000000000000000000000000000000000000000000000000002A"
} 
```
- Tell us what happened instead.
It returns:
```
 {
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x0000000000000000000000000000000000000000000000000000000000000000"
} 
```

Thanks!

PR #1523, opened 2023-05-18
fix eth_getStorageAt always returning 0

typestring:
# Description

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests
I've tested this PR following this issue [#1447](https://github.com/0xPolygon/polygon-edge/issues/1447)

# Documentation update

# Additional comments

The RLP Parsing is already being done by the Snapshot's GetStorage method

https://github.com/0xPolygon/polygon-edge/blob/ed61412225766b716ce49babd30014a14db87b89/state/immutable-trie/snapshot.go#L41-L53


typestring:

typestring:
recheck

typestring:
recheck

Stefan-Ethernal:
@typestring Tnx for your PR.

A couple of notes:
- seems like there are unused imports in `eth_endpoint.go` and because of that, build, and consequently all the other CI checks, are failing at the moment (https://github.com/0xPolygon/polygon-edge/actions/runs/5016718450/jobs/8997304506). So remove all the unused imports.
- there is an e2e test (https://github.com/0xPolygon/polygon-edge/blob/develop/e2e-polybft/e2e/jsonrpc_test.go#L148-L163), which tests `eth_getStorageAt` endpoint, however it does not try to alter `TestSimple.sol` contract's state and then read it, but just testing a scenario where nothing is written into storage and retrieving default value (`0x0`). It would be awesome if you could add a scenario with sending a transaction, which would alter state of `TestSimple.sol` (you would probably need to change `TestSimple` SC and provide a setter) and then assert that given value is exactly the one you have set.
- rename mentioned test from `eth_getStorage` to `eth_getStorageAt`.

Does it make sense?

PR #1530, opened 2023-05-21
E2e remote tests

sergerad:
NOTE: draft state, incomplete PR

# Description

We would like to re-use the existing e2e tests against remote clusters to validate our deployments.

In the past, we have implemented our own e2e tests for this purpose, but it involved duplicating some of the code in these tests. Ideally, we would not have to maintain a separate e2e test suite to this one.

This PR shows how we could update the e2e tests to work against both local test clusters and remote clusters by using `testCluster` and `testBridge` interfaces.

**Are these changes something the polygon-edge team would consider suitable?**

If not, are there any alternative implementations or suggestions regarding maintaining our own e2e test suite (with minimal code duplication). Thanks!


sergerad:
Managed to get the `TestE2E_Bridge_Transfers` tests to run against our remote cluster but the tests rely on the fact that test cluster L2 events are instantaneous (whereas for our cluster we would need to wait):

```
    bridge_test.go:144: 
                Error Trace:    /Users/sergeradinovich/Source/sergerad/polygon-edge/e2e-polybft/e2e/helpers_test.go:231
                                                        /Users/sergeradinovich/Source/sergerad/polygon-edge/e2e-polybft/e2e/bridge_test.go:144
                Error:          Not equal: 
                                expected: 0
                                actual  : 5
```

Not impossible to factor around but might start getting complicated with forking logic between local vs remote clusters.

sergerad:
Alternatively could make completely separate tests for remote e2e and do a small refactor so that the `polygon-edge` cli command code can be re-used, similar to how the PR has it now

vcastellm:
This is an interesting use case @sergerad, and I have mixed feelings about it, on one side I feel that e2e are not meant to test external clusters, only the application in GH CI, that should be enough to tests that everything is correct on each merge. So testing external clusters should be done using another tool.

On the other side we've implemented a good amount of tests here that can be also useful to test external deployments, but IMO they are unnecessarily extensive, for testing external cluster only a subset should be necessary.

I'm thinking that you can implement the necessary scenarios here https://github.com/0xPolygon/polygon-edge/tree/develop/loadtest/scenarios they will be useful for running Smoke testing for all users, including us, and maybe promote them to load tests when you need it.

Can you describe a bit more the necessary features that you want to test? maybe we can work together in this.

sergerad:
>  I feel that e2e are not meant to test external clusters, only the application in GH CI, that should be enough to tests that everything is correct on each merge. So testing external clusters should be done using another tool.

Yep the tests I am wanting wouldn't run on merge. We would run them after deploying a fresh network or an Edge upgrade on an existing network, for example.

> On the other side we've implemented a good amount of tests here that can be also useful to test external deployments, but IMO they are unnecessarily extensive, for testing external cluster only a subset should be necessary.

Yep agreed.

> I'm thinking that you can implement the necessary scenarios here https://github.com/0xPolygon/polygon-edge/tree/develop/loadtest/scenarios they will be useful for running Smoke testing for all users, including us, and maybe promote them to load tests when you need it.

Yep this could work. If you guys / community would benefit from some tests here, happy to put them there. Alternatively we can impl the tests in our own repo - I don't want to add code to polygon-edge that has no value for others.

> Can you describe a bit more the necessary features that you want to test? maybe we can work together in this.

Most important features to test are deposits and withdraw/exit.

Thanks!

sergerad:
This PR proved the existing e2e tests could be used for testing remote environments but the diffs were complicated and muddied the existing tests. Agreed that tests for remote env should be separate and a subset of what is in the e2e dirs today.

We have a private repo with some Go e2e tests that we could contribute back to this repo at some point. We also have a typescript SDK with helpful functionality for testing which we could use to contribute e2e remote tests also.

vcastellm:
> Yep this could work. If you guys / community would benefit from some tests here, happy to put them there.

They will be welcome +1 to add more scenarios

PR #1533, opened 2023-05-22
EVM-666: Fix `eth_getLogs` index issue

goran-ethernal:
# Description

This PR fixes issue where `eth_getLog` rpc call did not return correct values for `LogIndex` for block logs.

Before, the `LogIndex` field was set by transaction instead of block. For example, if block had two transactions with two logs each, the call would return log indexes: 0, 1, 0, 1, instead of: 0, 1, 2, 3.

This PR added a UT for this.

# Changes include

# Checklist

## Testing

### DIFF
--- a/jsonrpc/filter_manager.go
+++ b/jsonrpc/filter_manager.go
@@ -375,10 +375,13 @@ func (f *FilterManager) getLogsFromBlock(query *LogQuery, block *types.Block) ([
 		return nil, err
 	}
 
-	logs := make([]*Log, 0)
+	var (
+		logIdx int
+		logs   = make([]*Log, 0)
+	)
 
 	for idx, receipt := range receipts {
-		for logIdx, log := range receipt.Logs {
+		for _, log := range receipt.Logs {
 			if query.Match(log) {
 				logs = append(logs, &Log{
 					Address:     log.Address,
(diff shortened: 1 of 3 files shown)


rzmahmood:
Just confirming, this will fix for logs returned by both `eth_getLogs` and `eth_getBlockByNumber`?

jsonrpc/filter_manager.go
```diff
@@ -375,10 +375,13 @@ func (f *FilterManager) getLogsFromBlock(query *LogQuery, block *types.Block) ([
 		return nil, err
 	}
 
-	logs := make([]*Log, 0)
+	var (
+		logIdx int
+		logs   = make([]*Log, 0)
+	)
```
Stefan-Ethernal: Nitpick (use `uint64` and remove the `var` block, because we can initialize variables).
```suggestion
	logIdx := uint64(0)
	logs   := make([]*Log, 0)
```

Stefan-Ethernal (approved):
LGTM

jsonrpc/filter_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1533/commits/d6f055a498f815258c68974f1aeae9969bf0b1c7

Stefan-Ethernal:
> Just confirming, this will fix for logs returned by both `eth_getLogs` and `eth_getBlockByNumber`?

AFAIK, the `eth_getBlockByNumber` endpoint does not return logs. So this PR affects the `eth_getLogs` and `eth_getFilterLogs` endpoints.

igorcrevar (approved):

vcastellm (approved):
LGTM

Issue #1528, opened 2023-05-19
Root Chain Fund Command fails for any network except local geth Instance for v0.9.0


adirola:
Root Chain Fund Command fails for any network except local geth Instance for v0.9.0
## Description
The command` ./polygon-edge rootchain fund --data-dir ./test-chain-1 --amount 10000000000000000 --native-root-token <root_token_address> --private-key <hex_encoded_deployer_private_key> --json-rpc https://matic-mumbai.chainstacklabs.com --mint`
with the following result : no accounts registered

## Your environment
- Linux (ubuntu 20.04).
- Edge / Supernets Version v0.9.0.    
- release :  v0.9.0.
- Locally build from source.
- No validators are not running on any container service.

## Steps to reproduce
- On running the root chain fund command on any network except local geth instance using a --private-key flag it gives error: no accounts registered
- The issue seems to be in the file fund.go in directory command/rootchain/fund. In this file irrespective of the json-rpc provided it will call sendTransactionLocal function which tries to execute the transaction on local geth Instance.
the above call can be found on line no. : 131  
error line : receipt, err := txRelayer.SendTransactionLocal(txn) 
- `./polygon-edge rootchain fund --data-dir ./test-chain-1 --amount 10000000000000000 --native-root-token <root_token_address> --private-key <hex_encoded_deployer_private_key> --json-rpc https://matic-mumbai.chainstacklabs.com --mint`

## Expected behavior
- The expected behavior is: I should be able to mint native-root-token to validator account using the --private key of the deployer address used to deploy root contracts on the chain
- The result received was no accounts registered.

## Logs
<img width="2238" alt="Screenshot 2023-05-19 at 5 22 17 PM" src="https://github.com/0xPolygon/polygon-edge/assets/9098273/93f074c1-0091-45a8-9266-99f9ac00251f">

## Proposed solution
The solution seems to be: If --private-key flag is passed then don't issue a sendTransactionLocal call 

<img width="790" alt="Screenshot 2023-05-19 at 9 44 33 PM" src="https://github.com/0xPolygon/polygon-edge/assets/9098273/bbaefd25-cf57-4f0b-9ffc-4cc44b4d6597">

If this issue is not being actively worked upon i can work on it as well and raise a PR.

goran-ethernal:
Addressed through #1540

adirola:
[...]

PR #1540, opened 2023-05-24
Modify rootchain command to use the provided private key to fund validators

dusan-maksimovic:
# Description

This PR fixes the [issue](https://github.com/0xPolygon/polygon-edge/issues/1528) with rootchain fund command when the rootchain is not locally deployed geth instance. If the "private-key" flag is set, then it will be used to sign the transactions to send tokens to validators. Otherwise, validators will be funded by the coinbase account by sending an unsigned transaction, which is only possible with geth instance started in development mode.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/rootchain/fund/fund.go
+++ b/command/rootchain/fund/fund.go
@@ -91,19 +91,16 @@ func runCommand(cmd *cobra.Command, _ []string) {
 		return
 	}
 
-	var (
-		deployerKey    ethgo.Key
-		stakeTokenAddr types.Address
-	)
+	deployerKey, err := helper.GetRootchainPrivateKey(params.deployerPrivateKey)
+	if err != nil {
+		outputter.SetError(fmt.Errorf("failed to initialize deployer private key: %w", err))
 
-	if params.mintStakeToken || params.deployerPrivateKey != "" {
-		deployerKey, err = helper.GetRootchainPrivateKey(params.deployerPrivateKey)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


command/rootchain/fund/fund.go
```diff
@@ -96,7 +96,7 @@ func runCommand(cmd *cobra.Command, _ []string) {
 		rootTokenAddr types.Address
 	)
 
-	if params.mintRootToken {
+	if params.mintRootToken || params.deployerPrivateKey != "" {
```
Stefan-Ethernal: To me, it would be more correct to split this condition into two:
1. set `deployerKey` just in case non-empty string is provided for the deployer private key
2. set `rootTokenAddr` only if `mintRootToken` is set to true.

Stefan-Ethernal (approved):
LGTM aside from comment.

command/rootchain/fund/fund.go
dusan-maksimovic: Fixed through [33bab10](https://github.com/0xPolygon/polygon-edge/pull/1540/commits/33bab10d3e144ea3e3006fae3550dcf797506ba4)

Since in e2e we are providing --mint flag, but do not provide --private-key flag, this would not work. I agreed with Stefan that we should initialize deployerKey always, and rootTokenAddr only if --mint flag is set.

goran-ethernal (approved):

PR #1544, opened 2023-05-24
EVM-607 Hard-forking PoC

igorcrevar:
# Description

Forking Manager PoC

Example:
Let's say we have three forks: A, B, and C, with starting blocks 0, 10, and 20, respectively. For each fork, we have a simple handler that returns a string. We can achieve this using the code example below

```
fm := GetInstance()

const handlerDesc = "fn"

type handlerFn func() string

// define handlers
var (
    h1 handlerFn = func() string { return "Handler for fork A" }
    h2 handlerFn = func() string { return "Handler for fork B" }
    h3 handlerFn = func() string { return "Handler for fork C" }
)

// register forks
fm.RegisterFork("A")
fm.RegisterFork("B")
fm.RegisterFork("C")

// register handlers
assert.NoError(t, fm.RegisterHandler("A", handlerDesc, h1))
assert.NoError(t, fm.RegisterHandler("B", handlerDesc, h2))
assert.NoError(t, fm.RegisterHandler("C", handlerDesc, h3))

// activate forks
assert.NoError(t, fm.ActivateFork("A", 0))
assert.NoError(t, fm.ActivateFork("B", 10))
assert.NoError(t, fm.ActivateFork("C", 20))

// check if appropriate handler is called for some block number
for i := 0; i < 50; i++ {
    handler := fm.GetHandler(handlerDesc, uint64(i)).(handlerFn)
    
    if i < 10 {
	 assert.Equal(t, "Handler for fork A", handler())
    } else if i < 20 {
	 assert.Equal(t, "Handler for fork B", handler())
    } else {
	 assert.Equal(t, "Handler for fork C", handler())
    }
}
```

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


forkmanager/fork_manager.go
```diff
@@ -0,0 +1,233 @@
...
+
+	fm.forkMap[name] = &Fork{
+		Name:            name,
+		FromBlockNumber: 0,
+		IsActive:        false,
+		Handlers:        map[ForkHandlerName]interface{}{},
+	}
+}
+
+func (fm *forkManager) RegisterHandler(forkName ForkName, handlerName ForkHandlerName, handler interface{}) error {
```
Nemanja0x: Why this is not a part of `RegisterFork()`? I assume that when I register a fork I want to register all of its handlers at the same time.

forkmanager/fork.go
```diff
@@ -0,0 +1,43 @@
...
+
+type ForkInfo struct {
+	Name            ForkName
+	FromBlockNumber uint64
+}
+
+type Fork struct {
+	Name            ForkName
+	FromBlockNumber uint64
+	IsActive        bool
```
Nemanja0x: Why do I need this? Isn't it active as soon as it reaches predefined block number?

forkmanager/fork_manager.go
```diff
@@ -0,0 +1,233 @@
...
+var (
+	forkManagerInstance     *forkManager
+	forkManagerInstanceLock sync.Mutex
+)
+
+type forkManager struct {
+	lock sync.Mutex
+
+	forkMap     map[ForkName]*Fork
+	handlersMap map[ForkHandlerName][]ForkActiveHandler
```
Nemanja0x: Shouldn't this be a part of a fork? Why do we need all the handlers inside of the `ForkManager`? I assumed that all of the handlers will be assigned to a specific fork, or is this some kind of optimization.?

consensus/polybft/polybft.go
```diff
@@ -598,3 +599,38 @@ func (p *Polybft) GetBridgeProvider() consensus.BridgeDataProvider {
...
+			activeForks = append(activeForks, info)
+		}
+	}
+
+	return forkmanager.GetInstance().RegisterAll(availableForks, handlers, activeForks)
+}
+
+func init() {
+	// init default fork
+	availableForks := []forkmanager.ForkName{""}
```
Nemanja0x: Why do we need an empty fork?

forkmanager/fork_manager.go
igorcrevar: its optimization. binary search is used to determine which handler should be used.
For example there are 3 forks:
(Fork=a, StartingBlock=0)
(Fork=b, StartingBlock=10)
(Fork=c, StartingBlock=20)

lets say handlers are registered for fork a, b but not for c

If you want to retrieve handler for block < 10, then you should get  handler for fork a
If you want to retrieve handler for block >= 10 (even block lets say 66), then you should get  handler for fork b

forkmanager/fork.go
igorcrevar: Because fork can be registered but not activated. I did not want to use *uint64 for `FromBlockNumber` so I use two fields for same effect (but its more memory access friendly because its not indirect pointer access).

Remember that fork manager can be used even for dynamic forks from smart contract

forkmanager/fork_manager.go
igorcrevar: Because not every fork will have handlers for everything.
 Lets say there are forks A, B, C
 But it is possible to register "extra data" handlers only for forks A and C
 and to register "proposal selection" handlers for forks B and C

consensus/polybft/polybft.go
igorcrevar: We can remove this for now... But lets say we introduce new handlers for some fork which starts from block N. In that case there is a need for handler from block 0

igorcrevar: removed https://github.com/0xPolygon/polygon-edge/pull/1544/commits/5491b2211cdd2736d7ca5cbdfd27c758cedecf20

consensus/polybft/polybft.go
```diff
@@ -598,3 +599,38 @@ func (p *Polybft) GetBridgeProvider() consensus.BridgeDataProvider {
...
+func registerForksAndHandlers(forks *chain.Forks) error {
+	//nolint:godox
+	// TODO: update to real values and read from smart contract
+	availableForks := []forkmanager.ForkName{}
+	activeForks := []*forkmanager.ForkInfo{}
+	handlers := []forkmanager.ForkHandler{}
+
+	for name, block := range *forks {
+		if !chain.IsForkAvailable(name) {
+			panic(fmt.Errorf("fork is not available: %s", name)) //nolint:gocritic
```
Stefan-Ethernal: Maybe it would be better to propagate the error and prevent node starting up if possible?

consensus/polybft/polybft.go
```diff
@@ -598,3 +599,38 @@ func (p *Polybft) GetBridgeProvider() consensus.BridgeDataProvider {
...
+
+func init() {
+	// init default fork
+	availableForks := []forkmanager.ForkName{""}
+	activeForks := []*forkmanager.ForkInfo{
+		forkmanager.NewForkInfo("", 0),
+	}
+
+	if err := forkmanager.GetInstance().RegisterAll(availableForks, nil, activeForks); err != nil {
+		panic(err) //nolint:gocritic
```
Stefan-Ethernal: Propagate the error?

tests/testing.go
```diff
@@ -405,83 +405,83 @@ func (t *stTransaction) UnmarshalJSON(input []byte) error {
...
+		chain.Homestead:      chain.NewFork(0),
+		chain.EIP150:         chain.NewFork(0),
+		chain.EIP155:         chain.NewFork(0),
+		chain.EIP158:         chain.NewFork(0),
+		chain.Byzantium:      chain.NewFork(0),
+		chain.Constantinople: chain.NewFork(0),
+		chain.Petersburg:     chain.NewFork(0),
+		chain.Istanbul:       chain.NewFork(0),
+		chain.London:         chain.NewFork(0),
 	},*/
```
Stefan-Ethernal: Not sure why this is commented out. Should we uncomment it?

igorcrevar: I have removed commented out code https://github.com/0xPolygon/polygon-edge/pull/1544/commits/fa180dae7360c57990e1f28b196c2b7a866befa9

consensus/polybft/polybft.go
igorcrevar: I have refactored this a little bit https://github.com/0xPolygon/polygon-edge/pull/1544/commits/fa180dae7360c57990e1f28b196c2b7a866befa9

consensus/polybft/polybft.go
igorcrevar: I have refactored this a little bit https://github.com/0xPolygon/polygon-edge/pull/1544/commits/fa180dae7360c57990e1f28b196c2b7a866befa9

Nemanja0x (approved):

forkmanager/fork.go
```diff
@@ -0,0 +1,17 @@
+package forkmanager
+
+type ForkHandlerName string
+
+type ForkName string
```
Stefan-Ethernal: Why not simply using strings instead of type aliases? Though I agree that code is probably a bit more verbose this way.

consensus/polybft/polybft.go
```diff
@@ -361,6 +362,26 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 	}
 }
 
+func ForkManagerFactory(forks *chain.Forks) error {
+	fm := forkmanager.GetInstance()
+
+	for name, block := range *forks {
```
Stefan-Ethernal: Rename `block` to `forkBlock` maybe?

forkmanager/fork_manager.go
```diff
@@ -0,0 +1,222 @@
...
+func (fm *forkManager) GetHandler(name ForkHandlerName, blockNumber uint64) interface{} {
+	fm.lock.Lock()
+	defer fm.lock.Unlock()
+
+	handlers, exists := fm.handlersMap[name]
+	if !exists {
+		return nil
+	}
+
+	// binary search to find first position where handler.FromBlockNumber < blockNumber
```
Stefan-Ethernal:
```suggestion
	// binary search to find first position where handler.FromBlockNumber >= blockNumber
```

forkmanager/fork_manager.go
```diff
@@ -0,0 +1,222 @@
...
+		// keep everything in sorted order
+		index := sort.Search(len(handlers), func(i int) bool {
+			return handlers[i].FromBlockNumber >= blockNumber
+		})
+		handlers = append(handlers, ForkActiveHandler{})
+		copy(handlers[index+1:], handlers[index:])
+		handlers[index] = ForkActiveHandler{
+			FromBlockNumber: blockNumber,
+			Handler:         handler,
+		}
```
Stefan-Ethernal: Should we return an error if somebody tries to register handler for the block number that is already present in the `handlers` slice?

forkmanager/fork_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	fork.IsActive = false
+
+	for forkHandlerName := range fork.Handlers {
+		fm.removeHandler(forkHandlerName, fork.FromBlockNumber)
+	}
+
+	return nil
+}
+
+func (fm *forkManager) GetHandler(name ForkHandlerName, blockNumber uint64) interface{} {
```
Stefan-Ethernal: Add a comment?

forkmanager/fork_manager.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// GetForkBlock returns fork block if fork is registered and activated
+func (fm *forkManager) GetForkBlock(name ForkName) (uint64, error) {
+	fm.lock.Lock()
+	defer fm.lock.Unlock()
+
+	fork, exists := fm.forkMap[name]
+	if !exists {
+		return 0, fmt.Errorf("fork does not exists: %s", name)
```
Stefan-Ethernal:
```suggestion
		return 0, fmt.Errorf("fork does not exist: %s", name)
```

Stefan-Ethernal (approved):

forkmanager/fork_manager.go
igorcrevar: this is internal method. If someone calls RegisterHandler with existing fork and handler name, old handler will be overwritten

forkmanager/fork.go
igorcrevar: I wanted possibility for easy switching to other key type (for example`int`). But I am not even 90% sure. What other guys think?

forkmanager/fork_manager.go
igorcrevar: Fixed

consensus/polybft/polybft.go
```diff
@@ -361,6 +362,34 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 	}
 }
 
+func ForkManagerFactory(forks *chain.Forks) error {
```
vcastellm, line 364: Why having this in the consensus? look like this belongs to the root

forkmanager/fork.go
vcastellm: For me it's ok to use string instead of alias because with `Name` suffix I doubt we're gonna change the type. But I'm also ok with the alias.

vcastellm (commented):
Can you add a small code example on how a handler should be used?

vcastellm (commented):
Can you add a small code example on how a handler should be used in the PR body?

igorcrevar:
> 

Example is here https://github.com/0xPolygon/polygon-edge/pull/1567

consensus/polybft/polybft.go
igorcrevar: handlers can depend on consensus. Current code i have put there probably can be moved in root, but handler registration can't (look at the example pr)

forkmanager/fork.go
igorcrevar: I have removed fork name alias because 99,99% fork name will always be string. But i have kept `HandlerDesc` alias because we may change it from string to int for performance gain

consensus/polybft/polybft.go
igorcrevar: I have refactored this a little bit. Server do all the forks registration and activation but factory method register all the handlers. If there are no factory method then consensus does not uses fork manager at all

https://github.com/0xPolygon/polygon-edge/pull/1544/commits/37e021cfa5311fa38e4a120d16ad8fe9793e75be

chain/chain.go
```diff
@@ -41,8 +41,6 @@ type Chain struct {
 
 // Genesis specifies the header fields, state of a genesis block
 type Genesis struct {
-	Config *Params `json:"config"`
```
Stefan-Ethernal, line 44: :partying_face:

goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1564, opened 2023-05-29
EVM-675 Implement additional metrics

goran-ethernal:
# Description

Added additional metrics to `Edge` codebase:

### Consensus metrics:
- `consensus block_execution_time` - measures the time of block execution.
- `consensus block_building_time` - measures the block building time.
- `consensus block_space_used` - measures the gas used by block.

### Network metrics:
- `network egress_bytes` - measures the number of bytes sent over the P2P network in the gossip protocol.
- `network ingress_bytes` - measures the number of bytes received over the P2P network in the gossip protocol.
- `network bad_messages` - counter of bad messages received over the P2P network in the gossip protocol.

### TX Pool Metrics:
- `txpool slots_used` - amount of slots currently occupying the pool measured through time.
- `txpool invalid_tx_type` - counter of transactions that had invalid tx type.
- `txpool oversized_data_txs` - counter of transactions that had size more then allowed transaction max size.
- `txpool negative_value_tx` - counter of transactions that had negative `Value` field.
- `txpool invalid_signature_txs` - counter of transactions with invalid signature.
- `txpool invalid_sender_txs` - counter of transactions with invalid sender.
- `txpool contract_deploy_too_large_txs` - counter of contract deployment transactions where contract size is too large.
- `txpool underpriced_tx` - counter of underpriced transactions.
- `txpool fee_cap_too_high_dynamic_tx` - counter of dynamic (EIP-1559) transactions whose fee cap is too high.
- `txpool tip_too_high_dynamic_tx` - counter of dynamic (EIP-1559) transactions whose gas tip cap is too high.
- `txpool tip_above_fee_cap_dynamic_tx` - counter of dynamic (EIP-1559) transactions whose gas tip cap is above gas fee cap.
- `txpool nonce_too_low_tx` - counter of transactions whose nonce was too low.
- `txpool invalid_account_state_tx` - counter of transactions whose account state was invalid.
- `txpool insufficient_funds_tx` - counter of transactions whose sender did not have sufficient funds to execute them.
- `txpool invalid_intrinsic_gas_tx` - counter of transactions whose transaction gas cost could not be calculated.
- `txpool intrinsic_gas_low_tx` - counter of transactions whose gas is lower then calculated transaction gas cost.
- `txpool block_gas_limit_exceeded_tx` - counter of transactions whose gas cost exceeds the block gas limit.
- `txpool rejected_future_tx` - counter of rejected transaction s that have a future nonce.
- `txpool already_known_tx` - counter of transactions who were not added to the pool because they are already known transactions.
- `txpool added_tx` - measures the added transactions through time.

### JSON RPC metrics:
Each `JSON RPC` endpoint function has a metric that measures its execution time.
Each `JSON RPC` endpoint function has a metric that measures the number of errors that occurred in their execution (counter of all errors).

For example:
`json_rpc eth_getBlockByNumber_time` - measures the execution time of `eth_getBlockByNumber` endpoint.
`json_rpc eth_getBlockByNumber_errors` - measures the number of errors that happened in calls to given endpoint.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -110,7 +110,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 
 	_, root := transition.Commit()
 
-	metrics.SetGauge([]string{consensusMetricsPrefix, "block_execution_time"}, float32(time.Now().UTC().Sub(start).Seconds()))
+	metrics.SetGauge([]string{consensusMetricsPrefix, "block_execution_time"},
+		float32(time.Now().UTC().Sub(start).Seconds()))
 
 	if root != block.Header.StateRoot {
 		return nil, fmt.Errorf("incorrect state root: (%s, %s)", root, block.Header.StateRoot)
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"time"
 
+	metrics "github.com/armon/go-metrics"
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/0xPolygon/polygon-edge/blockchain"
@@ -87,6 +88,7 @@ func (p *blockchainWrapper) CommitBlock(block *types.FullBlock) error {
 func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Block,
 	callback func(*state.Transition) error) (*types.FullBlock, error) {
 	header := block.Header.Copy()
+	start := time.Now().UTC()
 
 	transition, err := p.executor.BeginTxn(parent.StateRoot, header, types.BytesToAddress(header.Miner))
 	if err != nil {
@@ -108,6 +110,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 
 	_, root := transition.Commit()
 
+	metrics.SetGauge([]string{consensusMetricsPrefix, "block_execution_time"}, float32(time.Now().UTC().Sub(start).Seconds()))
+
 	if root != block.Header.StateRoot {
 		return nil, fmt.Errorf("incorrect state root: (%s, %s)", root, block.Header.StateRoot)
 	}
@@ -118,6 +122,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Txns:     block.Transactions,
 		Receipts: transition.Receipts(),
 	})
+
+	metrics.SetGauge([]string{consensusMetricsPrefix, "block_space_used"}, float32(builtBlock.Header.GasUsed))
 
 	return &types.FullBlock{
 		Block:    builtBlock,
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -122,8 +122,6 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Receipts: transition.Receipts(),
 	})
 
-	updateBlockSpaceUsedMetric(builtBlock.Header.GasUsed)
-
 	return &types.FullBlock{
 		Block:    builtBlock,
 		Receipts: transition.Receipts(),
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -6,7 +6,6 @@ import (
 	"math/big"
 	"time"
 
-	metrics "github.com/armon/go-metrics"
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/0xPolygon/polygon-edge/blockchain"
@@ -110,8 +109,7 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 
 	_, root := transition.Commit()
 
-	metrics.SetGauge([]string{consensusMetricsPrefix, "block_execution_time"},
-		float32(time.Now().UTC().Sub(start).Seconds()))
+	updateBlockExecutionMetric(start)
 
 	if root != block.Header.StateRoot {
 		return nil, fmt.Errorf("incorrect state root: (%s, %s)", root, block.Header.StateRoot)
@@ -124,7 +122,7 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Receipts: transition.Receipts(),
 	})
 
-	metrics.SetGauge([]string{consensusMetricsPrefix, "block_space_used"}, float32(builtBlock.Header.GasUsed))
+	updateBlockSpaceUsedMetric(builtBlock.Header.GasUsed)
 
 	return &types.FullBlock{
 		Block:    builtBlock,
--- a/jsonrpc/types.go
+++ b/jsonrpc/types.go
@@ -9,7 +9,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
-const jsonRpcMetric = "json_rpc"
+const jsonRPCMetric = "json_rpc"
 
 // For union type of transaction and types.Hash
 type transactionOrHash interface {
--- a/jsonrpc/types.go
+++ b/jsonrpc/types.go
@@ -9,6 +9,8 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+const jsonRpcMetric = "json_rpc"
+
 // For union type of transaction and types.Hash
 type transactionOrHash interface {
 	getHash() types.Hash
--- a/network/gossip.go
+++ b/network/gossip.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/armon/go-metrics"
 	"github.com/hashicorp/go-hclog"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -61,6 +62,8 @@ func (t *Topic) Publish(obj proto.Message) error {
 		return err
 	}
 
+	metrics.SetGauge([]string{networkMetrics, "egress_bytes"}, float32(len(data)))
+
 	return t.topic.Publish(context.Background(), data)
 }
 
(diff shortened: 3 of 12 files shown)


consensus/polybft/blockchain_wrapper.go
```diff
@@ -108,6 +110,9 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 
 	_, root := transition.Commit()
 
+	metrics.SetGauge([]string{consensusMetricsPrefix, "block_execution_time"},
+		float32(time.Now().UTC().Sub(start).Seconds()))
```
Stefan-Ethernal: Move it to the `consensus_metrics.go` (`updateBlockMetrics` function).

consensus/polybft/blockchain_wrapper.go
```diff
@@ -119,6 +124,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Receipts: transition.Receipts(),
 	})
 
+	metrics.SetGauge([]string{consensusMetricsPrefix, "block_space_used"}, float32(builtBlock.Header.GasUsed))
+
```
Stefan-Ethernal: Move it to the `consensus_metrics.go` (`updateBlockMetrics` function).

jsonrpc/types.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+const jsonRPCMetric = "json_rpc"
```
Stefan-Ethernal, line 12: Just thinking out loud: maybe we should keep all metrics prefixes in a single place?

txpool/txpool.go
```diff
@@ -565,39 +575,55 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
...
 
 		// Check EIP-1559-related fields and make sure they are correct
 		if tx.GasFeeCap == nil || tx.GasTipCap == nil {
+			metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
+
 			return ErrUnderpriced
 		}
 
 		if tx.GasFeeCap.BitLen() > 256 {
+			metrics.IncrCounter([]string{txPoolMetrics, "fee_cap_to_high_dynamic_tx"}, 1)
```
Stefan-Ethernal:
```suggestion
			metrics.IncrCounter([]string{txPoolMetrics, "fee_cap_too_high_dynamic_tx"}, 1)
```

txpool/txpool.go
```diff
@@ -565,39 +575,55 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
...
+
 			return ErrFeeCapVeryHigh
 		}
 
 		if tx.GasTipCap.BitLen() > 256 {
+			metrics.IncrCounter([]string{txPoolMetrics, "tip_to_high_dynamic_tx"}, 1)
```
Stefan-Ethernal:
```suggestion
			metrics.IncrCounter([]string{txPoolMetrics, "tip_too_high_dynamic_tx"}, 1)
```

txpool/txpool.go
```diff
@@ -548,13 +554,17 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 	// Extract the sender
 	from, signerErr := p.signer.Sender(tx)
 	if signerErr != nil {
+		metrics.IncrCounter([]string{txPoolMetrics, "invalid_signature_txs"}, 1)
```
Stefan-Ethernal, line 557: IMO, we should unify all these invalid transactions under single metrics ("invalid_tx"). I would like to hear other opinions as well.

Stefan-Ethernal (commented):
We haven't implemented these metrics, related to Level DB:
| Resource | Type | Metric Name |
| ------------- | ------------- |--|
DB | Response | Trie Read/Write/Commit Times
DB | Utilization | leveldb IOPs
DB | Response | leveldb IO Latency
DB | Response | Compaction time

Also not sure if this can be deduced from one of the provided metrics:
| Resource | Type | Metric Name |
| ------------- | ------------- |--|
P2P | Utilization | Messages (TX, Headers, Bodies) per second

txpool/txpool.go
goran-ethernal: I think it is more useful to keep it separated. That way you have a more precise measure what the users are sending, or possibly identify a malicious actor.

jsonrpc/types.go
goran-ethernal: They are package related, so I don't think it will bring much benefit to have them in single place, but whatever we agree upon.

goran-ethernal:
> We haven't implemented these metrics, related to Level DB:
> 
> Resource	Type	Metric Name
> DB	Response	Trie Read/Write/Commit Times
> DB	Utilization	leveldb IOPs
> DB	Response	leveldb IO Latency
> DB	Response	Compaction time
> Also not sure if this can be deduced from one of the provided metrics:
> 
> Resource	Type	Metric Name
> P2P	Utilization	Messages (TX, Headers, Bodies) per second

As John said, those are just proposals. Implementing those `leveldb` metrics is not easy, and I am not sure how will they impact performance if we are to measure it by second. As per `P2P` metrics, yes, they can see it on DataDog, since the graph is representing measured values in time. You can zoom in basically to a second on a graph and see how many messages you received in that time.

txpool/txpool.go
Stefan-Ethernal: That's correct, but it has a couple of downsides as well:
- you don't have a cumulative number which is suggesting how many invalid transactions in total are in some time period. Having multiple metrics is great if you are really interested in details, but I'm afraid the essence is lost and you can't see the forest for the trees.
- @praetoriansentry needs to create a DD dashboard that holds ~20 metrics, which I'm not sure he is keen doing :smile: 

So to me, let's discuss it and find the most appropriate solution. 

IMO mitigation would be to keep these metrics (if needed) and to provide cumulative metrics (e.g. `invalid_txs`) as well, which would report all the invalid transactions and then we have the best out of both worlds.

goran-ethernal: Ok, agreed. My understanding is we want it by type of error, but I am maybe wrong.

consensus/polybft/blockchain_wrapper.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1564/commits/de7a671831ce441316a1fd172270314a90ad0c70

consensus/polybft/blockchain_wrapper.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1564/commits/de7a671831ce441316a1fd172270314a90ad0c70

txpool/txpool.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1564/commits/de7a671831ce441316a1fd172270314a90ad0c70

txpool/txpool.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1564/commits/de7a671831ce441316a1fd172270314a90ad0c70

txpool/txpool.go
vcastellm: I don't find necessary to provide cumulative metrics, that's easily done in the metrics UI, DD, Grafana, etc.

consensus/polybft/blockchain_wrapper.go
```diff
@@ -108,6 +109,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 
 	_, root := transition.Commit()
 
+	updateBlockExecutionMetric(start)
```
vcastellm, line 112: Shouldn't this be after line 119?

network/gossip.go
```diff
@@ -61,6 +62,8 @@ func (t *Topic) Publish(obj proto.Message) error {
 		return err
 	}
 
+	metrics.SetGauge([]string{networkMetrics, "egress_bytes"}, float32(len(data)))
```
vcastellm, line 65: I'd say to use a counter here, as a gauge does not accumulate the values over time, something that is interesting for this kind of metric.

consensus/polybft/blockchain_wrapper.go
goran-ethernal: I think we consider block executed after each transaction was applied and transition committed. That is why I put the update of the metric there.

syncer/client.go
```diff
@@ -380,18 +381,22 @@ func blockStreamToChannel(stream proto.SyncPeer_GetBlocksClient) (<-chan *types.
...
 
 			block, err := fromProto(protoBlock)
 			if err != nil {
+				metrics.IncrCounter([]string{syncerMetrics, "bad_block"}, 1)
 				errorCh <- err
 
 				break
 			}
 
+			metrics.SetGauge([]string{syncerMetrics, "ingress_bytes"}, float32(len(protoBlock.Block)))
```
vcastellm, line 398: Similar here, counter?

syncer/syncer.go
```diff
@@ -207,6 +208,10 @@ func (s *syncer) Sync(callback func(*types.FullBlock) bool) error {
 	return nil
 }
 
+func (s *syncer) syncerMetricsRun() {
+
+}
```
vcastellm: Why empty?

network/gossip.go
goran-ethernal: I added `SetGauge` here because it might be useful to see how much data was sent in which point in time. To me that tells more then a simple counter of how many total bytes we sent.

syncer/client.go
goran-ethernal: As per comment: https://github.com/0xPolygon/polygon-edge/pull/1564#discussion_r1210275782

syncer/syncer.go
goran-ethernal: Leftover. Will remove.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1564/commits/d5f43872a74ec167b66fadfaa15d6fc13944c013

Stefan-Ethernal (approved):
Probably we should document somewhere on the wiki all the available metrics.

network/gossip.go
vcastellm: In this case it's correct because the definiton of Gauge in metrics lib and Prometheus are equivalent, but they lack the Datadog COUNT type, that is what I was talking about, not monotonic counter.

vcastellm (approved):
LGTM

PR #1569, opened 2023-05-30
EVM-660 Use batch write inside blockchain.go (LevelDB/memory)

jelacamarko:
# Description

- In the previous implementation, `WriteFullBlock` and other methods of `Blockchain` structure wrote each part of the block separately. However, with the changes introduced in this PR, everything is written together as a single batch. This can improve performance and efficiency when writing data
- The `KeyValueStorage` is focused on read operations only + it provides a method [`func (s *KeyValueStorage) NewBatch() Batch `] to create a concrete implementation of a Batch (memory or LevelDB)
- The actual writing is handled by the `BatchWriter,` which holds the concrete `Batch` interface implementation. It contains methods such as `PutHeadHash` or `PutHeadNumber` etc that are putting desired changes into a batch queue. However, these methods do not immediately change the storage. The changes are only saved to the actual database or memory when the `WriteBatch` method is called.

LevelDB options that are changed: 
```
OpenFilesCacheCapacity = DefaultHandles
BlockCacheCapacity = DefaultCache / 2 * opt.MiB
WriteBuffer = DefaultCache / 4 * opt.MiB
```
LevelDB options are set to the same values that go-ethereum uses.  
```
DefaultHandles = 256
DefaultCache = 256
```

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


blockchain/storage/leveldb/batch.go
```diff
@@ -0,0 +1,32 @@
+package leveldb
+
+import (
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+type batchLevelDB struct {
+	db   *leveldb.DB
+	b    *leveldb.Batch
+	Size int
```
Stefan-Ethernal: Do we even need this field? It is just being aggregated in both `Delete` and `Put` functions, but its value is not used anywhere.

```suggestion
```

blockchain/storage/leveldb/leveldb.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/0xPolygon/polygon-edge/blockchain/storage"
 	"github.com/hashicorp/go-hclog"
 	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+)
+
+const (
+	DefaultCache   = int(256)
+	DefaultHandles = int(256)
```
Stefan-Ethernal, line 14: Casts are not necessary. Also does it make sense to leave just a single constant (e.g. `DefaultCache`)? Why do we need two of them with the same values?
```suggestion
	DefaultCache   = 256
	DefaultHandles = 256
```

blockchain/storage/leveldb/leveldb.go
```diff
@@ -25,7 +31,13 @@ func Factory(config map[string]interface{}, logger hclog.Logger) (storage.Storag
...
 // NewLevelDBStorage creates the new storage reference with leveldb
 func NewLevelDBStorage(path string, logger hclog.Logger) (storage.Storage, error) {
-	db, err := leveldb.OpenFile(path, nil)
+	options := opt.Options{}
+	// Set default options
+	options.OpenFilesCacheCapacity = DefaultHandles
+	options.BlockCacheCapacity = DefaultCache / 2 * opt.MiB
+	options.WriteBuffer = DefaultCache / 4 * opt.MiB // Two of these are used internally
+
+	db, err := leveldb.OpenFile(path, &options)
```
Stefan-Ethernal: Nitpick:

```suggestion
	options := &opt.Options{}
	// Set default options
	options.OpenFilesCacheCapacity = DefaultHandles
	options.BlockCacheCapacity = DefaultCache / 2 * opt.MiB
	options.WriteBuffer = DefaultCache / 4 * opt.MiB // Two of these are used internally

	db, err := leveldb.OpenFile(path, options)
```

blockchain/storage/memory/batch.go
```diff
@@ -0,0 +1,35 @@
+package memory
+
+import (
+	"github.com/0xPolygon/polygon-edge/helper/hex"
+)
+
+type batchMemory struct {
+	db           map[string][]byte
+	keysToDelete [][]byte
+	valuesToPut  [][2][]byte
```
Stefan-Ethernal, line 13: Is 2 here denoting key+value byte slices? So the first slice represents `key` and the second one represents `value`?

blockchain/storage/batch.go
```diff
@@ -0,0 +1,98 @@
...
+// 	return b.Delete(key)
+// }
+
+// func (b *BatchHelper) Write() error {
+// 	return b.batch.Write()
+// }
+
+// func (b *BatchHelper) Put(k []byte, data []byte) error {
+// 	return b.batch.Put(k, data)
+// }
```
Stefan-Ethernal: Leftovers, remove it?

blockchain/storage/batch.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/umbracle/fastrlp"
+)
+
+type Batch interface {
+	Delete(key []byte)
+	Write() error
+	Put(k []byte, data []byte)
+}
+
+type BatchHelper struct {
```
Stefan-Ethernal: Rename the file to `batch_helper.go`?

blockchain/blockchain.go
```diff
@@ -634,10 +632,18 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
...
-		if err := b.writeHeaderImpl(event, h); err != nil {
+
+		batch := b.db.NewBatch()
+		batchHelper := storage.NewBatchHelper(batch)
+
+		if err := b.writeHeaderImpl(batchHelper, event, h); err != nil {
+			return err
+		}
+
+		if err := batch.Write(); err != nil {
```
Stefan-Ethernal: Seems like this can be organized a bit better. `BatchHelper` should be better encapsulated IMO (it should have all the knowledge how to deal with batch instance).

I'd consider letting `batchHelper` to manipulate with all the batch operations. So in this case, `Write` could be called against `batchHelper` (consider maybe invoking it even inside `writeHeaderImpl`).

blockchain/storage/leveldb/leveldb_test.go
```diff
@@ -37,3 +48,227 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
...
+		return err
+	})
+
+	return size, err
+}
+
+func TestWriteFullBlock(t *testing.T) {
+	t.Helper()
+
+	s, _, path := newStorageP(t)
```
Stefan-Ethernal, line 208: Why is this function marked as test helper?

blockchain/storage/leveldb/leveldb_test.go
```diff
@@ -37,3 +48,227 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
...
+
+			batchHelper.WriteHeader(b.Block.Header)
+			batchHelper.WriteHeadNumber(uint64(i))
+			batchHelper.WriteHeadHash(b.Block.Header.Hash)
+			batchHelper.WriteReceipts(b.Block.Hash(), b.Receipts)
+
+			if err := batch.Write(); err != nil {
+				fmt.Println(err)
+				t.FailNow()
+			}
```
Stefan-Ethernal: Why not rely on assertions here (e.g. `require.NoError`)?

blockchain/storage/leveldb/leveldb_test.go
```diff
@@ -37,3 +48,227 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
...
+
+			fmt.Println("writing block", i)
+
+			size, _ := DirSize(path)
+			fmt.Println("\tldb file count:", countLdbFilesInPath(path))
+			fmt.Println("\tdir size", size/1_000_000, "MBs")
```
Stefan-Ethernal: Probably it is better to rely on `t.Log` (or `t.Logf`) if we are about to keep these outputs.

Stefan-Ethernal:
It would be probably interesting to implement some benchmark tests and see how much gain we get by using batch updates.

blockchain/storage/leveldb/leveldb_test.go
```diff
@@ -37,3 +48,227 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
...
+func generateRandomByteSlice(count int) []byte {
+	s := make([]byte, count)
+	for i := 0; i < count; i++ {
+		s[i] = byte(rand.Int())
+	}
+
+	return s
+}
+
+func generateFullTx(nonce uint64, from types.Address, to *types.Address, value *big.Int, dynamic bool, v *big.Int) *types.Transaction {
```
Stefan-Ethernal: Double check, maybe some of these helper functions are already implemented, so they can be reused.

blockchain/storage/memory/batch.go
igorcrevar: yes, exactly. `batchMemory` is used mostly for the tests. maybe we should add comment like:
`// slice representing (key, value) pairs that will be added/updated on Write call

blockchain/storage/leveldb/batch.go
igorcrevar: I think it is safe to remove this field. We have copied it from the geth code base, they are using that field for statistic we currently do not have

blockchain/blockchain.go
goran-ethernal: I have to agree with Stefan. Why can't the batch helper instantiate a new batch, by passing the db instance to `NewBatchHelper`, and as Stefan said, Write can be called on batch helper, not on batch. This then makes it easier because you are only dealing with the helper instance, and you don't have to worry about the batch as well.

blockchain/blockchain.go
```diff
@@ -975,6 +985,13 @@ func (b *Blockchain) WriteBlock(block *types.Block, source string) error {
 		logArgs = append(logArgs, "generation_time_in_seconds", diff)
 	}
 
+	b.writeLock.Lock()
```
goran-ethernal: Why is this lock moved here?

blockchain/storage/leveldb/leveldb.go
```diff
@@ -25,7 +31,13 @@ func Factory(config map[string]interface{}, logger hclog.Logger) (storage.Storag
 
 // NewLevelDBStorage creates the new storage reference with leveldb
 func NewLevelDBStorage(path string, logger hclog.Logger) (storage.Storage, error) {
```
goran-ethernal, line 33: Do we need another constructor which takes on `opt.Options` instance if someone wants to use another configuration? I am saying this because we already had a similar issue reported on github where someone wanted to use a different db, and a different set of options, but we didn't support it.

```suggestion
func NewLevelDBStorage(path string, logger hclog.Logger) (storage.Storage, error) {
      defaultOptions := &opt.Options{/*set default options*/ }
      
      return NewLevelDBStorageWithOpt(path, logger, defaultOptions)
}

func NewLevelDBStorageWithOpt(path string, logger hclog.Logger, opts opt.Options) (storage.Storage, error) {
   // do the magic in the existing constructor
}
```

blockchain/storage/leveldb/batch.go
jelacamarko: I've also used it when checking the size of the batch. It can be removed.

blockchain/storage/leveldb/leveldb.go
jelacamarko: Both values are used for options, and they determine the way how lbd files are created. That's why I left two constants. It's the same value because that's the default value on go-ethereum. Probably this can be a part of some configuration in the future.

blockchain/storage/leveldb/leveldb_test.go
jelacamarko: It should not be. Copy paste error

blockchain/blockchain.go
jelacamarko: That should be at the beginning of the function. I'll move it.

blockchain/storage/leveldb/leveldb.go
igorcrevar: Why don't we put initialization of fields in brackets? like:
```
options := &opt.Options{
  OpenFilesCacheCapacity: ...
  ...
}
```

blockchain/blockchain.go
jelacamarko: @igorcrevar I made this change. Can you please take a look? 

Pull Request CI / Tests that are failing are passing locally. I'll run CI again after E2E tests are finished.

blockchain/storage/keyvalue.go
```diff
@@ -355,3 +355,7 @@ func (s *KeyValueStorage) Close() error {
 func (s *KeyValueStorage) NewBatch() Batch {
 	return s.db.NewBatch()
 }
+
+func (s *KeyValueStorage) WriteBatch() error {
+	return s.db.NewBatch().Write()
```
igorcrevar: this is not good i think. You are creating new batch each time you are doing write

jelacamarko: I think that method isn't really necessary. I removed WriteBatch from the Storage interface because it is a helper method. I just forgot to remove it from the KeyValueStorage.

blockchain/storage/testing.go
```diff
@@ -733,3 +733,11 @@ func (m *MockStorage) Close() error {
 func (m *MockStorage) HookClose(fn closeDelegate) {
 	m.closeFn = fn
 }
+
+func (m *MockStorage) NewBatch() Batch {
+	return nil
+}
+
+func (m *MockStorage) WriteBatch() error {
```
igorcrevar: This func not needed

blockchain/storage/batch_helper.go
```diff
@@ -0,0 +1,92 @@
...
+	batch Batch
+}
+
+func NewBatchHelper(storage Storage) *BatchHelper {
+	batch := storage.NewBatch()
+
+	return &BatchHelper{batch: batch}
+}
+
+func (b *BatchHelper) WriteHeader(h *types.Header) {
```
igorcrevar: Maybe we should rename all these WriteXYZ methods to PutXYZ because actual writing to db occurs on `WriteBatch`

blockchain/blockchain.go
```diff
@@ -921,6 +929,8 @@ func (b *Blockchain) WriteBlock(block *types.Block, source string) error {
 	b.writeLock.Lock()
 	defer b.writeLock.Unlock()
 
+	batchHelper := storage.NewBatchHelper(b.db)
```
goran-ethernal: Nitpick: Move this before the line where you start using it.

blockchain/blockchain.go
```diff
@@ -857,11 +862,12 @@ func (b *Blockchain) executeBlockTransactions(block *types.Block) (*BlockResult,
 // This function is a copy of WriteBlock but with a full block which does not
 // require to compute again the Receipts.
 func (b *Blockchain) WriteFullBlock(fblock *types.FullBlock, source string) error {
-	block := fblock.Block
-
 	b.writeLock.Lock()
 	defer b.writeLock.Unlock()
 
+	block := fblock.Block
+	batchHelper := storage.NewBatchHelper(b.db)
```
goran-ethernal: Nitpick: Move this before the line where you start using it.

blockchain/storage/leveldb/leveldb_test.go
```diff
@@ -37,3 +48,223 @@ func newStorage(t *testing.T) (storage.Storage, func()) {
...
+			ticker.Stop()
+
+			return
+		case <-ticker.C:
+			ch <- b
+		}
+	}
+}
+
+func DirSize(path string) (int64, error) {
```
goran-ethernal: If this is only used in tests, I would say, don't expose it, pass `t *testing.T` to it and change it a bit:

```suggestion
func DirSize(t *testing.T, path string) int64 {
     t.Helper()
     
     var size int64

	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if !info.IsDir() {
			size += info.Size()
		}

		return err
	})
	
	require.NoError(t, err)
	
	return size
}
```

blockchain/storage/leveldb/leveldb.go
jelacamarko: https://github.com/0xPolygon/polygon-edge/pull/1569/commits/5ae832c5e3697bd99a48d8345ce5825cccd9328a

goran-ethernal (approved):
Except from the small comments, looks good.

blockchain/blockchain.go
jelacamarko: https://github.com/0xPolygon/polygon-edge/pull/1569/commits/cf46fe06de0cc1c83b672f510a0d29d53f71765a

blockchain/storage/keyvalue.go
jelacamarko: https://github.com/0xPolygon/polygon-edge/pull/1569/commits/4e193947f8c4ae38e7ad04b81a52a57bbe4cc54d

blockchain/storage/batch_helper.go
jelacamarko: https://github.com/0xPolygon/polygon-edge/pull/1569/commits/3cee4952af72992fe3d2446915c244722ea74574

stana-miric (approved):

blockchain/blockchain.go
igorcrevar: `BatchHelper` is renamed to `BatchWriter` and all write operations are now going through him - `KeyValueStorage` is used for reading operations and to retrieve `Batch` implementation used by `BatchWriter`

blockchain/storage/leveldb/leveldb_test.go
igorcrevar: There isn't anything exposed in common. we can do that in some other pr if necessary

blockchain/blockchain.go
```diff
@@ -634,10 +576,18 @@ func (b *Blockchain) WriteHeadersWithBodies(headers []*types.Header) error {
 		}
 	}
 
-	// Write the actual headers
-	for _, h := range headers {
+	// Write the actual headers in separate batches for now
```
vcastellm, line 574: Why in separate batches?

vcastellm (approved):
LGTM this should bring great improvement.

blockchain/blockchain.go
igorcrevar: it would  require a lot of code changes if we write everything in one batch (because writing new header requires reading of parent headers data). Good thing is that this function is only used by tests

PR #1576, opened 2023-06-01
EVM-677: Fix log index issue on `eth_getTransactionReceipt`

goran-ethernal:
# Description

`eth_getTransactionReceipt` did not return correct indexes of logs. For example, if we had a block with 2 transactions with 2 logs each, and if we want a receipt from the second transaction, the function would return log indexes 0, 1, instead of 2, 3, since there is a transaction in the block that had logs before the desired transaction.

This PR fixes the issue and adds new UT.

This is continuation of work from [PR 1533](https://github.com/0xPolygon/polygon-edge/pull/1533).

# Changes include

# Checklist

## Testing

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -327,8 +327,7 @@ func (e *Eth) GetTransactionReceipt(hash types.Hash) (interface{}, error) {
 
 		// accumulate receipt logs indexes from block transactions
 		// that are before the desired transaction
-		receipt := receipts[i]
-		logIndex += len(receipt.Logs)
+		logIndex += len(receipts[i].Logs)
 	}
 
 	if txIndex == -1 {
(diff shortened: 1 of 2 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -315,35 +315,41 @@ func (e *Eth) GetTransactionReceipt(hash types.Hash) (interface{}, error) {
...
-			indx = i
+			txIndex = i
 
 			break
 		}
+
+		// accumulate receipt logs indexes from block transactions
+		// that are before the desired transaction
+		receipt := receipts[i]
+		logIndex += len(receipt.Logs)
```
Stefan-Ethernal: Nitpick:
```suggestion
		logIndex += len(receipts[i].Logs)
```

stana-miric (approved):

Stefan-Ethernal (approved):
LGTM

jsonrpc/eth_endpoint.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1576/commits/738950d59e4b93144f5504dc2d3884a6fa97a58e

PR #1578, opened 2023-06-01
Prevent ACL admins to remove themselves

vcastellm:
# Description

This PR implements a protection to ACLs so going forward an admin can not remove himself from the admins list, it should be removed by another admin.

# Changes include

# Checklist

## Testing

### DIFF
--- a/e2e-polybft/e2e/acls_test.go
+++ b/e2e-polybft/e2e/acls_test.go
@@ -92,7 +92,7 @@ func TestE2E_AllowList_ContractDeployment(t *testing.T) {
 
 	{
 		// Step 4. 'adminAddr' sends a transaction to enable 'targetAddr'.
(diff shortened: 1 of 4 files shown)


igorcrevar (approved):

QEDK:
> This PR implements a protection to ACLs so going forward an admin can not remove himself from the admins list, it should be removed by another admin.

The admin should not be able to remove themselves when there's only one admin, otherwise I think they should be able to remove themselves, no?

Nemanja0x (approved):

begmaroman (approved):

Stefan-Ethernal (approved):
As a side note, @QEDK's comment makes sense as well.

vcastellm:
> The admin should not be able to remove themselves when there's only one admin, otherwise I think they should be able to remove themselves, no?

We can not count how many admins there are in the lists due to limitation of the state. This was the first idea but we needed to adapt it to this.

PR #1582, opened 2023-06-02
Invoke estimate gas limit if gas limit not already provided

jelacamarko:
# Description

Transaction relayer (aka `TxRelayer`) relies on `eth_estimateGasLimit` to calculate the gas limit and set it to the transaction, instead of using the hardcoded value before sending the transaction.

Hardcoded gas limit values are left only in the `bridge exit` CLI command and inside E2E test framework as a fallback in case gas estimation fails (for example account is not allow listed).

# Changes include

# Checklist

## Testing


goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1584, opened 2023-06-05
Consolidate bridge transaction result structs

Stefan-Ethernal:
# Description

This PR cleans up bridge CLI commands by consolidating all bridge transactions into a single one.

It also updates the SC repo to the latest revision.

# Changes include

# Checklist

## Testing


igorcrevar (approved):

command/bridge/common/params.go
```diff
@@ -155,3 +159,64 @@ func ExtractExitEventID(receipt *ethgo.Receipt) (*big.Int, error) {
 
 	return nil, errors.New("failed to find exit event log")
 }
+
+type BridgeTxResult struct {
+	Sender       string     `json:"sender"`
+	Receivers    []string   `json:"receivers"`
+	ExitEventIDs []*big.Int `json:"exitEventIDs"`
```
Nemanja0x, line 166: We are going to use only the first element of slice for certain commands?

Nemanja0x (approved):

command/bridge/deposit/erc20/deposit_erc20.go
```diff
@@ -197,6 +195,8 @@ func runCommand(cmd *cobra.Command, _ []string) {
 					return fmt.Errorf("receiver: %s, amount: %s", receiver, amount)
 				}
 
+				blockNumbers = append(blockNumbers, receipt.BlockNumber)
```
dusan-maksimovic: Could it happen, since there is no lock, that two go routines try to append at the same time, and the results for one of them get overwritten by the other one? Also, since there are multiple go routines, could it happen that the same block gets added to the slice multiple times, if different transactions enter the same block?

Stefan-Ethernal: Good catch. :monocle_face: :slightly_smiling_face: 

Fixed https://github.com/0xPolygon/polygon-edge/pull/1584/commits/79db3787ad127c4558abdd76c0369c6185e5c41a

Stefan-Ethernal: > could it happen that the same block gets added to the slice multiple times, if different transactions enter the same block?

IMO if this happens, it doesn't matter, it is expected that we print the same block multiple times (namely that receivers exit event ids, block numbers, amounts/token ids slices are the same length).

dusan-maksimovic (approved):

PR #1588, opened 2023-06-06
EVM-437 Batch calls over websockets not working properly

igorcrevar:
# Description

There was no support for batch requests over web socket in edge

```
HI Team - [@Jesse Lee](https://0xpolygon.slack.com/team/U044LBPUHUG) and I observed that polycli monitor didn't work properly when connected to Edge via websocket.  It works fine on a standard HTTP transport. The root cause seems to be that websocket RPC doesn't seem to support standard JSON RPC requests that are [batched.In](http://batched.in/) the example screenshot, I'm doing 4 things:


OK .Making a batch call over websocket to Alchemy. This works fine


NOT OK. Making a batch call over websocket to v0.7.1-alpha3 Nexon. This does not work.


NOT OK. Making a batch call over websocket to v0.6.2 Necon. This does not work.


OK. Making a batch call over HTTP to v0.7.1-alpha3 Nexon. This works fineJust wanted to check and see if we're doing something wrong or if this is expected behavior?
```

# Changes include

# Checklist

## Testing

### DIFF
--- a/jsonrpc/dispatcher.go
+++ b/jsonrpc/dispatcher.go
@@ -231,11 +231,45 @@ func (d *Dispatcher) RemoveFilterByWs(conn wsConn) {
 }
 
 func (d *Dispatcher) HandleWs(reqBody []byte, conn wsConn) ([]byte, error) {
+	// first try to unmarshal to batch request
+	// if there is an error try to unmarshal to single request
+	var batchReq BatchRequest
+	if err := json.Unmarshal(reqBody, &batchReq); err == nil {
+		const (
+			openSquareBracket  = 91 // [
+			closeSquareBracket = 93 // ]
+			comma              = 44 // ,
+		)
+
+		responses := make([][]byte, len(batchReq))
+
+		for i, req := range batchReq {
+			responses[i], err = d.handleWs(req, conn)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		var buf bytes.Buffer
+
+		// batch output should look like:
+		// [ { "requestId": "1", "status": 200 }, { "requestId": "2", "status": 200 } ]
+		buf.WriteByte(openSquareBracket)                // [
+		buf.Write(bytes.Join(responses, []byte{comma})) // join responses with the comma separator
+		buf.WriteByte(closeSquareBracket)               // ]
+
+		return buf.Bytes(), nil
+	}
+
 	var req Request
 	if err := json.Unmarshal(reqBody, &req); err != nil {
 		return NewRPCResponse(req.ID, "2.0", nil, NewInvalidRequestError("Invalid json request")).Bytes()
 	}
 
+	return d.handleWs(req, conn)
+}
+
+func (d *Dispatcher) handleWs(req Request, conn wsConn) ([]byte, error) {
 	// if the request method is eth_subscribe we need to create a
 	// new filter with ws connection
 	if req.Method == "eth_subscribe" {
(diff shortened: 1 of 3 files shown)


jsonrpc/dispatcher.go
```diff
@@ -231,11 +231,45 @@ func (d *Dispatcher) RemoveFilterByWs(conn wsConn) {
...
 
 func (d *Dispatcher) HandleWs(reqBody []byte, conn wsConn) ([]byte, error) {
+	// first try to unmarshal to batch request
+	// if there is an error try to unmarshal to single request
+	var batchReq BatchRequest
+	if err := json.Unmarshal(reqBody, &batchReq); err == nil {
+		const (
+			openSquareBracket  = 91 // [
+			closeSquareBracket = 93 // ]
+			comma              = 44 // ,
```
Stefan-Ethernal: Just curious could you have used strings instead of ASCII codes?

igorcrevar: fixed

Nemanja0x (approved):

jsonrpc/dispatcher.go
```diff
@@ -231,54 +237,90 @@ func (d *Dispatcher) RemoveFilterByWs(conn wsConn) {
...
-	// new filter with ws connection
-	if req.Method == "eth_subscribe" {
-		filterID, err := d.handleSubscribe(req, conn)
-		if err != nil {
-			return NewRPCResponse(req.ID, "2.0", nil, err).Bytes()
-		}
+	reqBody = bytes.TrimLeft(reqBody, " \t\r\n")
 
-		resp, err := formatFilterResponse(req.ID, filterID)
+	// if body begins with [ than consider this request as batch request
```
Stefan-Ethernal:
```suggestion
	// if body begins with [ consider it as a batch request
```

jsonrpc/dispatcher.go
```diff
@@ -231,54 +237,90 @@ func (d *Dispatcher) RemoveFilterByWs(conn wsConn) {
...
+	if len(reqBody) > 0 && reqBody[0] == openSquareBracket {
+		var batchReq BatchRequest
 
+		err := json.Unmarshal(reqBody, &batchReq)
 		if err != nil {
-			return NewRPCResponse(req.ID, "2.0", nil, err).Bytes()
+			return NewRPCResponse(nil, "2.0", nil,
```
Stefan-Ethernal, line 254: Should we provide request id as part of `RPCResponse` even in case of returning an error?

jsonrpc/dispatcher.go
```diff
@@ -231,54 +237,90 @@ func (d *Dispatcher) RemoveFilterByWs(conn wsConn) {
...
-	}
-
-	if req.Method == "eth_unsubscribe" {
-		ok, err := d.handleUnsubscribe(req)
-		if err != nil {
-			return nil, err
+		// if not disabled, avoid handling long batch requests
+		if d.params.isExceedingBatchLengthLimit(uint64(len(batchReq))) {
+			return NewRPCResponse(
+				nil,
```
Stefan-Ethernal, line 261: Provide request id here?

jsonrpc/dispatcher.go
igorcrevar: There is no id for batch request. only ids for each of requests inside batch

jsonrpc/dispatcher.go
igorcrevar: there is no id for batch request, only id for each of requests inside batch

Stefan-Ethernal (approved):

jsonrpc/dispatcher.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1588/commits/0f7548b633a3bb84aa337f837ad13bc71ddba4c4

goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1594, opened 2023-06-07
Extract reward token approve transaction to a separate function

Stefan-Ethernal:
# Description
Extract approve reward pool transaction as a separate function, to make code cleaner.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/polybft.go
+++ b/consensus/polybft/polybft.go
@@ -141,8 +141,12 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 			return err
 		}
 
-		if err = mintRewardTokensToWalletAddress(&polyBFTConfig, transition); err != nil {
-			return err
+		if !isNativeRewardToken(polyBFTConfig) {
+			// if reward token is a native erc20 token, we don't need to mint an amount of tokens
+			// for given wallet address to it since this is done in premine
+			if err = mintRewardTokensToWalletAddress(&polyBFTConfig, transition); err != nil {
+				return err
+			}
 		}
 
 		// initialize RewardPool SC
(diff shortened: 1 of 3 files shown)


consensus/polybft/polybft.go
```diff
@@ -141,8 +141,12 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 			return err
 		}
 
-		if err = mintRewardTokensToWalletAddress(&polyBFTConfig, transition); err != nil {
-			return err
+		if !isNativeRewardToken(polyBFTConfig) {
```
Nemanja0x: Nitpicking: Why don't just introduce this check `if !isNativeRewardToken(polyBFTConfig)` at the beginning of  `mintRewardTokensToWalletAddress`. Then `mintRewardTokensToWalletAddress` is safe? Maybe I am worng, just asking.

Nemanja0x (approved):

Nemanja0x:
Some test have failed, probably we should check those before merge.

consensus/polybft/polybft.go
Stefan-Ethernal: I was first thinking putting it there, but then I thought why it should even call the function if it is native reward token and immediately jump out of it. But since you have asked the same, I'd consider putting condition within the mint function itself (although it is expected to be called only on this place).

Stefan-Ethernal:
> Some test have failed, probably we should check those before merge.

You are right, I've made a mistake. We must send `approve` transaction to the reward token, independently which token is used for rewards (native or custom one). Since `approve` transaction allows reward pool to spend reward tokens on behalf of reward wallet. This is being done due to the fact that reward pool distributes rewards (see: https://github.com/0xPolygon/core-contracts/blob/dev/contracts/child/validator/RewardPool.sol#L77-L80).

Regarding this PR, functionality wise it is the same as before. However I find it necessary to extract reward token approval to a separate function.

consensus/polybft/polybft.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1594/commits/7557c1d623769385a0ebe48f3acbc6e4461fbbdb

goran-ethernal (approved):

PR #1598, opened 2023-06-08
[EVM-685]: Remove premining of validator balances on L2 for non-mintable tokens

goran-ethernal:
# Description

There was a misuse of `premine` flag in `genesis` if native child token is not mintable. Basically, we premined balances of validators, creating a situation where we minted tokens "out of thin air" on `childchain`. If native child token is not mintable, then validators need to bridge their tokens once the `childchain` is started.
Exceptions are zero address (since it is used to transfer tokens from it to the account address in bridge), and address of reward wallet, if native reward token is used as reward token.

If native child token is mintable, then we allow premining of any address.

Since we finalize stake of genesis validators on `rootchain` on `supernet` command, there was also a gap in initializing `CheckpointManager` contract, since it holds the current validator set for checkpoint submission, and `VotingPower` (`Stake`) is a crucial part of it, we must initialize `CheckpointManager` only after we finalize stake of genesis validators on `supernet` command. Because of that, `supernet` command is modified to initialize `CheckpointManager` with correct data (`VotingPower`) after it finalizes the genesis validators stake on `rootchain` (keep in mind that deployment of `CheckpointManager` is still in the `deploy` command. Only its `initialization` is in `supernet` command).

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/polybft_config.go
+++ b/consensus/polybft/polybft_config.go
@@ -104,6 +104,8 @@ type BridgeConfig struct {
 	StakeManagerAddr                  types.Address `json:"stakeManagerAddr"`
 	// only populated if stake-manager-deploy command is executed, and used for e2e tests
 	StakeTokenAddr types.Address `json:"stakeTokenAddr,omitempty"`
+	BLSAddress     types.Address `json:"blsAddr"`
+	BN256G2Address types.Address `json:"bn256G2Addr"`
 
 	JSONRPCEndpoint         string                   `json:"jsonRPCEndpoint"`
 	EventTrackerStartBlocks map[types.Address]uint64 `json:"eventTrackerStartBlocks"`
@@ -158,6 +160,8 @@ func (r *RootchainConfig) ToBridgeConfig() *BridgeConfig {
 		ChildMintableERC1155PredicateAddr: r.ChildMintableERC1155PredicateAddress,
 		CustomSupernetManagerAddr:         r.CustomSupernetManagerAddress,
 		StakeManagerAddr:                  r.StakeManagerAddress,
+		BLSAddress:                        r.BLSAddress,
+		BN256G2Address:                    r.BN256G2Address,
 	}
 }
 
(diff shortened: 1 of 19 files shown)


dusan-maksimovic (approved):

Nemanja0x (approved):

stana-miric (approved):

e2e-polybft/e2e/helpers_test.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+const nativeTokenMintableTestCfg = "Mintable Edge Coin:MEC:18:true"
```
Stefan-Ethernal: Don't forget to provide owner of mintable token, as it is changed in the and it is obligatory to provide an owner account when using mintable native token https://github.com/0xPolygon/polygon-edge/pull/1603.

consensus/polybft/polybft_config.go
```diff
@@ -104,6 +104,8 @@ type BridgeConfig struct {
 	StakeManagerAddr                  types.Address `json:"stakeManagerAddr"`
 	// only populated if stake-manager-deploy command is executed, and used for e2e tests
 	StakeTokenAddr types.Address `json:"stakeTokenAddr,omitempty"`
+	BLSAddress     types.Address `json:"blsAddr"`
+	BN256G2Address types.Address `json:"bn256G2Addr"`
```
Stefan-Ethernal, line 108: Why do you need these two addresses in the genesis? Were you thinking to put it just in case somebody is curios about them (since Edge is not using them)...

Stefan-Ethernal (approved):
LGTM, aside from minor comments

consensus/polybft/polybft_config.go
goran-ethernal: Because they are deployed in the `deploy` command, but used for initialization of `CheckpointManager` in `supernet` command. We weren't saving them to `genesis` before, but now we need to.

e2e-polybft/e2e/helpers_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1598/commits/afe1c185b74d5adc067947223834a605d1567f01

PR #1600, opened 2023-06-08
jesse/fix-nightly-v1.0.0-rc1

gatsbyz:


praetoriansentry (approved):

.github/workflows/deploy.nightly.devnet.yml
```diff
@@ -188,6 +202,9 @@ jobs:
 
           main
 
+          sudo apt update
+          sudo apt install nodejs npm
```
Stefan-Ethernal: Why do we need npm and Node JS?

Stefan-Ethernal (commented):
Can you provide a bit more context, what was wrong with the previous version of the nightly workflow?

.github/workflows/deploy.nightly.devnet.yml
gatsbyz: i got rid of this

gatsbyz:
> Can you provide a bit more context, what was wrong with the previous version of the nightly workflow?

@Stefan-Ethernal I updated the script to use the `--test` flag. The `stake` command was failing with the error `staking transaction failed on block`. And this is because a ERC20StakeToken needs to be deployed. We learned that the `--test` flag is doing this in the back, so we included it in this nightly script. Hope this explains.

vcastellm (approved):

PR #1603, opened 2023-06-09
Allow providing mintable native token owner account

Stefan-Ethernal:
# Description

The `genesis` command's `--native-token-config` flag is expanded with another parameter that allows specifying the owner of a mintable native token. Prior to this PR, the first validator from the initial validator set was set as the owner of the mintable native token, which means that it was able to do the minting and burning.

In case a non-mintable native token is used, the owner account address should be omitted (since it is not applicable in that case).

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/genesis/params.go
+++ b/command/genesis/params.go
@@ -43,7 +43,7 @@ const (
 	defaultNativeTokenName     = "Polygon"
 	defaultNativeTokenSymbol   = "MATIC"
 	defaultNativeTokenDecimals = uint8(18)
-	nativeTokenParamsNumber    = 4
+	minNativeTokenParamsNumber = 4
 )
 
 // Legacy flags that need to be preserved for running clients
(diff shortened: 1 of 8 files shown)


command/genesis/params.go
```diff
@@ -481,41 +481,55 @@ func (p *genesisParams) extractNativeTokenMetadata() error {
...
 	if err != nil {
 		return errInvalidTokenParams
 	}
 
+	// in case it is mintable native token, it is expected to have 5 parameters provided
+	if isMintable && len(params) != minNativeTokenParamsNumber+1 {
+		return errInvalidTokenParams
+	}
+
+	// owner address
```
igorcrevar, line 524: i think this will fail if `!isMintable` and fifth parameter is not provided

goran-ethernal (approved):

igorcrevar (approved):

command/genesis/params.go
Stefan-Ethernal: Spot on 🙌 : https://github.com/0xPolygon/polygon-edge/pull/1603/commits/b0a999f940e0f5fdcca105728aa7ccec0b45f2f7.

Implemented unit test as well: https://github.com/0xPolygon/polygon-edge/pull/1603/commits/1c008c19ed981455201429e5a61a48a411fc8851

dusan-maksimovic (approved):

vcastellm (approved):
If we don't want to introduce a global admin address in genesis, this is good enough, otherwise maybe we should think on it.

Stefan-Ethernal:
> If we don't want to introduce a global admin address in genesis, this is good enough, otherwise maybe we should think on it.

We can make sure to reiterate that in the future. For now, we are good with this approach.

PR #1609, opened 2023-06-12
TX pool debug level logging optimizations

dusan-maksimovic:
# Description

When there are a lot of txs being added to the tx pool, converting TX hash into string for logging takes a lot of time. If the log level isn't Debug or lower, this is unnecessary. Now we first check if the log level is Debug or lower before performing the conversion. Also added this check in couple places where sender address is logged on Debug level.

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -427,11 +427,14 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 	clearAccountQueue(dropped)
 
 	p.eventManager.signalEvent(proto.EventType_DROPPED, tx.Hash)
-	p.logger.Debug("dropped account txs",
-		"num", droppedCount,
-		"next_nonce", nextNonce,
-		"address", tx.From.String(),
-	)
+
+	if p.logger.IsDebug() {
+		p.logger.Debug("dropped account txs",
+			"num", droppedCount,
+			"next_nonce", nextNonce,
+			"address", tx.From.String(),
+		)
+	}
 }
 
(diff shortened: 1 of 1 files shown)


txpool/txpool.go
```diff
@@ -427,11 +427,14 @@ func (p *TxPool) Drop(tx *types.Transaction) {
...
-		"num", droppedCount,
-		"next_nonce", nextNonce,
-		"address", tx.From.String(),
-	)
+
+	if p.logger.IsDebug() {
+		p.logger.Debug("dropped account txs",
+			"num", droppedCount,
+			"next_nonce", nextNonce,
+			"address", tx.From.String(),
```
igorcrevar, line 435: I think we do not need to explicitly convert `tx.From.String()`. We can just use `tx.From` and in that case even `if` statement is unnecessary 

Same goes for all the other changes

dusan-maksimovic: @igorcrevar We could, but in that case the value will be converted to String by using reflection in the logger code, which is much slower. I considered that approach but eventually opted for this one. But I could change it if we agree it is better to have it that way.

dusan-maksimovic: I mean, it will be converted to String IF the log level is Debug or lower, so in that case it would be much slower than with if statement.

goran-ethernal (approved):

igorcrevar (approved):

Stefan-Ethernal (approved):

PR #1616, opened 2023-06-13
Print child token address as part of deposit commands

Stefan-Ethernal:
# Description

This PR introduces outputting of the child token address in the deposit commands.

# Changes include

# Checklist

## Testing


stana-miric (approved):

command/bridge/common/params.go
```diff
@@ -160,21 +161,58 @@ func ExtractExitEventID(receipt *ethgo.Receipt) (*big.Int, error) {
...
+	ChildTokenAddr *types.Address `json:"childTokenAddr"`
 
 	Title string `json:"title"`
 }
 
 func (r *BridgeTxResult) GetOutput() string {
 	var buffer bytes.Buffer
 
-	vals := make([]string, 0, 5)
+	vals := make([]string, 0, 7)
```
goran-ethernal, line 215:
```suggestion
	vals := make([]string, 0, 6)
```

goran-ethernal (approved):

consensus/polybft/contractsapi/helper.go
```diff
@@ -13,6 +14,16 @@ type StateTransactionInput interface {
 	DecodeAbi(b []byte) error
 }
 
+// Event is an abstraction for Ethereum events
+type Event interface {
```
goran-ethernal: Is this used?

command/bridge/common/params.go
Stefan-Ethernal: In the worst case scenario it would be 7 (for ERC1155 which have amounts and token ids set). So I'd leave it to 7 if you agree?

consensus/polybft/contractsapi/helper.go
Stefan-Ethernal: I thought it would be useful, but eventually, it isn't. I''ve removed it: https://github.com/0xPolygon/polygon-edge/pull/1616/commits/3cf102887c9beb0522a907a7b290b832927d7915

PR #1624, opened 2023-06-15
updated cluster script to install jq if not available

rachit77:
# Description

Modified the cluster script to include the installation of the `jq` package if it is not already installed on the system.

# Changes include

# Checklist

## Testing

### Manual tests

I have manually tested the script to start a cluster locally


vcastellm (changes requested):
IMO we should not include the installation of `jq`, just notify the user the requirement.

As it is now we would introduce a piece of code that's very difficult to ensure that it works with the wide variety of systems the users will try, therefore it will be de facto limited to certain platforms without a strong reason. For example, does it works with non sudoers? Does it work in NetBSD? Etc

PR #1627, opened 2023-06-16
Log contract deployment transaction details

jelacamarko:
# Description

Added Output Formatter to TxRelayer as TxRelayerOption. 
Log deployment contract transaction details: from, gas, and gas price.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/rootchain/deploy/result.go
+++ b/command/rootchain/deploy/result.go
@@ -13,13 +13,17 @@ type deployContractResult struct {
 	Name    string        `json:"name"`
 	Address types.Address `json:"address"`
 	Hash    types.Hash    `json:"hash"`
+	GasUsed uint64        `json:"gasUsed"`
 }
 
-func newDeployContractsResult(name string, address types.Address, hash ethgo.Hash) *deployContractResult {
+func newDeployContractsResult(name string,
+	address types.Address,
+	hash ethgo.Hash, gasUsed uint64) *deployContractResult {
 	return &deployContractResult{
 		Name:    name,
 		Address: address,
 		Hash:    types.BytesToHash(hash.Bytes()),
+		GasUsed: gasUsed,
 	}
 }
 
@@ -28,10 +32,11 @@ func (r deployContractResult) GetOutput() string {
 
 	buffer.WriteString("\n[ROOTCHAIN - DEPLOY CONTRACT]\n")
 
-	vals := make([]string, 0, 3)
+	vals := make([]string, 0, 4)
 	vals = append(vals, fmt.Sprintf("Name|%s", r.Name))
 	vals = append(vals, fmt.Sprintf("Contract (address)|%s", r.Address))
 	vals = append(vals, fmt.Sprintf("Transaction (hash)|%s", r.Hash))
+	vals = append(vals, fmt.Sprintf("Transaction (gas used)|%d", r.GasUsed))
 
 	buffer.WriteString(helper.FormatKV(vals))
 	buffer.WriteString("\n")
--- a/txrelayer/txrelayer.go
+++ b/txrelayer/txrelayer.go
@@ -137,7 +137,7 @@ func (t *TxRelayerImpl) sendTransactionLocked(txn *ethgo.Transaction, key ethgo.
 
 	if t.writer != nil {
 		_, _ = t.writer.Write([]byte(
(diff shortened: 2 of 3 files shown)


txrelayer/txrelayer.go
```diff
@@ -130,6 +135,12 @@ func (t *TxRelayerImpl) sendTransactionLocked(txn *ethgo.Transaction, key ethgo.
 		return ethgo.ZeroHash, err
 	}
 
+	if t.writer != nil {
+		_, _ = t.writer.Write([]byte(
+			fmt.Sprintf("[ROOTCHAIN - CONTRACTS DEPLOYMENT TRANSACTION DETAIL]\nFrom = %s \nGas = %d  \nGas Price = %d\n",
```
Stefan-Ethernal: @Nemanja0x would we like to log Input field as well?

Stefan-Ethernal: This is wrong. You should not have hardcoded logging this kind of message here, because `TxRelayer` is used not only by `rootchain deploy` command, but on multiple places.

I feel that you should provide some kind of logging callback, instead of explicitly hardcoding logging message here.

Stefan-Ethernal (dismissed):

txrelayer/txrelayer.go
jelacamarko: Can the logging message be just "Transaction detail" ?

jelacamarko: Or we can add a message argument to the WithWriter function.

stana-miric (approved):

txrelayer/txrelayer.go
Stefan-Ethernal: > Can the logging message be just "Transaction detail" ?

...or you can simply add [TxRelayer.SendTransaction] as prefix and then transaction details. Although check with @Nemanja0x, about this point (https://github.com/0xPolygon/polygon-edge/pull/1627#discussion_r1232097459). However you find it better.

Stefan-Ethernal: Removing extra space there:
```suggestion
			fmt.Sprintf("[ROOTCHAIN - CONTRACTS DEPLOYMENT TRANSACTION DETAIL]\nFrom = %s \nGas = %d \nGas Price = %d\n",
```

Nemanja0x: No need for input, we cannot read anything from it.

Nemanja0x: It is not bad idea to have a message argument, but however you decide. Just do not want to to  have hardcoded "ROOTCHAIN ..."

Stefan-Ethernal: If you have SC artifacts, then you could know which SC are you deploying.

Nemanja0x: Ok, then add that as well.

jelacamarko: We already have additional information. Log currently looks like this 

```
INFO [06-19|10:48:59.069] Submitted contract creation
hash=0xdf1ee50904451a1dcd7d75a6e5c1e4725856a0792d56630c4a8d49a5e5258c8e
from=0x1FC1411A3Bd09E63d4A306FD1EB838bA457ddA13 nonce=3
contract=0x436604426F31A05f905C64edc973E575BdB46471 value=0
[TxRelayer.SendTransaction]
From = 0x1FC1411A3Bd09E63d4A306FD1EB838bA457ddA13 
Gas = 5242880 
Gas Price = 477436156

```
And also later in log 

```
[ROOTCHAIN - DEPLOY CONTRACT]
Name                   = RootERC721Predicate
Contract (address)     = 0x436604426F31A05f905C64edc973E575BdB46471
Transaction (hash)     = 0xdf1ee50904451a1dcd7d75a6e5c1e4725856a0792d56630c4a8d49a5e5258c8e
Transaction (gas used) = 1458136
```

jelacamarko: I have added gas used from the receipt

Nemanja0x (approved):

command/rootchain/deploy/result.go
```diff
@@ -32,6 +36,7 @@ func (r deployContractResult) GetOutput() string {
 	vals = append(vals, fmt.Sprintf("Name|%s", r.Name))
```
Stefan-Ethernal, line 36: Nitpick: increase the capacity of this slice by 1:
```go
vals := make([]string, 0, 4)
```

PR #1629, opened 2023-06-16
[EVM-697]: Implement `eth_maxPriorityFeePerGas`

goran-ethernal:
# Description

This PR implements `eth_maxPriorityFeePerGas` json rpc endpoint.

**Parameters**:
- This method does not accept any parameters.

**Returns**:
- result string - The hex value of the priority fee needed to be included in a block (in wei).

**Request**:

```
curl https://link-to-some-node/ \
-X POST \
-H "Content-Type: application/json" \
--data '{"method":"eth_maxPriorityFeePerGas","params":[],"id":1,"jsonrpc":"2.0"}'
```
**Response**:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0xbebc200"
 }
```

The `eth_maxPriorityFeePerGas` rpc point returns an estimated fee per gas, indicating the amount you can pay as a priority fee (tip) in order to have a transaction in the current block.

Edge implementation follows the `go-ethereum` implementation.

# Changes include

# Checklist

## Testing

### DIFF
--- a/gasprice/gasprice.go
+++ b/gasprice/gasprice.go
@@ -14,7 +14,7 @@ import (
 
 const couldNotFoundBlockFormat = "could not find block. Number: %d, Hash: %s"
 
-// Default config for gas helper (as per ethereum)
+// DefaultGasHelperConfig is the default config for gas helper (as per ethereum)
 var DefaultGasHelperConfig = &Config{
 	NumOfBlocksToCheck: 20,
 	PricePercentile:    60,
@@ -130,8 +130,8 @@ func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
 		blockMiner := types.BytesToAddress(currentBlock.Header.Miner)
 		signer := crypto.NewSigner(g.backend.Config().Forks.At(currentBlock.Number()),
 			uint64(g.backend.Config().ChainID))
-
 		blockTxPrices := make([]*big.Int, 0)
+
 		for _, tx := range txSorter.txs {
 			tip := tx.EffectiveTip(baseFee)
 
@@ -193,6 +193,7 @@ func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
 	}
 
 	var price *big.Int
+
 	if len(allPrices) > 0 {
 		// sort prices from lowest to highest
 		sort.Sort(bigIntSorted(allPrices))
--- a/gasprice/gasprice.go
+++ b/gasprice/gasprice.go
@@ -188,12 +188,10 @@ func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
 
 	// at least amount of transactions to get
 	minNumOfTx := int(g.numOfBlocksToCheck) * 2
-	if len(allPrices) < minNumOfTx {
-		// collect some more blocks and transactions if not enough transactions were collected
-		for len(allPrices) < minNumOfTx && currentBlock.Number() > 0 {
-			if err := collectPrices(currentBlock); err != nil {
-				return nil, err
-			}
+	// collect some more blocks and transactions if not enough transactions were collected
+	for len(allPrices) < minNumOfTx && currentBlock.Number() > 0 {
+		if err := collectPrices(currentBlock); err != nil {
+			return nil, err
 		}
 	}
 
--- a/gasprice/gasprice.go
+++ b/gasprice/gasprice.go
@@ -123,12 +123,12 @@ func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
 	var allPrices []*big.Int
 
 	collectPrices := func(block *types.Block) error {
-		baseFee := currentBlock.Header.BaseFee
-		txSorter := newTxByEffectiveTipSorter(currentBlock.Transactions, baseFee)
+		baseFee := block.Header.BaseFee
+		txSorter := newTxByEffectiveTipSorter(block.Transactions, baseFee)
 		sort.Sort(txSorter)
 
-		blockMiner := types.BytesToAddress(currentBlock.Header.Miner)
-		signer := crypto.NewSigner(g.backend.Config().Forks.At(currentBlock.Number()),
+		blockMiner := types.BytesToAddress(block.Header.Miner)
+		signer := crypto.NewSigner(g.backend.Config().Forks.At(block.Number()),
 			uint64(g.backend.Config().ChainID))
 		blockTxPrices := make([]*big.Int, 0)
 
--- a/gasprice/gasprice.go
+++ b/gasprice/gasprice.go
@@ -42,8 +42,8 @@ type Config struct {
 	IgnorePrice *big.Int
 }
 
-// Backend is the interface representing blockchain and consensus backend
-type Backend interface {
+// Blockchain is the interface representing blockchain
+type Blockchain interface {
 	GetBlockByHash(hash types.Hash, full bool) (*types.Block, bool)
 	Header() *types.Header
 	Config() *chain.Params
@@ -73,7 +73,7 @@ type GasHelper struct {
 	// when collecting transactions
 	ignorePrice *big.Int
 	// backend is an abstraction of blockchain
-	backend Backend
+	backend Blockchain
 	// lastHeaderHash is the last header for which maxPriorityFeePerGas was returned
 	lastHeaderHash types.Hash
 
@@ -81,7 +81,7 @@ type GasHelper struct {
 }
 
 // NewGasHelper is the constructor function for GasHelper struct
-func NewGasHelper(config *Config, backend Backend) *GasHelper {
+func NewGasHelper(config *Config, backend Blockchain) *GasHelper {
 	pricePercentile := config.PricePercentile
 	if pricePercentile > 100 {
 		pricePercentile = 100
@@ -199,7 +199,9 @@ func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
 
 	if len(allPrices) > 0 {
 		// sort prices from lowest to highest
-		sort.Sort(bigIntSorted(allPrices))
+		sort.Slice(allPrices, func(i, j int) bool {
+			return allPrices[i].Cmp(allPrices[j]) < 0
+		})
 		// take the biggest price that is in the configured percentage
 		// by default it's 60, so it will take the price on that percentage
 		// of all prices in the array
@@ -249,20 +251,4 @@ func (t *txSortedByEffectiveTip) Less(i, j int) bool {
 	tip2 := t.txs[j].EffectiveTip(t.baseFee)
 
 	return tip1.Cmp(tip2) < 0
-}
-
-// bigIntSorted sorts big int slice from lowest to highest value
-type bigIntSorted []*big.Int
-
-// Len is implementation of sort.Interface
-func (b bigIntSorted) Len() int { return len(b) }
-
-// Less is implementation of sort.Interface
-func (b bigIntSorted) Less(i, j int) bool {
-	return b[i].Cmp(b[j]) < 0
-}
-
-// Swap is implementation of sort.Interface
-func (b bigIntSorted) Swap(i, j int) {
-	b[i], b[j] = b[j], b[i]
 }
--- a/gasprice/gasprice.go
+++ b/gasprice/gasprice.go
@@ -82,9 +82,14 @@ type GasHelper struct {
 
 // NewGasHelper is the constructor function for GasHelper struct
 func NewGasHelper(config *Config, backend Backend) *GasHelper {
+	pricePercentile := config.PricePercentile
+	if pricePercentile > 100 {
+		pricePercentile = 100
+	}
+
 	return &GasHelper{
 		numOfBlocksToCheck: config.NumOfBlocksToCheck,
-		pricePercentile:    config.PricePercentile,
+		pricePercentile:    pricePercentile,
 		sampleNumber:       config.SampleNumber,
 		ignorePrice:        config.IgnorePrice,
 		lastPrice:          config.LastPrice,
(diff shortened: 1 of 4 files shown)


gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+func (g *GasHelper) MaxPriorityFeePerGas() (*big.Int, error) {
+	currentHeader := g.backend.Header()
+
+	currentBlock, found := g.backend.GetBlockByHash(currentHeader.Hash, true)
+	if !found {
+		return nil, fmt.Errorf(couldNotFoundBlockFormat, currentHeader.Number, currentHeader.Hash)
+	}
+
+	g.lock.Lock()
+	lastPrice := g.lastPrice
```
igorcrevar, line 119: Maybe we can use *lru.Cache instead of this optimization?

goran-ethernal: Yes we can. But we are on a tight schedule, that is why I didn't wnat to implement that. I'll see on Monday if it will take too much time.

goran-ethernal: Added `lru.cache` as suggested: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/58c2de510e227952d09ff31151076a481279ae1f

goran-ethernal: After looking at the code once more, we decided to drop the commit with `lru.cache` since we only take the last value calculated, so no need for a cache.

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+		for len(allPrices) < minNumOfTx && currentBlock.Number() > 0 {
+			if err := collectPrices(currentBlock); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	var price *big.Int
+
+	if len(allPrices) > 0 {
```
igorcrevar, line 200: if `len(allPrices) == 0` price will stay nil. There will be panic on line 206... but even with another check on that line i am not sure if its good to return nil price

goran-ethernal: Good catch. Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/742140ae0555ff16139cb928df5e533493455bb8

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+
+	if currentHeader.Hash == lastHeader {
+		// small optimization, if we calculated already the price for given block
+		return new(big.Int).Set(lastPrice), nil
+	}
+
+	var allPrices []*big.Int
+
+	collectPrices := func(block *types.Block) error {
+		baseFee := currentBlock.Header.BaseFee
```
stana-miric: here function's parameter block should be used instead of currentBlock or it can be removed because the current block is updating in a loop when collectPrices is called(probably its cleaner to keep and use block parameter)?
same goes for all the places where currentBlock is used in collectPrices function.

goran-ethernal: Yes, but it doesn't matter, since `currentBlock` is updated on each iteration, but I will fix it since it's more cleaner with `currentBlock` passed to function.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/bf9af66bf04673d0bbe2fe7f312ab6e8cc070248

stana-miric (approved):

server/server.go
```diff
@@ -328,6 +332,9 @@ func NewServer(config *Config) (*Server, error) {
 		return nil, err
 	}
 
+	// here we can provide some other configuration
+	m.gasHelper = gasprice.NewGasHelper(gasprice.DefaultGasHelperConfig, m.blockchain)
```
Stefan-Ethernal, line 336: Should we expose gas helper config to server configuration as well? Does it make sense to have different configuration parameters per server instance. If it does, I don't see how we could provide different parameters, other than changing the code. 

On the other hand if those params should be network wide, then obviously we should put it as a part of on-chain governance.

To me, it makes sense either way, as I couldn't find in the specification is it expected that every single node in network must calculate the same value.

WDYT?

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+		// sort prices from lowest to highest
+		sort.Sort(bigIntSorted(allPrices))
+		// take the biggest price that is in the configured percentage
+		// by default its 60, so it will take the price on that percentage
+		// of all prices in the array
+		price = allPrices[(len(allPrices)-1)*int(g.pricePercentile)/100]
+	}
+
+	if price.Cmp(g.maxPrice) > 0 {
+		// if price is larger then the configured max price
```
Stefan-Ethernal:
```suggestion
		// if price is larger than the configured max price
```

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+
+	price := lastPrice
+
+	if len(allPrices) > 0 {
+		// sort prices from lowest to highest
+		sort.Sort(bigIntSorted(allPrices))
+		// take the biggest price that is in the configured percentage
+		// by default its 60, so it will take the price on that percentage
```
Stefan-Ethernal:
```suggestion
		// by default it's 60, so it will take the price on that percentage
```

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+		blockMiner := types.BytesToAddress(block.Header.Miner)
+		signer := crypto.NewSigner(g.backend.Config().Forks.At(block.Number()),
+			uint64(g.backend.Config().ChainID))
+		blockTxPrices := make([]*big.Int, 0)
+
+		for _, tx := range txSorter.txs {
+			tip := tx.EffectiveTip(baseFee)
+
+			if tip.Cmp(g.ignorePrice) == -1 {
+				// ignore transactions with tip lower then ignore price
```
Stefan-Ethernal:
```suggestion
				// ignore transactions with tip lower than ignore price
```

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		currentBlock, found = g.backend.GetBlockByHash(currentBlock.ParentHash(), true)
+		if !found {
+			return nil, fmt.Errorf(couldNotFoundBlockFormat, currentHeader.Number, currentHeader.Hash)
+		}
+	}
+
+	// at least amount of transactions to get
+	minNumOfTx := int(g.numOfBlocksToCheck) * 2
```
Stefan-Ethernal, line 190: Why you don't rely on `sampleNumber` parameter here instead?

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+
+	price := lastPrice
+
+	if len(allPrices) > 0 {
+		// sort prices from lowest to highest
+		sort.Sort(bigIntSorted(allPrices))
+		// take the biggest price that is in the configured percentage
+		// by default its 60, so it will take the price on that percentage
+		// of all prices in the array
+		price = allPrices[(len(allPrices)-1)*int(g.pricePercentile)/100]
```
Stefan-Ethernal, line 208: - I don't think that this downcasting is generally a good idea. 
Why `pricePercentile` is `uint64` and not `int` (or at least `uint`)?

- Should we validate `pricePercentile`? 

- It is not obvious to me why are you indexing this way (multiplying `pricePercentile` divided by 100 with last index of `allPrices`)? Seems like it can lead to index out of range in some cases.

gasprice/gasprice.go
```diff
@@ -0,0 +1,265 @@
...
+
+		currentBlock, found = g.backend.GetBlockByHash(currentBlock.ParentHash(), true)
+		if !found {
+			return nil, fmt.Errorf(couldNotFoundBlockFormat, currentHeader.Number, currentHeader.Hash)
+		}
+	}
+
+	// at least amount of transactions to get
+	minNumOfTx := int(g.numOfBlocksToCheck) * 2
+	if len(allPrices) < minNumOfTx {
```
Stefan-Ethernal: Seems like you don't need this condition, since for loop already checks it.

server/server.go
goran-ethernal: They can't be per server. It's a network wide configuration, probably provided through genesis, and on-chain governance. For now, I've left it to default values that `geth` uses, but in the future we should definitely provide a means to users to configure this through governance.

gasprice/gasprice.go
goran-ethernal: 1. Because I don't want percentile to be lower the 0, but ok, I can sanitize it before I set it in the `GasHelper` struct.
2. Answered in 1.
3. Not correct. It will never lead to index ouyt of range because percentile can be in the range of 0-100. As I said, I will sanitize the value provided by config before setting it to `GasHelper`.

gasprice/gasprice.go
goran-ethernal: If you are thinking about this loop:

`for i := uint64(0); i < g.numOfBlocksToCheck && currentBlock.Number() > 0; i++ {`

then it's not correct, because if you see that `collectPrices` function, there are cases when we don't collect any transaction from block (if block is empty, or all transactions are added by block miner).

gasprice/gasprice.go
goran-ethernal: `sampleNumber` is the amount of transactions we pool from the block. This `minNumOfTx` is from `geth` it basically means, if I didn't get at least 2 transactions per block, try to get more transactions from more blocks so that we can get a more precise calculation. If I had empty blocks, or blocks which have only transactions added by miner, then we need to look at more blocks to get enough transactions.

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/beeeda2e72ef1a9de256d45961f83db8fe7ff389

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/beeeda2e72ef1a9de256d45961f83db8fe7ff389

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/beeeda2e72ef1a9de256d45961f83db8fe7ff389

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/beeeda2e72ef1a9de256d45961f83db8fe7ff389

gasprice/gasprice.go
Stefan-Ethernal: Nope, I was referring to this for loop (obviously):
https://github.com/0xPolygon/polygon-edge/blob/EVM-697-Implement-eth_maxPriorityFee/gasprice/gasprice.go#L193

goran-ethernal: Ah, I see. Yep, I'll remove the check.

goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/74c482e73329fe7a785c718bf3ad57b831f36fb6

igorcrevar (approved):

gasprice/gasprice.go
Stefan-Ethernal: Ok, but still I'm thinking about why we would not provide a configuration parameter that would allow us to specify how many transactions per block we would like to gather.

It is a minor thing, but just asking out loud.

goran-ethernal: You have `sampleNumber`, but as I said, some blocks might be empty, some blocks can have only the miners transactions, so you will not get appropriate amount of transactions. This can also happen if the chain is young (has small amount of blocks), but in the case we do have enough blocks, we would need to look more in history to get more transactions, and this is more then enough.

To let someone configure how many transactions you need to get to have a more accurate calculation, is not something I would do, that is something the algorithm needs to figure out.

Stefan-Ethernal (approved):

gasprice/gasprice.go
```diff
@@ -0,0 +1,268 @@
...
+
+// Less is implementation of sort.Interface
+func (b bigIntSorted) Less(i, j int) bool {
+	return b[i].Cmp(b[j]) < 0
+}
+
+// Swap is implementation of sort.Interface
+func (b bigIntSorted) Swap(i, j int) {
+	b[i], b[j] = b[j], b[i]
+}
```
vcastellm: You can avoid this code using:

```
sort.Slice(allPrices, func(i, j int) bool {
    return allPrices[i].Cmp(allPrices[j]) > 0
})
```

gasprice/gasprice.go
```diff
@@ -0,0 +1,268 @@
...
+	// when collecting transactions
+	IgnorePrice *big.Int
+}
+
+// Backend is the interface representing blockchain and consensus backend
+type Backend interface {
+	GetBlockByHash(hash types.Hash, full bool) (*types.Block, bool)
+	Header() *types.Header
+	Config() *chain.Params
+}
```
vcastellm: Should the Backend interface be defined here?

goran-ethernal: Since I needed only these three functions from the blockchain, I didn't want to include some interfaces we already had in polybft. Plus, this makes mocking easier since I dont need to have mock implementations of all the functions in blockchain.

vcastellm: Why call it `Backend` if it's `Blockchain` interface? I vote to rename it then.

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/67ff422876baedb3ccee43e6e121f07428a028f2

gasprice/gasprice.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1629/commits/67ff422876baedb3ccee43e6e121f07428a028f2

vcastellm (approved):

PR #1638, opened 2023-06-19
[EVM-701]: Validate `ExitRootHash` in `CheckpointData`

goran-ethernal:
# Description

This PR introduces validation of provided `ExitRootHash` in `CheckpointData` in the proposal header `ExtraData`.

Since all validators collect `exit` events (needed to do L2-L1 transfers), they can validate the exit root hash provided by the block proposer in consensus, as an additional part of security. 

This PR also added appropriate UTs.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -38,8 +38,7 @@ type blockchainBackend interface {
 		txPool txPoolInterface, blockTime time.Duration, logger hclog.Logger) (blockBuilder, error)
 
 	// ProcessBlock builds a final block from given 'block' on top of 'parent'.
-	ProcessBlock(parent *types.Header, block *types.Block,
-		callback func(*state.Transition) error) (*types.FullBlock, error)
+	ProcessBlock(parent *types.Header, block *types.Block) (*types.FullBlock, error)
 
 	// GetStateProviderForBlock returns a reference to make queries to the state at 'block'.
 	GetStateProviderForBlock(block *types.Header) (contract.Provider, error)
@@ -83,8 +82,7 @@ func (p *blockchainWrapper) CommitBlock(block *types.FullBlock) error {
 }
 
 // ProcessBlock builds a final block from given 'block' on top of 'parent'
-func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Block,
-	callback func(*state.Transition) error) (*types.FullBlock, error) {
+func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Block) (*types.FullBlock, error) {
 	header := block.Header.Copy()
 	start := time.Now().UTC()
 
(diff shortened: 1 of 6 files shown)


consensus/polybft/fsm_test.go
```diff
@@ -710,6 +710,69 @@ func TestFSM_ValidateCommit_Good(t *testing.T) {
...
+
+	const (
+		accountsCount     = 5
+		parentBlockNumber = 25
+		signaturesCount   = 3
+	)
+
+	validators := validator.NewTestValidators(t, accountsCount)
+	parentExtra := createTestExtraObject(validators.GetPublicIdentities(), validator.AccountSet{}, 4, signaturesCount, signaturesCount)
+	parentExtra.Validators = nil
```
Stefan-Ethernal, line 724: Why it is important to set this to nil? I don't get it at the first glance.

Same here: https://github.com/0xPolygon/polygon-edge/pull/1638/files#diff-bf0766970c7dcf6e2f9546a5f524d03a9e2cefcb411b6838d3b74a3a500c15eeR736

consensus/polybft/blockchain_wrapper.go
```diff
@@ -83,8 +82,7 @@ func (p *blockchainWrapper) CommitBlock(block *types.FullBlock) error {
 }
 
 // ProcessBlock builds a final block from given 'block' on top of 'parent'
-func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Block,
-	callback func(*state.Transition) error) (*types.FullBlock, error) {
```
Stefan-Ethernal, line 87: Glad we got rid of this callback :raised_hands: 

I remember it was essential to have it validated after a block was executed since validators transitions were calculated by querying `ChildValidatorSet`, but now when tracking validators transitions is on the client side, it can be removed.

consensus/polybft/extra.go
```diff
@@ -459,6 +460,13 @@ func (c *CheckpointData) Validate(parentCheckpoint *CheckpointData,
 		return fmt.Errorf("epoch number should not change for epoch-ending block")
 	}
 
+	// exit root hash of proposer and
+	// validator that validates proposal have to match
+	if exitRootHash != c.EventRoot {
+		return fmt.Errorf("exit root hash not as expected. Expected: %s. Actual: %s",
```
Stefan-Ethernal: Nitpick: not sure if we should log expected and actual event roots. Could they be exploited somehow?
At least that was my mindset when I was validating current and next validators' hashes. I wasn't confident enough to log it due to some security considerations.

Stefan-Ethernal (approved):

consensus/polybft/extra.go
goran-ethernal: I don't think so, because they are `keccak256` hashes. I just put it there to see if they really don't match, but I can remove the log.

consensus/polybft/fsm_test.go
goran-ethernal: Because it's a non epoch ending block, and `ValidatorSetDelta` should be empty or nil, and I don't want the test to fail on that check, rather on the exit root hash check if it is as expected.

consensus/polybft/extra.go
Stefan-Ethernal: Then IMO either remove the hash from the error message here or expand the error message when we are checking validator set hashes to make error messages consistent.

stana-miric (approved):

consensus/polybft/extra.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1638/commits/ae88b4fee5735c76092d96575430705e62bcba9a

igorcrevar (approved):

PR #1645, opened 2023-06-21
EVM-713 Fork manager params

igorcrevar:
# Description

Forks inside `genesis.json` now will look like:
```
forks": {
            "Dummy1": {
                "block": 10,
                "params": {
                    "maxValidatorSetSize": 30
                }
            },
            "EIP150": {
                "block": 0,
            },
            ...
}
```

Currently there are 5 predefined parameters for forks

```
type ForkParams struct {
	// MaxValidatorSetSize indicates the maximum size of validator set
	MaxValidatorSetSize *uint64 `json:"maxValidatorSetSize,omitempty"`

	EpochSize *uint64 `json:"epochSize,omitempty"`

	// SprintSize is size of sprint
	SprintSize *uint64 `json:"sprintSize,omitempty"`

	// BlockTime is target frequency of blocks production
	BlockTime *common.Duration `json:"blockTime,omitempty"`

	// BlockTimeDrift defines the time slot in which a new block can be created
	BlockTimeDrift *uint64 `json:"blockTimeDrift,omitempty"`
}
```

We can specify some or all of those parameters for some fork in `genesis.json`. The parameters need to be pointers because some of them can be omitted. If some parameter is omitted for specific fork, then its values is picked from first fork with smaller block number or from initial fork parameters.

During the execution, parameter can be retrieved by using fork manager `func (fm *forkManager) GetParams(blockNumber uint64) *chain.ForkParams`

Initial parameters for polybft are set via `builtin.go` mapping function inside `polybft.go`
```
func ForkManagerInitialParamsFactory(config *chain.Chain) (*chain.ForkParams, error) {
	pbftConfig, err := GetPolyBFTConfig(config)
	if err != nil {
		return nil, err
	}

	return &chain.ForkParams{
		MaxValidatorSetSize: &pbftConfig.MaxValidatorSetSize,
		EpochSize:           &pbftConfig.EpochSize,
		SprintSize:          &pbftConfig.SprintSize,
		BlockTime:           &pbftConfig.BlockTime,
		BlockTimeDrift:      &pbftConfig.BlockTimeDrift,
	}, nil
}
```
If we want to add new parameter, then we need to add that parameter (as a pointer) to `ForkParams` and change its value in `genesis.json`. In that case its probably good to change `ForkManagerInitialParamsFactory` too

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/chain/params.go
+++ b/chain/params.go
@@ -5,6 +5,7 @@ import (
 	"math/big"
 	"sort"
 
+	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
@@ -117,17 +118,32 @@ func (f *Forks) At(block uint64) ForksInTime {
 	}
 }
 
+// ForkParams hard-coded fork params
+type ForkParams struct {
+	// MaxValidatorSetSize indicates the maximum size of validator set
+	MaxValidatorSetSize uint64 `json:"maxValidatorSetSize"`
+
+	EpochSize uint64 `json:"epochSize"`
+
+	// SprintSize is size of sprint
+	SprintSize uint64 `json:"sprintSize"`
+
+	// BlockTime is target frequency of blocks production
+	BlockTime common.Duration `json:"blockTime"`
+
+	// BlockTimeDrift defines the time slot in which a new block can be created
+	BlockTimeDrift uint64 `json:"blockTimeDrift"`
+}
+
 type Fork struct {
 	Block  uint64      `json:"block"`
-	Params interface{} `json:"params"`
+	Params *ForkParams `json:"params"`
 }
 
 func NewFork(n uint64) *Fork {
-	f := Fork{
+	return &Fork{
 		Block: n,
 	}
-
-	return &f
 }
 
 func (f Fork) Active(block uint64) bool {
--- a/chain/params.go
+++ b/chain/params.go
@@ -117,20 +117,25 @@ func (f *Forks) At(block uint64) ForksInTime {
 	}
 }
 
-type Fork uint64
+type Fork struct {
+	Block  uint64      `json:"block"`
+	Params interface{} `json:"params"`
+}
 
 func NewFork(n uint64) *Fork {
-	f := Fork(n)
+	f := Fork{
+		Block: n,
+	}
 
 	return &f
 }
 
 func (f Fork) Active(block uint64) bool {
-	return block >= uint64(f)
+	return block >= f.Block
 }
 
-func (f Fork) Int() *big.Int {
-	return big.NewInt(int64(f))
+func (f Fork) BigInt() *big.Int {
+	return new(big.Int).SetUint64(f.Block)
 }
 
 // ForksInTime should contain all supported forks by current edge version
--- a/chain/params.go
+++ b/chain/params.go
@@ -2,7 +2,6 @@ package chain
 
 import (
 	"errors"
-	"math/big"
 	"sort"
 
 	"github.com/0xPolygon/polygon-edge/helper/common"
@@ -89,18 +88,22 @@ const (
 )
 
 // Forks is map which contains all forks and their starting blocks from genesis
-type Forks map[string]*Fork
+type Forks map[string]Fork
 
 // IsActive returns true if fork defined by name exists and defined for the block
 func (f *Forks) IsActive(name string, block uint64) bool {
-	ff := (*f)[name]
+	ff, exists := (*f)[name]
 
-	return ff != nil && ff.Active(block)
+	return exists && ff.Active(block)
 }
 
 // SetFork adds/updates fork defined by name
-func (f *Forks) SetFork(name string, value *Fork) {
+func (f *Forks) SetFork(name string, value Fork) {
 	(*f)[name] = value
+}
+
+func (f *Forks) RemoveFork(name string) {
+	delete(*f, name)
 }
 
 // At returns ForksInTime instance that shows which supported forks are enabled for the block
@@ -123,6 +126,7 @@ type ForkParams struct {
 	// MaxValidatorSetSize indicates the maximum size of validator set
 	MaxValidatorSetSize *uint64 `json:"maxValidatorSetSize,omitempty"`
 
+	// EpochSize is size of epoch
 	EpochSize *uint64 `json:"epochSize,omitempty"`
 
 	// SprintSize is size of sprint
@@ -140,18 +144,12 @@ type Fork struct {
 	Params *ForkParams `json:"params,omitempty"`
 }
 
-func NewFork(n uint64) *Fork {
-	return &Fork{
-		Block: n,
-	}
+func NewFork(n uint64) Fork {
+	return Fork{Block: n}
 }
 
 func (f Fork) Active(block uint64) bool {
 	return block >= f.Block
-}
-
-func (f Fork) BigInt() *big.Int {
-	return new(big.Int).SetUint64(f.Block)
 }
 
 // ForksInTime should contain all supported forks by current edge version
--- a/chain/params.go
+++ b/chain/params.go
@@ -121,23 +121,23 @@ func (f *Forks) At(block uint64) ForksInTime {
 // ForkParams hard-coded fork params
 type ForkParams struct {
 	// MaxValidatorSetSize indicates the maximum size of validator set
-	MaxValidatorSetSize uint64 `json:"maxValidatorSetSize"`
+	MaxValidatorSetSize *uint64 `json:"maxValidatorSetSize,omitempty"`
 
-	EpochSize uint64 `json:"epochSize"`
+	EpochSize *uint64 `json:"epochSize,omitempty"`
 
 	// SprintSize is size of sprint
-	SprintSize uint64 `json:"sprintSize"`
+	SprintSize *uint64 `json:"sprintSize,omitempty"`
 
 	// BlockTime is target frequency of blocks production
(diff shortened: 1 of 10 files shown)


chain/params.go
```diff
@@ -85,6 +86,8 @@ const (
 	EIP150         = "EIP150"
 	EIP158         = "EIP158"
 	EIP155         = "EIP155"
+	Dummy1         = "Dummy1"
```
Nemanja0x: Remove these

chain/params.go
```diff
@@ -157,4 +183,6 @@ var AllForksEnabled = &Forks{
 	Petersburg:     NewFork(0),
 	Istanbul:       NewFork(0),
 	London:         NewFork(0),
+	Dummy1:         NewFork(0),
```
Nemanja0x: remove these

chain/params.go
igorcrevar: Fixed

chain/params.go
igorcrevar: Fixed

Nemanja0x (approved):

chain/params.go
```diff
@@ -117,20 +118,40 @@ func (f *Forks) At(block uint64) ForksInTime {
...
 
-type Fork uint64
+// ForkParams hard-coded fork params
+type ForkParams struct {
+	// MaxValidatorSetSize indicates the maximum size of validator set
+	MaxValidatorSetSize *uint64 `json:"maxValidatorSetSize,omitempty"`
 
-func NewFork(n uint64) *Fork {
-	f := Fork(n)
+	EpochSize *uint64 `json:"epochSize,omitempty"`
```
Stefan-Ethernal, line 130:
```suggestion
    // EpochSize is size of epoch
	EpochSize *uint64 `json:"epochSize,omitempty"`
```

forkmanager/fork_manager.go
```diff
@@ -218,12 +243,70 @@ func (fm *forkManager) removeHandler(handlerName HandlerDesc, blockNumber uint64
...
+	})
+
+	if index < len(fm.params) && fm.params[index].FromBlockNumber == blockNumber {
+		copy(fm.params[index:], fm.params[index+1:])
+		fm.params[len(fm.params)-1] = nil
+		fm.params = fm.params[:len(fm.params)-1]
+	}
+}
+
+func copyParams(dest, src *chain.ForkParams) {
```
Stefan-Ethernal, line 288: Maybe you could have made this generic function which would work with arbitrary objects and put it into common package (for example).

forkmanager/fork_manager.go
```diff
@@ -2,8 +2,11 @@ package forkmanager
...
 import (
 	"fmt"
+	"reflect"
 	"sort"
 	"sync"
+
+	"github.com/0xPolygon/polygon-edge/chain"
 )
 
 /*
```
Stefan-Ethernal: Let's maybe remove this comment, to me it does not have any value.

forkmanager/fork_manager.go
```diff
@@ -79,7 +84,7 @@ func (fm *forkManager) RegisterHandler(forkName string, handlerName HandlerDesc,
 }
 
 // ActivateFork activates fork from some block number
-// All handlers belong to this fork are also activated
+// All handlers and parameters belong to this fork are also activated
```
Stefan-Ethernal:
```suggestion
// All handlers and parameters belonging to this fork are also activated
```

forkmanager/fork.go
```diff
@@ -41,16 +53,20 @@ func ForkManagerInit(factory func(*chain.Forks) error, forks *chain.Forks) error
...
 	// Register forks
-	for name := range *forks {
+	for name, f := range *forks {
 		// check if fork is not supported by current edge version
 		if _, found := (*chain.AllForksEnabled)[name]; !found {
 			return fmt.Errorf("fork is not available: %s", name)
 		}
 
-		fm.RegisterFork(name)
+		if f != nil {
```
Stefan-Ethernal: When fork would be nil? Is it in case somebody manually deletes fork from the genesis?

consensus/polybft/polybft.go
```diff
@@ -435,6 +435,21 @@ func (p *Polybft) Initialize() error {
 	return nil
 }
 
+func ForkManagerInitialParamsFactory(config *chain.Chain) (*chain.ForkParams, error) {
+	pbftConfig, err := GetPolyBFTConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &chain.ForkParams{
```
Stefan-Ethernal, line 444: We have identified a couple of more parameteres when we were elaborating on-chain governance RFC: https://www.notion.so/polygontechnology/RFC-191-f40a27ea359b46f78f67e22c0627b0e6?pvs=4#a57804ec80b6433eaaecaa845b0c9507 (e.g. block-gas-limit/target and checkpoint-interval, which does not exist at the moment, but aim was to introduce it with on chain governance).

forkmanager/fork.go
```diff
@@ -18,21 +18,33 @@ type Fork struct {
 	Name string
 	// after the fork is activated, `FromBlockNumber` shows from which block is enabled
 	FromBlockNumber uint64
+	Params          *chain.ForkParams
```
Stefan-Ethernal: Nitpick: add some comment, even though it may be obvious what params stand for.

E.g.
```suggestion
    // fork consensus parameters
	Params          *chain.ForkParams
```

Stefan-Ethernal (approved):
I have a couple of questions:
- as I understand, this is just a temporary solution since we haven't implemented on chain governance yet. When and if we implement it, this feature would be redundant and we can remove it if necessary?
- we could change these parameters manually in the `genesis.json` and after server is restarted, those changes would take place. In that sense, it _may look_ redundant to it. Although convenience for forking support is that you can set the parameters and delay activation by specifying exact block and we are good. Anyways is forking mechanism going to override it if somebody chooses to change some value by hand?
- if I'm correct, this is only useful for us (Edge team) if we want to enfore some set of parameters along with their values, but not sure how it is going to be used by the end users? They would probably need to fork Edge repository and implement some fork on their own. Is that correct?

forkmanager/fork.go
igorcrevar: yup, exactly. but maybe we can make fork as value not pointer?

consensus/polybft/polybft.go
igorcrevar: we can easily add parameters later when we need them. personally i would leave only one or two in struct for now

forkmanager/fork.go
Stefan-Ethernal: Is it a valid use case in the first place? In case it isn't, maybe returning an error would be better. WDYT?

forkmanager/fork_manager.go
igorcrevar: this function is pretty specific so i think we should keep it here

forkmanager/fork.go
igorcrevar: Value inside forks map is now just Fork not *Fork. Fixed

chain/params.go
```diff
@@ -117,20 +122,41 @@ func (f *Forks) At(block uint64) ForksInTime {
...
 }
 
 func (f Fork) Active(block uint64) bool {
-	return block >= uint64(f)
+	return block >= f.Block
 }
 
-func (f Fork) Int() *big.Int {
-	return big.NewInt(int64(f))
+func (f Fork) BigInt() *big.Int {
```
goran-ethernal: Do we use this function anywhere?

goran-ethernal (approved):

chain/params.go
igorcrevar: removed

consensus/polybft/polybft.go
igorcrevar: We agree that we will add all  parameters needed through another pr

vcastellm:
> * as I understand, this is just a temporary solution since we haven't implemented on chain governance yet. When and if we implement it, this feature would be redundant and we can remove it if necessary?
> * we could change these parameters manually in the `genesis.json` and after server is restarted, those changes would take place. In that sense, it _may look_ redundant to it. Although convenience for forking support is that you can set the parameters and delay activation by specifying exact block and we are good. Anyways is forking mechanism going to override it if somebody chooses to change some value by hand?
> * if I'm correct, this is only useful for us (Edge team) if we want to enfore some set of parameters along with their values, but not sure how it is going to be used by the end users? They would probably need to fork Edge repository and implement some fork on their own. Is that correct?

Would be great to have answers for these questions @igorcrevar , how's that envisioned?

Nemanja0x:
> > * as I understand, this is just a temporary solution since we haven't implemented on chain governance yet. When and if we implement it, this feature would be redundant and we can remove it if necessary?

Correct.

> > * we could change these parameters manually in the `genesis.json` and after server is restarted, those changes would take place. In that sense, it _may look_ redundant to it. Although convenience for forking support is that you can set the parameters and delay activation by specifying exact block and we are good. Anyways is forking mechanism going to override it if somebody chooses to change some value by hand?

We need to add code (similar to handlers) which will determine what is the proper value of the parameter based on the block. 
For example, If epoch size is 50 before block 100 and 500 after block 1000 we need a handler to return a proper value based on the current block.


> > * if I'm correct, this is only useful for us (Edge team) if we want to enfore some set of parameters along with their values, but not sure how it is going to be used by the end users? They would probably need to fork Edge repository and implement some fork on their own. Is that correct?

Correct.

PR #1648, opened 2023-06-21
Deploy burn contract by default on EIP-1559 enabled

stana-miric:
# Description

## Context
This PR adds functionality of deploying default burn contract at genesis time if the native token is non mintable.
Burn contract related behavior is following:

### 1. Mintable native token
it is allowed to specify only `0x0` address for `--burn-contract` flag in this case. This is because, for mintable native token, default burn contract is not applicable (since tokens originate from the Supernets itself). However we are allowing transfers to `0x0` address (note that as of now, burning tokens would not imply in native token's total supply reduction).

### 2. Non-mintable native token
only one burn contract can be specified, and that address will be used to deploy the default contract. Also, destination address for default burn contract can be specified in --burn-contract flag and it will be used for initialization of the default contract. If destionation address is not set in --burn-contract flag, default burn contract will be initialized with zero address as a destination address.
If burn contract is not specified, EIP1559 feature will be disabled, london fork will be disabled, BaseFee and BaseFeeEM would not be set, tokens would not be burned.

Later, user can deploy some custom burn contract and update the burn contracts by:
1. Stoping the node
2. Changing genesis.json file by:
    2.1 setting the burnContracts using address of a custom deployed contract. Destination address is not needed here since it is only used when predeploying default burn contract at genesis time
    2.2 enabling london fork
3. Starting the node again 

# Changes include

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


command/genesis/genesis.go
```diff
@@ -74,7 +74,7 @@ func setFlags(cmd *cobra.Command) {
 		&params.burnContracts,
```
Stefan-Ethernal: Thinking out loud:
if genesis command is expected only once, when setting up Edge initially and not afterwards, is it worth considering changing flag type into ordinary string, instead of string array?

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1648/commits/b83d766bcc8c32a1ac15adfcdd82ded9a53c4902

goran-ethernal:
I have some questions here:

1. Now we can specify only one burn contract in `genesis` command, but you can deploy more burn contract contracts after you start the child chain, but you will have to change `genesis.json`? Am I right?
2. `BaseFeeEM` and other related EIP-1559 fields will not be set if you don't specify the burn contract, but when you deploy it, stop the node, and change `genesis.json` file, you will need to specify those fields as well (change them in json file)?

Stefan-Ethernal:
> I have some questions here:
> 
> 1. Now we can specify only one burn contract in `genesis` command, but you can deploy more burn contract contracts after you start the child chain, but you will have to change `genesis.json`? Am I right?

Correct.

> 2. `BaseFeeEM` and other related EIP-1559 fields will not be set if you don't specify the burn contract, but when you deploy it, stop the node, and change `genesis.json` file, you will need to specify those fields as well (change them in json file)?

Correct.

cc @DannyS03 for docs update.

consensus/polybft/polybft.go
```diff
@@ -360,6 +360,35 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 				contracts.NativeERC20TokenContract, input, "NativeERC20", transition); err != nil {
 				return err
 			}
+
+			// initialize EIP1559Burn SC
+			if config.Params.BurnContract != nil &&
+				len(config.Params.BurnContract) == 1 &&
+				!polyBFTConfig.NativeTokenConfig.IsMintable {
+				var contractAddress types.Address
+				for _, address := range config.Params.BurnContract {
```
vcastellm, line 369: Why do we need to maintain this as a map if there's only a single Burn contract?

Stefan-Ethernal: In absence of on chain governance, this is the way how one could configure different burn contracts in the runtime: by manually editing `genesis.json` and restarting the node(s).

vcastellm (approved):
LGTM

Nemanja0x (approved):

PR #1652, opened 2023-06-22
[EVM-703]: Saving events in boltdb should have retry mechanism

goran-ethernal:
# Description

Through this PR we implemented retry of inserting `bridge` events (`stateSync`, `exit`, and `transfer` events). Since errors on inserting events in `boltdb` can happen (db errors, disk errors, etc.), we needed to provide a way to get missed events (events that failed on `boltdb` insert). This is crucial, since if a `relayer` node fails inserting some `exit` or `stateSync` events, those deposits and withdrawals will never be executed.

This  PR implements `eventsGetter[T contractsapi.EventAbi]` struct, which gets desired events from transaction receipts by providing `from` and `to` block.

```
// eventsGetter is a struct for getting missed and current events
// of specified type from specified blocks
type eventsGetter[T contractsapi.EventAbi] struct {
	// blockchain is an abstraction of blockchain that provides necessary functions
	// for querying blockchain data (blocks, receipts, etc.)
	blockchain blockchainBackend
	// saveEventsFn is a plugin function used to return gotten events
	// and/or to save them to some db
	saveEventsFn func([]T) error
	// parseEventFn is a plugin function used to parse the event from transaction log
	parseEventFn func(*types.Header, *ethgo.Log) (T, bool, error)
	// isValidLogFn is a plugin function that validates the log
	// for example: if it was sent from the desired address
	isValidLogFn func(*types.Log) bool
}

// getFromBlocks gets events of specified type from specified blocks
// and saves them using the provided saveEventsFn
func (e *eventsGetter[T]) getFromBlocks(fromBlock, toBlock uint64) error 
```

`eventGetter` uses Go generics so it can be used on any event from edge `core-contracts`. It uses the plugin functions to have a more generic way of setting up events validation, events parsing and saving.

### Exit events retry
`checkpointManager` now uses `eventGetter[ExitEvent]` to get missed events. To know which block was last processed for exit events, a new bucket is added to `CheckpointStore`, called `lastProcessedBlock`. It is a bucket with single entry in it which indicates the last block that was processed for exit events. It is saved under a fixed key `lastProcessedBlockKey`.
On each block finalization by `consensus` or getting the block from `syncer`, `checkpointManager` will read the `lastSaved` bucket to know which block was last processed, and if it missed some blocks, it will use the `eventGetter` to gather all missed exit events and save them into `boltdb` through `eventGetter` plugin functions.

### Transfer events retry (for voting power changes)
`stakeManager` (which updates the voting power of validators on `childchain`, based on `Transfer` events from `ValidatorSet` contract, when their stake from `root` gets bridged to `child`), now uses `eventGetter[contractsapi.TransferEvent]` to get missed events. `StakeManagerStore` already had a mechanism in `boltdb` to know which block was last processed for `TransferEvents` so we just plugged in that logic to `eventGetter`. Same as in `checkpointManager`, on each block finalization by `consensuis` or getting the block from `syncer`, it will read the last processed block from its `boltdb` store, and will use `eventGetter` to collect any missed blocks so that it can update the voting power of validators correctly.

### State sync retry
Since `stateSync` events are events gotten from transactions on `rootchain`, we can not use the `eventGetter` for this, since it's limited to `childchain` use only. We get `stateSync` events from `evenTracker`, a third party library that syncs with `rootchain` and gets the desired logs. This is used on edge from the beginning of `polybft` consensus protocol, but it didn't propagate errors that might happen when client saves the gotten `stateSync` events to `db`. `AddLog` function now returns an error, and `eventTrackerStore` (used for storing `stateSyncEvents`), will propagate the error, and syncing with `rootchain` will retry, meaning that it will try to save the last log to client again.

# Changes include

# Checklist

## Testing


consensus/polybft/contractsapi/helper.go
```diff
@@ -13,6 +14,13 @@ type StateTransactionInput interface {
 	DecodeAbi(b []byte) error
 }
 
+// EventAbi is an interface representing an event generated in contractsapi
+type EventAbi interface {
+	Sig() ethgo.Hash
```
vcastellm, line 20: Nice to add a bit of description comments

consensus/polybft/checkpoint_manager.go
```diff
@@ -68,13 +68,53 @@ type checkpointManager struct {
...
+	// eventGetter gets exit events (missed or current) from blocks
+	eventGetter *eventsGetter[*ExitEvent]
 }
 
 // newCheckpointManager creates a new instance of checkpointManager
 func newCheckpointManager(key ethgo.Key, checkpointOffset uint64,
 	checkpointManagerSC types.Address, txRelayer txrelayer.TxRelayer,
 	blockchain blockchainBackend, backend polybftBackend, logger hclog.Logger,
 	state *State) *checkpointManager {
+	retry := &eventsGetter[*ExitEvent]{
```
vcastellm, line 80: Not sure if all this impl belongs to here...

goran-ethernal: It's a setup of the `eventGetter`, so it should be there when creating a new `checkpointManager`. Maybe I can extract these functions to seperate functions, and not be all inline like this, for readability.

vcastellm (approved):
Aside from the small comments LGTM

consensus/polybft/contractsapi/helper.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1652/commits/0121c9eb4c4f482b7730c59e59f6b19f65fbb340

consensus/polybft/checkpoint_manager.go
goran-ethernal: Moved that parsing logic to a separate function for improved readability:
 https://github.com/0xPolygon/polygon-edge/pull/1652/commits/e2abe2656c00ffd57be8f4a353e2d251676ba9a7

igorcrevar (approved):

consensus/polybft/state_event_getter.go
```diff
@@ -0,0 +1,93 @@
...
+	// parseEventFn is a plugin function used to parse the event from transaction log
+	parseEventFn func(*types.Header, *ethgo.Log) (T, bool, error)
+	// isValidLogFn is a plugin function that validates the log
+	// for example: if it was sent from the desired address
+	isValidLogFn func(*types.Log) bool
+}
+
+// getFromBlocks gets events of specified type from specified blocks
+// and saves them using the provided saveEventsFn
+func (e *eventsGetter[T]) getFromBlocks(lastProcessedBlock uint64, currentBlock *types.FullBlock) error {
```
igorcrevar: It would be better if you change signature of this method to
`func (e *eventsGetter[T]) getFromBlocks(
	lastProcessedBlock uint64, currentBlock *types.FullBlock) ([]T, error)`
and to remove `saveEventsFn func([]T) error` from the fields

goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1652/commits/fbe95a3aaa03547258e9b8825842079045f49b43

PR #1656, opened 2023-06-22
EVM-716 Check if the error is produced for duplicate TXs

dusan-maksimovic:
# Description

The PR includes the following changes:
-   Introduction of account level nonce to transaction lookup, along with the required methods: get, set, remove, and reset.
-   Change in the handling of enqueue requests (`addTx`). That process is now synchronous - first goes basic validation, then more complicated validations that require locks  and then all the changes in txpool structures
-   Modification of a unit test that validates the expected behavior when a cheaper transaction needs to be discarded. This change was necessary because transactions with the same nonce are no longer enqueued at all.
- Some additional unit tests

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -59,6 +59,8 @@ var (
 	ErrTipAboveFeeCap          = errors.New("max priority fee per gas higher than max fee per gas")
 	ErrTipVeryHigh             = errors.New("max priority fee per gas higher than 2^256-1")
 	ErrFeeCapVeryHigh          = errors.New("max fee per gas higher than 2^256-1")
+	ErrNonceExistsInPool       = errors.New("tx with the same nonce is already present")
+	ErrReplacementUnderpriced  = errors.New("replacement tx underpriced")
 )
 
 // indicates origin of a transaction
@@ -106,7 +108,8 @@ through their designated channels. */
 // This request is created for (new) transactions
 // that passed validation in addTx.
 type enqueueRequest struct {
-	tx *types.Transaction
+	tx      *types.Transaction
+	errchan chan<- error
 }
 
 // A promoteRequest is created each time some account
@@ -379,6 +382,8 @@ func (p *TxPool) Pop(tx *types.Transaction) {
 	// update metrics
 	p.updatePending(-1)
 
+	account.nonceToTx.remove(tx) // update the account nonce -> *tx map
+
 	// update executables
 	if tx := account.promoted.peek(); tx != nil {
 		p.executables.push(tx)
@@ -393,6 +398,7 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 
 	account.promoted.lock(true)
 	account.enqueued.lock(true)
+	account.nonceToTx.lock()
 
 	// num of all txs dropped
 	droppedCount := 0
@@ -409,6 +415,7 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 	defer func() {
 		account.enqueued.unlock()
 		account.promoted.unlock()
+		account.nonceToTx.unlock()
 	}()
 
 	// rollback nonce
@@ -425,6 +432,8 @@ func (p *TxPool) Drop(tx *types.Transaction) {
 	// drop enqueued
 	dropped = account.enqueued.clear()
 	clearAccountQueue(dropped)
+
+	account.nonceToTx.reset() // reset accounts nonce map
 
 	p.eventManager.signalEvent(proto.EventType_DROPPED, tx.Hash)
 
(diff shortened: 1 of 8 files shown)


txpool/txpool.go
```diff
@@ -59,6 +59,7 @@ var (
 	ErrTipAboveFeeCap          = errors.New("max priority fee per gas higher than max fee per gas")
 	ErrTipVeryHigh             = errors.New("max priority fee per gas higher than 2^256-1")
 	ErrFeeCapVeryHigh          = errors.New("max fee per gas higher than 2^256-1")
+	ErrNonceExists             = errors.New("transaction with nonce already exists in pool")
```
goran-ethernal: I think this error is not something `ethereum` has standardized and in their code. I know that some of the users (I think Nexon was the one), wanted these errors to be one on one with `ethereum`, and we even had a PR for that.

I know we don't have transaction replacement in `txpool` (like `ethereum`), but I am not sure if we want this issue done like this. Let's wait for product to decide.

dusan-maksimovic:
There is a problem with this approach, so please don't do the review just yet. Will let you know when is ready for review.

txpool/txpool.go
```diff
@@ -724,90 +739,111 @@ func (p *TxPool) pruneAccountsWithNonceHoles() {
...
-		if account := p.accounts.get(tx.From); account != nil &&
-			tx.Nonce > account.getNonce() {
+		if tx.Nonce > accountNonce {
 			metrics.IncrCounter([]string{txPoolMetrics, "rejected_future_tx"}, 1)
 
 			return ErrRejectFutureTx
 		}
 	}
 
+	// try to see if there is transaction with same nonce for this account
```
vcastellm:
```suggestion
	// try to find if there is transaction with same nonce for this account
```

txpool/txpool.go
```diff
@@ -724,90 +739,111 @@ func (p *TxPool) pruneAccountsWithNonceHoles() {
...
+		if oldTxWithSameNonce.Hash == tx.Hash {
+			metrics.IncrCounter([]string{txPoolMetrics, "already_known_tx"}, 1)
+
+			return ErrAlreadyKnown
+		} else if oldTxWithSameNonce.GasPrice.Cmp(tx.GasPrice) >= 0 {
+			// if tx with same nonce does exist and has same or better gas price -> return error
+			return ErrUnderpriced
+		}
+
+		slotsFree += slotsRequired(oldTxWithSameNonce) // add old tx slots
```
vcastellm, line 793: If we get to this point, means the new transaction is accepted and will replace previous one, then in line 806 we check `slotsFree` but containing old a new tx slots added, this is not correct because the new slots will substitute old slots, if I'm not wrong.

txpool/txpool.go
```diff
@@ -724,90 +739,111 @@ func (p *TxPool) pruneAccountsWithNonceHoles() {
...
 	// validate incoming tx
 	if err := p.validateTx(tx); err != nil {
 		return err
 	}
 
+	// calculate tx hash
+	tx.ComputeHash()
+
+	// initialize account for this address once or retrieve existing one
+	account, _ := p.createAccountOnce(tx.From)
```
vcastellm: Why adding the bool return value if we're ignoring it?

I see it's because of the testing, I don't think we should add a return param just because of testing, we should be able to do it other way, or basically do not test that property.

txpool/txpool.go
igorcrevar: we must increment available slots by number of slots occupied by old tx. thats why this line exists

igorcrevar: we still dont know at this point that new tx will be accepted because number of slots it used can be too high for example

txpool/txpool.go
igorcrevar: i think its good to know if the account is created or retrieved

txpool/txpool.go
vcastellm: Ah yes you're right I misread the line, we're reserving the space.

txpool/txpool.go
vcastellm: But we're not using it 🤔  why not remove the return value?

igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1656/commits/01883acad54a6b729cac8b6914df46a1f678a01a

txpool/txpool.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1656/commits/01883acad54a6b729cac8b6914df46a1f678a01a

txpool/txpool.go
```diff
@@ -724,90 +739,111 @@ func (p *TxPool) pruneAccountsWithNonceHoles() {
...
 		return err
 	}
 
+	// calculate tx hash
+	tx.ComputeHash()
+
+	// initialize account for this address once or retrieve existing one
+	account := p.getOrCreateAccount(tx.From)
+	// populate currently free slots
+	slotsFree := p.gauge.max - p.gauge.read()
```
goran-ethernal: Maybe create a function on gauge, and call it `freeSlots`?

igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1656/commits/f844092fedefa815589ae923a258730eb3dab090

stana-miric (approved):

goran-ethernal (approved):

vcastellm (approved):
LGTM good job!

PR #1667, opened 2023-06-26
Upgraded script to check both dependencies

criadoperez:
# Description

Upgraded cluster script to check both dependencies and display all missing ones before exiting.
It was checking for `jq` but not for `nc`.

# Changes include

- [x ] Bugfix (non-breaking change that solves an issue)
- [ x] New feature (non-breaking change that adds functionality)

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,4 +1,4 @@
-# Run servers form local binary
+# Run servers from local binary
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


Stefan-Ethernal (approved):

goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1669, opened 2023-06-26
[EVM-696] eth_FeeHistory

owenwahlgren:
# Description

Implementation of EIP1559 `eth_feeHistory` into the new `gasprice` package
Example call:
`curl http://localhost:10002/   -X POST   -H "Content-Type: application/json"   --data '{"method":"eth_feeHistory","params":[4, 5, [25, 75]],"id":1,"jsonrpc":"2.0"}'`

Response:
`{"jsonrpc":"2.0","id":1,"result":{"OldestBlock":"0x1","BaseFeePerGas":["0x34d670c0","0x2e3ba2a8","0x2e3ba2a8"],"GasUsedRatio":["0x0","0x0"],"Reward":[["0x0","0x0"],["0x0","0x0"]]}}`

# Changes include

# Checklist

## Testing

### DIFF
--- a/gasprice/feehistory.go
+++ b/gasprice/feehistory.go
@@ -42,7 +42,8 @@ type FeeHistoryReturn struct {
 	Reward        [][]uint64
 }
 
-func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (*FeeHistoryReturn, error) {
+func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (
+	*FeeHistoryReturn, error) {
 	if blockCount < 1 {
 		return &FeeHistoryReturn{0, nil, nil, nil}, ErrBlockCount
 	}
--- a/gasprice/feehistory.go
+++ b/gasprice/feehistory.go
@@ -95,7 +95,7 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 			continue
 		}
 
-		block, ok := g.backend.GetBlockByNumber(i, false)
+		block, ok := g.backend.GetBlockByNumber(i, true)
 		if !ok {
 			return nil, nil, nil, nil, ErrBlockNotFound
 		}
--- a/gasprice/feehistory.go
+++ b/gasprice/feehistory.go
@@ -35,16 +35,22 @@ type txGasAndReward struct {
 	reward  *big.Int
 }
 
-func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (*uint64, *[]uint64, *[]float64, *[][]uint64, error) {
+func (g *GasHelper) FeeHistory(
+	blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (
+	*uint64, *[]uint64, *[]float64, *[][]uint64, error) {
+
 	if blockCount < 1 {
 		return nil, nil, nil, nil, ErrBlockCount
 	}
+
 	if newestBlock > g.backend.Header().Number {
 		newestBlock = g.backend.Header().Number
 	}
+
 	if blockCount > maxBlockRequest {
 		blockCount = maxBlockRequest
 	}
+
 	if blockCount > newestBlock {
 		blockCount = newestBlock
 	}
@@ -53,6 +59,7 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 		if p < 0 || p > 100 {
 			return nil, nil, nil, nil, ErrInvalidPercentile
 		}
+
 		if i > 0 && p < rewardPercentiles[i-1] {
 			return nil, nil, nil, nil, ErrInvalidPercentile
 		}
@@ -64,16 +71,17 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 		gasUsedRatio  = make([]float64, blockCount)
 		reward        = make([][]uint64, blockCount)
 	)
+
 	if oldestBlock < 1 {
 		oldestBlock = 1
 	}
+
 	percentileKey := make([]byte, 8*len(rewardPercentiles))
 	for i, p := range rewardPercentiles {
 		binary.LittleEndian.PutUint64(percentileKey[i*8:(i+1)*8], math.Float64bits(p))
 	}
 
 	for i := oldestBlock; i <= newestBlock; i++ {
-
 		cacheKey := cacheKey{number: i, percentiles: string(percentileKey)}
 		//cache is hit, load from cache and continue to next block
 		if p, ok := g.historyCache.Get(cacheKey); ok {
@@ -81,11 +89,14 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 			if !isOk {
 				return nil, nil, nil, nil, errors.New("could not convert catched processed fee")
 			}
+
 			baseFeePerGas[i-oldestBlock] = processedFee.baseFee
 			gasUsedRatio[i-oldestBlock] = processedFee.gasUsedRatio
 			reward[i-oldestBlock] = processedFee.reward
+
 			continue
 		}
+
 		block, ok := g.backend.GetBlockByNumber(i, false)
 		if !ok {
 			return nil, nil, nil, nil, ErrBlockNotFound
@@ -98,6 +109,7 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 			//reward percentiles not requested, skip rest of this loop
 			continue
 		}
+
 		reward[i-oldestBlock] = make([]uint64, len(rewardPercentiles))
 		if len(block.Transactions) == 0 {
 			for j := range reward[i-oldestBlock] {
@@ -108,6 +120,7 @@ func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPerc
 		}
 
 		sorter := make([]*txGasAndReward, len(block.Transactions))
+
 		for j, tx := range block.Transactions {
 			cost := tx.Cost()
 			sorter[j] = &txGasAndReward{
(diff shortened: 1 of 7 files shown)


owenwahlgren:

gasprice/feehistory.go
```diff
@@ -0,0 +1,141 @@
...
+	"log"
+	"math"
+	"math/big"
+	"sort"
+)
+
+var (
+	ErrInvalidPercentile = errors.New("invalid percentile")
+	ErrBlockCount        = errors.New("blockCount must be greater than 0")
+	ErrBlockInfo         = errors.New("could not find block info")
```
goran-ethernal: To be more consistent with rest of the client:
 
```suggestion
	ErrBlockNotFound         = errors.New("could not find block")
```

gasprice/feehistory.go
```diff
@@ -0,0 +1,141 @@
...
+	if newestBlock > g.backend.Header().Number {
+		newestBlock = g.backend.Header().Number
+	}
+	if blockCount > maxBlockRequest {
+		blockCount = maxBlockRequest
+	}
+	if blockCount > newestBlock {
+		blockCount = newestBlock
+	}
+	for i, p := range rewardPercentiles {
```
goran-ethernal, line 63: Add new lines between the `if's` and `for's`.

gasprice/feehistory.go
```diff
@@ -0,0 +1,141 @@
...
+	)
+	if oldestBlock < 1 {
+		oldestBlock = 1
+	}
+	percentileKey := make([]byte, 8*len(rewardPercentiles))
+	for i, p := range rewardPercentiles {
+		binary.LittleEndian.PutUint64(percentileKey[i*8:(i+1)*8], math.Float64bits(p))
+	}
+
+	for i := oldestBlock; i < oldestBlock+blockCount; i++ {
```
goran-ethernal: I think we can just use `i <= newestBlock` instead of `i < oldestBlock+blockCount`.

gasprice/feehistory.go
```diff
@@ -0,0 +1,141 @@
...
+
+		cacheKey := cacheKey{number: i, percentiles: string(percentileKey)}
+		//cache is hit, load from cache and continue to next block
+		if p, ok := g.historyCache.Get(cacheKey); ok {
+			log.Println("cache hit", p)
```
goran-ethernal: Remove this log?

gasprice/feehistory.go
```diff
@@ -0,0 +1,141 @@
...
+			baseFeePerGas[i-oldestBlock] = p.(*processedFees).baseFee
```
goran-ethernal:
```suggestion
                        processedFee, isOk := p.(*processedFees)
                        if !isOk {
                              return errors.New("could not convert cached processed fee")
                        }
                        
			baseFeePerGas[i-oldestBlock] = processedFee.baseFee
			gasUsedRatio[i-oldestBlock] = processedFee.gasUsedRatio
			reward[i-oldestBlock] = processedFee.reward
			
			continue
```

gasprice/feehistory_test.go
```diff
@@ -0,0 +1,111 @@
...
+		args := b.Called(n, full)
+
+		return args.Get(0).(*types.Block), args.Get(1).(bool) //nolint:forcetypeassert
+	}
+
+	block, exists := b.blocks[Uint64ToHash(n)]
+
+	return block, exists
+}
+func Uint64ToHash(n uint64) types.Hash {
```
goran-ethernal:
```suggestion
func uint64ToHash(n uint64) types.Hash {
```

gasprice/gasprice.go
```diff
@@ -87,6 +92,11 @@ func NewGasHelper(config *Config, backend Blockchain) *GasHelper {
 		pricePercentile = 100
 	}
 
+	cache, err := lru.New(100)
+	if err != nil {
+		panic(err)
```
goran-ethernal: There is a `linter` rule that forbids using panics in code. Just return an error from the `NewGasHelper` function, and change the places where we call it to propagate the error.

goran-ethernal:
Left some comments. Please take a look at Lint errors it reported as well. It looks good.

vcastellm:
@owenwahlgren please check linting issues

gasprice/gasprice.go
```diff
@@ -53,6 +55,7 @@ type Blockchain interface {
 type GasStore interface {
 	// MaxPriorityFeePerGas calculates the priority fee needed for transaction to be included in a block
 	MaxPriorityFeePerGas() (*big.Int, error)
+	FeeHistory(blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (*uint64, *[]uint64, *[]float64, *[][]uint64, error)
```
vcastellm: For consistency, var names can be removed

gasprice/feehistory_test.go
```diff
@@ -0,0 +1,112 @@
...
+			NewestBlock:       30,
+			RewardPercentiles: []float64{15, 20},
+			GetBackend: func() Blockchain {
+				header := &types.Header{
+					Number: 0,
+					Hash:   types.StringToHash("some header"),
+				}
+				backend := new(backendMock)
+				backend.On("Header").Return(header)
+				backend.On("GetBlockByNumber", mock.Anything, true).Return(errors.New("block not found"))
```
goran-ethernal: I think you need to do this:

```suggestion
				backend.On("GetBlockByNumber", mock.Anything, mock.Anything).Return(nil, false)
```

owahlgren:
thanks for the help :)
my mock for GetBlockByNumber was incorrect hence the failing test :/
`--- PASS: TestGasHelper_FeeHistory (0.00s)
    --- PASS: TestGasHelper_FeeHistory/Block_does_not_exist (0.00s)
    --- PASS: TestGasHelper_FeeHistory/Invalid_rewardPercentile (0.03s)
    --- PASS: TestGasHelper_FeeHistory/Block_Range_<_1 (0.04s)
`

goran-ethernal (approved):

gasprice/gasprice.go
```diff
@@ -53,6 +55,7 @@ type Blockchain interface {
 type GasStore interface {
 	// MaxPriorityFeePerGas calculates the priority fee needed for transaction to be included in a block
 	MaxPriorityFeePerGas() (*big.Int, error)
+	FeeHistory(uint64, uint64, []float64) (*uint64, *[]uint64, *[]float64, *[][]uint64, error)
```
Stefan-Ethernal: Could you add a comment?

gasprice/feehistory.go
```diff
@@ -0,0 +1,160 @@
...
+	gasUsedRatio float64
+}
+
+type txGasAndReward struct {
+	gasUsed *big.Int
+	reward  *big.Int
+}
+
+func (g *GasHelper) FeeHistory(blockCount uint64, newestBlock uint64, rewardPercentiles []float64) (
+	*uint64, *[]uint64, *[]float64, *[][]uint64, error) {
```
Stefan-Ethernal: IMO, it is worth considering adding a `feeHistoryResult` instead of having multiple return values.

Also not sure why you would need pointers here? AFAIK common practice is that if error is not nil, other return values are not considered as valid.

jsonrpc/eth_endpoint.go
```diff
@@ -802,3 +802,78 @@ func (e *Eth) MaxPriorityFeePerGas() (interface{}, error) {
...
+}
+
+func convertFloat64SliceToArgUint64Slice(slice []float64) []argUint64 {
+	argSlice := make([]argUint64, len(slice))
+	for i, value := range slice {
+		argSlice[i] = argUint64(value)
+	}
+
+	return argSlice
+}
```
Stefan-Ethernal: I suppose it is better to move these helper functions into `jsonrpc/types.go`. Or maybe we could have even avoided it if `gasPrice.FeeHistory` was returning the correct types (argUint64, []argUint64 etc.)?

WDYT?

gasprice/feehistory_test.go
```diff
@@ -0,0 +1,107 @@
+package gasprice
+
+import (
+	"testing"
+
+	"github.com/0xPolygon/polygon-edge/types"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
```
Stefan-Ethernal, line 16: Could you add some happy path test(s) as well?

gasprice/feehistory.go
owahlgren: made mistake with the pointers :sweat_smile: 
I followed the geth spec where values are returned unpacked from `gasPrice.FeeHistory`
If you think returning `feeHistoryResult` from the package is better I can make that change :smile:

jsonrpc/eth_endpoint.go
owahlgren: moved to `jsonrpc/types.go` for time being.
I can either make `gasPrice.FeeHistory` return a `feeHistory` result or have it return values as `argUint64`s before packing and returning from `eth_endpoint.go`
Let me know what you think is better

gasprice/feehistory.go
Stefan-Ethernal: > If you think returning feeHistoryResult from the package is better I can make that change smile

Yep, I do think it is better that way because the caller is more aware of what each return value denotes (since it is encapsulated within the named field). It is especially convenient when there are more than 2-3 return values. Do you agree?

jsonrpc/eth_endpoint.go
Stefan-Ethernal: > I can either make gasPrice.FeeHistory return a feeHistory result or have it return values as argUint64s before packing and returning from eth_endpoint.go

Thinking out loud: if we are about to be consistent with other JSON RPC API functions out there, then we should probably keep the current approach (that is converting it to those derived types only after the algorithm is executed and just before returning a response). Therefore I'd say we are good with the current approach.

jsonrpc/types.go
```diff
@@ -344,3 +344,37 @@ type progression struct {
 	CurrentBlock  argUint64 `json:"currentBlock"`
 	HighestBlock  argUint64 `json:"highestBlock"`
 }
+
+type feeHistoryResult struct {
```
vcastellm, line 348: What's this type used for? I guess this should be used in place of `FeeHistoryReturn`

gasprice/feehistory.go
```diff
@@ -0,0 +1,171 @@
...
+	baseFee      uint64
+	gasUsedRatio float64
+}
+
+type txGasAndReward struct {
+	gasUsed *big.Int
+	reward  *big.Int
+}
+
+type FeeHistoryReturn struct {
```
vcastellm, line 38: Same properties as `feeHistoryResult` it can be discarded

vcastellm:
One linting issue pending

igorcrevar (approved):

igorcrevar:
LGTM, Just fix lint errors

PR #1676, opened 2023-06-28
Handle state sync execution failures properly

Stefan-Ethernal:
# Description

Failed state sync events execution is handled properly now. Instead of expecting only failed transaction receipt, we should attest whether there is `StateSyncResult` with the `Success` field set to `false`. Those state sync events are also considered as failed state sync events.

The reason behind it is that `StateReceiver` invokes a target contract (in this case `NativeERC20`), without having a require, but instead only marking it as failed and emitting appropriate `StateSyncResult`:
https://github.com/0xPolygon/core-contracts/blob/main/contracts/child/StateReceiver.sol#L165-L170

# Changes include

# Checklist

## Testing

### Manual tests

1. remove premining `0x0` address from cluster script and run it
2.  when the cluster it is up and running, send deposit transactions using the following command:
```go
go run . bridge deposit-erc20 --sender-key aa75e9a7d427efc732f8e4f1a5b7646adcc61fd5bae40f80d13c8419c9f43d6d \
	--amounts 1000000000000000000 \
	--receivers <receiver address on Supernets> \
	--root-token <native erc20 address retrieved from genesis.json> \
	--root-predicate <erc20 predicate address retrieved from genesis.json> \
	--json-rpc http://localhost:8545 \
	--minter-key aa75e9a7d427efc732f8e4f1a5b7646adcc61fd5bae40f80d13c8419c9f43d6d
```
3. notice that there is the following error displayed in the state sync relayer log:
"State sync execution failed. err=failed to execute state sync id..."

### DIFF
--- a/consensus/polybft/statesyncrelayer/state_sync_relayer.go
+++ b/consensus/polybft/statesyncrelayer/state_sync_relayer.go
@@ -138,7 +138,7 @@ func (r *StateSyncRelayer) AddLog(log *ethgo.Log) {
 		}
 
 		if err := r.executeStateSync(stateSyncProof); err != nil {
-			r.logger.Error("Failed to execute state sync", "err", err)
+			r.logger.Error("State sync execution failed", "err", err)
 
 			continue
 		}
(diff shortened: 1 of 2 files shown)


goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1678, opened 2023-06-28
Feat/script cluster

criadoperez:
# Description

* Fixed a bug checking dependencies. Missing a space on new line 26
* Added check if docker-compose is installed check `--docker` flag is used
* Added help instructions to use the script. Help is displayed when:
   * `--help` flag is on
   * when script is executed with no commands `./script/cluster`
   * a command has an incorrect value, like `./script/cluster xxx`

New help command display:
```bash
% ./scripts/cluster --help 
Usage: cluster {consensus} [{command}] [{flags}]
Consensus:
  ibft            Start Supernets test environment locally with ibft consensus
  polybft         Start Supernets test environment locally with polybft consensus
Commands:
  stop            Stop the running environment
  destroy         Destroy the running environment
Flags:
  --docker        Run using Docker (requires docker-compose).
  --help          Display this help information
Examples:
  cluster polybft -- Run the script with the polybft consensus
  cluster polybtt --docker -- Run the script with the polybft consensus using docker
  cluster polybft stop -- Stop the running environment
```

# Changes include


scripts/cluster
```diff
@@ -15,11 +15,36 @@ if [[ "$1" == "polybft" ]] && ! command -v nc >/dev/null 2>&1; then
...
+fi
+
+function showhelp(){
+  echo "Usage: cluster {consensus} [{command}] [{flags}]"
+  echo "Consensus:"
+  echo "  ibft            Start Supernets test environment locally with ibft consensus"
+  echo "  polybft         Start Supernets test environment locally with polybft consensus"
+  echo "Commands:"
+  echo "  stop            Stop the running environment"
+  echo "  destroy         Destroy the running environment"
```
Stefan-Ethernal, line 38: There is also a `write-logs` "command", although it should be probably converted to flag syntactically. What it does is that it redirects stdout and stderr to log files and it is applicable if running the script in non-docker mode. Worth documenting it here as well.

scripts/cluster
```diff
@@ -15,11 +15,36 @@ if [[ "$1" == "polybft" ]] && ! command -v nc >/dev/null 2>&1; then
...
+# Check if docker-compose is installed
+if [[ "$2" == "--docker" ]] && ! command -v docker-compose >/dev/null 2>&1; then
+  echo "docker-compose is not installed."
+  echo "Manual installation instructions: Visit https://docs.docker.com/compose/install/ for more information."
+  dp_error_flag=1
+fi
+
 # Stop script if any of the dependencies have failed
-if [[ "$dp_error_flag" -eq 1]]; then
+if [[ "$dp_error_flag" -eq 1 ]]; then
```
Stefan-Ethernal, line 26: Noticed this as well and fixed it through #1676 (though it is not merged yet). However, doesn't harm keeping this change here as well.

criadoperez: Cool. :-)

scripts/cluster
criadoperez: Good point, thanks. I added here the `write-logs` command.
I agree it would be better to convert it to a flag in the future, but until that's done here we can state how it currently works today.

Stefan-Ethernal (approved):

vcastellm (approved):

criadoperez:
Fixed the conflict. Ready to merge.

PR #1679, opened 2023-06-28
Docker requirements

criadoperez:
# Description
Updated docker requirements.

* Docker Desktop is not required. Docker engine is. 
* The minimum version of docker engine is no longer 17.12.0+ but 20.10.0. This is because the docker-compose file is using the format 3.9 that was introduced with docker 20.10.0
* Docker Compose requirement is version 1.29 not 2. Version 2 is no longer an independent binary and called with `docker compose`. With version 1.x of docker composer you call it with `docker-compose`. The scripts are currently using the older format of `docker-compose`.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,8 +1,8 @@
 # Deploying local docker cluster
 
 ## Prerequisites
-* [Docker Desktop](https://www.docker.com/products/docker-desktop/) - Docker 17.12.0+
-* [Docker compose 2+](https://github.com/docker/compose/releases/tag/v2.14.1)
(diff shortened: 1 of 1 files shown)


Stefan-Ethernal (approved):
Have you intended to adapt the cluster script to use `docker compose` instead of `docker-compose` through this or #1678 maybe?

Are any other changes needed?

criadoperez:
No,  PR #1678 doesn't touch that part, and keeps using docker-compose as it it now. I
The migration to Compose V2 should be a separate PR after #1678.

vcastellm (approved):

PR #1683, opened 2023-06-29
Fix/evm 719

rachit77:
# Description

This PR solved the `nil` pointer dereference issue for [PeerMap](https://github.com/0xPolygon/polygon-edge/blob/develop/syncer/peers.go#L27) 

Logs of the noticed behavior:
```
[WARN]  polygon.server.syncer.sync-peer-client: failed to get status from a peer, skip: id=16Uiu2HAmEKVGNXmVM6gMwL9XBLc1LmS1cCWMBdnwVvKJwmJegUSf err="rpc error: code = Unavailable desc = connection error: desc = \"error reading server preface: stream reset\""
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x15a5759]
```

Jira Issue:[Here](https://polygon.atlassian.net/browse/EVM-719)

# Changes include

# Checklist

## Testing

### DIFF
--- a/syncer/client.go
+++ b/syncer/client.go
@@ -161,6 +161,7 @@ func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
 			status, err := m.GetPeerStatus(peerID)
 			if err != nil {
 				m.logger.Warn("failed to get status from a peer, skip", "id", peerID, "err", err)
+
 				return //Skip appending nil status
 			}
 
--- a/syncer/client.go
+++ b/syncer/client.go
@@ -165,7 +165,9 @@ func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
 
 			syncPeersLock.Lock()
 
-			syncPeers = append(syncPeers, status)
+			if status != nil {
+				syncPeers = append(syncPeers, status)
+			}
 
 			syncPeersLock.Unlock()
 		}()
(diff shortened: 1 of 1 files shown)


syncer/client.go
```diff
@@ -161,6 +161,8 @@ func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
 			status, err := m.GetPeerStatus(peerID)
 			if err != nil {
 				m.logger.Warn("failed to get status from a peer, skip", "id", peerID, "err", err)
+
```
igorcrevar, line 164: What do you think about this solution? (one lock less) :D
```
func (m *syncPeerClient) GetConnectedPeerStatuses() []*NoForkPeer {
	var (
		ps        = m.network.Peers()
		syncPeers = make([]*NoForkPeer, len(ps))
		wg        sync.WaitGroup
	)

	for i, p := range ps {
		wg.Add(1)

		go func(peerID peer.ID, pointer **NoForkPeer) {
			defer wg.Done()

			status, err := m.GetPeerStatus(peerID)
			if err != nil {
				m.logger.Warn("failed to get status from a peer, skip", "id", peerID, "err", err)
			} else {
				*pointer = status
			}
		}(p.Info.ID, &syncPeers[i])
	}

	wg.Wait()

	return syncPeers
}
```

It just a suggestion, your solution is just fine.

But it would be good if you write a test case for this

rachit77: This will exhibit the same old behavior i.e. adding incorrect peer details in the `PeerMap`. Assuming the `GetPeerStatus` function returns an error for an `index i` then the `syncPeers` slice will contain the zero value of the slice data type at that specific `index i`.

igorcrevar: you are right, my mistake :)

igorcrevar: I'll approve, just make some test

igorcrevar (approved):

vcastellm (approved):
LGTM

PR #1685, opened 2023-06-29
Ensure premining reserve account (0x0 address)

Stefan-Ethernal:
# Description

This PR introduces a validation to the genesis command, which ensures that escrow account is premined with some tokens inside genesis. This is important for minting and burning native tokens.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/genesis/params.go
+++ b/command/genesis/params.go
@@ -61,7 +61,8 @@ var (
 	errInvalidEpochSize       = errors.New("epoch size must be greater than 1")
 	errInvalidTokenParams     = errors.New("native token params were not submitted in proper format " +
 		"(<name:symbol:decimals count:mintable flag:[mintable token owner address]>)")
-	errRewardWalletAmountZero = errors.New("reward wallet amount can not be zero or negative")
+	errRewardWalletAmountZero  = errors.New("reward wallet amount can not be zero or negative")
+	errEscrowAccMustBePremined = errors.New("it is mandatory to premine escrow account (0x0 address)")
 )
 
 type genesisParams struct {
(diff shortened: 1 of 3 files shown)


command/genesis/params.go
```diff
@@ -61,7 +61,8 @@ var (
 	errInvalidEpochSize       = errors.New("epoch size must be greater than 1")
 	errInvalidTokenParams     = errors.New("native token params were not submitted in proper format " +
 		"(<name:symbol:decimals count:mintable flag:[mintable token owner address]>)")
-	errRewardWalletAmountZero = errors.New("reward wallet amount can not be zero or negative")
+	errRewardWalletAmountZero  = errors.New("reward wallet amount can not be zero or negative")
+	errEscrowAccMustBePremined = errors.New("it is mandatory to premine escrow account (0x0 address)")
```
vcastellm: Why you calling it `escrow` account?

Stefan-Ethernal: Not sure how to call that account, so it reminded me of escrow account (some special account which holds tokens on its end until some conditions are met to distribute tokens to the other accounts).
Should we call it somehow other than 0x0 address? 🤔😊

vcastellm: IDK, sounds a bit confusing to me, that definition resembles to me as the owner of the tokens, more like a `reserve`, wdyt?

vcastellm (approved):
LGTM aside from terminology

command/genesis/params.go
goran-ethernal: `tokenReserver`, `reserve`, `treasury`, something like that.

stana-miric (approved):

goran-ethernal (approved):

igorcrevar (approved):

PR #1702, opened 2023-07-05
Fix TestE2E_Migration

goran-ethernal:
# Description

E2E test `TestE2E_Migration` was failing because of a bug introduced in https://github.com/0xPolygon/polygon-edge/pull/1685. Through given PR we moved parsing of `premine` flag in `genesis` command to `validatePremineInfos` function, but that function was only called if consensus is `polybft`.

Since given E2E test, tests the migration from an older consensus (in this case `dev` consensus) to `polybft`, premine flag was not parsed when starting cluster in `dev` consensus, so the test's checks failed.

This PR fixes the issue, by first parsing `premine` flag (without checking which consensus is selected), and then calling validate if `polybft` consensus is in place.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/genesis/params.go
+++ b/command/genesis/params.go
@@ -145,6 +145,10 @@ func (p *genesisParams) validateFlags() error {
 		return errValidatorsNotSpecified
 	}
 
+	if err := p.parsePremineInfo(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


vcastellm (approved):
LGTM

igorcrevar (approved):

PR #1705, opened 2023-07-06
Stop retrying event tracker on context cancellation and timeout

sergerad:
# Changes

* Check for context completion in retry forever loops (in case anonymous func doesn't stop on context done)
* Check for context completion in anonymous funcs passed into retry forever loops where we want to avoid printing an error log in such cases

# Background 

When we shutdown, the loops that retry forever treat context cancellation as a retryable error. Instead, these loops should exit.

We also log these as ERRORs when they should not be logged at all.

These are the relevant log lines:
```
message:2023-07-05T23:40:04.443Z [ERROR] polygon.server.polybft.state-sync-manager.event_tracker: failed to sync: error="context canceled"
```
and
```
message:2023-07-05T23:39:49.465Z [ERROR] polygon.server.polybft.state-sync-manager.event_tracker: failed to start blocktracker: error="context canceled"
```

# Local Tests

Running with fix, shutdown looks like:
```
[SIGNAL] Caught signal: terminated
Gracefully shutting down client...
...
```

Without fix (v1.0.1):
```
[SIGNAL] Caught signal: terminated
Gracefully shutting down client...
...
2023-07-06T21:34:05.747+1200 [ERROR] polygon.server.polybft.state-sync-manager.event_tracker: failed to start blocktracker: error="context canceled"
2023-07-06T21:34:05.747+1200 [ERROR] polygon.server.polybft.state-sync-manager.event_tracker: failed to sync: error="context canceled"

```

### DIFF
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -32,15 +32,25 @@ var (
 	errInvalidDuration = errors.New("invalid duration")
 )
 
-// RetryForever will execute a function until it completes without error
+// RetryForever will execute a function until it completes without error or
+// the context is cancelled or expired.
 func RetryForever(ctx context.Context, interval time.Duration, fn func(context.Context) error) {
 	_ = retry.Do(ctx, retry.NewConstant(interval), func(context.Context) error {
-		if err := fn(ctx); err != nil {
-			return retry.RetryableError(err)
+		// Execute function and end retries if no error or context done
+		err := fn(ctx)
+		if err == nil || IsContextDone(err) {
+			return nil
 		}
 
-		return nil
+		// Retry on all other errors
+		return retry.RetryableError(err)
 	})
+}
+
+// IsContextDone returns true if the error is due to the context being cancelled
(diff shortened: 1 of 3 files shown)


sergerad:
@Stefan-Ethernal could you please review? Fix is not critical but the false ERROR log is a bit annoying for ops. Thanks

igorcrevar (approved):

igorcrevar:
I have approved the pr. However, I have also found that currently we have a bug if `tt.Sync` fails, next time there will be panic because of closing of already closed channel. Please add  line (`tt.ReadyCh = make(chan struct{})`) in code before merge
```
// Sync concurrently, retrying indefinitely
go common.RetryForever(ctx, time.Second, func(ctx context.Context) error {
        tt.ReadyCh = make(chan struct{}) // re-initialize channel
       
             if common.IsContextDone(err) {
```

goran-ethernal (approved):

PR #1706, opened 2023-07-06
EVM-732 Tx hash calculation should include tx type and chainID

igorcrevar:
# Description

 Tx hash calculation should include tx type and correct chain id

# Changes include

# Breaking changes

Tx hash calculation and TxRootHash calculation

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -116,6 +116,10 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Receipts: transition.Receipts(),
 	})
 
+	if builtBlock.Header.TxRoot != block.Header.TxRoot {
+		return nil, fmt.Errorf("incorrect tx root: (%s, %s)", builtBlock.Header.TxRoot, block.Header.TxRoot)
+	}
+
 	return &types.FullBlock{
 		Block:    builtBlock,
 		Receipts: transition.Receipts(),
--- a/consensus/polybft/blockchain_wrapper.go
+++ b/consensus/polybft/blockchain_wrapper.go
@@ -117,7 +117,8 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 	})
 
 	if builtBlock.Header.TxRoot != block.Header.TxRoot {
-		return nil, fmt.Errorf("incorrect tx root: (%s, %s)", builtBlock.Header.TxRoot, block.Header.TxRoot)
+		return nil, fmt.Errorf("incorrect tx root (expected: %s, actual: %s)",
+			builtBlock.Header.TxRoot, block.Header.TxRoot)
 	}
 
 	return &types.FullBlock{
--- a/consensus/polybft/fsm.go
+++ b/consensus/polybft/fsm.go
@@ -232,7 +232,7 @@ func (f *fsm) createBridgeCommitmentTx() (*types.Transaction, error) {
 		return nil, fmt.Errorf("failed to encode input data for bridge commitment registration: %w", err)
 	}
 
-	return createStateTransactionWithData(f.parent.Number+1, contracts.StateReceiverContract, inputData), nil
+	return createStateTransactionWithData(f.Height(), contracts.StateReceiverContract, inputData), nil
 }
 
 // getValidatorsTransition applies delta to the current validators,
@@ -255,7 +255,7 @@ func (f *fsm) createCommitEpochTx() (*types.Transaction, error) {
 		return nil, err
 	}
 
-	return createStateTransactionWithData(f.parent.Number+1, contracts.ValidatorSetContract, input), nil
+	return createStateTransactionWithData(f.Height(), contracts.ValidatorSetContract, input), nil
 }
 
 // createDistributeRewardsTx create a StateTransaction, which invokes RewardPool smart contract
@@ -266,7 +266,7 @@ func (f *fsm) createDistributeRewardsTx() (*types.Transaction, error) {
 		return nil, err
 	}
 
-	return createStateTransactionWithData(f.parent.Number+1, contracts.RewardPoolContract, input), nil
+	return createStateTransactionWithData(f.Height(), contracts.RewardPoolContract, input), nil
 }
 
 // ValidateCommit is used to validate that a given commit is valid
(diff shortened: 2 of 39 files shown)


Nemanja0x (dismissed):

goran-ethernal (dismissed):

rachit77 (dismissed):

stana-miric (dismissed):

dusan-maksimovic (dismissed):

Nemanja0x:
There are some issues regarding backward compatibility. It must not be merged until we resolve them

vcastellm (approved):
LGTM

jsonrpc/eth_endpoint.go
```diff
@@ -349,7 +338,7 @@ func (e *Eth) GetTransactionReceipt(hash types.Hash) (interface{}, error) {
 			Data:        argBytes(elem.Data),
 			BlockHash:   block.Hash(),
 			BlockNumber: argUint64(block.Number()),
-			TxHash:      txn.Hash,
```
dusan-maksimovic, line 352: Why couldn't you keep using the txn.Hash? The transaction was previously found by hash, so it must be populated already. Am I missing something here?

dusan-maksimovic (approved):
LGTM, just have that one comment/question.

jsonrpc/eth_endpoint.go
igorcrevar: Yeah, it's same thing. `txn.Hash == hash`. This change is from before when I had two hashes (old and new one) in tx lookup table. I can revert this line if needed?

dusan-maksimovic: No need, just wanted to double check.

consensus/polybft/blockchain_wrapper.go
```diff
@@ -116,6 +116,10 @@ func (p *blockchainWrapper) ProcessBlock(parent *types.Header, block *types.Bloc
 		Receipts: transition.Receipts(),
 	})
 
+	if builtBlock.Header.TxRoot != block.Header.TxRoot {
+		return nil, fmt.Errorf("incorrect tx root: (%s, %s)", builtBlock.Header.TxRoot, block.Header.TxRoot)
```
Stefan-Ethernal:
```suggestion
		return nil, fmt.Errorf("incorrect tx root (expected: %s, actual: %s)", builtBlock.Header.TxRoot, block.Header.TxRoot)
```

consensus/polybft/fsm.go
```diff
@@ -232,7 +232,7 @@ func (f *fsm) createBridgeCommitmentTx() (*types.Transaction, error) {
 		return nil, fmt.Errorf("failed to encode input data for bridge commitment registration: %w", err)
 	}
 
-	return createStateTransactionWithData(contracts.StateReceiverContract, inputData), nil
+	return createStateTransactionWithData(f.parent.Number+1, contracts.StateReceiverContract, inputData), nil
```
Stefan-Ethernal: Optional nitpick (possible _slightly_ more readable)

```suggestion
	return createStateTransactionWithData(f.Height(), contracts.StateReceiverContract, inputData), nil
```

txpool/txpool.go
```diff
@@ -597,6 +604,14 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 			return ErrInvalidTxType
 		}
 
+		// DynamicFeeTx should be rejected if TxHashWithType fork is registered but not enabled for current block
+		blockNumber, err := forkmanager.GetInstance().GetForkBlock(chain.TxHashWithType)
```
Stefan-Ethernal, line 608: What if there is an error returned from `GetForkBlock`? Should we propagate it?

forkmanager/fork.go
```diff
@@ -19,7 +13,7 @@ type Fork struct {
 	// after the fork is activated, `FromBlockNumber` shows from which block is enabled
 	FromBlockNumber uint64
 	// fork consensus parameters
-	Params *chain.ForkParams
+	Params interface{}
```
Stefan-Ethernal: Why `chain.ForkParams` does not suit us?

server/server.go
```diff
@@ -1030,3 +1020,57 @@ func (s *Server) startPrometheusServer(listenAddr *net.TCPAddr) *http.Server {
...
+		params, err := factory(config)
+		if err != nil {
+			return err
+		}
+
+		initialParams = params
+	}
+
+	fm := forkmanager.GetInstance()
+	fm.Clear()
```
Stefan-Ethernal: Do you even need to call `Clear` function here? `GetInstance` is already doing so in case `forkManager` instance is `nil`. 

Also I would consider renaming it to `Init`, but as you wish.

txpool/txpool.go
igorcrevar: no, that means that fork is not register or not activated. In that case we should not do anything

forkmanager/fork.go
igorcrevar: circular dependencies. forkmanager should not depend on any package

consensus/polybft/blockchain_wrapper.go
igorcrevar: I have just copy pasted old error from blockchain.go. Do you want me to change there also?

server/server.go
igorcrevar: there is possibility that we need to (pre)init() some forks/handlers for testing purposes (and . This clear makes sure that those predefined things do not affect main app fork initialization process. But currently we do not have any init() method for that (we used to in some poc example) so i can remove it. wdyt?

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1706/commits/157ad5d10a9adf248f300edb8705d76d49f6a08c

consensus/polybft/blockchain_wrapper.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1706/commits/157ad5d10a9adf248f300edb8705d76d49f6a08c

forkmanager/fork.go
igorcrevar: I have placed params in forkmanager https://github.com/0xPolygon/polygon-edge/pull/1706/commits/f557e795ce279727937f755f9a0fac69a26838f5

Stefan-Ethernal (approved):

PR #1715, opened 2023-07-11
fix: renaming EIP 1559 fields to match spec and adding chainId

lolchocotaco:
# Description

# Changes include

# Breaking changes

Renames:
- gasFeeCap --> maxFeePerGas
- gasTipCap --> maxPriorityFeePerGas

Adds `chainId` to the jsonrpc transaction payload. 

# Checklist

## Testing

### Manual tests

Local environment is broken for tests


jsonrpc/types.go
jp-imx, line 1: Doesn't this need a matching change to `jsonrpc/testsuite/transaction-eip1559.json`?

begmaroman (approved):

jsonrpc/types.go
jp-imx: I think this is why `jsonrpc/types_test.go` is failing in the checks.

lolchocotaco: Oh yeah. my local dev env is messed up. Needs a bunch of linting/test changes i think

jsonrpc/types.go
```diff
@@ -34,6 +34,7 @@ type transaction struct {
 	BlockNumber *argUint64     `json:"blockNumber"`
 	TxIndex     *argUint64     `json:"transactionIndex"`
 	Type        argUint64      `json:"type"`
+	ChainId     *argUint64     `json:"chainId"`
```
jp-imx: This file is also missing a change to [toTransaction](https://github.com/0xPolygon/polygon-edge/blob/develop/jsonrpc/types.go#L54) to include the `chainId` in the JSON response.

Presently, trying to make the change is tricky because it requires a mapping from `*big.Int` to `*argUint64`.

Perhaps the type here should be changed to `argBig` so the conversion can become: `argBig(*t.ChainID)`

jp-imx:
Should the changes in this PR extend to `txpoolTransaction` defined in `jsonrpc/txpool_endpoint.go`?

jsonrpc/types.go
lolchocotaco: Im going to change the scope of this PR to only be focused on the renaming vs the addition of the other vars. the addition of ChainId will come with the other txn hash fixes.

PR #1723, opened 2023-07-14
[EVM-736]: Quorum calculation discrepancy between SC and Edge

goran-ethernal:
# Description

Quorum was not calculated correctly according to the IBFT paper. It was calculated by ceiling the 2*totalVotingPower/3. For example, if we have total voting power of 6, where each validator has one token staked, than:

```
CEILING(2*6/3) = 4 -> 2 faulty -> WRONG
FLOOR(2*6/3) + 1 = 5 -> 1 faulty -> CORRECT
```

Since Edge Supernets is in release, this is added through a hard fork, called `quorumcalcalignment`. If chain is deployed from scratch, this fork is enabled by default from 0 block. If supernet is already running on an older version, users need to add `quorumcalcalignment` fork to `genesis.json` file, specify a block from which it is active, and restart the nodes.

# Changes include

# Checklist

## Testing

### DIFF
--- a/chain/params.go
+++ b/chain/params.go
@@ -78,16 +78,16 @@ func (p *Params) GetEngine() string {
 
 // predefined forks
 const (
-	Homestead      = "homestead"
-	Byzantium      = "byzantium"
-	Constantinople = "constantinople"
-	Petersburg     = "petersburg"
-	Istanbul       = "istanbul"
-	London         = "london"
-	EIP150         = "EIP150"
-	EIP158         = "EIP158"
-	EIP155         = "EIP155"
-	NoviSad        = "novisad"
+	Homestead           = "homestead"
+	Byzantium           = "byzantium"
+	Constantinople      = "constantinople"
+	Petersburg          = "petersburg"
+	Istanbul            = "istanbul"
+	London              = "london"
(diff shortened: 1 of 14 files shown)


chain/params.go
```diff
@@ -87,6 +87,7 @@ const (
 	EIP150         = "EIP150"
 	EIP158         = "EIP158"
 	EIP155         = "EIP155"
+	NoviSad        = "novisad"
```
Stefan-Ethernal: WDYT about setting a more descriptive name for the hard fork? E.g. QuorumCalcAlignment (or if you have something better and shorter on your mind, even better)?

Stefan-Ethernal (approved):

dusan-maksimovic (approved):

chain/params.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1723/commits/802be00c72c670fb37b2327291f0868d261e5556

igorcrevar (approved):

goran-ethernal:
Tagging @DannyS03 to document the new fork on Edge that fixes the quorum calculation bug.

PR #1724, opened 2023-07-14
EVM-720 Implement json output for all commands

jelacamarko:
# Description

This PR includes setting JSON output for the rootchain deploy command. The result of the deploy contracts is returned and set using `outputter.SetCommandResult`

`WriteCommandResult` in `jsonOutput` is fixed to print JSON object.

Setting the `--json` flag on the command doesn't affect `OutputFormatter` `Write` method, and every output created using that method will be in the original format. 

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


command/rootchain/deploy/deploy.go
```diff
@@ -378,31 +378,33 @@ func runCommand(cmd *cobra.Command, _ []string) {
...
 			contractsDeploymentTitle),
 	})
+	outputter.SetCommandResult(command.Results(commandResults))
 }
 
 // deployContracts deploys and initializes rootchain smart contracts
 func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client, chainID int64,
-	initialValidators []*validator.GenesisValidator, cmdCtx context.Context) (*polybft.RootchainConfig, int64, error) {
+	initialValidators []*validator.GenesisValidator, cmdCtx context.Context) (
+	*polybft.RootchainConfig, int64, []command.CommandResult, error) {
```
Stefan-Ethernal: Can we introduce some struct which would have the following fields and return instance of that struct instead:
- `rootchainCfg`
- `chainID`
- `commandResults`.

You can call it `deploymentResultInfo` (or if you have some better name feel free to name it that way :wink:).

command/rootchain/deploy/deploy.go
```diff
@@ -378,31 +378,33 @@ func runCommand(cmd *cobra.Command, _ []string) {
 		return
 	}
 
-	outputter.SetCommandResult(&helper.MessageResult{
+	outputter.WriteCommandResult(&helper.MessageResult{
 		Message: fmt.Sprintf("%s finished. All contracts are successfully deployed and initialized.",
 			contractsDeploymentTitle),
 	})
+	outputter.SetCommandResult(command.Results(commandResults))
```
Stefan-Ethernal: Could we aggregate `MessageResult` together with `commandResults` (add it to the last place for example) and then just call `SetCommandResult` with commandResults as a parameter (in that case, we could remove `WriteCommandResult`).

command/rootchain/deploy/deploy.go
```diff
@@ -546,18 +549,18 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
 			}
```
Stefan-Ethernal, line 558: Be careful here, IMO we should add results into the `commandResults` slice. 

Also on line 542, we should probably change that to `MessageResult` and add it to `commandResults` slice as well, instead of printing it straight away to `outputter`.

Stefan-Ethernal (commented):
Leaving some feedback to consider, otherwise looking good.

command/rootchain/deploy/deploy.go
jelacamarko: We would need a new structure, with the message field. Maybe that message should be removed. If we get the response, that should mean that everything is successfully deployed.

Stefan-Ethernal: This code snippet should work just fine as well and in that case we can remove `outputter.WriteCommandResult`:

```go
deploymentResultInfo.CommandResults = append(deploymentResultInfo.CommandResults, &helper.MessageResult{
		Message: fmt.Sprintf("%s finished. All contracts are successfully deployed and initialized.",
			contractsDeploymentTitle),
	})
outputter.SetCommandResult(command.Results(deploymentResultInfo.CommandResults))
```

command/rootchain/deploy/deploy.go
```diff
@@ -544,18 +545,23 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
 	}
 
 	if err := g.Wait(); err != nil {
-		outputter.WriteCommandResult(&helper.MessageResult{
-			Message: "[ROOTCHAIN - DEPLOY] Successfully deployed the following contracts\n"})
+		results := []command.CommandResult{}
```
Stefan-Ethernal: Why do you need this slice, couldn't you have used `commandResults` instead?

command/rootchain/deploy/deploy.go
```diff
@@ -544,18 +545,23 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
...
 				// Filter those out.
-				outputter.WriteCommandResult(result)
+				commandResults[i] = result
 			}
 		}
 
-		return deploymentResultInfo{nil, 0, nil}, err
+		results = append(results, commandResults...)
+
+		return deploymentResultInfo{nil, 0, results}, err
```
Stefan-Ethernal: Can you add field names when setting them? Code readability is slightly better.

Stefan-Ethernal (approved):

command/rootchain/deploy/deploy.go
jelacamarko: I wanted to place a message at the beginning, but that can also be done without the results slice.

Stefan-Ethernal: Indeed 🙂

dusan-maksimovic (approved):

PR #1735, opened 2023-07-20
[RFC-191]: EVM-740 - Move distribute rewards to first block in epoch

goran-ethernal:
# Description

`ValidatorSet` contract on supernet is used as a wrapper around staked tokens by validators on `L2`. For governance, it is used as a voting token as well, which makes sense, since we are only allowing validators to propose and vote for governance proposals.

But, to be able to use `ValidatorSet` as vote token in governance, we needed to change it to implement `ERC20VotesUpgradable` implementation, which changed how the contract checkpoints its total supply at a certain point in time, and how it returns `totalSupplyAt`, which is used by `RewardPool` contract to accurately distribute rewards.
Now `totalSupplyAt` has a check which prevents calling it for the last checkpointed total supply, if they happen in the same block, and because of that, reward distribution failed.

This PR moves reward distribution to the first block of epoch. So when an epoch completes, and gets committed (which will checkpoint total supply at `ValidatorSet` contract), the new epoch will start, and in the first block of new epoch, proposer will add the distribute rewards transaction for the previously completed epoch.

This has three benefits:
1. It lowers gas consumption on epoch ending blocks.
2. It enables more accurate reward distribution.
3. We can use `ValidatorSet` as voting token for governance.

**IMPORTANT NOTE**:
Rewards will only be distributed in the first block of an epoch if `governance` fork is enabled. If it is not enabled, rewards will be distributed old way, at epoch ending block. 

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/consensus_runtime.go
+++ b/consensus/polybft/consensus_runtime.go
@@ -25,7 +25,7 @@ import (
 const (
 	maxCommitmentSize       = 10
 	stateFileName           = "consensusState.db"
-	commitEpochLookbackSize = 2 // number of blocks to calculate commit epoch info from the previous epoch
+	commitEpochLookbackSize = 1 // number of blocks to calculate commit epoch info from the previous epoch
 )
 
 var (
@@ -356,6 +356,7 @@ func (c *consensusRuntime) FSM() error {
 	// calculation of epoch and sprint end does not consider slashing currently
 	isEndOfSprint := c.isFixedSizeOfSprintMet(pendingBlockNumber, epoch)
 	isEndOfEpoch := c.isFixedSizeOfEpochMet(pendingBlockNumber, epoch)
+	isFirstBlockOfEpoch := pendingBlockNumber == epoch.FirstBlockInEpoch
 
 	valSet := validator.NewValidatorSet(epoch.Validators, c.logger)
 
@@ -365,18 +366,19 @@ func (c *consensusRuntime) FSM() error {
 	}
 
 	ff := &fsm{
-		config:            c.config.PolyBFTConfig,
-		parent:            parent,
-		backend:           c.config.blockchain,
-		polybftBackend:    c.config.polybftBackend,
-		exitEventRootHash: exitRootHash,
-		epochNumber:       epoch.Number,
-		blockBuilder:      blockBuilder,
-		validators:        valSet,
-		isEndOfEpoch:      isEndOfEpoch,
-		isEndOfSprint:     isEndOfSprint,
-		proposerSnapshot:  proposerSnapshot,
-		logger:            c.logger.Named("fsm"),
+		config:              c.config.PolyBFTConfig,
+		parent:              parent,
+		backend:             c.config.blockchain,
+		polybftBackend:      c.config.polybftBackend,
+		exitEventRootHash:   exitRootHash,
+		epochNumber:         epoch.Number,
+		blockBuilder:        blockBuilder,
+		validators:          valSet,
+		isEndOfEpoch:        isEndOfEpoch,
+		isEndOfSprint:       isEndOfSprint,
+		isFirstBlockOfEpoch: isFirstBlockOfEpoch,
+		proposerSnapshot:    proposerSnapshot,
+		logger:              c.logger.Named("fsm"),
 	}
 
 	if isEndOfSprint {
@@ -389,7 +391,7 @@ func (c *consensusRuntime) FSM() error {
 	}
 
 	if isEndOfEpoch {
-		ff.commitEpochInput, ff.distributeRewardsInput, err = c.calculateCommitEpochInput(parent, epoch)
+		ff.commitEpochInput = c.calculateCommitEpochInput(parent, epoch)
 		if err != nil {
 			return fmt.Errorf("cannot calculate commit epoch info: %w", err)
 		}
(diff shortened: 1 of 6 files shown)


consensus/polybft/consensus_runtime.go
```diff
@@ -490,66 +496,105 @@ func (c *consensusRuntime) restartEpoch(header *types.Header) (*epochMetadata, e
...
-	currentBlock *types.Header,
-	epoch *epochMetadata,
-) (*contractsapi.CommitEpochValidatorSetFn,
-	*contractsapi.DistributeRewardForRewardPoolFn, error) {
-	uptimeCounter := map[types.Address]int64{}
-	blockHeader := currentBlock
-	epochID := epoch.Number
-	totalBlocks := int64(0)
+// calculateCommitEpochInput calculates commit epoch input data
+func calculateCommitEpochInput(
```
igorcrevar: Maybe we should rename this helper function to something like `getCommitEpochInput` or similar and make that function outside of `consensusRuntime`. There is no calculation, so name is misleading

goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1735/commits/4ebb065d1692b278a0013f336380472289bb324d

consensus/polybft/governance_fork.go
```diff
@@ -0,0 +1,35 @@
...
+	oldRewardLookbackSize = uint64(2) // number of blocks to calculate commit epoch info from the previous epoch
+	newRewardLookbackSize = uint64(1)
+)
+
+// shouldAddOrValidateRewardDistribution indicates if reward distribution transaction
+// should be validated or added at given block
+// if governance fork is enabled, reward distribution is only present on the first block of epoch
+// and if we are not at the start of chain
+// if governance fork is not enabled, reward distribution is only present at the epoch ending block
+func shouldAddOrValidateRewardDistribution(isFirstBlockOfEpoch, isEndOfEpoch bool,
```
Stefan-Ethernal: WDYT?
```suggestion
func isRewardDistributionBlock(isFirstBlockOfEpoch, isEndOfEpoch bool,
```

consensus/polybft/consensus_runtime_test.go
```diff
@@ -389,11 +397,10 @@ func TestConsensusRuntime_FSM_EndOfEpoch_BuildCommitEpoch(t *testing.T) {
 	validatorAccounts := validator.NewTestValidatorsWithAliases(t, []string{"A", "B", "C", "D", "E", "F"})
 	validators := validatorAccounts.GetPublicIdentities()
 
-	lastBuiltBlock, headerMap := createTestBlocks(t, 9, epochSize, validators)
+	lastBuiltBlock := &types.Header{Number: 9}
```
Stefan-Ethernal: Nitpick, we can spare this line, by inlining it when instantiating `consensusRuntime` :wink:

consensus/polybft/fsm.go
```diff
@@ -36,11 +36,11 @@ var (
...
 	errDistributeRewardsTxDoesNotExist = errors.New("distribute rewards transaction is " +
-		"not found in the epoch ending block")
-	errDistributeRewardsTxNotExpected = errors.New("didn't expect distribute rewards transaction " +
-		"in a non epoch ending block")
+		"not found in the given block, but should be added")
+	errDistributeRewardsTxNotExpected = errors.New("distribute rewards transaction " +
+		"is not expected at this block")
 	errDistributeRewardsTxSingleExpected = errors.New("only one distribute rewards transaction is " +
-		"allowed in an epoch ending block")
+		"allowed at the expected block")
```
```suggestion
		"allowed in the given block")
```

consensus/polybft/fsm.go
```diff
@@ -36,11 +36,11 @@ var (
 	errCommitEpochTxSingleExpected = errors.New("only one commit epoch transaction is allowed " +
 		"in an epoch ending block")
 	errDistributeRewardsTxDoesNotExist = errors.New("distribute rewards transaction is " +
-		"not found in the epoch ending block")
-	errDistributeRewardsTxNotExpected = errors.New("didn't expect distribute rewards transaction " +
-		"in a non epoch ending block")
+		"not found in the given block, but should be added")
```
```suggestion
		"not found in the given block, though it is expected to be present")
```

Stefan-Ethernal (approved):
Overall looks good!

consensus/polybft/governance_fork.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1735/commits/7fa9afd2985856e7b552d15dc83aa367daa78411

consensus/polybft/fsm.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1735/commits/7fa9afd2985856e7b552d15dc83aa367daa78411

consensus/polybft/consensus_runtime_test.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1735/commits/7fa9afd2985856e7b552d15dc83aa367daa78411

consensus/polybft/fsm.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1735/commits/7fa9afd2985856e7b552d15dc83aa367daa78411

igorcrevar (approved):

PR #1741, opened 2023-07-24
EVM-746 Fix txpool_content and txpool_inspect for dynamic fee txs

igorcrevar:
# Description

`txpoolTransaction` structure has been deleted. `txpool_content` endpoint should use `transaction` from same package.
txpool_inspect should work for dynamic fee txs also

# Changes include

# Checklist

## Testing

### DIFF
--- a/jsonrpc/txpool_endpoint.go
+++ b/jsonrpc/txpool_endpoint.go
@@ -14,6 +14,9 @@ type txPoolStore interface {
 
 	// GetCapacity returns the current and max capacity of the pool in slots
 	GetCapacity() (uint64, uint64)
+
+	// GetBaseFee returns current base fee
+	GetBaseFee() uint64
 }
 
 // TxPool is the txpool jsonrpc endpoint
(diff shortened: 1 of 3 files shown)


vcastellm (approved):
LGTM

jsonrpc/txpool_endpoint.go
```diff
@@ -22,8 +25,8 @@ type TxPool struct {
 }
 
 type ContentResponse struct {
-	Pending map[types.Address]map[uint64]*txpoolTransaction `json:"pending"`
-	Queued  map[types.Address]map[uint64]*txpoolTransaction `json:"queued"`
+	Pending map[types.Address]map[uint64]*transaction `json:"pending"`
+	Queued  map[types.Address]map[uint64]*transaction `json:"queued"`
```
Stefan-Ethernal, line 29: Does it mind since transaction struct has a couple of more fields comparing to `txpoolTransation` previously had (
E.g. `V`, `R`, `S`, `Type`, `ChainID`)? Should we add `omitempty` tag maybe?

Stefan-Ethernal (approved):

jsonrpc/txpool_endpoint.go
igorcrevar: I have test some public geth nodes (like https://rpc-sepolia.rockx.com/). Most of public nodes do not allow txpool_content method, but the ones that do, they return all these fields.
Tested with postman 
```
{
  "jsonrpc": "2.0",
  "method": "txpool_content",
  "params": [
  ],
  "id": 1
}
```

PR #1748, opened 2023-07-25
Eip 2929

rachit77:
# Description

EIP-2929 implementation for Edge
Reference: https://eips.ethereum.org/EIPS/eip-2929

# Changes include

# Checklist

## Testing

### Manual tests

Manually verified the call stack for tests included


chain/params.go
```diff
@@ -89,6 +89,7 @@ const (
 	EIP155              = "EIP155"
 	QuorumCalcAlignment = "quorumcalcalignment"
 	TxHashWithType      = "txHashWithType"
+	EIP2929             = "EIP2929"
```
Stefan-Ethernal, line 92: Just curious, are we going to keep it this way: being able to enable EIPs independently, or to follow Ethereum hard fork namings? For example, EIP2929 is part of the Berlin hard fork (as well as 2718, 2565, and 2930, which we are working on as well).

state/runtime/precompiled/precompiled.go
```diff
@@ -36,6 +39,22 @@ func init() {
...
+		types.StringToAddress("4"),
+		types.StringToAddress("5"),
+		types.StringToAddress("6"),
+		types.StringToAddress("7"),
+		types.StringToAddress("8"),
+		types.StringToAddress("9"),
+		types.StringToAddress(contracts.NativeTransferPrecompile.String()),
+		types.StringToAddress(contracts.BLSAggSigsVerificationPrecompile.String()),
+	}
+}
```
Stefan-Ethernal: WDYT about merging these two init functions into a single?

state/runtime/evm/instructions.go
```diff
@@ -16,6 +16,12 @@ import (
 
 type instruction func(c *state)
 
+const (
+	ColdAccountAccessCostEIP2929 = uint64(2600) // COLD_ACCOUNT_ACCESS_COST
+	ColdSloadCostEIP2929         = uint64(2100) // COLD_SLOAD_COST
```
Stefan-Ethernal: Maybe rename it to `ColdStorageReadCostEIP2929` (just to be consistent with `WarmStorageReadCostEIP2929`). WDYT?

state/runtime/access_list.go
```diff
@@ -0,0 +1,77 @@
...
+// returns 'true' if the operation results in a change (i.e., the address was not already present in the list).
+func (al *AccessList) AddAddress(address types.Address) bool {
+	if _, exists := (*al)[address]; exists {
+		return false
+	}
+
+	(*al)[address] = make(map[types.Hash]struct{})
+
+	return true
+}
```
Stefan-Ethernal: WDYT (it would be easier to invoke it, no need for the for loops when `ActivePrecompiles` are being added for instance etc.).

Also I'm not sure do we need to return a boolean, since its return value is not used anywhere.
```suggestion
func (al *AccessList) AddAddresses(addresses ...types.Address) {
	for _, addr := range address {
		if _, exists := (*al)[addr]; exists {
			continue
		}

		(*al)[addr] = make(map[types.Hash]struct{})
	}
}
```

state/runtime/evm/instructions.go
```diff
@@ -658,7 +699,15 @@ func opExtCodeSize(c *state) {
...
 	var gas uint64
-	if c.config.EIP150 {
+
+	if c.config.EIP2929 {
+		if addressPresent := c.accessList.ContainsAddress(addr); addressPresent {
+			gas = WarmStorageReadCostEIP2929
+		} else {
+			gas = ColdAccountAccessCostEIP2929
+			c.accessList.AddAddress(addr)
+		}
```
Stefan-Ethernal: Extract this into function, because it is repeated a couple of times?

chain/params.go
rachit77: We will switch this to Berlin once all Ethereum Improvement Proposals (EIPs) related to Berlin have been fully implemented, adhering to the Ethereum hard fork naming convention.

state/runtime/evm/instructions.go
rachit77: The name used in the EIP-2929 document is `ColdAccountAccessCost`, but I am open to changing it to the name you suggested.

state/runtime/precompiled/precompiled.go
rachit77: To ensure a clean code structure, I separated these two into two different init functions.
Will merge these init into one.

Also do you think it would be better to take precompiles address from `Precompiled struct` for each transactions rather than initializing a new variable i.e `ActivePrecompiles`

state/runtime/evm/instructions_test.go
```diff
@@ -437,6 +437,7 @@ func TestCreate(t *testing.T) {
 	}
 }
 
+// EIP2929: check some test return error when triggered from base
```
Stefan-Ethernal: Could you add some unit tests that would test different gas consumptions per opcodes? I see that you have opted to test it out on a higher level of abstraction (namely on the EVM level), but I think it would be beneficial for us to have some "atomic" unit tests (namely the ones that would test gas consumption per specific op code, depending on which hard fork is enabled).
WDYT?

state/runtime/evm/instructions.go
Stefan-Ethernal: To me, it makes sense to make it consistent, and you have put in the comment how it is referred to in the EIP itself so I think we are good with such an approach.

state/runtime/precompiled/precompiled.go
Stefan-Ethernal: > Also do you think it would be better to take precompiles address from Precompiled struct for each transactions rather than initializing a new variable i.e ActivePrecompiles

Yep, I think it would be better. You could add a new slice called `contractAddrs []types.Address` to `Precompiled` struct and populate it lazily (every time the `register` function is being called). It is better that way than to have to worry about maintaining two data structures that have redundant data.

state/runtime/evm/instructions_test.go
rachit77: yeah that makes sense to me, I will add those unit test.

For these EVM level transactions I have also verified the cost of individual opcodes by printing call stack.

rachit77: added uint test cases for some opcodes and will add more with eip-2930 implementation

begmaroman (approved):

igorcrevar (approved):

PR #1749, opened 2023-07-25
[RFC-191]: EVM-558 - Update client configuration based on governance proposals

goran-ethernal:
# Description

This PR implements `governanceManager` on Edge client, which is a handler of governance proposal events.

To be more detailed, `governanceManager` does these things:
- on starting of node it will take the `genesis` client configuration and save it to `boltdb` if the chain is just started (or the node just started syncing from genesis block). If we are starting from some newer block (in cases when node was down some time), genesis configuration will already be saved to `boltdb`.
- on finalization of each block (either from consensus or syncer), iterates through transaction receipts to see if any governance proposal was executed in given block. If it was executed, it will save those events to `boltdb` under epoch in which proposal was executed.
- on starting of the new epoch, it will check if there are any found governance events from previous step that happened in the closing epoch. If we had any events we found and saved in `boltdb` it will iterate through them and update client configuration based on new values of changed network parameters through on chain governance. Updated client configuration will be saved to `boltdb`. This is done at the end of epoch, because we can not ensure stability if we start changing parameters at the end of epoch. For example, if we are to change epoch size in the middle of epoch, a lot of things like distributing rewards, and commiting an epoch will not work correctly. That is why, we will save executed proposal events to `boltdb` first, and then at the end of the epoch, update the client configuration based on them.
- provides a function to get the most current configuration to `consensus_runtime`, so that the client has the latest parameter values.

`consensus_runtime` is changed so that now each epoch has it's own client configuration, provided by the `governance_manager`. A new field is introduced to `epochMetadata` struct (which holds config and data from current epoch), called `CurrentClientConfig`, and it will use that field whenever some parameter from client configuration is needed in some part of code.

**IMPORTANT NOTE**: `governance_manager` will only collect on-chain governance proposal events if `governance` fork is enabled. If it is not enabled, then `genesis` client configuration will always be used.

# Changes include

# Checklist

## Testing


consensus/polybft/consensus_runtime.go
```diff
@@ -233,11 +246,28 @@ func (c *consensusRuntime) initStakeManager(logger hcf.Logger) error {
...
+		c.config.blockchain,
+	)
+
+	return nil
+}
+
+// initGovernanceManager initializes governance manager
+func (c *consensusRuntime) initGovernanceManager(logger hcf.Logger) error {
+	governanceManager, err := newGovernanceManager(
+		c.config.GenesisPolyBFTConfig, // this is config from genesis file
```
Stefan-Ethernal: Consider removing this comment and putting it next to the field itself (if you find it necessary, although name itself is already descriptive enough to me, so comment is redundant IMO).

consensus/polybft/consensus_runtime.go
```diff
@@ -345,7 +382,7 @@ func (c *consensusRuntime) FSM() error {
 		parent,
 		types.Address(c.config.Key.Address()),
 		c.config.txPool,
-		c.config.PolyBFTConfig.BlockTime.Duration,
+		epoch.CurrentClientConfig.BlockTime.Duration,
```
Stefan-Ethernal, line 385: Is it safe accessing `CurrentClientConfig` outside of lock? Probably it would be better if we added it to `guardedDataDTO` (btw the struct with most generic name ever).

consensus/polybft/governance_manager.go
```diff
@@ -0,0 +1,411 @@
...
+	default:
+		return nil, false, errUnknownGovernanceEvent
+	}
+}
+
+// unmarshalGovernanceEvent unmarshals given raw event to desired type
+func unmarshalGovernanceEvent[T contractsapi.EventAbi](rawEvent []byte) (T, error) {
+	var event T
+
+	err := json.Unmarshal(rawEvent[32:], &event)
```
Stefan-Ethernal: Extract 32 to const?

consensus/polybft/state_store_governance.go
```diff
@@ -0,0 +1,183 @@
...
+
+// initialize creates necessary buckets in DB if they don't already exist
+func (g *GovernanceStore) initialize(tx *bolt.Tx) error {
+	_, err := tx.CreateBucketIfNotExists(governanceEventsBucket)
+	if err != nil {
+		return fmt.Errorf("failed to create bucket=%s: %w", string(governanceEventsBucket), err)
+	}
+
+	_, err = tx.CreateBucketIfNotExists(clientConfigBucket)
+	if err != nil {
```
Stefan-Ethernal: Nitpick
```suggestion
	if _, err = tx.CreateBucketIfNotExists(clientConfigBucket); err != nil {
```

consensus/polybft/state_store_governance.go
```diff
@@ -0,0 +1,183 @@
...
+		return fmt.Errorf("failed to create bucket=%s: %w",
+			string(clientConfigBucket), err)
+	}
+
+	_, err = tx.CreateBucketIfNotExists(lastProcessedGovernanceEventsBucket)
+	if err != nil {
```
Stefan-Ethernal:
```suggestion
	if _, err = tx.CreateBucketIfNotExists(lastProcessedGovernanceEventsBucket); err != nil {
```

consensus/polybft/state_store_governance.go
```diff
@@ -0,0 +1,183 @@
...
+	errNoLastProcessedGovernanceBlock = errors.New("no last processed block for governance in db")
+	errClientConfigNotFound           = errors.New("client (polybft) config not found in db")
+)
+
+type eventsRaw [][]byte
+
+// Bolt db schema:
+//
+// governance events/
+// |--> epoch -> slice of contractsapi.EventAbi
```
Stefan-Ethernal, line 30: Add last processed governance events and client config to schema as well?

consensus/polybft/state_store_governance.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/0xPolygon/polygon-edge/helper/common"
+	bolt "go.etcd.io/bbolt"
+)
+
+var (
+	governanceEventsBucket              = []byte("governanceEvents")
+	clientConfigBucket                  = []byte("clientConfig")
+	clientConfigKey                     = []byte("clientConfigKey")
+	lastProcessedGovernanceEventsBucket = []byte("lastProcessedGovernanceEvents")
+	lastProcessedGovernanceBlockKey     = []byte("lastProcessedGovernanceBlockKey")
```
Stefan-Ethernal: I'd suggest to consolidate naming, because `lastProcessedGovernanceBlockKey` identifies `lastProcessedGovernanceEvents`. So I'd suggest to rename it `lastProcessedGovernanceBlock` and `lastProcessedGovernanceBlockKey` (or however you want, just for the clarity sake, pls add suffix `Key` to the bucket name).

consensus/polybft/governance_manager.go
```diff
@@ -0,0 +1,411 @@
...
+	eventsGetter := &eventsGetter[contractsapi.EventAbi]{
+		blockchain: blockhain,
+		isValidLogFn: func(l *types.Log) bool {
+			return l.Address == contracts.NetworkParamsContract
+		},
+		parseEventFn: parseGovernanceEvent,
+	}
+
+	config, err := state.GovernanceStore.getClientConfig()
+	if config == nil || errors.Is(err, errClientConfigNotFound) {
```
Stefan-Ethernal, line 94: What if `getClientConfig` returns an error? Should we propagate it from `newGovernanceManager` (in else if branch)?

consensus/polybft/consensus_runtime.go
goran-ethernal: But `epoch` from which we get the `CurrentClientConfig` is a part of `guardedDataDTO`.

consensus/polybft/consensus_runtime.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/governance_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/state_store_governance.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/state_store_governance.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/state_store_governance.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/governance_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1749/commits/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

consensus/polybft/consensus_runtime.go
igorcrevar: whole `func (c *consensusRuntime) OnBlockInserted(fullBlock *types.FullBlock)` is inside write lock `c.lock.Lock()`. That is way we do not nee to call getGuarded...

consensus/polybft/state_store_governance.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/commit/6bcffbbd344e5db00b863d22c1872ace2a23a9f6

Stefan-Ethernal (approved):

igorcrevar (approved):

PR #1756, opened 2023-07-27
Slashing double signing tracker

Stefan-Ethernal:
# Description

This PR implements a slashing double signing tracker. It contains an in-memory mapping of relayed IBFT messages for specific heights, rounds, and senders. Based on tracked messages it is able to build evidences for specific heights.

# Changes include

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/polybft/polybft.go
+++ b/consensus/polybft/polybft.go
@@ -50,11 +50,11 @@ func Factory(params *consensus.Params) (consensus.Consensus, error) {
 	setupHeaderHashFunc()
 
 	polybft := &Polybft{
-		config:      params,
-		closeCh:     make(chan struct{}),
-		logger:      logger,
-		txPool:      params.TxPool,
-		msgHandlers: []IBFTMessageHandler{},
+		config:          params,
+		closeCh:         make(chan struct{}),
+		logger:          logger,
+		txPool:          params.TxPool,
+		ibftMsgHandlers: []IBFTMessageHandler{},
 	}
 
 	// initialize polybft consensus config
@@ -120,8 +120,8 @@ type Polybft struct {
 	// tx pool as interface
 	txPool txPoolInterface
 
-	// msgHandlers contains IBFT consensus messages handlers
-	msgHandlers []IBFTMessageHandler
+	// ibftMsgHandlers contains IBFT consensus messages handlers
+	ibftMsgHandlers []IBFTMessageHandler
 }
 
 func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *state.Transition) error {
@@ -471,7 +471,7 @@ func (p *Polybft) Initialize() error {
 
 	p.ibft = newIBFTConsensusWrapper(p.logger, p.runtime, p)
 	// register IBFTConsensusWrapper as IBFT message handler
-	p.msgHandlers = append(p.msgHandlers, p.ibft)
+	p.ibftMsgHandlers = append(p.ibftMsgHandlers, p.ibft)
 
 	if err = p.subscribeToIbftTopic(); err != nil {
 		return fmt.Errorf("IBFT topic subscription failed: %w", err)
@@ -552,7 +552,7 @@ func (p *Polybft) initRuntime() error {
 
 	p.runtime = runtime
 	// register double signing tracker as IBFT messages handler
-	p.msgHandlers = append(p.msgHandlers, runtime.doubleSigningTracker)
+	p.ibftMsgHandlers = append(p.ibftMsgHandlers, runtime.doubleSigningTracker)
 
 	return nil
 }
--- a/consensus/polybft/polybft.go
+++ b/consensus/polybft/polybft.go
@@ -14,6 +14,7 @@ import (
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
+	"github.com/0xPolygon/polygon-edge/consensus/polybft/slashing"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/validator"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/wallet"
 	"github.com/0xPolygon/polygon-edge/contracts"
@@ -50,10 +51,11 @@ func Factory(params *consensus.Params) (consensus.Consensus, error) {
 	setupHeaderHashFunc()
 
 	polybft := &Polybft{
-		config:  params,
-		closeCh: make(chan struct{}),
-		logger:  logger,
-		txPool:  params.TxPool,
+		config:      params,
+		closeCh:     make(chan struct{}),
+		logger:      logger,
+		txPool:      params.TxPool,
+		msgHandlers: []IBFTMessageHandler{slashing.NewDoubleSigningTracker()},
 	}
 
 	// initialize polybft consensus config
@@ -118,6 +120,9 @@ type Polybft struct {
 
 	// tx pool as interface
 	txPool txPoolInterface
+
+	// msgHandlers contains IBFT consensus messages handlers
+	msgHandlers []IBFTMessageHandler
 }
 
 func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *state.Transition) error {
@@ -466,6 +471,8 @@ func (p *Polybft) Initialize() error {
 	}
 
 	p.ibft = newIBFTConsensusWrapper(p.logger, p.runtime, p)
+	// register IBFTConsensusWrapper as IBFT message handler
+	p.msgHandlers = append(p.msgHandlers, p.ibft)
 
 	if err = p.subscribeToIbftTopic(); err != nil {
 		return fmt.Errorf("IBFT topic subscription failed: %w", err)
--- a/consensus/polybft/polybft.go
+++ b/consensus/polybft/polybft.go
@@ -14,7 +14,6 @@ import (
 	"github.com/0xPolygon/polygon-edge/consensus"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	bls "github.com/0xPolygon/polygon-edge/consensus/polybft/signer"
-	"github.com/0xPolygon/polygon-edge/consensus/polybft/slashing"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/validator"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/wallet"
 	"github.com/0xPolygon/polygon-edge/contracts"
@@ -55,7 +54,7 @@ func Factory(params *consensus.Params) (consensus.Consensus, error) {
 		closeCh:     make(chan struct{}),
 		logger:      logger,
 		txPool:      params.TxPool,
-		msgHandlers: []IBFTMessageHandler{slashing.NewDoubleSigningTracker()},
+		msgHandlers: []IBFTMessageHandler{},
 	}
 
 	// initialize polybft consensus config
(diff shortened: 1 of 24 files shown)


consensus/polybft/polybft.go
```diff
@@ -118,6 +119,9 @@ type Polybft struct {
 
 	// tx pool as interface
 	txPool txPoolInterface
+
+	// msgHandlers contains IBFT consensus messages handlers
+	msgHandlers []IBFTMessageHandler
```
dusan-maksimovic: Maybe rename to `ibftMsgHandlers`?

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	sender := types.BytesToAddress(msg.From)
+	// ignore messages where signer and sender are not the same
+	if signer != sender {
+		t.logger.Warn("signer and sender of IBFT message are not the same, ignoring it...",
+			"message", msg)
+
+		return
+	}
```
dusan-maksimovic: We should also check if the sender is an active validator.

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,189 @@
...
+		commit:      make(MessagesMap),
+		roundChange: make(MessagesMap),
+	}
+}
+
+// Handle is implementation of IBFTMessageHandler interface, which handles IBFT consensus messages
+func (t *DoubleSigningTrackerImpl) Handle(msg *ibftProto.Message) {
+	signer, err := wallet.RecoverSignerFromIBFTMessage(msg)
+	if err != nil {
+		t.logger.Warn("failed to recover signer upon message receive",
```
dusan-maksimovic: Maybe we should log this on Debug or Info level? Since the error is not that important for system functioning. Same goes for line 100.

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,189 @@
...
+		preprepare:  make(MessagesMap),
+		prepare:     make(MessagesMap),
+		commit:      make(MessagesMap),
+		roundChange: make(MessagesMap),
+	}
+}
+
+// Handle is implementation of IBFTMessageHandler interface, which handles IBFT consensus messages
+func (t *DoubleSigningTrackerImpl) Handle(msg *ibftProto.Message) {
+	signer, err := wallet.RecoverSignerFromIBFTMessage(msg)
```
dusan-maksimovic: Add msg validation here, e.g. if msg.View is nil, msg.Type is one of valid values, and maybe other checks if we need any.

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/9699327c17a940f08b201d0d0408cab6879d50f4

consensus/polybft/polybft.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/9699327c17a940f08b201d0d0408cab6879d50f4

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: This is still to be discussed since a validator can have a stake greater than 0, but it could happen that it is not elected for the current epoch validator (sub)set if the max validator set size is already reached. Also we might want to slash non-active validator if it is a double-signer anyways, no matter it is not in the active validator set.

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/c993edbb0702b37452327f40a60d9b63f47c3801

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,218 @@
...
+
+// PruneMsgsUntil deletes all messages maps until the specified height
+func (t *DoubleSigningTrackerImpl) PruneMsgsUntil(height uint64) {
+	for _, msgType := range ibftProto.MessageType_value {
+		msgs := t.resolveMessagesStorage(ibftProto.MessageType(msgType))
+		if msgs == nil {
+			continue
+		}
+
+		for msgHeight := range msgs.content {
```
dusan-maksimovic, line 161: Should we also acquire msgs.mux here?

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,218 @@
...
+	return &DoubleSignEvidence{signer: signer, round: round, messages: messages}
+}
+
+type Messages struct {
+	content MessagesMap
+	mux     sync.RWMutex
+}
+
+// getSenderMsgsLocked returns commit messages for given height, round and sender
+func (m *Messages) getSenderMsgsLocked(view *ibftProto.View, sender types.Address) []*ibftProto.Message {
```
stana-miric: missing lock?

dusan-maksimovic: Caught my eye as well. He locks in the outside method. Should we just rename this method?
Also, the comment says "commit messages", "commit" is now redundant.

Stefan-Ethernal: In fact `Locked` suffix implies that the function should be called from a context that already holds a lock. Any suggestion on a better name or how to refactor it?

> Also, the comment says "commit messages", "commit" is now redundant.

Indeed, I will fix it.

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,218 @@
...
+
+				firstSig := msgs[0].Signature
+				for _, msg := range msgs[1:] {
+					if !bytes.Equal(firstSig, msg.Signature) {
+						if evidence == nil {
+							evidence = newDoubleSignEvidence(address, round, []*ibftProto.Message{})
+							evidences = append(evidences, evidence)
+						}
+
+						evidence.messages = append(evidence.messages, msg)
```
stana-miric: we should add the first message as well

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,218 @@
...
+		for round, senderMsgs := range roundMsgs {
+			for address, msgs := range senderMsgs {
+				if len(msgs) <= 1 {
+					continue
+				}
+
+				firstSig := msgs[0].Signature
+				for _, msg := range msgs[1:] {
+					if !bytes.Equal(firstSig, msg.Signature) {
+						if evidence == nil {
```
dusan-maksimovic: This seems wrong. If we have two validators that double signed on the same height and round, wouldn't it accumulate messages of the second validator to the ones that were already collected for the first one? Since `evidence` is nil only at the beginning of the execution.

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	sender := types.BytesToAddress(msg.From)
+
+	msgMap := t.resolveMessagesStorage(msg.GetType())
+	msgMap.addMessage(msg.View, sender, msg)
+}
+
+// PruneMsgsUntil deletes all messages maps until the specified height
+func (t *DoubleSigningTrackerImpl) PruneMsgsUntil(height uint64) {
```
stana-miric, line 152: we should review this method again once we agree upon final design. if it will be possible for user to submit evidence for random height that we will prune only that height. if it is requirement that slashing is done for consecutive blocks than it is safe to prune until the specified block without risk of pruning block that is not slashed

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Yep, figured it out during implementation of unit tests as well.

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Good catch. Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/f34d131a162e6fe45a6bcdbf5d0ebbf1b96ce58f

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fix https://github.com/0xPolygon/polygon-edge/pull/1756/commits/f34d131a162e6fe45a6bcdbf5d0ebbf1b96ce58f

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: As agreed, we are going to leave it as it is at the moment and we can update it once we agree on all the design decisions.

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed comment content https://github.com/0xPolygon/polygon-edge/pull/1756/commits/77a254cbcfdde9f21c4e3e81203fab8aa2a02215

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/f34d131a162e6fe45a6bcdbf5d0ebbf1b96ce58f

consensus/polybft/slashing/double_signing_tracker.go
```diff
@@ -0,0 +1,221 @@
...
+						if evidence == nil {
+							evidence = newDoubleSignEvidence(address, round, []*ibftProto.Message{})
+							evidences = append(evidences, evidence)
+						}
+
+						evidence.messages = append(evidence.messages, msg)
+					}
+				}
+
+				evidence.messages = append([]*ibftProto.Message{firstMsg}, evidence.messages...)
```
dusan-maksimovic: Evidence could be nil here if some byzantine validator submitted the same message and signature two or more times.
I think we should also filter such messages in Handle() to protect from possible spam attacks that could (if someone is persistent) crash the node when it runs out of memory. If we filter them in Handle() then we can move evidence instantiation before the inner most loop. Does it make sense to you?

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed by #1767.

consensus/polybft/slashing/double_signing_tracker.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/67879095f8b35a6828e5210a417f012af4af6949

dusan-maksimovic (approved):

consensus/polybft/state_store_stake.go
```diff
@@ -56,3 +57,12 @@ func (s *StakeStore) getFullValidatorSet() (validatorSetState, error) {
 
 	return fullValidatorSet, err
 }
+
+func (s *StakeStore) GetValidators() (validator.AccountSet, error) {
```
goran-ethernal:
```suggestion
// GetAllValidators returns all validators regardless if they are active or not
func (s *StakeStore) GetAllValidators() (validator.AccountSet, error) {
```

Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/1756/commits/cdfe58d65ce641ae94cb38566f864d265a8b0eb2

stana-miric (approved):

PR #1765, opened 2023-08-01
Move `Encode*` and `Parse*` functions from `types` to `common` package

Stefan-Ethernal:
# Description

Moves `types.Parse*` (e.g. `ParseUint64OrHex` etc.) and `types.Encode*` (e.g. `EncodeUint64` etc.) to the `common` package in order to remove redundant `ParseUint64OrHex` function.

# Changes include

# Checklist

## Testing

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -132,13 +132,13 @@ func (g *Genesis) MarshalJSON() ([]byte, error) {
 	var enc Genesis
 	enc.Nonce = hex.EncodeToHex(g.Nonce[:])
 
-	enc.Timestamp = types.EncodeUint64(g.Timestamp)
-	enc.ExtraData = types.EncodeBytes(g.ExtraData)
+	enc.Timestamp = common.EncodeUint64(g.Timestamp)
+	enc.ExtraData = common.EncodeBytes(g.ExtraData)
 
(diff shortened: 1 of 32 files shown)


vcastellm (approved):
🔥

Nemanja0x:
`types` package keeps additional functions: `ParseUint256orHex`, `ParseBytes`, ...  Isn't it better to keep `ParseUint64orHex` inside `types` and remove it from `common`?   Otherwise, `ParseUint256orHex` will be a part of `types` and `ParseUint64orHex` will be a part of `common`

Stefan-Ethernal:
> `types` package keeps additional functions: `ParseUint256orHex`, `ParseBytes`, ... Isn't it better to keep `ParseUint64orHex` inside `types` and remove it from `common`? Otherwise, `ParseUint256orHex` will be a part of `types` and `ParseUint64orHex` will be a part of `common`

Yes it would be better, however it is not possible because of this function:
https://github.com/0xPolygon/polygon-edge/blob/develop/helper/common/common.go#L83-L97 (it uses `ParseUint64OrHex` and it is not possible to move it to `types` package because of circular imports).


If you want, we could move all the other `Parse*` functions from the `types` to the `common` package. WDYT?

begmaroman (approved):

PR #1776, opened 2023-08-03
Allow 0 premined reward wallet

sergerad:
# Description

Update logic that prevents 0 premine for reward wallet. Allow 0 premine if epoch rewards are also 0.

# Changes include

# Checklist

## Testing

### Manual tests

Ran local cluster against Sepolia rootchain. Verified that bridging works. Verified that legacy and EIP1559 L2 functionality works.

### DIFF
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -98,7 +98,7 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
 		rewardTokenAddr     = contracts.NativeERC20TokenContract
 	)
 
-	if p.rewardTokenCode == "" {
+	if p.rewardTokenCode == "" && p.epochReward > 0 {
 		// native token is used as a reward token, and reward wallet is not a zero address
 		// so we need to add that address to premine map
 		premineBalances[walletPremineInfo.address] = walletPremineInfo
--- a/command/genesis/polybft_params.go
+++ b/command/genesis/polybft_params.go
@@ -98,10 +98,12 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
 		rewardTokenAddr     = contracts.NativeERC20TokenContract
 	)
 
-	if p.rewardTokenCode == "" && p.epochReward > 0 {
+	if p.rewardTokenCode == "" {
 		// native token is used as a reward token, and reward wallet is not a zero address
-		// so we need to add that address to premine map
-		premineBalances[walletPremineInfo.address] = walletPremineInfo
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal (approved):

command/genesis/polybft_params.go
```diff
@@ -98,7 +98,7 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
 		rewardTokenAddr     = contracts.NativeERC20TokenContract
 	)
 
-	if p.rewardTokenCode == "" {
+	if p.rewardTokenCode == "" && p.epochReward > 0 {
```
goran-ethernal: If epoch reward is 0, this will go to `else` condition, which will fail.

I suggest this:

```
if p.rewardTokenCode == "" {
      if p.epochReward > 0 {
            // native token is used as a reward token, and reward wallet is not a zero address
	    // so we need to add that address to premine map
	    premineBalances[walletPremineInfo.address] = walletPremineInfo
      }
} else {
      bytes, err := hex.DecodeString(p.rewardTokenCode)
      if err != nil {
	   return fmt.Errorf("could not decode reward token byte code '%s' : %w", p.rewardTokenCode, err)
      }

      rewardTokenByteCode = bytes
      rewardTokenAddr = contracts.RewardTokenContract
}
```

goran-ethernal (changes requested):
I added a comment that needs to be fixed.

Also, one more thing that should be fixed, while I was looking at this PR, which is not directly affected by this PR, but it should be fixed since it is in regards to the reward wallet. It is the `validateRewardWallet` function in `command/genesis/params.go`. It doesn't validate the reward wallet flag correctly. It should do `parsePremineInfo` before checking if reward wallet address is zero address.

The function should look like this:

```
// validateRewardWallet validates reward wallet flag
func (p *genesisParams) validateRewardWallet() error {
	if p.rewardWallet == "" {
		return errors.New("reward wallet address must be defined")
	}

	premineInfo, err := parsePremineInfo(p.rewardWallet)
	if err != nil {
		return err
	}

	if premineInfo.address == types.ZeroAddress {
		return errors.New("reward wallet address must not be zero address")
	}

	if premineInfo.amount.Cmp(big.NewInt(0)) < 1 {
		return errRewardWalletAmountZero
	}

	return nil
}
```

Or check if the `epochReward` is 0 and `rewardTokenCode` is an empty string, then just don't validate the reward wallet at all.

sergerad:
@goran-ethernal Thanks - for the latter comment I went with leaving the epoch reward check after zero addr check to be safe

goran-ethernal (approved):

PR #1783, opened 2023-08-06
Calculate EIP1559 gas price in JSON-RPC

mchmatt:
# Description

Tools like https://github.com/blockscout/blockscout operate under the premise that field "gasPrice" will be present even if the TX is an EIP 1559 type, and thus don't work correctly when connected to Polygon-Edge's RPC. 
This PR calculates the gas price on the fly for requests done to the RPC, through the existing function GetGasPrice.

# Changes include

# Additional comments

This fixes issues https://github.com/blockscout/blockscout/issues/8128 and https://github.com/blockscout/blockscout/issues/8130. All other RPCs I have tested (Llama RPC, Alchemy, Ankr) include the gasPrice field for EIP1559 transactions.

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -237,6 +237,7 @@ func (e *Eth) GetTransactionByHash(hash types.Hash) (interface{}, error) {
 		// Find the transaction within the block
 		if txn, idx := types.FindTxByHash(block.Transactions, hash); txn != nil {
 			txn.GasPrice = txn.GetGasPrice(block.Header.BaseFee)
+
 			return toTransaction(
 				txn,
 				argUintPtr(block.Number()),
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -236,6 +236,7 @@ func (e *Eth) GetTransactionByHash(hash types.Hash) (interface{}, error) {
 
 		// Find the transaction within the block
 		if txn, idx := types.FindTxByHash(block.Transactions, hash); txn != nil {
+			txn.GasPrice = txn.GetGasPrice(block.Header.BaseFee)
 			return toTransaction(
 				txn,
 				argUintPtr(block.Number()),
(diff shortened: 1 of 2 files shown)


mchmatt:

mchmatt:
recheck

xxMattewxx:

Stefan-Ethernal (commented):
Hey @mchmatt thanks for the PR, generally looks good to me.

However, WDYT about setting a gas price in the `txpool_content` JSON RPC function as well (https://github.com/mchmatt/polygon-edge/blob/develop/jsonrpc/txpool_endpoint.go#L54)? I was thinking about adding this line there as well  `tx.GasPrice = tx.GetGasPrice(t.store.GetBaseFee())`.

Also, pay attention to the linting error and fix it, so we can merge it eventually.

mchmatt:
> However, WDYT about setting a gas price in the `txpool_content` JSON RPC function as well (https://github.com/mchmatt/polygon-edge/blob/develop/jsonrpc/txpool_endpoint.go#L54)? I was thinking about adding this line there as well `tx.GasPrice = tx.GetGasPrice(t.store.GetBaseFee())`.

I'm not sure how that would work as my understanding is that EIP 1559 TXs in the pool aren't assigned to a block yet, but maybe I'm missing something? Not really familiar with blockchain dev, sorry.

> Also, pay attention to the linting error and fix it, so we can merge it eventually.

*Should* be fixed now, really appreciate the quick reply.

Stefan-Ethernal:
> I'm not sure how that would work as my understanding is that EIP 1559 TXs in the pool aren't assigned to a block yet, but maybe I'm missing something?

Yes, that is correct, in the tx pool, tx is not yet included in the block. And since the Blockscout indexes only blocks, then we should be ok with your approach. You have my approval. :slightly_smiling_face:

Stefan-Ethernal (approved):

polygomic:
Please accept this pull request we have big problems with blockscout

rachit77 (approved):

Stefan-Ethernal:
We are going to merge this PR, although we have noticed that the `GetGasPrice` function does not calculate gas prices the same way as Geth does. Geth for instance just returns a gas fee cap for dynamic fee transactions, whereas Edge has some logic, which does not seem entirely correct.

PR #1788, opened 2023-08-07
[RFC-191]: EVM-760 - Utilize `ForkParams` contract

goran-ethernal:
# Description

This PR utilizes `ForkParams` contract to activate or deactivate forks based on on-chain governance proposals for fork activation/deactivation.

To activate a fork using on-chain governance users need to:
1. Send a governance proposal to activate a fork, where proposal `calldata` must be a call to `addNewFeature` function on `ForkParams` contract. In given `calldata` user must specify fork name and block from which it should be active  (**fork name in the proposal must match the name of that fork in [code]( https://github.com/0xPolygon/polygon-edge/blob/d5be69a735a347bf8746a844e33ee352e5cb2469/chain/params.go#L80):**).
2. Proposal must have enough votes, and executed.
3. After proposal gets executed, fork will be automatically added to edge client forks where it becomes active from specified block.

If a user didn't specify that fork in `genesis`, that fork will automatically be turned off when chain starts, and it can be activated by later specifying it in the genesis and setting some future block for it (restart of all nodes is required), or by using on-chain governance.

If a user specified a fork in `genesis`, and that fork gets activated when its block is reached, there is one way to deactivate it, and it's by using on-chain governance. Basically, user will need to create a proposal that changes it's block number to some block that will either never be reached, or it will be reached after a long period of time (they can always change this using `ForkParams` contract).

To deactivate a fork using on-chain governance users need to:
1. Send a governance proposal to deactivate a fork, where proposal `calldata` must be a call to `addNewFeature` function on `ForkParams` contrac (if it is not already added to that contract), or `updateFeatureBlock` (if it was already added to that contract). In given `calldata` user must specify fork name and some big block number that will not be reached in a long time (**fork name in the proposal must match the name of that fork in [code]( https://github.com/0xPolygon/polygon-edge/blob/d5be69a735a347bf8746a844e33ee352e5cb2469/chain/params.go#L80):**).
2. Proposal must have enough votes, and executed.
3. After proposal gets executed, fork will be automatically deactivated on edge client.

Code changes in this PR are as follows:
- `governanceManager` is expanded to collect both `NewFeature` and `UpdatedFeature` events that get emitted by `ForkParams` functions `addNewFeature` and `updateFeatureBlock` respectively.
- Those events will be saved to a separate bucket called `forkParamsEventsBucket`, where key is hash of fork name and value is block from which that feature is active.
- On finalization of each block (either from `consensus` or `syncer`), `governanceManager` will get those events from receipts and save them in `boltdb`. Note, that `governanceManager` has a means to get missed events (if some error occurred while saving those events in previous block.
- On each block that had these `ForkParams` events, it will register and activate/deactivate those forks in `forkManager`, so that the rest of the code knows if that fork is active or not, and from which block.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state_store_governance.go
+++ b/consensus/polybft/state_store_governance.go
@@ -55,11 +55,6 @@ func (g *GovernanceStore) initialize(tx *bolt.Tx) error {
 			string(clientConfigBucket), err)
 	}
 
-	if _, err := tx.CreateBucketIfNotExists(clientConfigBucket); err != nil {
-		return fmt.Errorf("failed to create bucket=%s: %w",
-			string(clientConfigBucket), err)
-	}
-
 	if _, err := tx.CreateBucketIfNotExists(lastProcessedGovernanceBlockBucket); err != nil {
 		return fmt.Errorf("failed to create bucket=%s: %w",
 			string(lastProcessedGovernanceBlockBucket), err)
--- a/consensus/polybft/state_store_governance.go
+++ b/consensus/polybft/state_store_governance.go
@@ -5,14 +5,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/big"
 
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	"github.com/0xPolygon/polygon-edge/helper/common"
+	"github.com/0xPolygon/polygon-edge/types"
 	bolt "go.etcd.io/bbolt"
 )
 
 var (
-	governanceEventsBucket             = []byte("governanceEvents")
+	networkParamsEventsBucket          = []byte("networkParamsEvents")
+	forkParamsEventsBucket             = []byte("forkParamsEvents")
 	clientConfigBucket                 = []byte("clientConfig")
 	clientConfigKey                    = []byte("clientConfigKey")
 	lastProcessedGovernanceBlockBucket = []byte("lastProcessedGovernanceBlock")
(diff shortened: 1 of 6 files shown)


consensus/polybft/state_store_governance.go
```diff
@@ -36,17 +40,27 @@ type GovernanceStore struct {
...
 	}
 
-	if _, err = tx.CreateBucketIfNotExists(clientConfigBucket); err != nil {
+	if _, err := tx.CreateBucketIfNotExists(clientConfigBucket); err != nil {
 		return fmt.Errorf("failed to create bucket=%s: %w",
 			string(clientConfigBucket), err)
 	}
 
-	if _, err = tx.CreateBucketIfNotExists(lastProcessedGovernanceBlockBucket); err != nil {
+	if _, err := tx.CreateBucketIfNotExists(clientConfigBucket); err != nil {
```
dusan-maksimovic: Should use lastProcessedGovernanceBlockBucket instead of clientConfigBucket?

dusan-maksimovic: Actually, the `clientConfigKey`.

goran-ethernal: It was redundant: https://github.com/0xPolygon/polygon-edge/pull/1788/commits/80f1b9cbe34b29cd0aa674ba9e536e107f1a5a21

dusan-maksimovic (approved):

stana-miric (approved):

PR #1789, opened 2023-08-07
fixed txpool maxEnqueued issue

### LINKED ISSUE #1782: Txpool problem
When one account sent above 128 transaction all capacity of wallet reached
and we got [ERROR] polygon.txpool: failed to add broadcast tx: err="maximum number of enqueued transactions reached"

the only way to solve problem is stop all nodes and start all again
If I stop adding new transactions again I got [ERROR] polygon.txpool: failed to add broadcast tx: err="maximum number of enqueued transactions reached"

if blockchain work in public how cloud we do stop all nodes and start them again to clear enqueued transactions
we should have command to clear enqueued transactions or if we stop new transaction remain 128 transaction most be handled by txpool and pool most be got clean

I think this is bug.

## Your environment
- OS and version.
- The version of the Polygon Edge.    
  (*Confirm the version of your Polygon edge client by running the following command: `polygon-edge version --grpc-address GRPC_ADDRESS`*)
- The branch that causes this issue.
- Locally or Cloud hosted (which provider).
- Please confirm if the validators are running under containerized environment (K8s, Docker, etc.).

## Steps to reproduce
- Tell us how to reproduce this issue.
- Where the issue is, if you know.
- Which commands triggered the issue, if any.
- Provide us with the content of your genesis file.
- Provide us with commands that you used to start your validators.
- Provide us with the peer list of each of your validators by running the following command: `polygon-edge peers list --grpc-address GRPC_ADDRESS`.
- Is the chain producing blocks and serving customers atm?

## Expected behavior
- Tell us what should happen.
- Tell us what happened instead.

## Logs
Provide us with debug logs from all of your validators by setting logging to `debug` output with: `server --log-level debug`

## Proposed solution
If you have an idea on how to fix this issue, please write it down here, so we can begin discussing it

rachit77:
# Description

Currently, the Txpool refuses to process transactions for a specific address once the `maxEnqueued` limit for that address is reached. This situation results in a stagnant state for the address, preventing it from accommodating any further transactions, including those with the `nextNonce`. With the introduction of this PR, addresses will have the capability to accept transactions with the `nextNonce`,  even when the `maxEnqueued` limit for an account has been reached.

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -814,7 +814,7 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 
 		slotsFree += slotsRequired(oldTxWithSameNonce) // add old tx slots
 	} else {
-		if account.enqueued.length() == account.maxEnqueued {
+		if account.enqueued.length() == account.maxEnqueued && tx.Nonce != accountNonce {
 			return ErrMaxEnqueuedLimitReached
 		}
 
--- a/txpool/txpool_test.go
+++ b/txpool/txpool_test.go
@@ -778,7 +778,7 @@ func TestEnqueueHandler(t *testing.T) {
 			assert.NoError(t, err)
 			pool.SetSigner(&mockSigner{})
 
-			//	mock full enqueued
+			// mock full enqueued
 			pool.accounts.maxEnqueuedLimit = 10
 
 			// add 10 transaction in txpool i.e. max enqueued transactions
--- a/txpool/txpool_test.go
+++ b/txpool/txpool_test.go
@@ -768,6 +768,43 @@ func TestEnqueueHandler(t *testing.T) {
 			acc.nonceToTx.unlock()
 		},
 	)
+
+	t.Run(
+		"accept new tx with nextNonce when enqueued is full",
+		func(t *testing.T) {
+			t.Parallel()
+
+			pool, err := newTestPool()
+			assert.NoError(t, err)
+			pool.SetSigner(&mockSigner{})
+
+			//	mock full enqueued
+			pool.accounts.maxEnqueuedLimit = 10
+
+			// add 10 transaction in txpool i.e. max enqueued transactions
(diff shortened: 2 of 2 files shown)


txpool/txpool.go
```diff
@@ -814,7 +814,7 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 
 		slotsFree += slotsRequired(oldTxWithSameNonce) // add old tx slots
 	} else {
-		if account.enqueued.length() == account.maxEnqueued {
+		if account.enqueued.length() == account.maxEnqueued && tx.Nonce != accountNonce {
 			return ErrMaxEnqueuedLimitReached
```
Stefan-Ethernal, line 818: Basically, we are going to return the error only in case the max enqueued limit is reached and the nonce-gapped transaction is submitted. I'm wondering do we then even need the `maxEnqueued` parameter in this case, and if not is it safe to remove it (though not sure if it is a desirable change and how to protect from spammers in that case)?

txpool/txpool_test.go
```diff
@@ -768,6 +768,43 @@ func TestEnqueueHandler(t *testing.T) {
...
+	t.Run(
+		"accept new tx with nextNonce when enqueued is full",
+		func(t *testing.T) {
+			t.Parallel()
+
+			pool, err := newTestPool()
+			assert.NoError(t, err)
+			pool.SetSigner(&mockSigner{})
+
+			//	mock full enqueued
```
Stefan-Ethernal:
```suggestion
			// mock full enqueued
```

txpool/txpool.go
rachit77: If `maxEnqueued` parameter is not present there than how will you check the number of nonce-gapped transaction for a address and return error accordingly.

Stefan-Ethernal (approved):

begmaroman (approved):

PR #1795, opened 2023-08-08
Send dynamic fee transactions to a rootchain

Stefan-Ethernal:
# Description

This PR sends a dynamic fee transaction for all of the root chain transactions, since there were some issues when Supernets was having the root chain set to the Sepolia network, as checkpoint submission transactions were underpriced.
If sending dynamic fee transaction fails, it fallbacks to legacy transaction and tries resending it.

It also fixes the transaction hash calculation, because it was not correct for access list transactions.

# Changes include

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments


goran-ethernal (approved):

txpool/txpool.go
```diff
@@ -614,9 +615,9 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 	if tx.Type == types.DynamicFeeTx {
 		// Reject dynamic fee tx if london hardfork is not enabled
 		if !forks.London {
-			metrics.IncrCounter([]string{txPoolMetrics, "invalid_tx_type"}, 1)
+			metrics.IncrCounter([]string{txPoolMetrics, "tx_type"}, 1)
 
-			return ErrInvalidTxType
```
Nemanja0x, line 619: Why changing this. Nexon just got used to `invaild_tx_type` :)

txrelayer/txrelayer.go
```diff
@@ -86,6 +87,14 @@ func (t *TxRelayerImpl) Call(from ethgo.Address, to ethgo.Address, input []byte)
 func (t *TxRelayerImpl) SendTransaction(txn *ethgo.Transaction, key ethgo.Key) (*ethgo.Receipt, error) {
 	txnHash, err := t.sendTransactionLocked(txn, key)
 	if err != nil {
+		if txn.Type != ethgo.TransactionLegacy && errors.Is(err, types.ErrTxTypeNotSupported) {
```
Nemanja0x: This part is a fallback from the situation if a blockchain does not accept dynamic transactions. We check if error is `ErrTxTypeNotSupported` and resend a legacy transaction. How do we know if some rootchain is going to return us this error if it does not support EIP-1559?

Stefan-Ethernal: We cannot be sure, but this was the best guess. Geth returns this error for instance, although it is wrapped (https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L63-L65). Bor returns this exact error message, since it is a fork of Geth (https://github.com/maticnetwork/bor/blob/develop/core/txpool/txpool.go#L754-L756).

So I'd say we have two options here:
- if any error is returned, try with `LegacyTx` fallback (though not sure is it ok to do it),
- if error message contains `ErrTxTypeNotSupported` string, fallback to `LegacyTx`

What do you think?

Nemanja0x: Leave it as it is and add another task into the backlog to check this.

Nemanja0x (approved):

PR #1805, opened 2023-08-11
Header.BaseFee is set to non-zero for london fork disabled

begmaroman:
# Description

Return zero basefee if the london hardfork is disabled

# Changes include

# Checklist

## Testing

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1364,6 +1364,10 @@ func (b *Blockchain) CalculateBaseFee(parent *types.Header) uint64 {
 	// Check if this is the first London hardfork block.
 	// Should return chain.GenesisBaseFee ins this case.
 	if parent.BaseFee == 0 {
+		if b.config.Genesis.BaseFee > 0 {
+			return b.config.Genesis.BaseFee
+		}
+
 		return chain.GenesisBaseFee
 	}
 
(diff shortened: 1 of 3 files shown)


blockchain/blockchain.go
```diff
@@ -1356,7 +1356,14 @@ func (b *Blockchain) Close() error {
...
 // CalculateBaseFee calculates the basefee of the header.
 func (b *Blockchain) CalculateBaseFee(parent *types.Header) uint64 {
+	// Return zero base fee is a london hardfork is not enabled
 	if !b.config.Params.Forks.IsActive(chain.London, parent.Number) {
+		return 0
+	}
+
+	// Check if this is the first London hardfork block.
+	// Should return chain.GenesisBaseFee ins this case.
+	if parent.BaseFee == 0 {
```
goran-ethernal, line 1366: This will never be 0, check this part of code: https://github.com/0xPolygon/polygon-edge/blob/db29c8a475c1f62a33213dbef23adee6e76e721c/chain/chain.go#L99

begmaroman: Good catch

begmaroman:
Tested manually. Sent several transactions with disabled london hardfork via Metamask.

stana-miric (approved):

Nemanja0x (approved):

Nemanja0x:
One thing to remember, maybe I'll create a ticket. If we have london fork disabled and we enable it at block 100: It can happen  that a node which is started after block 100 has different hash of genesis block. This is because genesis header will use the new configuration where the baseFee is not 0. Nevertheless, if this is assumption is correct, it is a new bug. The one reported here is solved.

PR #1809, opened 2023-08-13
Upgraded Makefile to check if all the required tools are installed

criadoperez:
# Description

Made some improvements on the Makefile
* It now checks the dependencies it has of like `git`, `go`, `protoc`, etc and informs the user to install them if he is running a make target that requires them and are not installed.
* Added a new `make help` command that displays all positions targets.

As a note, default behavior or running just `make` was running `make download-submodules`.  I respected this default behavior and left it as is, but we might consider changing default to `make help` or even something else.

# Changes include

## Testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,35 @@
-
 .PHONY: download-submodules
-download-submodules:
+download-submodules: check-git
 	git submodule init
 	git submodule update
 
+.PHONY: check-git
+check-git:
+	@which git > /dev/null || (echo "git is not installed. Please install and try again."; exit 1)
+
+.PHONY: check-go-tools
+check-go-tools:
+	@which go > /dev/null || (echo "Go is not installed.. Please install and try again."; exit 1)
+	@which go-bindata > /dev/null || (echo "go-bindata is not installed. Please install and try again."; exit 1)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,15 +1,35 @@
...
 	git submodule update
 
+.PHONY: check-git
+check-git:
+	@which git > /dev/null || (echo "git is not installed. Please install and try again."; exit 1)
+
+.PHONY: check-go-tools
+check-go-tools:
+	@which go > /dev/null || (echo "Go is not installed.. Please install and try again."; exit 1)
+	@which go-bindata > /dev/null || (echo "go-bindata is not installed. Please install and try again."; exit 1)
```
Stefan-Ethernal: Unit tests and E2E tests are failing because this prerequisite is not met.

Btw, I think that it is legacy, as I'm unaware that `make bindata` is used (it certainly isn't for polybft consensus protocol), so it probably can be removed.

criadoperez: Ok. Removed it then.

Stefan-Ethernal (approved):

igorcrevar (approved):

PR #974, opened 2022-11-28
Make "pending" block param act as "latest"

lazartravica:
Fixes EVM-231.

# Description

Changed the behaviour of functions that do:
 - fetching blocks by Number/String/Hash
 - decoding JSON-RPC requests
 - fetching blocks by number

# Changes include

- [x] Bugfix (non-breaking change that solves an issue)
- [x] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [x] I have updated the official documentation
- [x] I have added sufficient documentation in code

## Testing

- [ ] I have tested this code with the official test suite
- [x] I have tested this code manually

### Manual tests

Ran various JSON-RPC queries with "pending" as the block parameter.

# Documentation update

`//TODO`

PR #1816, opened 2023-08-16
NOTICKET: parse pending block correctly for BlockNumberOrHash

jp-imx:
# Description

It looks like this bug was intentionally introduced, so a bit more follow up might be required: https://github.com/0xPolygon/polygon-edge/pull/974

# Changes include

# Breaking changes

Technically it can be considered a breaking change as it changes the semantics of public endpoints that relly on unmarshalling `BlockNumberOrHash`.

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/codec.go
+++ b/jsonrpc/codec.go
@@ -176,7 +176,9 @@ func stringToBlockNumber(str string) (BlockNumber, error) {
 
 	str = strings.Trim(str, "\"")
 	switch str {
-	case pending, latest:
+	case pending:
+		return PendingBlockNumber, nil
+	case latest:
 		return LatestBlockNumber, nil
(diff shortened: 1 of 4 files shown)


jp-imx:
In trying to fix the unit tests I found that there are other places where the assumption that Pending === Latest has been encoded. For example: https://github.com/0xPolygon/polygon-edge/blob/develop/jsonrpc/helper.go#L26-L32

I think I need some advise on the wider impact of these changes.

Stefan-Ethernal:
The changes from this PR are included in https://github.com/0xPolygon/polygon-edge/pull/1853, which is already merged. Closing this one.

PR #1817, opened 2023-08-16
EVM-778 Debug Transaction endpoint - use one structure for logs

igorcrevar:
# Description

Use only one structure for logs to reduce memory consumption. 

Prior to these changes, there was an additional slice of objects that stored information about each step's execution. That slice was later on converted into json response slice of different objects. That caused additional memory consumption without any special reason.

Additionally, `omitempty` is added for fields
```
Stack           []string          `json:"stack,omitempty"`
Memory       []string          `json:"memory,omitempty"`
Storage       map[string]string `json:"storage,omitempty"`
```
and that will reduce json output in case when those fields are empty or disabled by json request input parameters (`disableStack, disableStorage, enableMemory`).

# Changes include

# Checklist

## Testing

- [X] I have tested this code manually

### Manual tests

Start cluster, deploy smart contract(s) and transaction that calls some of those sc. `0x09e60436fc2344c03ef2c721dcb179d2e9cb4cecf535f1217f25abe24ea37dc6` is hash of that transaction and that is the tx you want to debug.
Try to execute following endpoint request. You should get the same json response as with old code (minus that in newest version empty fields will be omitted)
```
[{
	"jsonrpc":"2.0",
	"method":"debug_traceTransaction",
	"params":[
        "0x09e60436fc2344c03ef2c721dcb179d2e9cb4cecf535f1217f25abe24ea37dc6",
        {
            "timeout": "30s",
            "disableStack": false,
            "disableStorage": false,
            "enableMemory": true,
            "enableReturnData": true
        }
	],
	"id":1
}]
```

### DIFF
--- a/state/runtime/tracer/structtracer/tracer.go
+++ b/state/runtime/tracer/structtracer/tracer.go
@@ -21,26 +21,17 @@ type Config struct {
 }
 
 type StructLog struct {
-	Pc            uint64                    `json:"pc"`
-	Op            string                    `json:"op"`
-	Gas           uint64                    `json:"gas"`
-	GasCost       uint64                    `json:"gasCost"`
-	Memory        []byte                    `json:"memory,omitempty"`
-	MemorySize    int                       `json:"memSize"`
-	Stack         []*big.Int                `json:"stack"`
-	ReturnData    []byte                    `json:"returnData,omitempty"`
-	Storage       map[types.Hash]types.Hash `json:"storage"`
-	Depth         int                       `json:"depth"`
-	RefundCounter uint64                    `json:"refund"`
-	Err           error                     `json:"err"`
-}
-
-func (l *StructLog) ErrorString() string {
-	if l.Err != nil {
-		return l.Err.Error()
-	}
-
-	return ""
+	Pc            uint64            `json:"pc"`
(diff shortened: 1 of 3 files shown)


state/runtime/tracer/structtracer/tracer.go
```diff
@@ -175,7 +164,7 @@ func (t *StructTracer) captureMemory(
 	t.currentMemory[len(t.currentMemory)-1] = currentMemory
 
 	if opCode == evm.CALL || opCode == evm.STATICCALL {
-		t.currentMemory = append(t.currentMemory, make([]byte, len(memory)))
+		t.currentMemory = append(t.currentMemory, nil)
```
begmaroman, line 167: Does it make sense to add a lock here?

stana-miric (approved):

state/runtime/tracer/structtracer/tracer.go
igorcrevar: execution of transaction is not parallel, so only one routine executes this code and therefor locks are not necessary :)

Nemanja0x (approved):

begmaroman (approved):

PR #1821, opened 2023-08-16
EVM-779 Throttling with concurrency

igorcrevar:
# Description

Instead of using requests per second measure to refuse request, use different approach:
Only `--concurrent-requests-debug` (default `32`) requests are allowed to be executed concurrently for all the debug endpoints 

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/jsonrpc/throttling.go
+++ b/jsonrpc/throttling.go
@@ -24,7 +24,7 @@ func NewThrottling(maximumConcurrentRequests uint64, timeout time.Duration) *Thr
 	}
 }
 
-// AttemptRequest returns an error if there is more thab
+// AttemptRequest returns an error if more than the maximum concurrent requests are currently being executed
 func (t *Throttling) AttemptRequest(
 	parentCtx context.Context,
 	requestHandler func() (interface{}, error)) (interface{}, error) {
--- a/jsonrpc/throttling.go
+++ b/jsonrpc/throttling.go
@@ -1,70 +1,41 @@
 package jsonrpc
 
 import (
-	"container/heap"
+	"context"
 	"errors"
-	"sync"
 	"time"
+
+	"golang.org/x/sync/semaphore"
 )
 
 var errRequestLimitExceeded = errors.New("request limit exceeded")
 
+// Throttling provides functionality which limits number of concurrent requests
 type Throttling struct {
-	requestsPerSeconds int
-	requests           timeQueue
-	lock               sync.Mutex
+	sem     *semaphore.Weighted
+	timeout time.Duration
 }
 
-func NewThrottling(requestsPerSeconds int) *Throttling {
+// NewThrottling creates new throttling and limits number of concurrent requests to maximumConcurrentRequests
+func NewThrottling(maximumConcurrentRequests uint64, timeout time.Duration) *Throttling {
 	return &Throttling{
-		requestsPerSeconds: requestsPerSeconds,
+		sem:     semaphore.NewWeighted(int64(maximumConcurrentRequests)),
+		timeout: timeout,
 	}
 }
 
-func (t *Throttling) AttemptRequest() error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+// AttemptRequest returns an error if there is more thab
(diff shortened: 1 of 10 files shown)


jsonrpc/throttling.go
```diff
@@ -28,43 +27,62 @@ func (t *Throttling) AttemptRequest() error {
...
+		items: make([]time.Time, capacity),
+		count: 0,
+		index: 0,
+	}
 }
 
-func (t *timeQueue) Push(x interface{}) {
-	if time, ok := x.(time.Time); ok {
-		*t = append(*t, time)
+func (t *timeQueue) push(value time.Time) bool {
```
Nemanja0x: add comment and explanation how this works

jsonrpc/throttling.go
```diff
@@ -28,43 +27,62 @@ func (t *Throttling) AttemptRequest() error {
...
+	t.index = (t.index + 1) % len(t.items)
+	t.count++
+
+	return true
 }
 
-func (t *timeQueue) Pop() interface{} {
-	x := (*t)[len(*t)-1]
-	*t = (*t)[0 : len(*t)-1]
+func (t timeQueue) first() time.Time {
```
Nemanja0x: add comment and explanation how this works

jsonrpc/throttling.go
```diff
@@ -28,43 +27,62 @@ func (t *Throttling) AttemptRequest() error {
...
+	index := (t.index - t.count + len(t.items)) % len(t.items)
```
Nemanja0x: add comment and explanation how this works

jsonrpc/throttling.go
```diff
@@ -28,43 +27,62 @@ func (t *Throttling) AttemptRequest() error {
...
+
+	return t.items[index]
+}
+
+func (t *timeQueue) pop() {
```
Nemanja0x: `pop()` but nothing is returned?

jsonrpc/throttling.go
```diff
@@ -28,43 +27,62 @@ func (t *Throttling) AttemptRequest() error {
...
+	t.requests.push(currTime)
 
 	return nil
 }
 
-type timeQueue []time.Time
-
-func (t *timeQueue) Len() int {
-	return len(*t)
+type timeQueue struct {
```
Nemanja0x: Add description of the structure

jsonrpc/throttling.go
```diff
@@ -1,70 +1,41 @@
...
 
-func (t *Throttling) AttemptRequest() error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	currTime := time.Now().UTC()
+// AttemptRequest returns an error if there is more thab
+func (t *Throttling) AttemptRequest(
+	parentCtx context.Context,
+	requestHandler func() (interface{}, error)) (interface{}, error) {
```
sergerad, line 30: Functional approach is handy when we don't know which functions are going to be throttled. In this situation, we implement every use case of `AttemptRequest()`, so the functional approach is not necessary.

Understood that the alternative is not that much better re verbosity because we would have to make the same call in each debug related function (including error checks).

Just a thought. Not suggesting changes. Both ways work 👍

Nemanja0x (approved):

jsonrpc/throttling.go
```diff
@@ -1,70 +1,41 @@
...
+		timeout: timeout,
 	}
 }
 
-func (t *Throttling) AttemptRequest() error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	currTime := time.Now().UTC()
+// AttemptRequest returns an error if there is more thab
```
Stefan-Ethernal: Finish this sentence :wink:

Check if this is good enough:

```suggestion
// AttemptRequest returns an error if there is more than maximum concurrent requests are currently being executed
```

Stefan-Ethernal (approved):

jsonrpc/throttling.go
igorcrevar: fixed

PR #1823, opened 2023-08-16
Expose JSON RPC function for slash exit events retrieval

Stefan-Ethernal:
# Description

This PR changes Bolt DB storage in a way that slashing exit event ids are stored in a separate bucket. Slashing exit events per se, are persisted in the same bucket where all the other exit events are. 
Also, a new JSON RPC function, called `bridge_getPendingSlashProofs` is exposed. When invoked, it queries the slashing exit events storage, which returns existing slashing exit event ids, and then for each one, a proof is created, by calling the existing `GenerateExitProof` function.

# Changes include

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/consensus/polybft/state_store_exit_event.go
+++ b/consensus/polybft/state_store_exit_event.go
@@ -52,7 +52,8 @@ type ExitEvent struct {
 Bolt DB schema:
 
 exit events/
-|--> (epoch+id+blockNumber) -> *ExitEvent (json marshalled)
+|--> (epoch+exit event id+blockNumber) -> *ExitEvent (json marshalled)
+|--> (exit event id) -> nil (slashing exit events)
 |--> (exitEventID) -> epochNumber
 |--> (lastProcessedBlockKey) -> block number
 */
--- a/consensus/polybft/state_store_exit_event.go
+++ b/consensus/polybft/state_store_exit_event.go
@@ -7,20 +7,27 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
-	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/umbracle/ethgo"
 	bolt "go.etcd.io/bbolt"
+
+	"github.com/0xPolygon/polygon-edge/chain"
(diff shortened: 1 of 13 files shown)


consensus/polybft/state_store_exit_event.go
```diff
@@ -54,11 +62,17 @@ type ExitEventStore struct {
 }
 
 // initialize creates necessary buckets in DB if they don't already exist
-func (s *ExitEventStore) initialize(tx *bolt.Tx) error {
+func (s *ExitEventStore) initialize(tx *bolt.Tx, blockNumber uint64) error {
 	if _, err := tx.CreateBucketIfNotExists(exitEventsBucket); err != nil {
 		return fmt.Errorf("failed to create bucket=%s: %w", string(exitEventsBucket), err)
 	}
 
+	if forkmanager.GetInstance().IsForkEnabled(chain.DoubleSignSlashing, blockNumber) {
```
dusan-maksimovic: What happens if the fork is not yet activated on node startup, but it exists in code? Bucket for `slashingExitEventsBucket` will not be created. Do we need to worry about this?

Nemanja0x: Why do we need to initialize store through the fork manager? Isn't it enough to initialize it anyway and do not use it if we do not need it?

stana-miric (approved):

consensus/polybft/state_store_exit_event.go
Stefan-Ethernal: Good point and you are right, storage logic should be as simple as possible.
Fixed
https://github.com/0xPolygon/polygon-edge/pull/1823/commits/429357f227843802a249e8fdb494b950bdfb638c

Stefan-Ethernal: Per @Nemanja0x comment, we have decided to remove checking of the hard fork and create a slashing bucket unconditionally.
https://github.com/0xPolygon/polygon-edge/pull/1823/commits/429357f227843802a249e8fdb494b950bdfb638c

dusan-maksimovic (approved):

Nemanja0x (approved):

PR #1839, opened 2023-08-21
Evm 785: enable and disable access lists in the runtime

Nemanja0x:
# Description
1. Being able to enable/disable access list for contracts/txs during runtime
2. Access list does not have to be defined during genesis but it can be enabled later on if access lists owner is defined during genesis.
3. Access lists owner can only be defined during genesis creation (`--access-lists-owner` flag).
  Genesis command example:
  ```
  ./polygon-edge genesis 
     --consensus polybft \
    --block-gas-limit 10000000 \
    --premine 0x85da99c8a7c2c95964c8efd687e95e632fc533d6:1000000000000000000000 \
    --premine 0x0000000000000000000000000000000000000000 \
    --epoch-size 10 \
    --reward-wallet 0xDEADBEEF:1000000 \
    --access-lists-owner 0xF8fC8D3D271ACACfcA412984f76e18Ad616BdCBa
    --burn-contract 0:0x0000000000000000000000000000000000000000
```    
4. Access lists owner (new role which will be introduced) and admin can enable/disable an access list
5. Access lists owner or admin can modify an access list before it was enabled
6. Smart contract that uses bridge allow/block lists has its own logic for enabling/disabling those lists and that logic is still the same as it was before.
7. If access lists owner is specified all access lists smart contracts will be deployed during the genesis phase and access list owner will be owner/creator of all of them
  7.1 That will be the case even if bridge allow list and bridge block list are empty
8. E2E test to support this feature

# Changes include

# Checklist

## Testing

# Documentation update

### CODE REVIEW

**Review by rachit77 — APPROVED** [review]
LGTM

File: chain/params.go [hunk 1]
```diff
@@ -21,6 +21,7 @@ type Params struct {
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
 
 	// Access control configuration
+	SuperAdminAllowBlock      *types.Address     `json:"superAdminAllowBlock,omitempty"`
```
Stefan-Ethernal:
```suggestion
	SuperAdminAccessLists      *types.Address     `json:"superAdminAccessLists,omitempty"`
```

state/runtime/addresslist/addresslist.go
```diff
@@ -86,13 +88,56 @@ func (a *AddressList) runInputCall(caller types.Address, input []byte,
...
+
+		return vl, gasUsed, nil
+	}
+
+	// SetEnabledList receives bool as input parameter which in abi get codified
+	// as a 32 bytes array
+	// all the other functions have the same input (i.e. tuple(address)) which
+	// in abi gets codified as a 32 bytes array with the first 20 bytes
+	// encoding the address
+	if len(inputBytes) != 32 {
```
Stefan-Ethernal:
```suggestion
	if len(inputBytes) != types.HashLength {
```

chain/params.go
Nemanja0x: Can we call it `SuperAdminAllowLists` ?

state/runtime/addresslist/addresslist.go
```diff
@@ -146,6 +195,17 @@ func (a *AddressList) GetRole(addr types.Address) Role {
 	return Role(res)
 }
 
+func (a *AddressList) IsEnabled() bool {
+	return a.state.GetStorage(a.addr, enabledKeyHash) != types.ZeroHash
+}
+
+func (a *AddressList) SetEnabled(value bool) {
+	vl, _ := abi.MustNewType("bool").Encode(value)
```
Stefan-Ethernal: Extract `abiBoolTrue` and `abiBoolFalse` to variables and use it as necessary (same thing was done in `precompiled.go`, so maybe we can extract those variables to some common package and use it from there).

state/runtime/addresslist/addresslist.go
```diff
@@ -86,13 +88,56 @@ func (a *AddressList) runInputCall(caller types.Address, input []byte,
 		return nil
 	}
 
+	// GetEnabledList does not have any parameters and returns bool value
```
Stefan-Ethernal:
```suggestion
	// GetListEnabledFunc does not have any parameters and returns bool value
```

state/executor.go
```diff
@@ -218,31 +218,37 @@ func (e *Executor) BeginTxn(
 		PostHook:    e.PostHook,
 	}
 
-	// enable contract deployment allow list (if any)
-	if e.config.ContractDeployerAllowList != nil {
-		txn.deploymentAllowList = addresslist.NewAddressList(txn, contracts.AllowListContractsAddr)
+	// allow/block list should be possible if
```
Stefan-Ethernal: Finish comment.

Check if this is good enough:
```suggestion
	// contract deployment access list should are enabled either if access lists super admin is defined or access lists are defined in the genesis
```

Stefan-Ethernal (approved):
Generally looking good to me. However I'd consider renaming `SuperAdminAllowBlock` to `SuperAdminAccessLists` everywhere.

chain/params.go
Stefan-Ethernal: Probably `SuperAdminAccessLists` suits the best, as it is applicable to both allow and block lists.

state/runtime/addresslist/addresslist.go
igorcrevar: This is only used for enabling/disabling list which will not be done frequently.
I will just make an internal func
```
func getAbiBoolValue(value bool) []byte {
	encodedValue, _ := abi.MustNewType("bool").Encode(value)

	return encodedValue
}
```
inside `state/runtime/addresslist/addresslist.go`. I am not sure where should I put that generic function, we can change that later

state/executor.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/0387a8b41875f716ef490c2f24f2b2e2ad086266

state/runtime/addresslist/addresslist.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/0387a8b41875f716ef490c2f24f2b2e2ad086266

state/runtime/addresslist/addresslist.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/0387a8b41875f716ef490c2f24f2b2e2ad086266

chain/params.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/0387a8b41875f716ef490c2f24f2b2e2ad086266

igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/0387a8b41875f716ef490c2f24f2b2e2ad086266

e2e-polybft/e2e/acls_test.go
```diff
@@ -413,3 +413,132 @@ func TestE2E_AddressLists_Bridge(t *testing.T) {
...
...6101a9576040519150601f19603f3d011682016040523d82523d6000602084013e6101ae565b606091505b50915091506000818060200190518101906101c99190610378565b9050809350505050919050565b600073ffffffffffffffffffffffffffffffffffffffff82169050919050565b6000610201826101d6565b9050919050565b610211816101f6565b82525050565b600060208201905061022c6000830184610208565b92915050565b600080fd5b610240816101f6565b811461024b57600080fd5b50565b60008135905061025d81610237565b92915050565b60006020828403121561027957610278610232565b5b60006102878482850161024e565b91505092915050565b6000819050919050565b6102a381610290565b82525050565b60006020820190506102be600083018461029a565b92915050565b600081519050919050565b600081905092915050565b60005b838110156102f85780820151818401526020810190506102dd565b60008484015250505050565b600061030f826102c4565b61031981856102cf565b93506103298185602086016102da565b80840191505092915050565b60006103418284610304565b915081905092915050565b61035581610290565b811461036057600080fd5b50565b6000815190506103728161034c565b92915050565b60006020828403121561038e5761038d610232565b5b600061039c84828501610363565b9150509291505056fea264697066735822122035f391b5c3dbf9a5e31072167a4ada71e9ba762650849f6320bc6150fa45aa9564736f6c63430008110033")
```
stana-miric: maybe to extract this as a constant since is use that in other places as well

chain/params.go
```diff
@@ -21,6 +21,7 @@ type Params struct {
 	BlockGasTarget uint64                 `json:"blockGasTarget"`
 
 	// Access control configuration
+	SuperAdminAccessLists     *types.Address     `json:"superAdminAccessLists,omitempty"`
```
vcastellm: Nitpick but `AccessListsSuperAdmin` reads better to me

e2e-polybft/e2e/acls_test.go
igorcrevar: Is it ok to have some `func getDummyContractByteCode()` ?

stana-miric (approved):

e2e-polybft/e2e/acls_test.go
igorcrevar: I have found even better solution (thanks to Goran): https://github.com/0xPolygon/polygon-edge/pull/1839/commits/a48f74606b6022f33c53e51c45515457f25dc5e6

state/executor.go
```diff
@@ -218,31 +218,40 @@ func (e *Executor) BeginTxn(
 		PostHook:    e.PostHook,
 	}
 
-	// enable contract deployment allow list (if any)
-	if e.config.ContractDeployerAllowList != nil {
-		txn.deploymentAllowList = addresslist.NewAddressList(txn, contracts.AllowListContractsAddr)
+	// contract deployment access list should be enabled
+	// either if access lists super admin is defined or access lists are defined in the genesis
```
vcastellm, line 222: So, all AccessLists will be enabled if SuperAdmin is defined? This should be stated in the docs at it has an impact in Gas costs for each transaction

state/runtime/addresslist/genesis.go
```diff
@@ -13,14 +13,20 @@ func ApplyGenesisAllocs(chain *chain.Genesis, addressListAddr types.Address, con
...
-	// enabled addr
-	for _, addr := range config.EnabledAddresses {
-		allocList.SetRole(addr, EnabledRole)
-	}
-
-	// admin addr
-	for _, addr := range config.AdminAddresses {
-		allocList.SetRole(addr, AdminRole)
+	if config == nil {
+		allocList.SetEnabled(false)
```
vcastellm, line 17:
```suggestion
		allowList.SetEnabled(false)
```

state/runtime/addresslist/genesis.go
```diff
@@ -13,14 +13,20 @@ func ApplyGenesisAllocs(chain *chain.Genesis, addressListAddr types.Address, con
...
+	} else {
+		allocList.SetEnabled(true)
```
vcastellm, line 19:
```suggestion
		allowList.SetEnabled(true)
```

vcastellm (commented):
Where's the superadmin stored?

igorcrevar:
> Where's the superadmin stored?

In genesis file in configuration. If we need to store that info in storage I think it would be easier to put superadmin for each list (otherwise we need to create another storage address just for that superadmin)

state/executor.go
igorcrevar: Yes, because super admin can enable/disable list at any time. By default if no role is specified, list will be disabled and there will be no role checking on transaction execution/smart contract deploy.

Yes, we should probably update the docs.  I have used the same predefined read/write gas costs for these new functions:
- `function setListEnabled(bool)`
- `function getListEnabled() returns (bool)`

chain/params.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1839/commits/ace82ed426a7aa68d3974648c9821584ad8b359e

state/executor.go
igorcrevar: I have changed this code. superadmin is kept in storage for each access list now. This means that currently we have same superadmin address for each list stored but in the future that will allow to have different super admin for each list. And it was easier to implement this way :)

igorcrevar: I have just realized, what was your first question. No, access list will be enabled if `getListEnabled` value in storage is set to true. If just superadmin is specified, then initial value will be `setListEnabled(false)`. But we need to create list in both cases in order to allow superadmin to change enabled/disabled value, add admin/normal roles etc.

vcastellm (changes requested):
Overall LGTM but there's some consistency details:

1. Bridge SCs function naming is different see https://github.com/0xPolygon/core-contracts/blob/main/contracts/lib/AccessList.sol#L20, this can be confusing, we can adapt it in both sides, even though I like this naming better
2. Consider emitting the event too
3. Now that we're introducing the concept of super admin would be great to make him the owner of the bridge ACL, I have doubt on this tho.

vcastellm:
Some of the considerations can be implemented in follow-up PRs but the minimum set of changes to be implemented here are:

1. If there's a superAdmin defined, we deploy Predicate ACLs contracts with both lists enabled (useBridgeAllowList/useBridgeBlockList) hardcoded to `true`, and superAdmin as owner. https://github.com/0xPolygon/polygon-edge/blob/develop/consensus/polybft/polybft.go#L182-L191 as we will control the usage using the precompile interface.
2. Assign the superAdmin to be the lists admin here https://github.com/0xPolygon/polygon-edge/blob/develop/consensus/polybft/polybft.go#L168-L178
3. Include new ACLs interface in the docs https://wiki.polygon.technology/docs/supernets/operate/deploy/access-control/supernets-allowlist-add-remove it's currently missing, and usage examples cc/ @DannyS03

server/server.go
```diff
@@ -217,40 +217,28 @@ func NewServer(config *Config) (*Server, error) {
 	}
 
 	// apply allow list contracts deployer genesis data
-	if m.config.Chain.Params.ContractDeployerAllowList != nil {
-		addresslist.ApplyGenesisAllocs(m.config.Chain.Genesis, contracts.AllowListContractsAddr,
-			m.config.Chain.Params.ContractDeployerAllowList)
-	}
+	addresslist.ApplyGenesisAllocs(m.config.Chain.Genesis, contracts.AllowListContractsAddr,
```
vcastellm, line 220: Enabling the ACL at genesis time unconditionally is not a good idea, it implies extra gas cots for all users regardless if they want to use the ACLs or not

igorcrevar: First line in function is
```
	if superAdmin == nil && config == nil {
		return
	}
	```

Comment (igorcrevar) [reply server/server.go:220]: I prefer to have checks on one place instead on six, but i can change that if needed

File: e2e-polybft/e2e/acls_test.go [hunk 12]
```diff
@@ -448,22 +448,22 @@ func TestE2E_AllowList_SuperAdmin(t *testing.T) {
...
 
-		saTxn = cluster.MethodTxn(t, superadmin, contracts.AllowListContractsAddr, input)
+		saTxn = cluster.MethodTxn(t, owner, contracts.AllowListContractsAddr, input)
 		require.NoError(t, saTxn.Wait())
 		require.False(t, saTxn.Failed())
 	}
 
 	{
-		// Step 3. no one can send normal transaction or deploy smart contract because lists are enabled and empty
+		// Step 3. future admin or rnUser can not send transaction or deploy smart contract because they are not in the lists yet
```
vcastellm: `rnUser`?

vcastellm (approved):
After latest modifications and several conversations with Igor, we've implemented the minimum set of changes that makes sense from the UX perspective, fixed several issues and refactored some parts, this PR should be ready to be merged.

Some extra improvements will be described in an RFC and implemented in the following weeks.

LGTM

PR #1840, opened 2023-08-21
cluster script fix

dusannosovic-ethernal:
# Description
The current problem with cluster script is the ability to  introduce `relayer` even if `IBFT` is active as a consensus mechanism. `Relayer` may be invoked only if `PolyBFT` is used. 

This PR fix cluster script fixes the above mentioned problem: 
Script will check what kind of consensus mechanism was invoked and consequently will introduce `relayer` or not.

# Changes include

# Checklist

## Testing

### DIFF
--- a/scripts/cluster
+++ b/scripts/cluster
@@ -158,7 +158,7 @@ function initRootchain() {
 function startServerFromBinary() {
   if [ "$1" == "write-logs" ]; then
     echo "Writing validators logs to the files..."
-    if [ "$1" == "polybft" ]; then
+    if [ "$0" == "polybft" ]; then
       ./polygon-edge server --data-dir ./test-chain-1 --chain genesis.json \
         --grpc-address :10000 --libp2p :30301 --jsonrpc :10002 --relayer \
         --num-block-confirmations 2 --seal --log-level DEBUG 2>&1 | tee ./validator-1.log &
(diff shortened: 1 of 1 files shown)


scripts/cluster
```diff
@@ -158,9 +158,15 @@ function initRootchain() {
 function startServerFromBinary() {
   if [ "$1" == "write-logs" ]; then
     echo "Writing validators logs to the files..."
-    ./polygon-edge server --data-dir ./test-chain-1 --chain genesis.json \
-      --grpc-address :10000 --libp2p :30301 --jsonrpc :10002 --relayer \
-      --num-block-confirmations 2 --seal --log-level DEBUG 2>&1 | tee ./validator-1.log &
+    if [ "$1" == "polybft" ]; then
```
Stefan-Ethernal: This is not correct, because we are sending `write-logs` flag value to this function (so "$1" contains value for it and not the consensus type).

Figure out if possible setting some global variable denoting consensus type being used or checking the correct flag value.

Note that cluster script can be run in a following ways
`scripts/cluster polybft [write-logs]`, meaning that write-logs is optional parameter.

vcastellm (commented):
It's not clear what's the difference between both commands given the conditional, can we separate params composition to a variable and assign or leave the var empty given the conditions?

Stefan-Ethernal:
Superseded by #1971.

PR #1846, opened 2023-08-23
Perform gas price limit check for dynamic tx

begmaroman:
# Description

Currently, there is a check that compares the gas price of a legacy tx with the gas price limit specified per node argument.
This check should be applicable for dynamic tx as well.

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -661,8 +661,6 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 		}
 	}
 
-	fmt.Println("tx.GetGasPrice(baseFee)", tx.GetGasPrice(baseFee), baseFee)
-
 	// Check if the given tx is not underpriced
 	if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
 		metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -662,7 +662,7 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 	}
 
 	// Check if the given tx is not underpriced
-	if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
+	if tx.GetGasPrice(baseFee).Cmp(new(big.Int).SetUint64(p.priceLimit)) < 0 {
 		metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
 
 		return ErrUnderpriced
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -661,6 +661,8 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 		}
 	}
 
+	fmt.Println("tx.GetGasPrice(baseFee)", tx.GetGasPrice(baseFee), baseFee)
+
 	// Check if the given tx is not underpriced
 	if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
 		metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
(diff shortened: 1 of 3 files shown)


txpool/txpool.go
```diff
@@ -659,13 +659,15 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
...
 		}
-	} else {
-		// Legacy approach to check if the given tx is not underpriced
-		if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
-			metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
+	}
 
-			return ErrUnderpriced
-		}
+	fmt.Println("tx.GetGasPrice(baseFee)", tx.GetGasPrice(baseFee), baseFee)
```
rachit77: Do we want this print statement here?

begmaroman: Removed :)

txpool/txpool.go
```diff
@@ -659,13 +659,13 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
...
 		}
-	} else {
-		// Legacy approach to check if the given tx is not underpriced
-		if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
-			metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
+	}
 
-			return ErrUnderpriced
-		}
+	// Check if the given tx is not underpriced
```
Nemanja0x, line 664: Just to check, we are using baseFee for both legacy and dynamic transactions when we want to check if is it underpriced.

begmaroman: Not really. Base fee value is needed for dynamic fee tx. Take a look at the function implementation to understand how the gas price is calculated for different tx types.

igorcrevar: `GetGasPrice` on the top has this:
```
```
so, for legacy `baseFee` parameter does not change anything

txpool/txpool.go
```diff
@@ -659,13 +659,13 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
 
 			return ErrUnderpriced
 		}
-	} else {
-		// Legacy approach to check if the given tx is not underpriced
-		if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
-			metrics.IncrCounter([]string{txPoolMetrics, "underpriced_tx"}, 1)
```
igorcrevar, line 665: Do we still need on line 657:
```
// Reject underpriced transactions
if tx.GasFeeCap.Cmp(new(big.Int).SetUint64(baseFee)) < 0 {
```

begmaroman: Yes

begmaroman: base fee != gas price limit

txpool/txpool.go
igorcrevar: Can we have tests for both cases in `func Test_TxPool_validateTx(t *testing.T)` ?

Nemanja0x: Now I do not understand where is the part of the code which checks if a legacy transaction is underpriced?

rachit77: `GetGasPrice` has the below condition, when can baseFee be 0?

```
} else if baseFee == 0 {
		return new(big.Int)
	}
```
and Please update the comment on top of `GetGasPrice` method from `- or calculate a value with formula: min(gasFeeCap, gasTipCap * baseFee);` to `- or calculate a value with formula: min(gasFeeCap, gasTipCap + baseFee);`

begmaroman: Updated the comment @rachit77

vcastellm (approved):
LGTM

txpool/txpool.go
```diff
@@ -659,13 +659,13 @@ func (p *TxPool) validateTx(tx *types.Transaction) error {
...
+	}
 
-			return ErrUnderpriced
-		}
+	// Check if the given tx is not underpriced
+	if tx.GetGasPrice(baseFee).Cmp(big.NewInt(0).SetUint64(p.priceLimit)) < 0 {
```
igorcrevar: why not `new(big.Int).SetUInt64(p.PriceLimit)`?

igorcrevar (approved):

rachit77 (approved):

txpool/txpool.go
begmaroman: Same question :)

PR #1855, opened 2023-08-28
Deploy proxy contracts on genesis

stana-miric:
# Description

This PR introduces integration of edge code with proxy smart contracts.

# Changes include

# Breaking changes

# Checklist

## Testing

## Documentation
@DannyS03 parameter proxy-contracts-admin for genesis command is added in this PR. This is address of the account that will be set as an admin from proxy contract and this address will be able to update address implementation(logic) contract


command/genesis/polybft_params.go
```diff
@@ -365,78 +370,78 @@ func (p *genesisParams) deployContracts(
 	polybftConfig *polyCommon.PolyBFTConfig,
 	chainConfig *chain.Chain,
 	burnContractAddr types.Address) (map[types.Address]*chain.GenesisAccount, error) {
-	type contractInfo struct {
-		artifact *artifact.Artifact
-		address  types.Address
+	var proxyAddresses []types.Address
+	for proxyAddr := range contracts.GetProxyImplementationMapping() {
+		proxyAddresses = append(proxyAddresses, proxyAddr)
 	}
```
Stefan-Ethernal: In order to fix the linting warning (define a slice capacity and use append).
WDYT?
```suggestion
	proxyToImplAddrMap := contracts.GetProxyImplementationMapping()
	proxyAddresses := make([]types.Address, 0, len(proxyToImplAddrMap))

	for proxyAddr := range proxyToImplAddrMap {
		proxyAddresses = append(proxyAddresses, proxyAddr)
	}
```

consensus/polybft/polybft.go
```diff
@@ -805,3 +810,40 @@ func (p *Polybft) GetBridgeProvider() consensus.BridgeDataProvider {
 func (p *Polybft) FilterExtra(extra []byte) ([]byte, error) {
```
Stefan-Ethernal, line 814: Unrelated to this PR, but caught my attention...
Can you just fix copy paste error in the comment, it states `GetBridgeProvider` instead of `FilterExtra`.

So it should be like this:
```go
// FilterExtra is an implementation of Consensus interface
```

command/genesis/polybft_params.go
```diff
@@ -615,3 +626,16 @@ func stringSliceToAddressSlice(addrs []string) []types.Address {
...
+
+	for i, proxyAddress := range inputs {
+		res[i] = &contractInfo{
+			artifact: contractsapi.GenesisProxy,
+			address:  proxyAddress,
+		}
+	}
+
+	return res
+}
```
Stefan-Ethernal: I'd consider renaming the function (for example):

```suggestion
func getProxyContractsInfo(addrs []types.Address) []*contractInfo {
	res := make([]*contractInfo, len(addrs))

	for i, proxyAddress := range addrs {
		res[i] = &contractInfo{
			artifact: contractsapi.GenesisProxy,
			address:  proxyAddress,
		}
	}

	return res
}
```

consensus/polybft/polybft.go
```diff
@@ -397,31 +412,21 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 			}
 
 			// initialize EIP1559Burn SC
-			if config.Params.BurnContract != nil &&
-				len(config.Params.BurnContract) == 1 &&
-				!polyBFTConfig.NativeTokenConfig.IsMintable {
-				var contractAddress types.Address
-				for _, address := range config.Params.BurnContract {
-					contractAddress = address
+			if burnContractAddress, ok := getBurnContractAddress(config, polyBFTConfig); ok {
```
goran-ethernal: Can we use here already gotten `burnContractAddress` and `ok` variables, that we got on line 155?

command/genesis/polybft_params.go
stana-miric: fix 8dfce2f

consensus/polybft/polybft.go
stana-miric: fix 8dfce2f

command/genesis/polybft_params.go
stana-miric: fix 8dfce2f

consensus/polybft/polybft.go
stana-miric: fix 8dfce2f

e2e-polybft/framework/test-cluster.go
```diff
@@ -590,6 +598,12 @@ func NewTestCluster(t *testing.T, validatorsCount int, opts ...ClusterOption) *T
 				strings.Join(sliceAddressToSliceString(cluster.Config.BridgeBlockListEnabled), ","))
 		}
 
+		proxyAdminAddr := cluster.Config.ProxyContractsAdmin
+		if proxyAdminAddr == "" {
+			proxyAdminAddr = "0x5aaeb6053f3e94c9b9a09f33669435e7ef1beaed"
```
goran-ethernal: It would be better to put the address of the first validator in cluster as `proxyAdminAddr` if it was not provided by that `WithProxyContractsAdmin`, since we already do this for bunch of other stuff.

stana-miric: We didn't use validator's address because of the possibility that validator might need to call some proxified contract directly and in that case call will fail since the owner of the proxy contract can only call functions related to upgrading admin and implementation address. We can set address of a validator if you think that such calls will not happen?

goran-ethernal: No, no, you are right. I forgot about that. Then I would say, just move this address to some global variable in test-cluster.

command/genesis/params.go
```diff
@@ -554,6 +561,19 @@ func (p *genesisParams) validateBurnContract() error {
...
 
+func (p *genesisParams) validateProxyContractsAdmin() error {
+	if strings.TrimSpace(p.proxyContractsAdmin) == "" {
+		return errors.New("proxy contracts admin address must be set")
+	}
+
+	proxyContractsAdminAddr := types.StringToAddress(p.proxyContractsAdmin)
+	if proxyContractsAdminAddr == types.ZeroAddress {
+		return errors.New("proxy contracts admin address must not be zero address")
+	}
```
Stefan-Ethernal, line 587: Should we prevent the `SystemCaller` address as well? What is your opinion?

consensus/polybft/polybft.go
```diff
@@ -150,6 +150,24 @@ func GenesisPostHookFactory(config *chain.Chain, engineName string) func(txn *st
 			return errMissingBridgeConfig
 		}
 
+		proxyAddr := contracts.GetProxyImplementationMapping()
+
+		var burnContractAddress types.Address
+
+		var isBurnContractSet bool
+
+		if burnContractAddress, isBurnContractSet = getBurnContractAddress(config, polyBFTConfig); isBurnContractSet {
```
Stefan-Ethernal: Can we shorten this a bit (I'm writing by heart, haven't checked in the IDE):
```suggestion
		burnContractAddress, isBurnContractSet := getBurnContractAddress(config, polyBFTConfig)

		if isBurnContractSet {
```

e2e-polybft/framework/test-cluster.go
stana-miric: fix 60fb909

command/genesis/params.go
stana-miric: fix 60fb909

consensus/polybft/polybft.go
stana-miric: fix 60fb909

Stefan-Ethernal (approved):

chain/params.go
```diff
@@ -33,6 +33,10 @@ type Params struct {
 	BurnContract map[uint64]types.Address `json:"burnContract"`
 	// Destination address to initialize default burn contract with
 	BurnContractDestinationAddress types.Address `json:"burnContractDestinationAddress,omitempty"`
+
+	// ProxyContractsAdmin is the address that will have the privilege to change both the proxy
+	// implementation address and the admin
+	ProxyContractsAdmin types.Address `json:"proxyContractsAdmin,omitempty"`
```
goran-ethernal: I would move this to `PolyBFTConfig` struct. This is not something that is shared with `IBFT`, it's pure Polybft thing.

stana-miric: fix 5fc3f68

goran-ethernal (approved):

PR #1857, opened 2023-08-29
Build executable heap with baseFee when London fork is enabled

rachit77:
# Description

This PR intends to solve the issue explained [here](https://polygon.atlassian.net/browse/EVM-809)

# Changes include

# Checklist

## Testing

### DIFF
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -1,7 +1,6 @@
 package txpool
 
 import (
-	"math/big"
 	"sync"
 	"sync/atomic"
 
@@ -45,7 +44,7 @@ func (m *accountsMap) exists(addr types.Address) bool {
 
 // getPrimaries collects the heads (first-in-line transaction)
 // from each of the promoted queues.
-func (m *accountsMap) getPrimaries(baseFee uint64) (primaries []*types.Transaction) {
+func (m *accountsMap) getPrimaries() (primaries []*types.Transaction) {
 	m.Range(func(key, value interface{}) bool {
 		addressKey, ok := key.(types.Address)
 		if !ok {
@@ -58,7 +57,7 @@ func (m *accountsMap) getPrimaries(baseFee uint64) (primaries []*types.Transacti
 		defer account.promoted.unlock()
 
 		// add head of the queue
-		if tx := account.promoted.peek(); tx != nil && tx.GetGasFeeCap().Cmp(new(big.Int).SetUint64(baseFee)) >= 0 {
+		if tx := account.promoted.peek(); tx != nil {
 			primaries = append(primaries, tx)
 		}
 
--- a/txpool/account.go
+++ b/txpool/account.go
@@ -1,6 +1,7 @@
 package txpool
 
 import (
+	"math/big"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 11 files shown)


Stefan-Ethernal (commented):
Can you figure out some kind of tests?

Also, I believe the currently failing E2E tests are related to this change (at least the tx pool test and migration one: https://github.com/0xPolygon/polygon-edge/actions/runs/6009379628/job/16298691007?pr=1857).

begmaroman:
I suggest to make the PR title more usefull

txpool/account.go
```diff
@@ -57,7 +58,7 @@ func (m *accountsMap) getPrimaries() (primaries []*types.Transaction) {
 		defer account.promoted.unlock()
 
 		// add head of the queue
-		if tx := account.promoted.peek(); tx != nil {
+		if tx := account.promoted.peek(); tx != nil && tx.GetGasFeeCap().Cmp(new(big.Int).SetUint64(baseFee)) >= 0 {
```
igorcrevar: what do you think to move this check `tx.GetGasFeeCap().Cmp(new(big.Int).SetUint64(baseFee))` to some function inside `transaction.go` - something like `CheckGasFeeCap(baseFee) bool`? then you can replace calculations on bunch of places. and maybe have `*big.Int` as parameter instead of `uint64` (later on we can change `baseFee` to `*big.Int `on other places)

igorcrevar (approved):
write some tests and check my other comment (changes for that are not mandatory).
otherwise lgtm

txpool/account.go
rachit77: This condition just used twice and it will be even simpler after EIP-2718. I think we should not extract this to a function.

dusan-maksimovic (approved):

txpool/txpool.go
```diff
@@ -335,7 +335,7 @@ func (p *TxPool) AddTx(tx *types.Transaction) error {
 // ready for execution. (primaries)
 func (p *TxPool) Prepare() {
 	// fetch primary from each account
-	primaries := p.accounts.getPrimaries()
+	primaries := p.accounts.getPrimaries(p.GetBaseFee())
 
 	// create new executables queue with base fee and initial transactions (primaries)
 	p.executables = newPricesQueue(p.GetBaseFee(), primaries)
```
Stefan-Ethernal: Nit: retrieve `baseFee` to local var and reuse it.
```suggestion
        baseFee := p.GetBaseFee()
	primaries := p.accounts.getPrimaries(baseFee)
	// create new executables queue with base fee and initial transactions (primaries)
	p.executables = newPricesQueue(baseFee, primaries)
```

txpool/account.go
```diff
@@ -44,7 +45,7 @@ func (m *accountsMap) exists(addr types.Address) bool {
 
 // getPrimaries collects the heads (first-in-line transaction)
 // from each of the promoted queues.
-func (m *accountsMap) getPrimaries() (primaries []*types.Transaction) {
+func (m *accountsMap) getPrimaries(baseFee uint64) (primaries []*types.Transaction) {
 	m.Range(func(key, value interface{}) bool {
 		addressKey, ok := key.(types.Address)
```
Stefan-Ethernal: As a side note, not related to this PR:
I think that we could just make type conversion of `value` parameter into `*account`, instead of calling 
`account := m.get(addressKey)` unnecesarily. WDYT, am I missing out something?

Stefan-Ethernal (approved):
Generally LGTM, although I'd like to see some new tests implemented if possible.

vcastellm (approved):
LGTM

txpool/account.go
rachit77: that's a good point, i also think that conversion is a redundant overhead here

igorcrevar (approved):

begmaroman (approved):

igorcrevar (approved):

PR #1863, opened 2023-08-30
Return all exit event IDs in the bridge commands

jelacamarko:
# Description

`ExtractExitEventID` function is changed, so now it returns an array of all exit events IDs, that are matched from receipt logs.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/command/bridge/common/params.go
+++ b/command/bridge/common/params.go
@@ -142,13 +142,13 @@ func (bp *ERC1155BridgeParams) Validate() error {
 	return nil
 }
 
-// ExtractExitEventID tries to extract exit event id from provided receipt
-func ExtractExitEventID(receipt *ethgo.Receipt) ([]*big.Int, error) {
-	var exitEvent contractsapi.L2StateSyncedEvent
-
+// ExtractExitEventIDs tries to extract all exit event ids from provided receipt
+func ExtractExitEventIDs(receipt *ethgo.Receipt) ([]*big.Int, error) {
 	exitEventIDs := make([]*big.Int, 0, len(receipt.Logs))
 
 	for _, log := range receipt.Logs {
+		var exitEvent contractsapi.L2StateSyncedEvent
+
 		doesMatch, err := exitEvent.ParseLog(log)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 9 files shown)


command/bridge/common/params.go
```diff
@@ -155,7 +158,11 @@ func ExtractExitEventID(receipt *ethgo.Receipt) (*big.Int, error) {
 			continue
 		}
 
-		return exitEvent.ID, nil
+		exitEventIDs = append(exitEventIDs, exitEvent.ID)
```
Stefan-Ethernal, line 161: You should not reuse the `exitEvent` instance, because then the next iteration sets the next exit event id and it gets rewritten in the slice (because the ID field is a `*big.Int`, namely a pointer).

So to me, you have two options:

1. reuse the same exitEvent instance, but create a new instance of big.Int when adding it into the `exitEventIDs` slice:
```go
exitEventIDs = append(exitEventIDs, new(big.Int).Set(exitEvent.ID))
```
2. make sure that each iteration holds its `exitEvent` reference within the for loop so that each iteration works with the fresh reference

This is the output extracted from the `bridge.log` of `TestE2E_Bridge_ChildChainMintableTokensTransfer`
```go
[DEPOSIT ERC 20]
Sender                  = 0x735CCeFD8dd837A99957a2a8833369dc9AfC342e
Receivers               = 0x735CCeFD8dd837A99957a2a8833369dc9AfC342e
Amounts                 = 100
Exit Event IDs          = 2, 2 // <-- here should be 1, 2
Inclusion Block Numbers = 12
Child Token Address     = 0xf2111e7bC4707A7E830F52C45FfFd8fF97E3707E
```

command/bridge/common/params.go
```diff
@@ -143,8 +143,11 @@ func (bp *ERC1155BridgeParams) Validate() error {
 }
 
 // ExtractExitEventID tries to extract exit event id from provided receipt
-func ExtractExitEventID(receipt *ethgo.Receipt) (*big.Int, error) {
+func ExtractExitEventID(receipt *ethgo.Receipt) ([]*big.Int, error) {
```
Stefan-Ethernal: Rename it and update the comment accordingly

```suggestion
func ExtractExitEventIDs(receipt *ethgo.Receipt) ([]*big.Int, error) {
```

stana-miric (approved):

Stefan-Ethernal (approved):

dusan-maksimovic (approved):

PR #1887, opened 2023-09-07
Introduce block tracker poll interval

goran-ethernal:
# Description

Some of the clients experienced an issue with block tracker (which is a part of `ethgo` third party library), where on all validator nodes, it polled rootchain rpc endpoints `eth_getBlockByNumber` and `eth_getBlockByHash` each second, causing them to spend all their allowed limit for querying rpc points on a main net such as Ethereum. This sequentially caused all other bridge actions to fail such as state sync events for deposits, or checkpointing. This issue only persisted for block trackers on validator nodes that track the state sync events (which transfer data from L1 to L2).

Block tracker in `ethgo` has a `PollInterval` (in `time.Duration`), which determines how often it calls given rpc endpoints, but it's configuration was not exposed explicitly, so we needed to expand our `event_tracker` to receive an interval, which then is set to the block tracker through its field.

This PR introduces two new flags:

1. In `genesis` command - we introduced a new flag `--block-tracker-poll-interval`, which is defined in time duration (for example, `--block-tracker-poll-interval "1s"` or `block-tracker-poll-interval "12s"`). This is part of the `genesis` command and `genesis.json` file because, all validator nodes are tracking state sync events. Block proposers are the ones that add state sync commitment transactions at the end of a sprint, while the rest of the validators validate the given transaction, hence all validators need to track state sync events from rootchain. For a state sync commitment transaction to be valid, quorum of validators must be reached, meaning at lest 2/3+1 validators must have the same state sync events tracked. So if we are to set this flag on a server level, that means that each validator can have a different poll time, and will have different number of state sync events tracked at a point of time, which is not a desired behavior.
2. In `server` command - we introduced a new flag `--relayer-poll-interval`, which is defined in time duration (for example, example, `--relayer-poll-interval "1s"` or `--relayer-poll-interval "12s"`). Since `relayer` node uses block tracker as well, but for child chain (tracking when a state sync commitment is submitted to child contract), we added this flag, so that the user can configure their `relayer` nodes as well to poll given rpc endpoints on child chain at desired intervals.

Tagging @DannyS03 to update the documentation.

# Changes include

# Checklist

## Testing


command/server/server.go
```diff
@@ -242,6 +242,13 @@ func setFlags(cmd *cobra.Command) {
 		"maximum size in bytes for a message read from the peer by websocket",
 	)
 
+	cmd.Flags().DurationVar(
+		&params.rawConfig.RelayerTrackerPollInterval,
+		relayerTrackerPollIntervalFlag,
+		defaultConfig.RelayerTrackerPollInterval,
+		"interval (number of seconds) at which relayer's tracker polls for latest block at childchain",
```
igorcrevar, line 249: isn't this for rootchain?

goran-ethernal: We have two flags. This particular flag you are referring is for relayer node, which uses block tracker to track `NewCommitment` events on child chain.

igorcrevar (approved):

tracker/event_tracker.go
```diff
@@ -36,6 +37,7 @@ func NewEventTracker(
 	numBlockConfirmations uint64,
 	startBlock uint64,
 	logger hcf.Logger,
+	pollInterval common.Duration,
```
Stefan-Ethernal: Why you haven't put `time.Duration`? `common.Duration` is there solely for JSON (de)serialization
```suggestion
	pollInterval time.Duration,
```

consensus/polybft/state_sync_manager.go
```diff
@@ -53,22 +54,23 @@ func (n *dummyStateSyncManager) Close()      {}
...
-	numBlockConfirmations uint64
+	stateSenderAddr          types.Address
+	stateSenderStartBlock    uint64
+	jsonrpcAddr              string
+	dataDir                  string
+	topic                    topic
+	key                      *wallet.Key
+	maxCommitmentSize        uint64
+	numBlockConfirmations    uint64
+	blockTrackerPollInterval common.Duration
```
Stefan-Ethernal: WDYT?
```suggestion
	blockTrackerPollInterval time.Duration
```

command/server/config/config.go
```diff
@@ -37,6 +38,8 @@ type Config struct {
 
 	ConcurrentRequestsDebug uint64 `json:"concurrent_requests_debug" yaml:"concurrent_requests_debug"`
 	WebSocketReadLimit      uint64 `json:"web_socket_read_limit" yaml:"web_socket_read_limit"`
+
+	RelayerTrackerPollInterval time.Duration `json:"relayer_tracker_poll_interval" yaml:"relayer_tracker_poll_interval"`
```
Stefan-Ethernal: Group it together with `Relayer` and `NumBlockConfirmations` fields, because it is related to relayer node(s). WDYT?

Also should this be `common.Duration` type maybe?

consensus/polybft/common/polybft_config.go
```diff
@@ -63,6 +63,10 @@ type PolyBFTConfig struct {
 
 	// GovernanceConfig defines on chain governance configuration
 	GovernanceConfig *GovernanceConfig `json:"governanceConfig,omitempty"`
+
+	// BlockTrackerPollInterval specifies interval
+	// at which block tracker polls for blocks on a rootchain
+	BlockTrackerPollInterval common.Duration `json:"blockTrackerPollInterval,omitempty"`
```
Stefan-Ethernal, line 74: Just thinking out loud, is it an overkill to rename and expand `EventTrackerStartBlocks` <address>:<event_tracker_config> (EventTrackerConfig would be a new struct, that would hold `StartingBlock` and `PollInterval` fields), so that we are able to configure different poll intervals for different contracts?

That way we could even hold parameter for relayer poll interval.

consensus/polybft/statesyncrelayer/state_sync_relayer.go
```diff
@@ -30,6 +32,7 @@ type StateSyncRelayer struct {
 	txRelayer              txrelayer.TxRelayer
 	key                    ethgo.Key
 	closeCh                chan struct{}
+	pollInterval           common.Duration
```
Stefan-Ethernal: time.Duration?

```suggestion
	pollInterval           time.Duration
```

command/server/server.go
```diff
@@ -242,6 +242,13 @@ func setFlags(cmd *cobra.Command) {
 		"maximum size in bytes for a message read from the peer by websocket",
 	)
 
+	cmd.Flags().DurationVar(
+		&params.rawConfig.RelayerTrackerPollInterval,
```
Stefan-Ethernal, line 246: Should we validate it, no need to specify it in case `relayer` flag is not provided?

Stefan-Ethernal (approved):
Generally LGTM

command/server/config/config.go
goran-ethernal: I don't think this needs to be common.Duration. I am not even sure if this is even saved to some file.

command/server/server.go
goran-ethernal: I think it is enough to validate it only for relayer node. If it is not a relayer node, it doesn't matter what the user specifies, because it won't be used.

tracker/event_tracker.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1887/commits/f47c9ae2d3128f5fa8fa5bd00b49ef665ae38f18

consensus/polybft/state_sync_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1887/commits/f47c9ae2d3128f5fa8fa5bd00b49ef665ae38f18

command/server/config/config.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1887/commits/f47c9ae2d3128f5fa8fa5bd00b49ef665ae38f18

consensus/polybft/statesyncrelayer/state_sync_relayer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1887/commits/f47c9ae2d3128f5fa8fa5bd00b49ef665ae38f18

command/server/server.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1887/commits/f47c9ae2d3128f5fa8fa5bd00b49ef665ae38f18

consensus/polybft/common/polybft_config.go
goran-ethernal: Hmmm, we would break the already run chains. Let's wait for a confirmation if 1.2 is a breaking change. If not, then I'd say it's safe to change this.

PR #1888, opened 2023-09-07
Root chain proxy contracts deployment

dusan-maksimovic:
# Description

This PR deploys proxy contract on the root chain, for all our contracts except token templates and StateSender contract. StateSender is not intended to be upgradeable (info from the SCs team), so it is not proxified.

# Changes include

# Breaking changes

# Checklist

## Testing


command/helper/helper.go
```diff
@@ -260,3 +262,16 @@ func ParseAmount(amount string) (*big.Int, error) {
 
 	return result, nil
 }
+
+func ValidateProxyContractsAdmin(proxyContractsAdmin string) error {
+	if strings.TrimSpace(proxyContractsAdmin) == "" {
```
goran-ethernal:
```suggestion
	if err := types.IsValidAddress(p.proxyContractsAdmin); err != nil {
		return fmt.Errorf("proxy contracts admin address is not valid: %w", err)
	}
```

command/rootchain/deploy/deploy.go
```diff
@@ -546,43 +573,72 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
...
-				commandResults[i] = result
-			}
+		return collectResultsOnError(results), err
+	}
+
+	proxyAdmin := types.StringToAddress(params.proxyContractsAdmin)
+
+	g, ctx = errgroup.WithContext(cmdCtx)
+
+	for _, contract := range allContracts {
```
goran-ethernal: Just curious. Is deploying proxies separately, in another routine, done because of an optimization, or some other reason?

dusan-maksimovic: Yes- proxy contract constructor accepts implementation contract address as argument, so we need to deploy implementation contracts first.

command/helper/helper.go
dusan-maksimovic: fixed 0ef41b8e1de5eb3745e115ad043f730d112f4537

command/rootchain/deploy/deploy.go
goran-ethernal: I was thinking can we deploy a contract and its proxy in single routine. I don't know what are the differences in execution time if we deploy proxy in the same routine as contract, and separately. If the difference is significant then leave it like this, if it is not, then it will be simpler to just deploy it in the same routine as contract.

dusan-maksimovic: Ah, I didn't understand your question properly. Yes we could do that. I guess the difference wouldn't be noticeable, so I will do it.

dusan-maksimovic: Fixed 63b8373897db3bd9a21f2245d0dfe9aab726f29e

goran-ethernal (approved):

Nemanja0x (approved):

command/rootchain/supernet/stakemanager/params.go
```diff
@@ -30,5 +31,9 @@ func (s *stakeManagerDeployParams) validateFlags() error {
 	// validate jsonrpc address
 	_, err := helper.ParseJSONRPCAddress(s.jsonRPC)
 
+	if err := helper.ValidateProxyContractsAdmin(s.proxyContractsAdmin); err != nil {
+		return err
+	}
+
 	return err
```
Stefan-Ethernal: This part is incorrect, because we are not returning an error in case `helper.ParseJSONRPCAddress` fails

```suggestion
	_, err := helper.ParseJSONRPCAddress(s.jsonRPC)
    if err != nil {
        return err
    }

	return helper.ValidateProxyContractsAdmin(s.proxyContractsAdmin)
```

command/rootchain/deploy/params.go
```diff
@@ -49,5 +51,9 @@ func (ip *deployParams) validateFlags() error {
 		return errors.New("stake token address is not provided")
 	}
 
+	if err := helper.ValidateProxyContractsAdmin(ip.proxyContractsAdmin); err != nil {
+		return err
+	}
+
 	return nil
```
Stefan-Ethernal: Nitpick:
```suggestion
	    return helper.ValidateProxyContractsAdmin(ip.proxyContractsAdmin)
```

Stefan-Ethernal:
Should we premine/fund rootchain proxy contract admin in cluster script (and other deployment scripts that are maintained by dev tools team)?

command/rootchain/deploy/deploy.go
```diff
@@ -506,17 +530,18 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
...
+			hasProxy: true,
 		},
 	}
 
 	allContracts = append(tokenContracts, allContracts...)
 
 	g, ctx := errgroup.WithContext(cmdCtx)
-	results := make([]*deployContractResult, len(allContracts))
-	commandResults := make([]command.CommandResult, len(allContracts))
+	results := make(map[string]*deployContractResult, len(allContracts))
```
Stefan-Ethernal, line 540: Any reason why this is a map now? As far as I can see, we are not relying on contracts names in the results. 

The previous approach with slice was enabling us that each routine writes within the single index thus we could avoid mutex usage...

Stefan-Ethernal (approved):

command/rootchain/deploy/deploy.go
dusan-maksimovic: We changed it to map when we were deploying proxies in separate goroutine, so we used the contract name to get result entry to obtain the address of the deployed implementation contract. But then Goran suggested we could do this in the same goroutine, so we changed that, but we left out the map. In any case, allContracts now doesn't contain all contracts- since proxies don't have their instances in this slice.
EDIT: The later fact makes it impossible to switch back to slice without maintaining and locking some additional structure where we would update indices that were taken by proxy results as they are deployed. If you agree I would keep the map?

command/rootchain/supernet/stakemanager/params.go
dusan-maksimovic: Fixed 18372dcc66dbf29850da56b0164cc578f19bbdbc
It's strange that linter didn't catch this one.

command/rootchain/deploy/params.go
dusan-maksimovic: Fixed 18372dcc66dbf29850da56b0164cc578f19bbdbc

PR #1890, opened 2023-09-07
Base Fee Configs

rachit77:
# Description

This PR adds the functionality to provide initial baseFee, baseFeeEM at the time of genesis file creation.

# Changes include

# Checklist

## Testing


command/genesis/polybft_params.go
```diff
@@ -361,7 +361,7 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
 	if p.isBurnContractEnabled() {
 		// only populate base fee and base fee multiplier values if burn contract(s)
 		// is provided
-		chainConfig.Genesis.BaseFee = command.DefaultGenesisBaseFee
+		chainConfig.Genesis.BaseFee = p.baseFee
 		chainConfig.Genesis.BaseFeeEM = command.DefaultGenesisBaseFeeEM
```
igorcrevar: Why don't specify flag for `BaseFeeEM` also? Maybe we should allow `--genesis-base-fee` flag to specify both values like `baseFee[:baseFeeEM]` (two values separated with :, by default `baseFeeEM` is not required). what do you think?

rachit77: That's a really good point, will change the code accordingly.

command/genesis/genesis.go
```diff
@@ -77,6 +77,13 @@ func setFlags(cmd *cobra.Command) {
 		"the burn contract block and address (format: <block>:<address>[:<burn destination>])",
 	)
 
+	cmd.Flags().Uint64Var(
+		&params.baseFee,
+		genesisBaseFeeFlag,
+		command.DefaultGenesisBaseFee,
+		"initial base fee for the chain in wei, Default BaseFee is 1 Gwei (London fork is disabled when baseFee is 0)",
```
Stefan-Ethernal: I didn't know that London fork is disabled if `baseFee` is set to 0. I thought that London fork is enabled only if `burn-contract` flag is provided. Should we add validation that `baseFee` must be greater than 0 in case London fork is enabled?

Stefan-Ethernal (approved):
Aside from Igor's comment and minor comment from my end LGTM.

command/genesis/utils.go
```diff
@@ -164,6 +164,37 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polyCommon.BurnContract
...
+	baseFee   uint64
+	baseFeeEM uint64
+}
+
+// parseBaseFeeConfig parses provided base fee configuration and returns baseFeeInfo
+func parseBaseFeeConfig(baseFeeConfigRaw string) *baseFeeInfo {
+	baseFeeInfo := &baseFeeInfo{command.DefaultGenesisBaseFee, command.DefaultGenesisBaseFeeEM}
+
+	baseFeeConfig := strings.Split(baseFeeConfigRaw, ":")
+	if len(baseFeeConfig) != 2 {
```
stana-miric: what if we only want to set BaseFee without setting BaseFeeFM when specifying genesis-base-fee-config? do we want to support that as well or the user must specify BaseFeeFM  even if it is equal to default?

stana-miric (approved):

command/genesis/utils.go
```diff
@@ -164,6 +164,41 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polyCommon.BurnContract
...
+	baseFee   uint64
+	baseFeeEM uint64
+}
+
+// parseBaseFeeConfig parses provided base fee configuration and returns baseFeeInfo
+func parseBaseFeeConfig(baseFeeConfigRaw string) (*baseFeeInfo, error) {
+	baseFeeInfo := &baseFeeInfo{command.DefaultGenesisBaseFee, command.DefaultGenesisBaseFeeEM}
+
+	baseFeeConfig := strings.Split(baseFeeConfigRaw, ":")
+	if len(baseFeeConfig) > 2 {
```
igorcrevar: I think we should check if there is more than one `:` (two values) in validate

igorcrevar: And ` &baseFeeInfo{command.DefaultGenesisBaseFee, command.DefaultGenesisBaseFeeEM}` we should return only if flag is not specified

igorcrevar (approved):

command/genesis/params.go
```diff
@@ -576,6 +583,27 @@ func (p *genesisParams) validateBurnContract() error {
...
+	if p.baseFeeConfig == "" {
+		return errors.New("genesis base fee config must be set")
+	}
+
+	baseFeeInfo, err := parseBaseFeeConfig(p.baseFeeConfig)
+	if err != nil {
+		return err
+	}
+
+	if baseFeeInfo.baseFee <= 0 {
```
dusan-maksimovic: Since it is uint64 it can't be less than 0. Same for `baseFeeInfo.baseFeeEM`. Plus `parseBaseFeeConfig()` would return an error if negative values are provided.

command/genesis/genesis.go
```diff
@@ -77,6 +77,15 @@ func setFlags(cmd *cobra.Command) {
 		"the burn contract block and address (format: <block>:<address>[:<burn destination>])",
 	)
 
+	cmd.Flags().StringVar(
+		&params.baseFeeConfig,
+		genesisBaseFeeConfigFlag,
+		command.DefaultGenesisBaseFeeConfig,
+		`initial base fee(in wei) and baseFeeEM (format: <baseFee>[:<baseFeeEM>]).
```
stana-miric: I would make both of these values optional parameters since we can set default values if the user does not specify any of them and remove the mandatory validation that the field must be set (which is added in validateGenesisBaseFeeConfig). I don't know if others agree?

rachit77: @stana-miric Do you suggest that if a user provides an empty string as value to `genesis-base-fee-config` flag than we should use default values instead of throwing error?

That check in `validateGenesisBaseFeeConfig` is for throwing error when uses gives a empty string as input.

stana-miric: yes, in my opinion that would be ok since we have default values that can be used. I would make the flag mandatory only in case when we don't know what to set if the value is not specified but that is not the case with this flag. Don't know if others agree with that?

command/genesis/polybft_params.go
```diff
@@ -367,8 +367,9 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
 	if p.isBurnContractEnabled() {
 		// only populate base fee and base fee multiplier values if burn contract(s)
 		// is provided
-		chainConfig.Genesis.BaseFee = command.DefaultGenesisBaseFee
-		chainConfig.Genesis.BaseFeeEM = command.DefaultGenesisBaseFeeEM
+		baseFeeInfo, _ := parseBaseFeeConfig(p.baseFeeConfig)
```
Stefan-Ethernal: Consider logging an error message here and `p.baseFeeConfig` raw string, so that we are aware why we are running default configuration in case of some errors in `parseBaseFeeConfig`.

command/genesis/genesis.go
Stefan-Ethernal: I'd suggest slight rename here:

```suggestion
		`initial base fee(in wei) and base fee elasticity multiplier (provided in the following format: <baseFee>[:<baseFeeEM>]). 
```

command/genesis/params.go
```diff
@@ -438,8 +444,9 @@ func (p *genesisParams) initGenesisConfig() error {
 
 	// burn contract can be set only for non mintable native token
 	if p.isBurnContractEnabled() {
-		chainConfig.Genesis.BaseFee = command.DefaultGenesisBaseFee
-		chainConfig.Genesis.BaseFeeEM = command.DefaultGenesisBaseFeeEM
+		baseFeeInfo, _ := parseBaseFeeConfig(p.baseFeeConfig)
```
Stefan-Ethernal: Consider logging an error message here and `p.baseFeeConfig` raw string, so that we are aware why we are running default configuration in case of some errors in `parseBaseFeeConfig`.

rachit77: @Stefan-Ethernal I have logged the error message  in `validateGenesisBaseFeeConfig`. 
Default configuration is only selected when `genesis-base-fee-config` flag is not used while genesis creation and when value of either baseFee or baseFeeEM is not provided in the flag.
Do you think is it still necessary to log error here?

command/genesis/params.go
```diff
@@ -576,6 +583,27 @@ func (p *genesisParams) validateBurnContract() error {
 	return nil
 }
 
+func (p *genesisParams) validateGenesisBaseFeeConfig() error {
+	if p.baseFeeConfig == "" {
+		return errors.New("genesis base fee config must be set")
+	}
```
Stefan-Ethernal: What is the point in insisting `baseFeeConfig` to be mandatory, when it is related to EIP-1559 (which means that if London hard fork is not enabled, we don't need base fee). IMO it should not be a mandatory flag.

On the other hand, I'm not sure should we default it in that case and persist it in the `genesis.json` or completely omit it?

command/genesis/params.go
Stefan-Ethernal: I see, can you just return provided `p.baseFeeConfig` string together with error when propagating an error from `validateGenesisBaseFeeConfig`?

command/genesis/params.go
rachit77: `baseFeeConfig` is not mandatory. This is the case when user gives a empty string as input in that case we should log an error and not use default values. Default values will be used in case no `genesis-base-fee-config` flag is used while genesis creation.

The error is string is quite misleading, I will change it.

Stefan-Ethernal:
Not sure why we have not merged this PR? 

@rachit77 Anyways, now that #1934 got merged, it is worth considering should we provide `baseFeeChangeDenom` configuration parameter with `baseFeeConfig` as well. WDYT?

command/genesis/params.go
```diff
@@ -34,6 +34,7 @@ const (
 	epochRewardFlag              = "epoch-reward"
 	blockGasLimitFlag            = "block-gas-limit"
 	burnContractFlag             = "burn-contract"
+	genesisBaseFeeConfigFlag     = "genesis-base-fee-config"
```
Stefan-Ethernal: Should we shorten the flag name just to `base-fee-config`?

command/genesis/params.go
```diff
@@ -566,6 +568,31 @@ func (p *genesisParams) validateBurnContract() error {
 	return nil
 }
 
+func (p *genesisParams) validateGenesisBaseFeeConfig() error {
+	if p.baseFeeConfig == "" {
+		return errors.New("invalid input(empty string) for genesis base fee config flag")
+	}
+
+	baseFeeInfo, err := parseBaseFeeConfig(p.baseFeeConfig)
```
Stefan-Ethernal, line 576: Does it make sense to store this instance in the `genesisParams` field, so that we do not parse it twice, but just here and re-use it when command is being run?

command/genesis/genesis.go
```diff
@@ -77,6 +77,17 @@ func setFlags(cmd *cobra.Command) {
 		"the burn contract block and address (format: <block>:<address>[:<burn destination>])",
 	)
 
+	cmd.Flags().StringVar(
+		&params.baseFeeConfig,
+		genesisBaseFeeConfigFlag,
+		command.DefaultGenesisBaseFeeConfig,
+		`initial base fee(in wei) and base fee elasticity multiplier
```
Stefan-Ethernal:
```suggestion
		`initial base fee (in wei), base fee elasticity multiplier, and base fee change denominator
```

command/genesis/genesis.go
```diff
@@ -77,6 +77,17 @@ func setFlags(cmd *cobra.Command) {
...
+		(provided in the following format: [<baseFee>][:<baseFeeEM>][:<baseFeeChangeDenom>]). 
+		BaseFeeChangeDenom represents the value to bound the amount the base fee can change between blocks.
+		Default BaseFee is 1 Gwei, BaseFeeEM is 2 and BaseFeeChangeDenom is 8.
+		Note: BaseFee, BaseFeeEM BaseFeeChangeDenom and should be greater than 0.`,
```
Stefan-Ethernal:
```suggestion
		Note: BaseFee, BaseFeeEM, and BaseFeeChangeDenom should be greater than 0.`,
```

command/genesis/params.go
```diff
@@ -174,6 +171,10 @@ func (p *genesisParams) validateFlags() error {
 			return err
 		}
 
+		if err := p.validateGenesisBaseFeeConfig(); err != nil {
+			return err
+		}
```
Stefan-Ethernal: This should be done for the IBFT as well, right?

command/genesis/polybft_params.go
```diff
@@ -308,8 +307,10 @@ func (p *genesisParams) generatePolyBftChainConfig(o command.OutputFormatter) er
...
+		chainConfig.Genesis.BaseFee = baseFeeInfo.baseFee
+		chainConfig.Genesis.BaseFeeEM = baseFeeInfo.baseFeeEM
```
Stefan-Ethernal: Wanted to check (and this probably needs broader consensus within the team) should we put `BaseFeeEM` to the `chain.Params` struct? It seems like it is more natural to have it there, since it is rather a configuration parameter. However it still needs checking whether it is feasible to do and are there any unwanted side effects.

rachit77: Added `BaseFeeChangeDenom` to `chain.Genesis` struct to make it consistent with `BaseFeeEM` and `BaseFee`

command/genesis/utils.go
```diff
@@ -163,6 +164,55 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polybft.BurnContractInf
...
+		baseFeeEM, err := strconv.ParseUint(baseFeeConfig[1], 10, 64)
+		if err != nil {
+			return baseFeeInfo, err
+		}
+
+		baseFeeInfo.baseFeeEM = baseFeeEM
+	}
+
+	if len(baseFeeConfig) == 3 && baseFeeConfig[2] != "" {
+		baseFeeChangeDenom, err := strconv.ParseUint(baseFeeConfig[2], 10, 64)
```
Stefan-Ethernal: Optional nitpick:

```suggestion
		baseFeeChangeDenom, err := common.ParseUint64orHex(&baseFeeConfig[2])
```

command/genesis/utils.go
```diff
@@ -163,6 +164,55 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polybft.BurnContractInf
...
+		command.DefaultGenesisBaseFeeEM,
+		command.DefaultGenesisBaseFeeChangeDenom,
+	}
+
+	baseFeeConfig := strings.Split(baseFeeConfigRaw, ":")
+	if len(baseFeeConfig) > 3 {
+		return baseFeeInfo, errors.New("invalid number of arguments for base fee configuration")
+	}
+
+	if baseFeeConfig[0] != "" {
```
Stefan-Ethernal: Prevent index out of range:

```suggestion
	if len(baseFeeConfig) >= 1 && baseFeeConfig[0] != "" {
```

command/genesis/utils.go
```diff
@@ -163,6 +164,55 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polybft.BurnContractInf
...
+		baseFee, err := strconv.ParseUint(baseFeeConfig[0], 10, 64)
+		if err != nil {
+			return baseFeeInfo, err
+		}
+
+		baseFeeInfo.baseFee = baseFee
+	}
+
+	if len(baseFeeConfig) == 2 && baseFeeConfig[1] != "" {
```
Stefan-Ethernal: Beware, in case `baseFeeConfig` consists of the 3 parts (base fee, base fee EM and base fee change denominator), only `baseFeeChangeDenom` is going to be populated (this if would not be satisfied):

```suggestion
	if len(baseFeeConfig) >= 2 && baseFeeConfig[1] != "" {
```

command/genesis/utils.go
```diff
@@ -163,6 +164,55 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polybft.BurnContractInf
...
+		command.DefaultGenesisBaseFeeChangeDenom,
+	}
+
+	baseFeeConfig := strings.Split(baseFeeConfigRaw, ":")
+	if len(baseFeeConfig) > 3 {
+		return baseFeeInfo, errors.New("invalid number of arguments for base fee configuration")
+	}
+
+	if baseFeeConfig[0] != "" {
+		baseFee, err := strconv.ParseUint(baseFeeConfig[0], 10, 64)
```
Stefan-Ethernal: Optional nitpick:

```suggestion
		baseFee, err := common.ParseUint64orHex(&baseFeeConfig[0])
```

command/genesis/utils.go
```diff
@@ -163,6 +164,55 @@ func parseBurnContractInfo(burnContractInfoRaw string) (*polybft.BurnContractInf
...
+		if err != nil {
+			return baseFeeInfo, err
+		}
+
+		baseFeeInfo.baseFee = baseFee
+	}
+
+	if len(baseFeeConfig) == 2 && baseFeeConfig[1] != "" {
+		baseFeeEM, err := strconv.ParseUint(baseFeeConfig[1], 10, 64)
```
Stefan-Ethernal: Optional nitpick:

```suggestion
		baseFeeEM, err := common.ParseUint64orHex(&baseFeeConfig[1])
```

PR #1891, opened 2023-09-07
fix region to eu-west-1 for nightly

gatsbyz:


.github/workflows/nightly.yml
```diff
@@ -48,7 +48,8 @@ jobs:
     name: Build Devnet
     uses: ./.github/workflows/deploy.nightly.devnet.yml
     secrets:
-      AWS_REGION: ${{ secrets.AWS_REGION }}
```
gatsbyz: seems like the `secrets.AWS_REGION` is set as `us-west-2` for the whole repo, so i can't change the actual secret value. wondering if i should create a temporary `AWS_REGION_NIGHTLY` for nightly builds

Stefan-Ethernal: @vcastellm do you have permission to administer secrets and if yes, is it safe to change it now to `eu-west-1`?

Stefan-Ethernal (commented):
@vcastellm @gatsbyz there are other GH workflows that rely on `AWS_REGION` env variable, so should we change it there as well?

Stefan-Ethernal (approved):
Other from comments LGTM.

vcastellm (approved):

PR #1899, opened 2023-09-08
EVM-831 Implement state sync relayer without blocktracker

### LINKED ISSUE #1811: Failed state sync executions are not retried
## Description

We have observed that failed state sync executions are not retried by Edge. Its hard to tell based on the src whether this is intended or not.

Is this expected behaviour? I understand that some errors around state sync execution cannot be retried successfully without some intervention first, but some errors (like the one below) could be transient in nature.

The price limit cause is just an example and I expect we would see the same behaviour if we caused this behaviour by different means. I believe we saw the same behaviour for a token that we had not mapped (stateSyncResult event contained msg UNMAPPED_TOKEN)

## Your environment
- Edge in Docker based on image from this repo
- v1.1.0
- Both local and on k8s

## Steps to reproduce

This is the process to recreate this behaviour (we did this on cluster and locally to verify):
1. Bootstrap the supernet as per usual w/ polygon-edge CLI
2. Perform deposit to relayer addr
3. Run validators with --price-limit 100000000 
4. Observe initial state sync execution errors due to price limit (check balance did not update also):
```
[ERROR] polygon.server.relayer: State sync execution failed: err="failed to send execute state sync transaction for id 1: {\"code\":-32600,\"message\":\"insufficient funds for gas * price + value\"}"
[ERROR] polygon.server.relayer: State sync execution failed: err="failed to send execute state sync transaction for id 2: {\"code\":-32600,\"message\":\"insufficient funds for gas * price + value\"}
```
5. Restart validators with --price-limit 0
6. Wait for epochs to tick over, observe that state sync 1 and 2 are not retried
7. Perform deposit to relayer addr again
8. Observe that state sync 3 is processed successfully (check balance is updated also)
```
[INFO]  polygon.server.relayer: Execute commitment: Block=460 StartID=3 EndID=3
[INFO]  polygon.server.relayer: State sync executed: ID=3
```

I believe the relayer does not retry execute transactions against StateReceiver.sol either (the above is just rejected by RPC). 

#### Relayer txs to execute against StateReceiver.sol rejected by RPC:
I would expect the failing txs to be retried. In the above example, they are rejected by RPC on the basis of gas.

#### Relayer txs to execute against StateReceiver.sol fail during execution e.g. UNMAPPED_TOKEN etc:
Unsure whether this should be retried automatically or not given the category of errors that occur here.

## Logs

Relevant logs are in the steps to reproduce

## Proposed solution

Relayer should keep track of failed execute txs and retry? Might depend on the type of error as suggested above in expected behaviour section.

igorcrevar:
# Description

State sync relayer should not depend on blocktracker but should be part of consensus runtime like all the other similar parts.

Consensus runtime should call state sync relayer `PostBlock(req *common.PostBlockRequest) error` every time new block is inserted. State sync relayer should parse all the events and should submit transaction for each of those. It also internally remembers next event ID and next block number that should be processed (because sometimes something can go wrong).

Flag `--relayer-poll-interval` will be removed because its not used anymore

# Changes include

# Checklist

## Testing


consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+
+func (d *dummyStakeSyncRelayer) PostBlock(req *PostBlockRequest) error { return nil }
+
+func (d *dummyStakeSyncRelayer) Init() error { return nil }
+func (d *dummyStakeSyncRelayer) Close()      {}
+
+var _ StateSyncRelayer = (*stateSyncRelayerImpl)(nil)
+
+// StateSyncRelayerStateData is used to keep state of state sync relayer
+type StateSyncRelayerStateData struct {
```
goran-ethernal: Since we only have one field in this struct, I would remove it completely, and just use the `uint64`.

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+	logger     hclog.Logger
+	blockchain blockchainBackend
+
+	notifyCh chan struct{}
+	closeCh  chan struct{}
+
+	config *stateSyncRelayerConfig
+
+	// eventsGetter gets contractsapi.NewCommitmentEvent from blocks
+	eventsGetter *eventsGetter[*contractsapi.NewCommitmentEvent]
```
goran-ethernal: You can use one `eventsGetter` for multiple events. Check out how it was done for governance: https://github.com/0xPolygon/polygon-edge/blob/f2b24e79912802b1b6c8fb3ba1abdbcc9eec94c0/consensus/polybft/governance_manager.go#L99

This will optimize the code, plus it would remove a lot of lines.

command/genesis/params.go
```diff
@@ -534,16 +534,6 @@ func (p *genesisParams) validatePremineInfo() error {
 	return errReserveAccMustBePremined
 }
 
-// validateBlockTrackerPollInterval validates block tracker block interval
-// which can not be 0
-func (p *genesisParams) validateBlockTrackerPollInterval() error {
```
Stefan-Ethernal: How come it is valid to specify the poll interval to 0? What is semantic when it is set to 0? Does it mean it would constantly query the target endpoint?
Bear in mind that this value is used in the state sync manager, which retrieves state syncs from the L1 chain.

consensus/consensus.go
```diff
@@ -64,6 +64,12 @@ type Config struct {
 
 	// Path is the directory path for the consensus protocol to store information
 	Path string
+
+	// IsRelayer is true if node is relayer
+	IsRelayer bool
+
+	// RPCEndpoint
+	RPCEndpoint string
```
Stefan-Ethernal, line 72: Does it make sense to abstract these two fields in a separate struct (e.g. `RelayerConfig`) and put it inside `PolyBFTConfig` instead (namely it would be encoded inside `Config`)?

consensus/polybft/consensus_runtime.go
```diff
@@ -82,6 +82,8 @@ type runtimeConfig struct {
 	txPool                txPoolInterface
 	bridgeTopic           topic
 	numBlockConfirmations uint64
+	IsRelayer             bool
+	RPCEndpoint           string
```
Stefan-Ethernal: Does it make sense to put these fields into `PolyBFTConfig` and capture it in a separate struct (the one that I have advised to introduce in the `consensus.go` can be reused)?

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+
+		if err = json.Unmarshal(raw, &objs[i]); err != nil {
+			return fmt.Errorf("failed to unmarshal event %d: %w", event.EventID, err)
+		}
+
+		proofs[i] = proof.Data
+	}
+
+	input, err := (&contractsapi.BatchExecuteStateReceiverFn{
```
Stefan-Ethernal, line 245: :rocket:

e2e-polybft/framework/test-cluster.go
```diff
@@ -506,11 +499,6 @@ func NewTestCluster(t *testing.T, validatorsCount int, opts ...ClusterOption) *T
 				cluster.Config.BlockTime.String())
 		}
 
-		if cluster.Config.RelayerTrackerPollInterval != 0 {
-			args = append(args, "--block-tracker-poll-interval",
-				cluster.Config.RelayerTrackerPollInterval.String())
-		}
```
Stefan-Ethernal: Shouldn't this consult `RootTrackerPollInterval` and provide that value instead?

P.S. Interestingly enough `RootTrackerPollInterval` was introduced but never was used, in terms that some flag was provided based on it. :slightly_smiling_face:

e2e-polybft/e2e/bridge_test.go
```diff
@@ -1531,75 +1531,3 @@ func TestE2E_Bridge_Transfers_AccessLists(t *testing.T) {
 		}
 	})
 }
-
-func TestE2E_Bridge_Transfers_WithBlockTrackerPollInterval(t *testing.T) {
```
Stefan-Ethernal, line 1535: Keep this e2e test, but just remove the non-existing option `framework.WithRelayerTrackerPollInterval(relayerPollInterval)`?

consensus/polybft/state_sync_relayer.go
```diff
@@ -286,38 +287,53 @@ func (ssr *stateSyncRelayerImpl) GetLogFilters() map[types.Address][]types.Hash
...
+			ssr.logger.Info("event has been processed", "block", header.Number, "event", eventID)
+
+			return ssr.state.StateSyncStore.updateStateSyncRelayerEvents(nil, []uint64{eventID}, dbTx)
 		}
-	}
 
-	ssr.logger.Info("state sync relayer updated state", "block", header.Number,
-		"new", newEvents, "processed", processedEventIDs,
-		"failed", failedEventID, "reasons", failedReason)
+		// maybe add failed events as new ones (that would overwrite in db)
```
goran-ethernal: Remove this comment.

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,347 @@
...
+type stateSyncRelayerConfig struct {
+	maxBlocksToWaitForResend uint64
+	maxAttemptsToSend        uint64
+	maxEventsPerBatch        uint64
+}
+
+type stateSyncRelayerImpl struct {
+	txRelayer  txrelayer.TxRelayer
+	key        ethgo.Key
+	store      stateSyncProofRetriever
```
goran-ethernal:
```suggestion
	proofRetriever      stateSyncProofRetriever
```

goran-ethernal (approved):

dusan-maksimovic (approved):

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+	Init() error
+	Close()
+}
+
+// stateSyncProofRetriever is an interface that exposes function for retrieving state sync proof
+type stateSyncProofRetriever interface {
+	GetStateSyncProof(stateSyncID uint64) (types.Proof, error)
+}
+
+var _ StateSyncRelayer = (*dummyStakeSyncRelayer)(nil)
```
Stefan-Ethernal: Typo:
`dummyStakeSyncRelayer` -> `dummyStateSyncRelayer`

consensus/polybft/consensus_runtime.go
```diff
@@ -238,6 +248,34 @@ func (c *consensusRuntime) initCheckpointManager(logger hcf.Logger) error {
 	return nil
 }
 
+// initStateSyncRelayer initializes state sync relayer
+// if not enabled, then a dummy checkpoint manager will be used
+func (c *consensusRuntime) initStateSyncRelayer(logger hcf.Logger) error {
+	if c.config.IsRelayer {
+		// enable checkpoint manager
```
Stefan-Ethernal:
```suggestion
```

consensus/polybft/consensus_runtime.go
```diff
@@ -238,6 +248,34 @@ func (c *consensusRuntime) initCheckpointManager(logger hcf.Logger) error {
 	return nil
 }
 
+// initStateSyncRelayer initializes state sync relayer
+// if not enabled, then a dummy checkpoint manager will be used
```
Stefan-Ethernal:
```suggestion
// if not enabled, then a dummy state sync relayer will be used
```

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+	notifyCh chan struct{}
+	closeCh  chan struct{}
+
+	config *stateSyncRelayerConfig
+}
+
+func NewStateSyncRelayer(
+	txRelayer txrelayer.TxRelayer,
+	stateReceiverAddr types.Address,
+	state *State,
```
Stefan-Ethernal: Nitpick:
Provide `StateSyncStore` reference instead?

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+
+		for i := commitEvent.StartID.Uint64(); i <= commitEvent.EndID.Uint64(); i++ {
+			newEvents = append(newEvents, &StateSyncRelayerEventData{EventID: i})
+		}
+
+		ssr.logger.Info("new events has been arrived", "block", header.Number, "events", newEvents)
+
+		return ssr.state.StateSyncStore.updateStateSyncRelayerEvents(newEvents, nil, dbTx)
+
+	case stateSyncResultEvent.Sig():
```
Stefan-Ethernal: Minor consideration:
WDYT about extracting `commitEvent.Sig()` and `stateSyncResultEvent.Sig()` into some global variable and reusing it?

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,344 @@
...
+
+	switch log.Topics[0] {
+	case commitEvent.Sig():
+		_, err := commitEvent.ParseLog(log)
+		if err != nil {
+			return err
+		}
+
+		newEvents := []*StateSyncRelayerEventData{}
+
```
Stefan-Ethernal, line 301: Initialize with `make` and provide capacity?

```suggestion
		newEvents := make([]*StateSyncRelayerEventData, 0, commitEvent.StartID.Uint64()-commitEvent.StartID.Uint64()+1)
```

igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/polybft/state_sync_relayer.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/polybft/state_sync_relayer.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/polybft/consensus_runtime.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/polybft/consensus_runtime.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/polybft/state_sync_relayer.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/648a4f5c02d020891a5e3f1d3c9be482dc8c8639

consensus/consensus.go
igorcrevar: I will just have `consensusConfig       *consensus.Config` inside `type runtimeConfig struct`
No need to have one more structure

consensus/polybft/consensus_runtime.go
igorcrevar: I will just have `consensusConfig       *consensus.Config` inside `type runtimeConfig struct`
No need to have one more structure

Stefan-Ethernal: Nope, If you put it inside `PolyBFTConfig`, where it belongs IMO, you would already have it there, because runtimeConfig contains `PolyBFTConfig` reference. However, your call.

e2e-polybft/e2e/bridge_test.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/b78a6e22aaf18ebcaf78d1b07c8404df19124717

e2e-polybft/framework/test-cluster.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1899/commits/b78a6e22aaf18ebcaf78d1b07c8404df19124717

command/genesis/params.go
igorcrevar: returned back :) my bad https://github.com/0xPolygon/polygon-edge/pull/1899/commits/b78a6e22aaf18ebcaf78d1b07c8404df19124717

consensus/polybft/consensus_runtime.go
igorcrevar: we need to move 
```
	// IsRelayer is true if node is relayer
	IsRelayer bool `json:"isRelayer,omitempty"`

	// RPCEndpoint
	RPCEndpoint string `json:"rpcEndpoint,omitempty"`
```	
to `PolybftConfig` and set those at genesis. We can do that in separate PR

Stefan-Ethernal (approved):

consensus/polybft/state_store_state_sync.go
```diff
@@ -39,6 +41,12 @@ commitments/
 
 stateSyncProofs/
 |--> stateSyncProof.StateSync.Id -> *StateSyncProof (json marshalled)
+
+relayerData/
+|--> data -> *StateSyncRelayerStateData (json marshalled)
```
Stefan-Ethernal: Is this a leftover?

```suggestion
```

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,347 @@
...
+		for eventID := firstID; eventID <= lastID; eventID++ {
+			newEvents[eventID-firstID] = &StateSyncRelayerEventData{EventID: eventID}
+		}
+
+		ssr.logger.Info("new events has been arrived", "block", header.Number, "events", newEvents)
+
+		return ssr.state.updateStateSyncRelayerEvents(newEvents, nil, dbTx)
+
+	case stateSyncResultEventSignnature:
```
Stefan-Ethernal, line 310: Typo
```suggestion
	case stateSyncResultEventSignature:
```

consensus/polybft/state_sync_relayer.go
```diff
@@ -0,0 +1,347 @@
...
+	// defaultMaxEventsPerBatch specifies maximum events per one batchExecute tx
+	defaultMaxEventsPerBatch = 10
+)
+
+var (
+	errFailedToExecuteStateSync     = errors.New("failed to execute state sync")
+	errUnknownStateSyncRelayerEvent = errors.New("unknown event")
+
+	commitmentEventSignature       = new(contractsapi.NewCommitmentEvent).Sig()
+	stateSyncResultEventSignnature = new(contractsapi.StateSyncResultEvent).Sig()
```
Stefan-Ethernal: Typo
```suggestion
	stateSyncResultEventSignature = new(contractsapi.StateSyncResultEvent).Sig()
```

PR #1906, opened 2023-09-12
Speed up root chain contracts deployment

dusan-maksimovic:
# Description

Results write lock is now held shorter, so more things can be done in parallel than it was the case before.

# Changes include

# Checklist

## Testing

### DIFF
--- a/command/rootchain/deploy/deploy.go
+++ b/command/rootchain/deploy/deploy.go
@@ -560,7 +560,7 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client, 
 					return fmt.Errorf("deployment of %s contract failed", contract.name)
 				}
 
-				var deployResults []*deployContractResult
+				deployResults := make([]*deployContractResult, 0, 2)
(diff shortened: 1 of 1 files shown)


command/rootchain/deploy/deploy.go
```diff
@@ -560,15 +560,13 @@ func deployContracts(outputter command.OutputFormatter, client *jsonrpc.Client,
 					return fmt.Errorf("deployment of %s contract failed", contract.name)
 				}
 
-				resultsLock.Lock()
-				defer resultsLock.Unlock()
-
+				var deployResults []*deployContractResult
```
Stefan-Ethernal: Nitpick:
the worst-case scenario is to have 2 items in the `deployResults` slice, so I'd suggest specifying a capacity

```suggestion
				deployResults := make([]*deployContractResult, 0, 2)
```

Stefan-Ethernal (approved):
Aside from the comment, LGTM.

Nemanja0x (approved):

PR #1907, opened 2023-09-14
txpool slotGauge uint64 underflow

ohijiho:
# Description

txpool.slotGauge.height can exceed its max and the return value of freeSlots() can underflow. slotGauge.increase function is called by addTx function. The operation including free slots check, decrease and increase is not atomic, so in the following scenario, too many transactions are accepted so that slotGauge.heigh exceeds max.

1. slotGauge has only one free slot left. That is, height is one lower than max.
2. first addTx is called and executed right until gauge.increase()
3. second addTx is concurrently called and executed to the end. gauge.increase() is called and now slotGauge is full (height = max).
4. in the first addTx, gauge.increase() will be successfully executed since it already passed free slots check.
5. height exceeds the max (height = max + 1).

Instead of non-atomic load - check - decrease - increase, the updated version follows these steps:
1. Atomically check and increase height with net increment using slotGauge.increasedWithinLimit function. If height exceeds the max when increased, increaseWithinLimit will fail, and then ErrTxpoolOverflow.
2. Execute txpool updates which can fail. If it fails, decrease the height with the exact amount increased.
3. Decrease height with net decrement. Only one of 1 or 3 is executed depending on whether the height is increased or decreased.
4. Rest of the txpool updates which never fail.

# Changes include

# Checklist

## Testing


txpool/txpool.go
```diff
@@ -837,22 +838,41 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
 		return ErrTxPoolOverflow
 	}
 
+	slotsAllocated := slotsRequired(tx)
```
igorcrevar, line 841: I think its safe to remove check
```
if slotsRequired(tx) > slotsFree {
		return ErrTxPoolOverflow
	}
```

from the lines 845-848 and 

```
	// populate currently free slots
	slotsFree := p.gauge.freeSlots()
```
from the lines 790-791

and 

`slotsFree += slotsRequired(oldTxWithSameNonce) // add old tx slots` from the line 831

Slots availability check implemented in this is PR is enough I think:
```
if slotsAllocated > slotsFreed {
		slotsIncreased = slotsAllocated - slotsFreed
		if !p.gauge.increaseWithinLimit(slotsIncreased) {
			return ErrTxPoolOverflow
		}
	}
```


Otherwise, this pr looks good!

ohijiho: I just removed them. Thank you.

igorcrevar (approved):
LGTM!

txpool/txpool.go
```diff
@@ -832,27 +829,41 @@ func (p *TxPool) addTx(origin txOrigin, tx *types.Transaction) error {
...
+		if !p.gauge.increaseWithinLimit(slotsIncreased) {
+			return ErrTxPoolOverflow
+		}
 	}
 
 	// add to index
 	if ok := p.index.add(tx); !ok {
 		metrics.IncrCounter([]string{txPoolMetrics, "already_known_tx"}, 1)
 
+		p.gauge.decrease(slotsIncreased)
```
igorcrevar: I have already approved, but I think it would be good to have if here:
```
if slotsIncreased > 0 {
   p.gauge.decrease(slotsIncreased)
}
```

ohijiho: I fixed it and resolved merge conflicts.

Stefan-Ethernal (approved):

PR #1908, opened 2023-09-14
Use `ErrReplacementUnderpriced`

ohijiho:
# Description

When a new transaction's price is not higher than the previously existing transaction, return ErrReplacementUnderpriced instead of ErrUnderpriced. I believe that this error was originally intended for this use. It can be useful when one has to tell whether the transaction with the requested nonce exists in the txpool after sendTransaction call.

# Changes include

# Breaking changes

Not sure, but it could be a breaking change in some sense.

The error message the client gets will change.

# Checklist

## Testing

### DIFF
--- a/txpool/slot_gauge.go
+++ b/txpool/slot_gauge.go
@@ -3,8 +3,9 @@ package txpool
 import (
 	"sync/atomic"
 
-	"github.com/0xPolygon/polygon-edge/types"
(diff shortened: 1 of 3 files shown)


igorcrevar (approved):

Stefan-Ethernal (approved):
Tnx for your contribution!

PR #1915, opened 2023-09-18
Updating proposer calculator on restart

stana-miric:
# Description

This PR fixes the inconsistent state of the proposer calculator on node restart. The inconsistent state occurs when the proposer calculator's height is less than or equal to the current block number. It can occur when the node is stopped at a moment when a block is inserted into the blockchain, but the post-block function for the proposer calculator doesn't update the snapshot database. Later, when the node is started again, if the majority of valid nodes have a consistent state, everything will be okay, as the bad node will receive the new block from the syncer, which will fix the problem. However, if the majority of validators have an inconsistent state, consensus for the new block cannot be reached, and new blocks won't be produced. The solution for this is to align the proposer calculator on node startup by updating the proposer calculator's height based on the height of the current block (the last inserted block).

# Changes include

# Checklist

## Testing

### Manual tests
I've tested this scenario by changing the code by adding additional time. Sleep in order to be able to reproduce the bug and thus confirm that the scenario is for sure the one described above. This  code that reproduces problematic scenario each time cannot be used as an official test and it was only used to confirm the problem and verify the fix. Also the test for a bulk drop is adapted and it probably can cause this scenario on rare occasions.


igorcrevar (approved):
Approved, we should also do exactly the same thing for each component that saves its internal state `OnBlockInsert`

Stefan-Ethernal (approved):

PR #1566, opened 2023-05-29
Return Revert result as Data

vcastellm:
# Description

This allows returning custom revert errors from solidity, they are ABI encoded functions that can represent rich and gas efficient revert messages.

Details of custom errors: https://blog.soliditylang.org/2021/04/21/custom-errors/

This is the implementation expected by ethers.js and web3.js and the format go-ethereum chose https://github.com/ethereum/go-ethereum/issues/19027#issuecomment-640500608

image

This PR implements it following this format.

Fixes https://github.com/0xPolygon/polygon-edge/issues/1350

# Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Checklist

- [x] I have assigned this PR to myself
- [x] I have added at least 1 reviewer
- [x] I have added the relevant labels
- [ ] I have updated the official documentation
- [ ] I have added sufficient documentation in code

## Testing

- [x] I have tested this code with the official test suite
- [ ] I have tested this code manually

PR #1922, opened 2023-09-20
Propagate custom Solidity errors for `eth_estimateGas` JSON RPC function

### LINKED ISSUE #1801: Custom Errors return null and error message unable to apply transaction even for the highest gas limit 50000000
# Custom Errors return null and error message unable to apply transaction even for the highest gas limit 50000000
## Description
Hi, I have a smart contract with a custom error.
```
  function _checkHasAccess(bytes32 _senderRole) internal view virtual {
        if ( _hasAccess(__senderRole))) {
           revert PermissionedContractAccessDenied();
        }
    }
```
On ethersJs, when I submit a transaction with an unauthorized role. It should return the customer error on the `data` property, but instead, it returns `data = null` and a message. 
```
{message: 'unable to apply transaction even for the highest gas limit 50000000: execution reverted'}
```
On Mumbai, it gives me the correct error message.
```
"message": "Custom error: PermissionedContractAccessDenied",
```

## Your environment
- macOS Ventura 13.4.1
-  Polygon version v1.0.1 
- v1.0.1 
- Cloud hosted 

## Steps to reproduce
-  Add any smart contract with a customer error
- Use Ethersjs to submit a tx
- The error data will be null without any bytes value to decode

## Expected behavior
- Expected reverted transaction object should look like this
```
    code: 'CALL_EXCEPTION',
    action: 'estimateGas',
    data: '0x8cd01c38',
    reason: null,
    transaction: {
      to: '0x81cA45368Af26a91A74B91Bf4Ae5017c5AaB0703',
      data: '0x06244ace000000000000000000000000000000000000000000000000000000000000000036346434633731373430393564626534613963633034656300000000000000003634643334663335326163666462343463373362396431360000000000000000',
      from: '0x877427CCBd3061Affd5c6518bc87799B9Cf3C408'
    },
    invocation: null,
    revert: null,
    info: {
    error: { code: 3, message: 'execution reverted', data: '0x8cd01c38' },
    payload: {
      method: 'eth_estimateGas',
      params: [Array],
      id: 4,
      jsonrpc: '2.0'
    }
  },
 
}
```
- What happened instead.
```
   code: 'CALL_EXCEPTION',
    action: 'estimateGas',
    data: null,
    reason: null,
    transaction: {
      to: '0x8aCB619584817f3D9d587c3806b8c18B95Fb33fe',
      data: '0x06244ace46554e445f41444d494e5f524f4c45000000000000000000000000000000000036346434633930653835303035343635303735353766396500000000000000003634643334663335326163666462343463373362396431360000000000000000',
      from: '0x877427CCBd3061Affd5c6518bc87799B9Cf3C408'
    },
    invocation: null,
    revert: null,
  info: {
    error: {
      code: -32603,
      message: 'unable to apply transaction even for the highest gas limit 50000000: execution reverted'
    },
    payload: {
      method: 'eth_estimateGas',
      params: [Array],
      id: 4,
      jsonrpc: '2.0'
    }
  }
  },
 
}
```

jelacamarko:
# Description

If a custom error occurs, the `result.ReturnValue` will represent ABI encoded custom error in the `eth_estimateGas` JSON RPC function.
The same thing was done through this PR, but only for `Call` (https://github.com/0xPolygon/polygon-edge/pull/1566). This particular PR fixes the `eth_estimateGas` function and propagates the execution result return value (which should be a hex-encoded custom error in case it reverts with it).

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/jsonrpc/eth_endpoint.go
+++ b/jsonrpc/eth_endpoint.go
@@ -653,7 +653,7 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 
 	// Run the transaction with the specified gas value.
 	// Returns a status indicating if the transaction failed and the accompanying error
-	testTransaction := func(gas uint64, shouldOmitErr bool) (bool, error) {
+	testTransaction := func(gas uint64, shouldOmitErr bool) (bool, interface{}, error) {
 		transaction.Gas = gas
 
 		result, applyErr := e.store.ApplyTxn(header, transaction, nil)
@@ -665,10 +665,10 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 				// Specifying the transaction failed, but not providing an error
 				// is an indication that a valid error occurred due to low gas,
 				// which will increase the lower bound for the search
-				return true, nil
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), nil
 			}
 
-			return true, applyErr
+			return true, []byte(hex.EncodeToString(result.ReturnValue)), applyErr
 		}
 
 		// Check if an out of gas error happened during EVM execution
@@ -677,30 +677,30 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 				// Specifying the transaction failed, but not providing an error
 				// is an indication that a valid error occurred due to low gas,
 				// which will increase the lower bound for the search
-				return true, nil
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), nil
 			}
 
 			if isEVMRevertError(result.Err) {
 				// The EVM reverted during execution, attempt to extract the
 				// error message and return it
-				return true, constructErrorFromRevert(result)
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), constructErrorFromRevert(result)
 			}
 
-			return true, result.Err
+			return true, []byte(hex.EncodeToString(result.ReturnValue)), result.Err
 		}
 
-		return false, nil
+		return false, []byte(hex.EncodeToString(result.ReturnValue)), nil
 	}
 
 	// Start the binary search for the lowest possible gas price
 	for lowEnd < highEnd {
 		mid := lowEnd + ((highEnd - lowEnd) >> 1) // (lowEnd + highEnd) / 2 can overflow
 
-		failed, testErr := testTransaction(mid, true)
+		failed, retVal, testErr := testTransaction(mid, true)
 		if testErr != nil && !isEVMRevertError(testErr) {
 			// Reverts are ignored in the binary search, but are checked later on
 			// during the execution for the optimal gas limit found
-			return 0, testErr
+			return retVal, testErr
 		}
 
 		if failed {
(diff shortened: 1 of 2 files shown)


jsonrpc/eth_endpoint.go
```diff
@@ -677,30 +677,30 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
...
-				return true, constructErrorFromRevert(result)
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), constructErrorFromRevert(result)
 			}
 
-			return true, result.Err
+			return true, []byte(hex.EncodeToString(result.ReturnValue)), result.Err
 		}
 
-		return false, nil
+		return false, []byte(hex.EncodeToString(result.ReturnValue)), nil
```
Stefan-Ethernal: We should not return return value in this case, because it is not being propagated. WDYT?

```suggestion
		return false, nil, nil
```

jsonrpc/eth_endpoint.go
```diff
@@ -677,30 +677,30 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 				// Specifying the transaction failed, but not providing an error
 				// is an indication that a valid error occurred due to low gas,
 				// which will increase the lower bound for the search
-				return true, nil
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), nil
```
Stefan-Ethernal: Capture this into a local variable and reuse it when necessary:
`errorData := []byte(hex.EncodeToString(result.ReturnValue))`

Stefan-Ethernal (commented):
Generally, it looks good. Tested it manually and the behavior is expected. Could we add some more unit tests maybe?

jsonrpc/eth_endpoint.go
```diff
@@ -653,7 +653,7 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 
 	// Run the transaction with the specified gas value.
 	// Returns a status indicating if the transaction failed and the accompanying error
```
stana-miric: update description by adding new return parameter

jsonrpc/eth_endpoint.go
```diff
@@ -665,10 +665,10 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
 				// Specifying the transaction failed, but not providing an error
 				// is an indication that a valid error occurred due to low gas,
 				// which will increase the lower bound for the search
-				return true, nil
+				return true, []byte(hex.EncodeToString(result.ReturnValue)), nil
```
dusan-maksimovic: If I am correct, if the `applyErr` is not nil, it could happen that `result` is nil, so it is not safe to use it inside `if applyErr != nil {` branch without checking for nil first.

jsonrpc/eth_endpoint.go
```diff
@@ -665,10 +665,10 @@ func (e *Eth) EstimateGas(arg *txnArgs, rawNum *BlockNumber) (interface{}, error
...
 			}
 
-			return true, applyErr
+			return true, []byte(hex.EncodeToString(result.ReturnValue)), applyErr
```
dusan-maksimovic: Same as above.

jsonrpc/eth_endpoint.go
jelacamarko: So there can only happen gas error and revert error? And in any other failed case we won't have value for ReturnValue?

Stefan-Ethernal (approved):
LGTM

stana-miric (approved):

dusan-maksimovic (approved):

PR #1924, opened 2023-09-21
Improve Dockerfile to build core-contracts

FournyP:
# Description

I've updated the Dockerfile inside of `docker/local` folder because it didn't build the `core-contracts` folder.

# Changes include

## Testing

### Manual tests

I've build the Dockerfile successfully :
image

### DIFF
--- a/docker/local/Dockerfile
+++ b/docker/local/Dockerfile
@@ -1,13 +1,25 @@
-FROM golang:1.20-alpine AS builder
+FROM golang:1.20-alpine AS go-builder
 
 WORKDIR /polygon-edge
 
-ADD go.mod go.sum ./
+COPY go.mod ./go.mod 
+COPY go.sum ./go.sum
 RUN go mod download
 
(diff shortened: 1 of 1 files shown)


FournyP:

Stefan-Ethernal:
Tnx for your PR @FournyP. However, since Edge (namely polybft consensus protocol) is shipped with a pre-compiled set of smart contracts, I don't think this step is necessary. Polybft uses these contracts (https://github.com/0xPolygon/polygon-edge/blob/develop/consensus/polybft/contractsapi/gen_sc_data.go), which is generated by compiling `core-contracts`. That is why it is not a pre-requisite to clone `core-contracts` and compile them before running Edge.

Have you witnessed some kind of error, which made you think this step is required?

Stefan-Ethernal:
This step is unnecessary, as Edge is shipped with already compiled smart contracts as part of its binary.

PR #1925, opened 2023-09-21
Provide race and shuffle flags when running unit tests

Stefan-Ethernal:
# Description

This PR provides race and shuffle flags when running unit tests to discover any outstanding race conditions and potential hidden dependencies between different tests (respectively).

Also, fix race conditions in unit tests that are reported by race flag.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1005,7 +1005,7 @@ func (b *Blockchain) writeBody(batchWriter *storage.BatchWriter, block *types.Bl
 
 	// Write txn lookups (txHash -> block)
 	for _, txn := range block.Transactions {
-		batchWriter.PutTxLookup(txn.Hash, block.Hash())
+		batchWriter.PutTxLookup(txn.GetHash(), block.Hash())
 	}
 
 	return nil
@@ -1049,7 +1049,7 @@ func (b *Blockchain) recoverFromFieldsInTransactions(transactions []*types.Trans
 
 		sender, err := b.txSigner.Sender(tx)
 		if err != nil {
-			b.logger.Warn("failed to recover from address in Tx", "hash", tx.Hash, "err", err)
+			b.logger.Warn("failed to recover from address in Tx", "hash", tx.GetHash(), "err", err)
 
 			continue
 		}
--- a/command/bridge/deposit/erc20/deposit_erc20.go
+++ b/command/bridge/deposit/erc20/deposit_erc20.go
@@ -195,9 +195,7 @@ func runCommand(cmd *cobra.Command, _ []string) {
 					return fmt.Errorf("failed to create tx input: %w", err)
 				}
 
-				var (
-					receipt *ethgo.Receipt
-				)
+				var receipt *ethgo.Receipt
 
 				receipt, err = txRelayer.SendTransaction(depositTxn, depositorKey)
 				if err != nil {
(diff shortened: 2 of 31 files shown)


blockchain/blockchain.go
```diff
@@ -1005,7 +1005,7 @@ func (b *Blockchain) writeBody(batchWriter *storage.BatchWriter, block *types.Bl
 
 	// Write txn lookups (txHash -> block)
 	for _, txn := range block.Transactions {
-		batchWriter.PutTxLookup(txn.Hash, block.Hash())
+		batchWriter.PutTxLookup(txn.GetHash(), block.Hash())
```
Stefan-Ethernal, line 1008: Unsure about accessing the hash in a thread-safe way everywhere in this commit. It is probably not necessary, but it was necessary for this commit https://github.com/0xPolygon/polygon-edge/pull/1925/commits/f64ced2046fa11bda5b4488218038218e0c2de6c that's for sure.

vcastellm: Why not? I think it's worth it

command/bridge/deposit/erc20/deposit_erc20.go
```diff
@@ -195,6 +195,10 @@ func runCommand(cmd *cobra.Command, _ []string) {
 					return fmt.Errorf("failed to create tx input: %w", err)
 				}
 
+				var (
+					receipt *ethgo.Receipt
+				)
+
```
vcastellm:
```suggestion
				var receipt *ethgo.Receipt

```

blockchain/blockchain.go
Stefan-Ethernal: I was thinking because, most of the time it is from a single goroutine and we face multiple goroutines in unit test execution since we are executing them in parallel. So, although performance overhead is probably negligible, we may acquire a read lock even when we do not need it.

vcastellm: I assume that this change comes from a fix in CI and as performance is negligible and change is future proof, for me it's worth it

vcastellm (approved):
LGTM

blockchain/blockchain.go
Stefan-Ethernal: > I assume that this change comes from a fix in CI.

In fact, this change was not initiated (https://github.com/0xPolygon/polygon-edge/pull/1925/commits/760cfe5f2fba4faaebb9f9450decf44ee6c34b78), because CI was complaining, but I thought that maybe we should access it in a thread-safe manner, just to be future-proof as you said.
So eventually I don't mind dropping the commit.

jelacamarko (approved):

types/transaction_fork_hash.go
```diff
@@ -47,7 +47,9 @@ func (th *TransactionHashForkV1) ComputeHash(t *Transaction) {
 	t.ChainID = big.NewInt(0)
 
 	v := t.MarshalRLPWith(ar)
-	hash.WriteRlp(t.Hash[:0], v)
+	hashTmp := ZeroHash
```
igorcrevar, line 50: can we have here `var hashTmp types.Hash`? We don't need unnecessary copy

stana-miric (approved):

types/transaction_fork_hash.go
Stefan-Ethernal: Isn't it the same in the end? `ZeroHash` is a global variable and I don't understand which copy are you referring to?

igorcrevar: Yeah, I have checked with chatgpt it seams that both approaches has same performance. I thought that hashTmp := ZeroHash is less performant

igorcrevar (approved):

PR #1928, opened 2023-09-21
EVM-843 Fix block post processing

igorcrevar:
# Description

If node is stopped before stake manager has updated its internal state, there could be problems with consensus after node starts again. So, to prevent this, stake manager should update its internal state when node is started again.

Stake sync manager and checkpoint manager do not need to update their states because they do not have anything of importance for `go-ibft` consensus.
On the other hand, governance manager does have and that logic should be implemented if we decide to return back governance implementation

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/stake_manager.go
+++ b/consensus/polybft/stake_manager.go
@@ -67,7 +67,7 @@ func newStakeManager(
 	validatorSetAddr, supernetManagerAddr types.Address,
 	blockchain blockchainBackend,
 	maxValidatorSetSize int,
-) *stakeManager {
+) (*stakeManager, error) {
 	eventsGetter := &eventsGetter[*contractsapi.TransferEvent]{
 		blockchain: blockchain,
 		isValidLogFn: func(l *types.Log) bool {
@@ -81,7 +81,7 @@ func newStakeManager(
 		},
 	}
 
-	return &stakeManager{
+	sm := &stakeManager{
 		logger:                  logger,
 		state:                   state,
 		rootChainRelayer:        rootchainRelayer,
@@ -90,6 +90,12 @@ func newStakeManager(
 		maxValidatorSetSize:     maxValidatorSetSize,
 		eventsGetter:            eventsGetter,
 	}
+
+	if err := sm.init(blockchain); err != nil {
+		return nil, err
+	}
+
+	return sm, nil
 }
 
 // PostEpoch saves the initial validator set to db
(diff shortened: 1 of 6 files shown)


consensus/polybft/stake_manager.go
```diff
@@ -115,12 +121,23 @@ func (s *stakeManager) PostBlock(req *PostBlockRequest) error {
 		return err
 	}
 
-	s.logger.Debug("Stake manager on post block", "block", req.FullBlock.Block.Number(),
-		"last saved", fullValidatorSet.BlockNumber, "last updated", fullValidatorSet.UpdatedAtBlockNumber)
+	s.logger.Debug("Stake manager on post block",
+		"block", req.FullBlock.Block.Number(),
```
Stefan-Ethernal: Nitpick: extract `blockNumber` to a local var and reuse it:
`blockNumber := req.FullBlock.Block.Number()`

consensus/polybft/stake_manager.go
```diff
@@ -115,12 +121,23 @@ func (s *stakeManager) PostBlock(req *PostBlockRequest) error {
...
+	// get transfer currentBlockEvents from current block
+	transferEvents, err := s.eventsGetter.getFromBlocks(fullValidatorSet.BlockNumber, req.FullBlock)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not get transfer events from current block. Error: %w", err)
+	}
+
+	// we should save new state even if number of events is zero
+	// because otherwise next time we will process more blocks
+	if len(transferEvents) > 0 {
```
Stefan-Ethernal: Couldn't you add early return inside `updateWithReceipt` function in case `transferEvents` are empty? Then you can get rid off this check on two places.

Stefan-Ethernal (approved):
LGTM aside from nitpicks.

Stefan-Ethernal (approved):
LGTM aside from nitpicks.

consensus/polybft/stake_manager.go
igorcrevar: I wanted to save state in any case (comment above explains it)

igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1928/commits/7459127e696d97266ca729acc9ffa7de07762236

consensus/polybft/stake_manager.go
igorcrevar: fixed https://github.com/0xPolygon/polygon-edge/pull/1928/commits/7459127e696d97266ca729acc9ffa7de07762236

goran-ethernal (approved):

PR #1931, opened 2023-09-25
Added option to disable structLogs for debug_trace* rpc methods.

ohijiho:
# Description

The easiest reliable way to get return data of a transaction is debug_traceTransaction rpc, but it comes with huge structLogs data, which uses a lot of network traffic.

This is a simple solution for that: Just add an option to disable it. 

# Changes include

# Checklist

## Testing

### Manual tests

1. Start the chain
2. Make a transaction
3. Call trace_debugTransaction with disableLogs set to true and then false.
4. Check that structLogs field is filled only when disableLogs is set to false.

# Additional comments

I removed some test code lines that skips certain test suites. I don't know exactly why they have been made skipped originally, but they seems to work well now.

### DIFF
--- a/jsonrpc/debug_endpoint.go
+++ b/jsonrpc/debug_endpoint.go
@@ -77,11 +77,12 @@ func NewDebug(store debugStore, requestsPerSecond uint64) *Debug {
 }
 
 type TraceConfig struct {
-	EnableMemory     bool    `json:"enableMemory"`
-	DisableStack     bool    `json:"disableStack"`
-	DisableStorage   bool    `json:"disableStorage"`
-	EnableReturnData bool    `json:"enableReturnData"`
-	Timeout          *string `json:"timeout"`
+	EnableMemory      bool    `json:"enableMemory"`
+	DisableStack      bool    `json:"disableStack"`
+	DisableStorage    bool    `json:"disableStorage"`
+	EnableReturnData  bool    `json:"enableReturnData"`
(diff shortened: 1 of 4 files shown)


vcastellm (approved):
LGTM

dusan-maksimovic (approved):
LGTM. Perhaps we should align all the flags in `TraceConfig` so we don't have half of them be EnableXYZ and the other half DisableXYZ. But this can be done through some other PR.

Stefan-Ethernal:
> so we don't have half of them be EnableXYZ and the other half DisableXYZ

@ohijiho I agree entirely with @dusan-maksimovic here. Also then it makes more sense to expose the `EnableMemory`, `EnableStorage`, and `EnableStack` fields, and no need to have `DisableStructLogs` in general. WDYT?

ohijiho:
> > so we don't have half of them be EnableXYZ and the other half DisableXYZ
> 
> @ohijiho I agree entirely with @dusan-maksimovic here. Also then it makes more sense to expose the `EnableMemory`, `EnableStorage`, and `EnableStack` fields, and no need to have `DisableStructLogs` in general. WDYT?

Then, structLogs is disabled by default and enabled when one of memory, storage or stack is enabled? If that's the case, the config will be restricted so that there's no option to enable only the structLogs.

FYI, the main purpose of this PR is adding an option to entirely disable the structLogs, which is different from disabling all three extra informations.

Stefan-Ethernal (approved):

PR #1935, opened 2023-09-26
EVM-846 Stake manager post epoch problem

igorcrevar:
# Description

Stake manager creates (and saves) its initial state on PostEpoch call. This could lead to a problem if save fails

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/stake_manager.go
+++ b/consensus/polybft/stake_manager.go
@@ -3,6 +3,7 @@ package polybft
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/big"
 	"sort"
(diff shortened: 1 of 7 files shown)


stana-miric (approved):

consensus/polybft/stake_manager.go
```diff
@@ -193,6 +175,33 @@ func (s *stakeManager) init(blockchain blockchainBackend) error {
...
 }
 
+func (s *stakeManager) getOrInitValidatorSet() (validatorSetState, error) {
+	validatorSet, err := s.state.StakeStore.getFullValidatorSet()
+	if err != nil {
+		if !errors.Is(err, errNoFullValidatorSet) {
+			return validatorSetState{}, err
+		}
+
+		validators, err := s.polybftBackend.GetValidators(0, nil)
```
goran-ethernal, line 185: I would return an error here if the validators set returned here is empty. Just in case.

igorcrevar: I will add this check and error inside validator snapshot because that is the place for it

igorcrevar: https://github.com/0xPolygon/polygon-edge/pull/1935/commits/ad50306767f1a30b17d751bbaa3a379e3890df1b

goran-ethernal (approved):

PR #1938, opened 2023-09-26
[EVM-842]: New block/event tracker

goran-ethernal:
# Description

New implementation of block/event tracker in the `polybft` package.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/eventtracker/event_tracker.go
+++ b/consensus/polybft/eventtracker/event_tracker.go
@@ -1,4 +1,4 @@
-package polybft
+package eventtracker
 
 import (
 	"context"
@@ -24,10 +24,10 @@ type BlockProvider interface {
 	GetLogs(filter *ethgo.LogFilter) ([]*ethgo.Log, error)
 }
 
-// PolybftTrackerConfig is a struct that holds configuration of a PolybftEventTracker
-type PolybftTrackerConfig struct {
-	// RpcEndpoint is the full json rpc url on some node on a tracked chain
-	RpcEndpoint string
+// EventTrackerConfig is a struct that holds configuration of a EventTracker
+type EventTrackerConfig struct {
+	// RPCEndpoint is the full json rpc url on some node on a tracked chain
+	RPCEndpoint string
 
 	// StartBlockFromConfig represents a starting block from which tracker starts to
 	// track events from a tracked chain.
@@ -78,9 +78,9 @@ type PolybftTrackerConfig struct {
 	EventSubscriber EventSubscriber
 }
 
-// PolybftEventTracker represents a tracker for events on desired contracts on some chain
-type PolybftEventTracker struct {
-	config *PolybftTrackerConfig
+// EventTracker represents a tracker for events on desired contracts on some chain
+type EventTracker struct {
+	config *EventTrackerConfig
 
 	closeCh chan struct{}
 	once    sync.Once
@@ -89,11 +89,11 @@ type PolybftEventTracker struct {
 	blockContainer *TrackerBlockContainer
 }
 
-// NewPolybftEventTracker is a constructor function that creates a new instance of the PolybftEventTracker struct.
+// NewEventTracker is a constructor function that creates a new instance of the EventTracker struct.
 //
 // Example Usage:
 //
-//	config := &PolybftEventTracker{
+//	config := &EventTracker{
 //		RpcEndpoint:           "http://some-json-rpc-url.com",
 //		StartBlockFromConfig:  100_000,
 //		NumBlockConfirmations: 10,
@@ -109,14 +109,14 @@ type PolybftEventTracker struct {
 //			IDs:       []ethgo.Hash{idHashOfSomeEvent},
 //		},
 //	}
-//		t := NewPolybftEventTracker(config)
+//		t := NewEventTracker(config)
 //
 // Inputs:
-//   - config (TrackerConfig): configuration of PolybftEventTracker.
+//   - config (TrackerConfig): configuration of EventTracker.
 //
 // Outputs:
-//   - A new instance of the PolybftEventTracker struct.
-func NewPolybftEventTracker(config *PolybftTrackerConfig) (*PolybftEventTracker, error) {
+//   - A new instance of the EventTracker struct.
+func NewEventTracker(config *EventTrackerConfig) (*EventTracker, error) {
 	lastProcessedBlock, err := config.Store.GetLastProcessedBlock()
 	if err != nil {
 		return nil, err
@@ -131,7 +131,7 @@ func NewPolybftEventTracker(config *PolybftTrackerConfig) (*PolybftEventTracker,
 		definiteLastProcessedBlock = lastProcessedBlock
 	}
 
-	return &PolybftEventTracker{
+	return &EventTracker{
 		config:         config,
 		closeCh:        make(chan struct{}),
 		blockTracker:   blocktracker.NewJSONBlockTracker(config.BlockProvider),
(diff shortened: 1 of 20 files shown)


consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+
+	blockTracker   blocktracker.BlockTrackerInterface
+	blockContainer *TrackerBlockContainer
+}
+
+// NewEventTracker is a constructor function that creates a new instance of the EventTracker struct.
+//
+// Example Usage:
+//
+//	config := &EventTracker{
```
stana-miric: should be EventTrackerConfig instead of EventTracker

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+//
+// Outputs:
+//   - A new instance of the EventTracker struct.
+func NewEventTracker(config *EventTrackerConfig) (*EventTracker, error) {
+	lastProcessedBlock, err := config.Store.GetLastProcessedBlock()
+	if err != nil {
+		return nil, err
+	}
+
+	definiteLastProcessedBlock := lastProcessedBlock
```
stana-miric: do we need definiteLastProcessedBlock? we could just update the lastProcessedBlock directly if there is no need of using the original value of lastProcessedBlock

goran-ethernal: Hmmm, good point. I'll add that.

stana-miric:
LGTM

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+	blockContainer *TrackerBlockContainer
+}
+
+// NewEventTracker is a constructor function that creates a new instance of the EventTracker struct.
+//
+// Example Usage:
+//
+//	config := &EventTracker{
+//		RpcEndpoint:           "http://some-json-rpc-url.com",
+//		StartBlockFromConfig:  100_000,
```
dusan-maksimovic: `EventTrackerConfig` doesn't have this field.

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+	PollInterval time.Duration
+
+	// Logger is the logger instance for event tracker
+	Logger hcf.Logger
+
+	// LogFilter defines which events are tracked and from which contracts on the tracked chain
+	LogFilter map[ethgo.Address][]ethgo.Hash
+
+	// Store is the store implementation for data that tracker saves (lastProcessedBlock and logs)
+	Store EventTrackerStore
```
dusan-maksimovic: It feels a bit strange to have the store as part of the configuration. WDYT about passing it directly to the `NewEventTracker()`?

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+//	defer tracker.Close()
+//
+// Inputs: None
+//
+// Flow:
+//  1. The Close() method is called on an instance of EventTracker.
+//  2. The closeCh channel is closed, which signals the goroutines to stop.
+//
+// Outputs: None
+func (e *EventTracker) Close() {
```
dusan-maksimovic, line 163: `Close()` is not called to stop the `EventTracker` when `stateSyncManager` is stopped.

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+				return err
+			}
+
+			if err := e.blockContainer.AddBlock(block); err != nil {
+				return err
+			}
+		}
+
+		// now process logs from confirmed blocks if any
+		if err := e.processLogs(); err != nil {
```
dusan-maksimovic, line 338: Any strong reason why the logs are processed within `for` loop and not all together after all the blocks are fetched? It would make sense if you were fetching in one go routine and processing in another, with some kind of lock on data.

consensus/polybft/eventtracker/event_tracker.go
```diff
@@ -0,0 +1,464 @@
...
+			return err
+		}
+	}
+
+	// add latest block
+	if err := e.blockContainer.AddBlock(latestBlock); err != nil {
```
dusan-maksimovic, line 344: Not fetching the latest block since we already have it adds a bit of code complexity, but I'm not sure it is saving enough execution time to make it justified. If you also fetched the latest block then you could get rid of these two `AddBlock()` and `processLogs()` calls. WDYT?

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1938/commits/8e5ec7f6841e9b2c1c8120d23b62d90fc709c71b

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Done: https://github.com/0xPolygon/polygon-edge/pull/1938/commits/8e5ec7f6841e9b2c1c8120d23b62d90fc709c71b

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: To reduce impact on memory. While the tracker is getting the batch, we can check if it has any confirmed blocks to process. This way we are processing everything we can, and cleaning memory much faster.

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Hmm, I am not keen to pinging tracked chain to get it again if I already have it. If we can reduce even one call to the rpc node then I would say it is justified (we saw this on ethereum, when the old tracker sent too much calls to the node, and the node refused further calls from the caller, meaning bridge got stuck). It does add some code complexity, I agree, but it's pretty small impact on code.

dusan-maksimovic: It shouldn't require additional call to RPC endpoint, since it would be retrieved in previous for loop. You would just need to strip the -1 in calculation. And it would make the code simpler. Not a mandatory thing, just a suggestion. :)

goran-ethernal: Latest block is not gotten in the for loop, rather from the Track function on the jsonrpc block tracker.

dusan-maksimovic: It could be retrieved in the for loop (again) if there wasn't for -1.

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1938/commits/c96202f1a85b8cbf27b24250dae9bfa37217d146

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1938/commits/c96202f1a85b8cbf27b24250dae9bfa37217d146

consensus/polybft/eventtracker/event_tracker.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1938/commits/c96202f1a85b8cbf27b24250dae9bfa37217d146

dusan-maksimovic (approved):

PR #1941, opened 2023-09-27
EVM-850 Validator rootchain balance metrics

igorcrevar:
# Description

Egde should periodically retrieve validator balance from the root chain and output that information as a Prometheus metric.
A new server flag `--metrics-interval` has been introduced. By default, this flag returns `8s` for interval. A value `0s` will disable metrics.

Also `checkpoint_block_number` metric is moved from `checkpoint_manager` to the polybft and it is being published by each validator node.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state_stats.go
+++ b/consensus/polybft/state_stats.go
@@ -1,10 +1,14 @@
 package polybft
 
 import (
+	"math/big"
 	"time"
 
 	"github.com/armon/go-metrics"
+	"github.com/hashicorp/go-hclog"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/umbracle/ethgo"
+	"github.com/umbracle/ethgo/jsonrpc"
 )
 
 // startStatsReleasing starts the process that releases BoltDB stats into prometheus periodically.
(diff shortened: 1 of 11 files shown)


jelacamarko (approved):

consensus/polybft/state_stats.go
```diff
@@ -154,3 +158,44 @@ func (s *State) startStatsReleasing() {
...
+		case <-ticker.C:
+			balance, err := rpcClient.Eth().GetBalance(validatorAddress, ethgo.Latest)
+			if err != nil {
+				logger.Error("metrics get balance call failed", "err", err)
+
+				continue
+			}
+
+			balanceInGwei := new(big.Int).Div(balance, gweiPerWei).Uint64()
+			metrics.SetGauge([]string{"bridge", "validator_balance_gwei", validatorAddress.String()},
```
goran-ethernal:
```suggestion
			metrics.SetGauge([]string{"bridge", "validator_root_balance_gwei", validatorAddress.String()},
```

goran-ethernal (approved):

consensus/polybft/state_stats.go
```diff
@@ -154,3 +158,44 @@ func (s *State) startStatsReleasing() {
...
+			return
+		case <-ticker.C:
+			balance, err := rpcClient.Eth().GetBalance(validatorAddress, ethgo.Latest)
+			if err != nil {
+				logger.Error("metrics get balance call failed", "err", err)
+
+				continue
+			}
+
+			balanceInGwei := new(big.Int).Div(balance, gweiPerWei).Uint64()
```
Stefan-Ethernal: Why not floating point division by relying on `big.Float`s?

stana-miric (approved):

consensus/polybft/state_stats.go
```diff
@@ -154,3 +158,44 @@ func (s *State) startStatsReleasing() {
 		prev = stats
 	}
 }
+
+func polybftMetrics(rootnodeURL string,
```
Stefan-Ethernal: WDYT about renaming this to `publishRootchainMetrics` and publishing it from `checkpoint_manager`? The reasoning is that checkpointManager already contains rootchainTxRelayer (that exposes `jsonrpc.Client`), which can be reused. Then, I would also consider moving the following metrics there as well (currently, only checkpoint proposers are publishing it but on checkpoint blocks only, but this way all validators would publish it on a more regular basis):

`metrics.SetGauge([]string{"bridge", "checkpoint_block_number"}, float32(header.Number))`

consensus/polybft/state_stats.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1941/commits/61e9e8d0044a2482ee232649943a7e9861e56c58

consensus/polybft/state_stats.go
igorcrevar: Fixed https://github.com/0xPolygon/polygon-edge/pull/1941/commits/61e9e8d0044a2482ee232649943a7e9861e56c58

jelacamarko (approved):

begmaroman (approved):

vcastellm:
Late LGTM

PR #1949, opened 2023-10-02
Skip account balance check for eth_call

### LINKED ISSUE #1927: Price limit error “not enough funds to cover gas costs” for getter calls
Simple non state changing call errors with “not enough funds to cover gas costs” when `--price-limit` is above 0

## Description

Account without tokens (balance is 0) could not perform simple smart contract getter call (non changing state call should not require any gas).
Seems like the problem is in function `func (t *Transition) subGasLimitPrice(msg *types.Transaction) error` which gets executed even for non state change client calls.

## Your environment
- OS and version:
`Ubuntu 22.04`
- The version of the Polygon Edge.    
Built from commit 
```
f2b24e7 2023-09-14 11:17 +0200 Dusan Nosovic          o [HEAD] {origin/v1.2.0-rc1} Introduce `ErrNonceUintOverflow` as error for nonce overflow (#1905)
```
- Locally or Cloud hosted (which provider).
`Locally`

## Steps to reproduce

Smart contract and scripts used are here:
[smart_contract_and_scripts.zip](https://github.com/0xPolygon/polygon-edge/files/12684684/smart_contract_and_scripts.zip)

```
Used origin/v1.2.0-rc1 branch

Generate keys, create a genesis file and run a network:

$ polygon-edge secrets init --insecure --data-dir test-chain-1
$ polygon-edge genesis --consensus ibft --ibft-validators-prefix-path test-chain- --bootnode /ip4/127.0.0.1/tcp/10001/p2p/16Uiu2HAmCwvnmjdsSyjJJJTbesHf6C9JvVp4RahqnvAfGnmfCNcV --premine 0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac:0x123123123123123

Run network with price limit 100
$ polygon-edge server --data-dir ./test-chain-1 --chain genesis.json  --price-limit=100

// Premined account is 0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac with private key 0x5fb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da702133

// Deploy simple TestContract2 smart contract provided (using ethers.js)

$ ts-node ./deploy-test-contract2.ts http://127.0.0.1:8545

In my case contract TestContact2 is deployed to address 0xc01Ee7f10EA4aF4673cFff62710E1D7792aBa8f3

Now run very simple script, that would only perform view getter call (getUintValue()), with the account that has 0 tokens on its balance (used account 0x5283ac54A7B9669F6415168DC7a5FcEe05019E45 private key 0x11eddfae7abe45531b3f18342c8062969323a7131d3043f1a33c40df74803cc7

$ ts-node check.ts http://127.0.0.1:8545 0xc01Ee7f10EA4aF4673cFff62710E1D7792aBa8f3 0x11eddfae7abe45531b3f18342c8062969323a7131d3043f1a33c40df74803cc7 


Simple call fails with error "not enough funds to cover gas costs","error":{"code":-32600}. That is a non state changing call, so it should be executed.


Runnin a script with a account that has premined tokens succeeds:

$ ts-node check.ts http://127.0.0.1:8545 0xc01Ee7f10EA4aF4673cFff62710E1D7792aBa8f3 0x5fb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da702133 

Retrieved uint `a` value is= BigNumber { _hex: '0x13', _isBigNumber: true }

```

## Expected behavior
- Simple smart getter call should be executed when account without tokens performs the call
- Error: 
```
"not enough funds to cover gas costs","error":{"code":-32600}
```
## Proposed solution
This seems like a serious issue, should be addressed immediately.

begmaroman:
# Description

No needed to check an account balance for eth_call. Accounts with zero balance still should be able to execute smart contract functions.

# Changes include

# Breaking changes

# Checklist

## Testing


state/runtime/runtime.go
```diff
@@ -21,6 +21,7 @@ type TxContext struct {
 	ChainID      int64
 	Difficulty   types.Hash
 	Tracer       tracer.Tracer
+	NoBaseFee    bool
```
Stefan-Ethernal: I'd consider renaming this to `Payable`, `IsPayable` (or something like that) for 2 reasons:
- `BaseFee` in the name implies that is is related only to EIP-1559 only, which is not the case
- `NoBaseFee` then implies that logic is inverted (namely transaction is paid in case `NoBaseFee` is set to `false`, double negation basically)

state/executor.go
```diff
@@ -1044,6 +1044,11 @@ func (t *Transition) SetCodeDirectly(addr types.Address, code []byte) error {
 	return nil
 }
 
+// SetNoBaseFee sets no base fee flag to the transition runtime context
+func (t *Transition) SetNoBaseFee(noBaseFee bool) {
```
Stefan-Ethernal: Consider renaming accordingly, in case the field name gets changed.

Stefan-Ethernal (commented):
Btw what should happen in the `TraceCall`, `TraceTxn`, and `TraceBlock` functions in the `server.go`? Should those executions be payable, since those are related to `debug_*` JSON RPC endpoints. I'm afraid that currently, they are payable, but it probably should not be from a logical standpoint. WDYT?

state/runtime/runtime.go
```diff
@@ -64,6 +65,7 @@ type Host interface {
 	GetStorage(addr types.Address, key types.Hash) types.Hash
 	SetStorage(addr types.Address, key types.Hash, value types.Hash, config *chain.ForksInTime) StorageStatus
 	SetState(addr types.Address, key types.Hash, value types.Hash)
+	SetNoBaseFee(noBaseFee bool)
```
Stefan-Ethernal: Consider renaming accordingly, in case the field name gets changed.

state/runtime/runtime.go
begmaroman: Agree. One thing is that I would have a payable state by default with an option to disable it. In this case, the flag should have `nonPayable` name.

state/executor.go
begmaroman: Done

state/runtime/runtime.go
begmaroman: Done

begmaroman:
@Stefan-Ethernal re debug endpoints. The logic inside those endpoints implies using an already existing state stored within a block. If a transaction was mined into the block, it means there were no issues with the balance. In this case, it just does not make sense to switch on this flag since a tx sender had enough balance within a specific block so the balance check is going to pass when tracing.

state/executor.go
```diff
@@ -1044,6 +1044,12 @@ func (t *Transition) SetCodeDirectly(addr types.Address, code []byte) error {
 	return nil
 }
 
+// SetNonPayable sets activates an indicator that deactivates a check of
+// tx cost against tx executor balance.
```
vcastellm:
```suggestion
// SetNonPayable deactivates the check of tx cost against tx executor balance.
```

stana-miric (approved):

dusan-maksimovic (approved):

state/executor.go
begmaroman: Done

vcastellm (approved):
LGTM

Stefan-Ethernal (approved):
LGTM. 

Btw would be awesome if we could include some e2e test (e.g. have dummy contract, account with 0 balance and make sure it is able to query the state of SC). We can do it through a separate PR if that makes sense to you.

begmaroman:
@Stefan-Ethernal added e2e test

igorcrevar (approved):
Can we have estimateGas and maybe even debugTransaction calls inside e2e test?

Approved

begmaroman:
@igorcrevar added e2e tests for those two endpoints.

PR #1965, opened 2023-10-06
Update web3_clientVersion endpoint to return a correct value

begmaroman:
# Description

Update the given endpoint in order to return a correct value as defined per https://ethereum.org/en/developers/docs/apis/json-rpc/#web3_clientversion

# Changes include

# Breaking changes

# Checklist

## Testing


secrets/helper/helper.go
```diff
@@ -104,6 +102,21 @@ func InitBLSValidatorKey(secretsManager secrets.SecretsManager) ([]byte, error)
 		return nil, err
 	}
 
+	// This section converts the given "blsSecretKeyEncoded" hex encoded private key
+	// into a raw big integer bytes in order to be consistent with the account object
+	// that is used to work with big numbers.
```
begmaroman, line 107: This fixes issues related to ibft like:
```
math/big: cannot unmarshal "..." into a *big.Int
```

vcastellm (approved):
LGTM

jsonrpc/web3_endpoint.go
```diff
@@ -13,15 +14,27 @@ type Web3 struct {
...
+	if versioning.Version != "" {
+		version = versioning.Version
+	} else if versioning.Commit != "" {
+		version = versioning.Commit[:8]
+	}
+
 	return fmt.Sprintf(
 		clientVersionTemplate,
 		w.chainName,
 		w.chainID,
```
Stefan-Ethernal: It doesn't seem that any other Ethereum client implementation embeds `chain id` into the return value, so I'd probably remove it.

jsonrpc/web3_endpoint_test.go
```diff
@@ -61,11 +62,13 @@ func TestWeb3EndpointClientVersion(t *testing.T) {
 
 	assert.NoError(t, expectJSONResult(resp, &res))
 	assert.Contains(t, res,
-		fmt.Sprintf(
-			clientVersionTemplate,
+		fmt.Sprintf("%s-%d/%s/%s-%s/%s",
```
Stefan-Ethernal: Why you couldn't use `clientVersionTemplate`?


```suggestion
		fmt.Sprintf(clientVersionTemplate,
```

rachit77 (approved):

jsonrpc/web3_endpoint.go
begmaroman: The official spec says that the endpoint should return a client version. It does not say what specific information it should return so I guess it is up to devs. Currently, this string includes a chain ID so just decided to keep it in a new format as well. Do you have any doubts why the chain ID should be removed from there?

Stefan-Ethernal: Based on some examples I've encountered and in the docs (https://docs.infura.io/networks/ethereum/json-rpc-methods/web3_clientversion#response), it does not provide chain id in the response, so in that context, we are not aligned with the other Ethereum-based clients implementations. 

Not that I'm totally against it, but I more thinking about there might be a tool that expects exactly this format of the client version (also mind the separators used / instead of -), like e.g. Geth or Erigon is returning it:
```go
<client name>/<client version>/<OS>-<platform>/<go version>
```

begmaroman: Ok, done

jsonrpc/web3_endpoint_test.go
begmaroman: Because this test should expect a specific format returning by endpoint. If `clientVersionTemplate` has a mistake, this test should fail. If we use this variable here, it won't fail in case of a mistake.

Stefan-Ethernal (approved):

igorcrevar (approved):

PR #1972, opened 2023-10-09
PostBlock optimization

goran-ethernal:
# Description

This PR optimizes `PostBlock` and `OnBlockInserted` handlers on `consensus_runtime` and its components (`state_sync_manager`, `checkpoint_manager`, `stake_manager` and `proposer_calculator`).

Before, each of the given components was iterating through the receipts of given finalized block, in `OnBlockInserted` function, to get the required events (for example, `checkpoint_manager` was getting `Exit` events, `stake_manager` was getting `Transfer` events, etc). To optimize this behavior, and remove iterating through the same receipts multiple times, this PR introduces these concepts:

### EventProvider
`EventProvider` is a component which retrieves receipts of a range of blocks or a single block to get desired events from them.
It requires components to `Subscribe` to it, using the `Subscribe(subscriber EventSubscriber)` function. 
Each of components that subscribe to the `EventProvider` need to implement the `EventSubscriber` interface:
```
// EventSubscriber specifies functions needed for a component to subscribe to eventProvider
type EventSubscriber interface {
	GetLogFilters() map[types.Address][]types.Hash
	AddLog(header *types.Header, log *ethgo.Log, dbTx DBTransaction) error
}
```
`GetLogFilters` function returns the map, where key is the address of a contract on which we want to get some type of events, and value is the signature or ID of the event on that contract that the subscriber needs. This map is used to filter out events from receipts so that it can return correct results to the given subscriber.

`AddLog` function is the handler function on the subscriber. When the `EventProvider` encounters an event that matches one of the filters in the `GetLogFilters` function for that subscriber, it will call the `AddLog` function on the subscriber, so that he can handle the given event (store it in memory, to a db, etc).

This way, we provide only one iteration through the blocks and receipts to get the desired events of different components, plus, if `AddLog` returns an error on any component, `EventProvider` immediately stops going through receipts, meaning that state of one component is tied to the state of other components, keeping the state more atomic, where either all components get updated, or no component does. This helps the consensus to have a more cleaner state. If one component is more advanced than the other, it may cause serious issues to running a more correct consensus.

### Global DB Transaction in `OnBlockInserted`
In `state.go` we added a `func (s *State) beginTx(isWriteTx bool)` function so that it can open a manually maintained transaction. This is needed for keeping the correct state of each component as mentioned in the previous section. This means that `OnBlockInserted` function has more control over its components. If either of them fail to save their events, or to do `PostBlock` functions, global `db` transaction is rolledback, and state of all components for that block is cleaned and thrown away, keeping the consensus more accurate, and error prone.

At the beginning of the `OnBlockInserted` function in `conesnsus_runtime`, a global `db` transaction is opened, and only after all components insert their events (gotten from `EventProvider`) and do `PostBlock` succesfuly, only then we will commit the transaction and save the components state to db. If any of them fail in these actions, transaction is rolled back.

### Exceptions to previously mentioned concepts
`PostBlock` function on `checkpoint_manager` is done after committing the global `db` transaction, since it does not change any state in `boltDB`, it only sends a checkpoint in a separate thread.

`validator_snapshot.go` - `GetSnapshot` function will open a `db` transaction if it is not passed to it. The reason behind this is that, `GetSnapshot` function is called from multiple go routines: `OnBlockInserted`, starting a new sequence in `polybft`, validating blocks in `fsm`, etc., and what can happen is that one routine can take its lock, but in the mean time, `OnBlockInserted` opens a global `db` transaction, which can result in a deadlock, since in `OnBlockInserted` proposer calculator, updates the proposer snapshot, by calling this `GetSnapshot` function. So the first routine gets the lock on `validatorSnapshot`, but it can not do anything with the `db` since `OnBlockInserted` holds the global `db` transaction. And `OnBlockInserted` can not update the proposer snapshot, because other routine holds the `validatorSnapshot` lock.
This is mitigated by opening `db` transaction on `validatorSnapshot` if it is not already passed.

# Changes include

# Checklist

## Testing


consensus/polybft/state_event_getter.go
```diff
@@ -39,8 +181,8 @@ func (e *eventsGetter[T]) getFromBlocks(lastProcessedBlock uint64,
 	return allEvents, nil
 }
 
-// getEventsFromAllBlocks gets events of specified type from all the blocks specified [from, to]
-func (e *eventsGetter[T]) getEventsFromAllBlocks(from, to uint64) ([]T, error) {
+// getEventsFromBlocks gets events of specified type from all the blocks specified [from, to]
+func (e *eventsGetter[T]) getEventsFromBlocks(from, to uint64) ([]T, error) {
```
Stefan-Ethernal:
```suggestion
// getEventsFromBlocksRange gets events of specified type from the provided blocks range [from, to]
func (e *eventsGetter[T]) getEventsFromBlocksRange(from, to uint64) ([]T, error) {
```

consensus/polybft/state.go
```diff
@@ -3,10 +3,25 @@ package polybft
...
+)
+
+type DBTransaction interface {
+	Commit() error
+	Rollback() error
+	Bucket(bucket []byte) *bolt.Bucket
+	DeleteBucket(bucket []byte) error
+	CreateBucket(bucket []byte) (*bolt.Bucket, error)
+	CreateBucketIfNotExists(bucket []byte) (*bolt.Bucket, error)
+}
```
Stefan-Ethernal: This interface is too tight with bolt db implementation, so not sure do we even need it? How it is currently, seems like we don't have any benefit of this interface (you could easily used concrete implementation `bolt.Tx`).

consensus/polybft/state_event_getter.go
```diff
@@ -1,12 +1,154 @@
...
+	}
+}
+
+// Subscribe subscribes given EventSubscriber to desired logs (events)
+func (e *EventProvider) Subscribe(subscriber EventSubscriber) {
+	e.lock.Lock()
+	e.subscriberIDCounter++
+	subscriberID := e.subscriberIDCounter
+	e.subscribers[e.subscriberIDCounter] = subscriber
+	e.lock.Unlock()
```
Stefan-Ethernal: Why do you need lock here? From what I can see, `Subscribe` is called from a single goroutine.

Also use `subscriberID` variable since you have introduced it:

```suggestion
	e.subscriberIDCounter++
	subscriberID := e.subscriberIDCounter
	e.subscribers[subscriberID] = subscriber
```

consensus/polybft/state_event_getter.go
```diff
@@ -1,12 +1,154 @@
...
+
+// EventProvider represents an event provider in a blockchain system
+// that returns desired events to subscribers from finalized blocks
+// Note that this provider needs to be called manually on each block finalization.
+// The EventProvider struct has the following fields:
+// - blockchain: A blockchainBackend type that represents the blockchain backend used by the event provider.
+// - subscribers: A map[string]eventSubscriber that stores the subscribers of the event provider.
+// - allFilters: A map[types.Address]map[types.Hash][]string that stores the filters for event logs.
+type EventProvider struct {
+	lock                sync.Mutex
```
Stefan-Ethernal: Do you need lock?

consensus/polybft/state_event_getter.go
```diff
@@ -1,12 +1,154 @@
...
+			existingAddressFilters[f] = append(existingAddressFilters[f], subscriberID)
+		}
+	}
+}
+
+// GetEventsFromBlocks gets all desired logs (events) for each subscriber in given block range
+//
+// Inputs:
+// - lastProcessedBlock - last finalized block that was processed for desired events
+// - latestBlock - latest finalized block
```
Stefan-Ethernal, line 74: You haven't documented `dbTx` parameter.

consensus/polybft/state_event_getter.go
```diff
@@ -1,12 +1,154 @@
...
+}
+
+// EventProvider represents an event provider in a blockchain system
+// that returns desired events to subscribers from finalized blocks
+// Note that this provider needs to be called manually on each block finalization.
+// The EventProvider struct has the following fields:
+// - blockchain: A blockchainBackend type that represents the blockchain backend used by the event provider.
+// - subscribers: A map[string]eventSubscriber that stores the subscribers of the event provider.
+// - allFilters: A map[types.Address]map[types.Hash][]string that stores the filters for event logs.
+type EventProvider struct {
```
Stefan-Ethernal, line 29: `eventsGetter` and `EventProvider` share some common code (extracting certain logs from receipts for range of blocks). Can you somehow abstract that code in order not to have code repetitions, because those two are fundamentally similar (although eventsGetter filters logs of interest by `isValidLogFn`, whereas `EventProvider` does the similar thing, but determines logs of interest based on event signatures and additionally it has subscribers).

consensus/polybft/state_event_getter.go
```diff
@@ -1,12 +1,154 @@
...
+
 	"github.com/0xPolygon/polygon-edge/blockchain"
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	"github.com/0xPolygon/polygon-edge/types"
 	"github.com/umbracle/ethgo"
 )
 
+// EventSubscriber specifies functions needed for a component to subscribe to eventProvider
+type EventSubscriber interface {
+	GetLogFilters() map[types.Address][]types.Hash
```
Stefan-Ethernal, line 16:
```suggestion
    // GetRelevantLogs returns mapping of smart contract addresses to relevant event signatures
	GetRelevantLogs() map[types.Address][]types.Hash
```

consensus/polybft/state.go
```diff
@@ -80,10 +95,86 @@ func (s *State) initStorages() error {
 			return err
 		}
```
```suggestion
```

consensus/polybft/stake_manager.go
```diff
@@ -371,6 +376,49 @@ func (s *stakeManager) getBlsKey(address types.Address) (*bls.PublicKey, error)
 	return pubKey, nil
 }
 
+// EventSubscriber implementation
```
Stefan-Ethernal, line 382:
```suggestion
```

consensus/polybft/state_event_getter.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state_event_getter.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state_event_getter.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state_event_getter.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state_event_getter.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/stake_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/12cd8d0769aeb15c9635ee45aed989c59e3aa7ff

consensus/polybft/state.go
goran-ethernal: Idea behind this abstraction was to use it in tests, and to not expose the underlying db implementation. Let me do something to remove these four functions:

```
Bucket(bucket []byte) *bolt.Bucket
DeleteBucket(bucket []byte) error
CreateBucket(bucket []byte) (*bolt.Bucket, error)
CreateBucketIfNotExists(bucket []byte) (*bolt.Bucket, error)
```

goran-ethernal: Ok, I've removed the `DBTransaction` completely, and through some other PR we will figure out how to hide the underlying `db` implementation behind some interface in a more correct and optimal manner.

Fix: https://github.com/0xPolygon/polygon-edge/pull/1972/commits/c997b8670ef994e08864cf4eb0f5c590d5219667

consensus/polybft/state_event_getter.go
goran-ethernal: I think I can have some common code between them. Let me do something.

goran-ethernal: I've managed to pull out some common code for both `eventsGetter` and `eventProvider`:

https://github.com/0xPolygon/polygon-edge/pull/1972/commits/ee6f73e1d972d154e055b433b0b22724e1eed906

dusan-maksimovic (approved):

igorcrevar (approved):

rachit77 (approved):

Stefan-Ethernal (approved):

PR #1977, opened 2023-10-10
Remove executed state sync events and proofs from the store

dusan-maksimovic:
# Description

This PR removes state sync events and their proofs after they have been successfully executed. Removal is performed after each new block is inserted. Any errors returned are just logged, since the removal of this data from bolt db is not essential for node functioning.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/state_store_state_sync.go
+++ b/consensus/polybft/state_store_state_sync.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hashicorp/go-hclog"
+
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	"github.com/0xPolygon/polygon-edge/helper/common"
 	bolt "go.etcd.io/bbolt"
@@ -42,7 +44,8 @@ stateSyncProofs/
 */
 
 type StateSyncStore struct {
-	db *bolt.DB
+	db     *bolt.DB
+	logger hclog.Logger
 }
 
 // initialize creates necessary buckets in DB if they don't already exist
@@ -73,6 +76,28 @@ func (s *StateSyncStore) insertStateSyncEvent(event *contractsapi.StateSyncedEve
 		bucket := tx.Bucket(stateSyncEventsBucket)
 
 		return bucket.Put(common.EncodeUint64ToBytes(event.ID.Uint64()), raw)
+	})
+}
+
+// removeStateSyncEventsAndProofs removes state sync events and their proofs from the buckets in db
+func (s *StateSyncStore) removeStateSyncEventsAndProofs(stateSyncEventIDs []uint64) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		eventsBucket := tx.Bucket(stateSyncEventsBucket)
+		proofsBucket := tx.Bucket(stateSyncProofsBucket)
+
+		for _, stateSyncEventID := range stateSyncEventIDs {
+			stateSyncEventIDKey := common.EncodeUint64ToBytes(stateSyncEventID)
+
+			if err := eventsBucket.Delete(stateSyncEventIDKey); err != nil {
+				s.logger.Info(fmt.Sprintf("failed to remove state sync event (ID=%d): %v", stateSyncEventID, err))
+			}
+
+			if err := proofsBucket.Delete(stateSyncEventIDKey); err != nil {
+				s.logger.Info(fmt.Sprintf("failed to remove state sync event proof (ID=%d): %v", stateSyncEventID, err))
+			}
+		}
+
+		return nil
 	})
 }
 
--- a/consensus/polybft/state_store_state_sync.go
+++ b/consensus/polybft/state_store_state_sync.go
@@ -86,11 +86,11 @@ func (s *StateSyncStore) removeStateSyncEventsAndProofs(stateSyncEventIDs []uint
 			stateSyncEventIDKey := common.EncodeUint64ToBytes(stateSyncEventID)
 
 			if err := eventsBucket.Delete(stateSyncEventIDKey); err != nil {
-				return fmt.Errorf("failed to remove state sync event (ID=%d): %v", stateSyncEventID, err)
+				return fmt.Errorf("failed to remove state sync event (ID=%d): %w", stateSyncEventID, err)
 			}
 
 			if err := proofsBucket.Delete(stateSyncEventIDKey); err != nil {
-				return fmt.Errorf("failed to remove state sync event proof (ID=%d): %v", stateSyncEventID, err)
+				return fmt.Errorf("failed to remove state sync event proof (ID=%d): %w", stateSyncEventID, err)
 			}
 		}
 
--- a/consensus/polybft/state_store_state_sync.go
+++ b/consensus/polybft/state_store_state_sync.go
@@ -5,8 +5,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/hashicorp/go-hclog"
-
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/contractsapi"
 	"github.com/0xPolygon/polygon-edge/helper/common"
 	bolt "go.etcd.io/bbolt"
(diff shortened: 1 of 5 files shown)


stana-miric (approved):

consensus/polybft/state_store_state_sync.go
```diff
@@ -76,6 +79,28 @@ func (s *StateSyncStore) insertStateSyncEvent(event *contractsapi.StateSyncedEve
...
+func (s *StateSyncStore) removeStateSyncEventsAndProofs(stateSyncEventIDs []uint64) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		eventsBucket := tx.Bucket(stateSyncEventsBucket)
+		proofsBucket := tx.Bucket(stateSyncProofsBucket)
+
+		for _, stateSyncEventID := range stateSyncEventIDs {
+			stateSyncEventIDKey := common.EncodeUint64ToBytes(stateSyncEventID)
+
+			if err := eventsBucket.Delete(stateSyncEventIDKey); err != nil {
+				s.logger.Info(fmt.Sprintf("failed to remove state sync event (ID=%d): %v", stateSyncEventID, err))
```
goran-ethernal: I wouldn't log it here. Just return an error, and log the error in `state_sync_manager`.

consensus/polybft/state_sync_manager.go
```diff
@@ -602,3 +634,16 @@ func (s *stateSyncManager) multicast(msg interface{}) {
 		s.logger.Warn("failed to gossip bridge message", "err", err)
 	}
 }
+
+// parseStateSyncResultEvent parses StateSyncResult event from the provided log
+func parseStateSyncResultEvent(h *types.Header, l *ethgo.Log) (*contractsapi.StateSyncResultEvent, bool, error) {
+	var stateSyncResultEvent contractsapi.StateSyncResultEvent
+
+	if len(l.Topics) == 0 || l.Topics[0] != stateSyncResultEvent.Sig() {
```
goran-ethernal: No need for this check. I would also remove this function completely, and just add this parse code inline where you instantiate `eventGetter`.
```suggestion
	var stateSyncResultEvent contractsapi.StateSyncResultEvent
	doesMatch, err := stateSyncResultEvent.ParseLog(l)

	return &stateSyncResultEvent, doesMatch, err
```

consensus/polybft/state_sync_manager.go
```diff
@@ -399,14 +412,33 @@ func (s *stateSyncManager) PostEpoch(req *PostEpochRequest) error {
 }
 
 // PostBlock notifies state sync manager that a block was finalized,
-// so that it can build state sync proofs if a block has a commitment submission transaction
+// so that it can build state sync proofs if a block has a commitment submission transaction.
+// Additionally, it will remove any processed state sync events and their proofs from the store.
 func (s *stateSyncManager) PostBlock(req *PostBlockRequest) error {
+	events, err := s.eventsGetter.getEventsFromReceipts(req.FullBlock.Block.Header, req.FullBlock.Receipts)
```
goran-ethernal, line 423: This would mean we will only remove executed state syncs from this block. If an error occurs while deleting them, or if `eventsGetter` returns an error, executed state syncs from this block will never be deleted. We need that retry mechanism that stake manager and checkpoint manager have, but maybe it doesn't matter, because, this PR changes all that: https://github.com/0xPolygon/polygon-edge/pull/1972.

Your call.

Stefan-Ethernal (approved):

consensus/polybft/state_store_state_sync.go
Stefan-Ethernal: Dušan and I discussed this, we don't see state syncs and/or proofs removal as a critical operation so we wanted to proceed and execute the rest of the function (commitment insertion and proofs building). That's why an error is not propagated to a caller.

goran-ethernal: I understand, but my point is to not place a logger in the store, rather have the store return an error, and just log it in the caller (`stateSyncManager`).

Stefan-Ethernal: Sorry, I made an oversight. Makes sense.

dusan-maksimovic: If I returned the first error that was encountered, then other data (which might get removed during the next loop iteration) would not be removed.

goran-ethernal: That is why we have that retry mechanism in other components and in this PR https://github.com/0xPolygon/polygon-edge/pull/1972. You need it if you are to remove all executed state sync events. As it is now, if an error occurs, they are left in db forever.

dusan-maksimovic: I don't think the retry mechanism will help in this situation. The db either has or hasn't this data under the given key. If some key is provided and the removal fails the first time, it will also fail during retry calls. There will just be N unnecessary calls which are also going to fail.

consensus/polybft/state_sync_manager.go
dusan-maksimovic: This is similar to the comment regarding logging or returning error.

Regarding the `eventsGetter` failure, do we have a reason to believe that, if parsing of logs for some block failed it could be successful on retry to parse that same log?

Edit: Current mechanism in checkpoint manager, which uses `eventsGetter.getFromBlocks()` doesn't work well, since last processed block is saved after each `PostBlock()`, so trying to get events from earlier blocks will do nothing since it has a `for` loop which would look like this: `for i := 7; i <= 6; i++ {` and wouldn't do anything. That is why I didn't use `eventsGetter.getFromBlocks()` but directly `eventsGetter.getEventsFromReceipts` which is called inside, and the result is the same.

consensus/polybft/state_store_state_sync.go
goran-ethernal: Yes, but `Delete` doesn't fail if it doesn't have the key in the bucket. It fails because tx is closed, it is not a write transaction, or a file error occurs. That then, requires a retry.

dusan-maksimovic: But we know we are within a open and writable transaction, since we opened it by calling `s.db.Update()` previously. We wouldn't be executing `Delete()` call if that wasn't the case.

goran-ethernal: Yes we are, but as I said, chances are slim, but `db` can still return an error (file is no longer available, `os` is returning some errors). I am fine if you don't want a retry mechanism here, deleting executing events is not a critical nor an important thing, they can exist in `db` forever, as far as I am concerned. I just wanted this to follow the same pattern we have in other components.

consensus/polybft/state_sync_manager.go
goran-ethernal: `Regarding the eventsGetter failure, do we have a reason to believe that, if parsing of logs for some block failed it could be successful on retry to parse that same log` - if you are only using `getEventsFromReceipts` then no. If we are to implement the retry mechanism, than yes.

`t: Current mechanism in checkpoint manager, which uses eventsGetter.getFromBlocks() doesn't work well, since last processed block is saved after each PostBlock()` - this is not true. It does save the last processed block in `PostBlock`, but it only saves it after `eventsGetter` successfully returns the events from given blocks, and after they are saved in `db` successfully. Plus, the `eventsGetter`, even if it doesn't have missed blocks, it will still process events from the current block passed in `PostBlock`, when using `getFromBlocks(lastProcessedBlock uint64, currentBlock *types.FullBlock)`.

This PR https://github.com/0xPolygon/polygon-edge/pull/1972 goes even further, and introduces the concept of global db transaction, so if one component fails to update its state, all of them fail, and nothing will be stored in db.

consensus/polybft/state_store_state_sync.go
dusan-maksimovic: returning errors instead of logging: 118901573a263bb979712ae7984a8c22d6246db5

consensus/polybft/state_sync_manager.go
dusan-maksimovic: fixed: 118901573a263bb979712ae7984a8c22d6246db5

PR #1982, opened 2023-10-12
Do not send checkpoint if node is syncing

goran-ethernal:
# Description

A user of Edge had an issue where he was testing Disaster Recovery simulation, where he stopped a validator node, and deleted all its data, to see if the node will sync up, recover, and rejoin consensus.

He noticed that while syncing, node was trying to send checkpoints on epoch ending blocks where he was a minter before it was shutdown and its data got deleted.

Since those checkpoints were already sent to `rootchain`, this PR introduces a fix, where on `submitCheckpoint`, it checks if given checkpoint was already sent, and simply returns if it is.

# Changes include

# Checklist

## Testing

### DIFF
--- a/consensus/polybft/checkpoint_manager.go
+++ b/consensus/polybft/checkpoint_manager.go
@@ -126,6 +126,16 @@ func (c *checkpointManager) submitCheckpoint(latestHeader *types.Header, isEndOf
 		return err
 	}
 
+	if lastCheckpointBlockNumber > latestHeader.Number {
+		// this node is syncing after its data was cleaned up
+		// what happened is that, node was the proposer on this block,
+		// and it already sent this checkpoint, but since its data got cleaned
+		// it started syncing from scratch, even its own minted blocks
+		// so there is no need to send a checkpoint, since it is already sent
+		// CheckpointManager will fail the transaction, so no need to spend tokens on it
+		return nil
+	}
+
 	c.logger.Debug("submitCheckpoint invoked...",
 		"latest checkpoint block", lastCheckpointBlockNumber,
 		"checkpoint block", latestHeader.Number)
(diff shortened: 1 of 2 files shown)


consensus/polybft/checkpoint_manager.go
```diff
@@ -126,6 +126,16 @@ func (c *checkpointManager) submitCheckpoint(latestHeader *types.Header, isEndOf
 		return err
 	}
 
+	if lastCheckpointBlockNumber > latestHeader.Number {
+		// this node is syncing after its data was cleaned up
+		// what happened is that, node was the proposer on this block,
+		// and it already sent this checkpoint, but since its data got cleaned
+		// it started syncing from scratch, even its own minted blocks
+		// so there is no need to send a checkpoint, since it is already sent
+		// CheckpointManager will fail the transaction, so no need to spend tokens on it
```
Stefan-Ethernal: Let's shorten the comment (mind the linter, you would probably need to split it into several lines):

```suggestion
		// node is out of sync (doesn't have reached a tip of the chain), so even though it is a proposer, it would checkpoint block that is already checkpointed and transaction would fail anyway
```

consensus/polybft/checkpoint_manager_test.go
```diff
@@ -39,83 +39,109 @@ func TestCheckpointManager_SubmitCheckpoint(t *testing.T) {
 
 	validators := validator.NewTestValidatorsWithAliases(t, aliases)
```
Stefan-Ethernal: Could it lead to race conditions, since both subtests are using the same validators slice (though probably they are just reading them), but still think it would be a good idea that each one has its own copy. Wdyt?

Stefan-Ethernal (approved):

consensus/polybft/checkpoint_manager.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1982/commits/c4720038620cb297c87aca3d93003e334f237070

consensus/polybft/checkpoint_manager_test.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/1982/commits/c4720038620cb297c87aca3d93003e334f237070

rachit77 (approved):

begmaroman (approved):

PR #1985, opened 2023-10-12
Track progression while syncing blocks

stana-miric:
# Description

This RP fixing progression tracking when syncing block in syncer. Earlier, progression instance existed in syncer, but was not used at all.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -188,7 +188,7 @@ func (s *syncer) Sync(callback func(*types.FullBlock) bool) error {
 		}
 
 		// fetch block from the peer
-		lastNumber, shouldTerminate, err := s.bulkSyncWithPeer(bestPeer.ID, callback)
+		lastNumber, shouldTerminate, err := s.bulkSyncWithPeer(bestPeer.ID, bestPeer.Number, callback)
 		if err != nil {
 			s.logger.Warn("failed to complete bulk sync with peer, try to next one", "peer ID", "error", bestPeer.ID, err)
 		}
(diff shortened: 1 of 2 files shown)


syncer/syncer.go
```diff
@@ -218,11 +219,18 @@ func (s *syncer) bulkSyncWithPeer(peerID peer.ID, newBlockCallback func(*types.F
...
+
 	defer func() {
 		err := s.syncPeerClient.CloseStream(peerID)
 		if err != nil {
 			s.logger.Error("Failed to close stream: ", err)
 		}
+
+		// Stop monitoring the sync progression upon exit
+		s.syncProgression.StopProgression()
 	}()
```
Stefan-Ethernal, line 237: IMO we should expose the `UpdateCurrentProgression(currentBlock uint64)` function to the `Progression` interface and invoke it whenever `blockCh` gets notified, e.g. here after this line https://github.com/0xPolygon/polygon-edge/blob/2351bc8597b0c6a127074cd17f345d1a0c57b6f5/syncer/syncer.go#L266. It is because `eth_syncing` exposes current block number as well as the information and we would not have it populated at the moment. WDYT?

dusan-maksimovic (approved):

syncer/syncer.go
stana-miric: When starting progression, ProgressionWrapper runs a loop in which subscriber is notified each time when the block is inserted and updates the CurrentBlock. No need to explicitly call UpdateCurrentProgression because PogressionWrapper will do it for us in RunUpdateLoop fn. Additionally, UpdateCurrentProgression if not member of Progression interface as it should be. 
If calling UpdateCurrentProgression separately that only means that we will update CurrentBlock twice as RunUpdateLoop will do it anyway.

Stefan-Ethernal: Sorry, I missed that out. I was looking into restore functionality and there they are explicitly invoking `UpdateCurrentProgression`.

Stefan-Ethernal (approved):

igorcrevar (approved):

PR #1986, opened 2023-10-13
Added option to specify a root path for IBFT validators dir

begmaroman:
# Description

Currently, the command that generates validators set does not support an option to use a specific root dir so the current one is used always. 
This PR introduces a root validators path parameter.

# Changes include

# Breaking changes

# Checklist

## Testing

### DIFF
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -3,14 +3,14 @@ package genesis
 import (
 	"fmt"
 
-	"github.com/0xPolygon/polygon-edge/helper/common"
+	"github.com/spf13/cobra"
 
 	"github.com/0xPolygon/polygon-edge/command"
 	"github.com/0xPolygon/polygon-edge/command/genesis/predeploy"
 	"github.com/0xPolygon/polygon-edge/command/helper"
 	"github.com/0xPolygon/polygon-edge/consensus/ibft"
+	"github.com/0xPolygon/polygon-edge/helper/common"
 	"github.com/0xPolygon/polygon-edge/validators"
-	"github.com/spf13/cobra"
 )
 
 func GetCommand() *cobra.Command {
--- a/command/genesis/genesis.go
+++ b/command/genesis/genesis.go
@@ -3,11 +3,12 @@ package genesis
 import (
 	"fmt"
 
+	"github.com/0xPolygon/polygon-edge/helper/common"
+
 	"github.com/0xPolygon/polygon-edge/command"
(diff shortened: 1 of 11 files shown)


docker/local/polygon-edge.sh
```diff
@@ -46,7 +46,8 @@ case "$1" in
                   rm -f /data/genesis.json
 
                   createGenesisConfig "$2" "$secrets" \
-                  --ibft-validators-prefix-path data-
+                    --ibft-validators-path /data \
+                    --ibft-validators-prefix-path data-
```
Stefan-Ethernal: WDYT about removing `ibft-validators-path` and `ibft-validators-prefix-path` flags from the `genesis` command and using `validators-path` and `validators-prefix` flags for both IBFT and Polybft? We can do that through a separate PR, if you find it possible.

command/genesis/genesis.go
```diff
@@ -125,6 +125,14 @@ func setFlags(cmd *cobra.Command) {
 			"the type of validators in IBFT",
 		)
 
+		cmd.Flags().StringVar(
+			&params.validatorRootPath,
```
Stefan-Ethernal: Should we validate if the path exists and stop the execution otherwise?

command/ibft/switch/ibft_switch.go
```diff
@@ -65,6 +65,14 @@ func setFlags(cmd *cobra.Command) {
 
 	{
 		// PoA Configuration
+		cmd.Flags().StringVar(
+			&params.ibftValidatorRootPath,
```
Stefan-Ethernal: Validate existence of a path?

Stefan-Ethernal (approved):
Overal LGTM, tnx for fixing this. Leaving a couple of considerations though.

docker/local/polygon-edge.sh
begmaroman: I'll do that within this PR

begmaroman: Done

rachit77 (approved):

command/ibft/switch/ibft_switch.go
begmaroman: Done

command/genesis/genesis.go
begmaroman: It is done already: https://github.com/0xPolygon/polygon-edge/blob/develop/command/genesis/params.go#L206-L210

stana-miric (approved):

PR #1991, opened 2023-10-16
Implement unscubscribe on event stream

stana-miric:
# Description

This PR implements unsubscribe on event stream, that will be called when subscription is closed. Earlier event stream held slice of update channels which belonged to subscriptions and were used to notify subscription when the event is received (push fn). The problem here was if some subscription is closed, its update channel was not removed from the event stream, and since it is buffered channel of 5 events, after the buffer was full and  events kept coming, push function was blocked on writing to that channel preventing other subscriptions to receive events. Earlier there was default branch, but we [removed it](https://github.com/0xPolygon/polygon-edge/pull/1361) to avoid loosing events.
With this fix  events stream is changed to have map of subscription and when unsubscribe is called on event stream that will remove the subscription and close subscription channel.

# Changes include

# Checklist

## Testing

### DIFF
--- a/blockchain/subscription.go
+++ b/blockchain/subscription.go
@@ -13,7 +13,6 @@ type void struct{}
 type Subscription interface {
 	GetEventCh() chan *Event
 	GetEvent() *Event
-	Close()
 }
 
 // FOR TESTING PURPOSES //
@@ -33,15 +32,11 @@ func NewMockSubscription() *MockSubscription {
 func (m *MockSubscription) Push(e *Event) {
 	m.updateCh <- e
 }
-func (m *MockSubscription) Close() {
-	close(m.closeCh)
-}
 
 // subscription is the Blockchain event subscription object
 type subscription struct {
-	updateCh    chan *Event  // Channel for update information
-	closeCh     chan void    // Channel for close signals
-	eventStream *eventStream // event stream for unsubscribing
+	updateCh chan *Event // Channel for update information
+	closeCh  chan void   // Channel for close signals
 }
 
 // GetEventCh creates a new event channel, and returns it
@@ -58,11 +53,6 @@ func (s *subscription) GetEvent() *Event {
 	case <-s.closeCh:
 		return nil
 	}
-}
-
-// Close closes the subscription
-func (s *subscription) Close() {
-	s.eventStream.unsubscribe(s)
 }
 
 type EventType int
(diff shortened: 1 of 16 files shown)


dusan-maksimovic (approved):

blockchain/subscription.go
```diff
@@ -127,33 +131,44 @@ func (b *Blockchain) SubscribeEvents() Subscription {
...
+}
 
-	e.updateCh = append(e.updateCh, ch)
+func (e *eventStream) unsubscribe(sub *subscription) {
+	e.Lock()
+	defer e.Unlock()
 
-	return ch
+	// Check if the subscription exists in the map
+	if _, exists := e.subscriptions[sub]; exists {
```
Stefan-Ethernal: `unsubscribe` can be called multiple times for the same `subscription` instance?

Do we need this check because of closing potentially already closed `closeCh` in the `subscription`?

I'd try removing this check and try to fix the issue in case `unsubscribe` is called multiple times, because I would not expect that. WDYT?

blockchain/subscription.go
```diff
@@ -127,33 +131,44 @@ func (b *Blockchain) SubscribeEvents() Subscription {
 	return b.stream.subscribe()
 }
 
-// eventStream is the structure that contains the event list,
-// as well as the update channel which it uses to notify of updates
+// eventStream is the structure that contains the subscribers
+// which it uses to notify of updates
 type eventStream struct {
 	sync.Mutex
```
Stefan-Ethernal: Is it worth considering converting it into `RWMutex`?

blockchain/subscription.go
```diff
@@ -33,11 +33,15 @@ func NewMockSubscription() *MockSubscription {
...
+	close(m.closeCh)
+}
 
 // subscription is the Blockchain event subscription object
 type subscription struct {
-	updateCh chan *Event // Channel for update information
-	closeCh  chan void   // Channel for close signals
+	updateCh    chan *Event  // Channel for update information
+	closeCh     chan void    // Channel for close signals
+	eventStream *eventStream // event stream for unsubscribing
```
Stefan-Ethernal: Hide `eventStream` behind some interface (only `Unsubscribe` function should be exposed)?

blockchain/subscription_test.go
```diff
@@ -109,3 +109,120 @@ func TestSubscription_BufferedChannel_MultipleSubscriptions(t *testing.T) {
...
+	t.Parallel()
+
+	var (
+		e     = newEventStream()
+		event = &Event{
+			NewChain: []*types.Header{
+				{
+					Number: 100,
+				},
+			},
```
Stefan-Ethernal: Nitpick: we don't need this (in case you don't want to make an assertion against it).
```suggestion
```

blockchain/subscription.go
stana-miric: Agree, I'll change it and leave unconditional closing as it was before with addition of removing subscription from the map

stana-miric: fix ebd7d22

blockchain/subscription.go
stana-miric: fix ebd7d22

blockchain/subscription.go
stana-miric: fix ebd7d22

blockchain/subscription_test.go
stana-miric: fix ebd7d22

Stefan-Ethernal (approved):

goran-ethernal (approved):

PR #2001, opened 2023-10-20
CallTracer implementation for debug endpoints

goran-ethernal:
# Description

This PR implements the `callTracer` for `debug` rpc endpoints. `debug_` rpc endpoints config is now expanded with `tracer` field to specify which tracer you want to be used. Currently, Edge has two tracers: `structTracer` and `callTracer`. If no tracer is specified when calling some `debug` endpoint, `structTracer` will be used by default.

This implementation is provided by the Gateway.fm friends, who implemented this on their own Edge fork.

#### How to use `callTracer`

1. Start Edge nodes.
2. Run this curl to get some epoch ending block:
```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber", "params": ["10", false]}' http://localhost:10002`
```
3. After `eth_getBlockByNumber` returns a result, take some transaction hash from `transactions` item in json result.
4. Run this curl to get call `debug_traceTransaction`:
```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"debug_traceTransaction", "params": ["transactionHash", {"tracer": "callTracer"}]}' http://localhost:10002
```
5. Notice that result looks something like this:
```
{
  "jsonrpc": "2.0",
  "id": null,
  "result": {
    "type": "CALL",
    "from": "0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE",
    "to": "0x0000000000000000000000000000000000000101",
    "value": "0x0",
    "gas": "0xeedd4",
    "gasUsed": "0x13966",
    "input": "0x0f50287c00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000",
    "output": "0x",
    "calls": [
      {
        "type": "DELEGATECALL",
        "from": "0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE",
        "to": "0x0000000000000000000000000000000000000101",
        "value": "0x0",
        "gas": "0xea7be",
        "gasUsed": "0x12eb7",
        "input": "0x0f50287c00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000",
        "output": "0x"
      }
    ]
  }
}
```

# Changes include

# Checklist

## Testing

### DIFF
--- a/jsonrpc/debug_endpoint.go
+++ b/jsonrpc/debug_endpoint.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/0xPolygon/polygon-edge/helper/hex"
 	"github.com/0xPolygon/polygon-edge/state/runtime/tracer"
+	"github.com/0xPolygon/polygon-edge/state/runtime/tracer/calltracer"
 	"github.com/0xPolygon/polygon-edge/state/runtime/tracer/structtracer"
 	"github.com/0xPolygon/polygon-edge/types"
 )
@@ -83,6 +84,7 @@ type TraceConfig struct {
 	EnableReturnData  bool    `json:"enableReturnData"`
 	DisableStructLogs bool    `json:"disableStructLogs"`
 	Timeout           *string `json:"timeout"`
+	Tracer            string  `json:"tracer"`
 }
 
 func (d *Debug) TraceBlockByNumber(
@@ -248,13 +250,19 @@ func newTracer(config *TraceConfig) (
 		}
 	}
 
-	tracer := structtracer.NewStructTracer(structtracer.Config{
-		EnableMemory:     config.EnableMemory && !config.DisableStructLogs,
-		EnableStack:      !config.DisableStack && !config.DisableStructLogs,
-		EnableStorage:    !config.DisableStorage && !config.DisableStructLogs,
-		EnableReturnData: config.EnableReturnData,
-		EnableStructLogs: !config.DisableStructLogs,
-	})
+	var tracer tracer.Tracer
+
+	if config.Tracer == "callTracer" {
+		tracer = calltracer.NewCallTracer()
+	} else {
+		tracer = structtracer.NewStructTracer(structtracer.Config{
+			EnableMemory:     config.EnableMemory && !config.DisableStructLogs,
+			EnableStack:      !config.DisableStack && !config.DisableStructLogs,
+			EnableStorage:    !config.DisableStorage && !config.DisableStructLogs,
+			EnableReturnData: config.EnableReturnData,
+			EnableStructLogs: !config.DisableStructLogs,
+		})
+	}
 
 	timeoutCtx, cancel := context.WithTimeout(context.Background(), timeout)
 
--- a/jsonrpc/debug_endpoint.go
+++ b/jsonrpc/debug_endpoint.go
@@ -255,7 +255,7 @@ func newTracer(config *TraceConfig) (
 	var tracer tracer.Tracer
 
 	if config.Tracer == callTracerName {
-		tracer = calltracer.NewCallTracer()
+		tracer = &calltracer.CallTracer{}
 	} else {
 		tracer = structtracer.NewStructTracer(structtracer.Config{
 			EnableMemory:     config.EnableMemory && !config.DisableStructLogs,
--- a/jsonrpc/debug_endpoint.go
+++ b/jsonrpc/debug_endpoint.go
@@ -13,6 +13,8 @@ import (
 	"github.com/0xPolygon/polygon-edge/types"
 )
 
+const callTracerName = "callTracer"
+
 var (
 	defaultTraceTimeout = 5 * time.Second
 
@@ -252,7 +254,7 @@ func newTracer(config *TraceConfig) (
 
 	var tracer tracer.Tracer
 
-	if config.Tracer == "callTracer" {
+	if config.Tracer == callTracerName {
 		tracer = calltracer.NewCallTracer()
 	} else {
 		tracer = structtracer.NewStructTracer(structtracer.Config{
--- a/state/runtime/tracer/calltracer/call_tracer.go
+++ b/state/runtime/tracer/calltracer/call_tracer.go
@@ -41,15 +41,9 @@ type CallTracer struct {
 	activeGas          uint64
 	activeAvailableGas uint64
 
-	cancelLock *sync.Mutex
+	cancelLock sync.Mutex
 	reason     error
 	stop       bool
-}
-
-func NewCallTracer() *CallTracer {
-	return &CallTracer{
-		cancelLock: &sync.Mutex{},
-	}
 }
 
 func (c *CallTracer) Cancel(err error) {
@@ -73,6 +67,13 @@ func (c *CallTracer) Clear() {
 }
 
 func (c *CallTracer) GetResult() (interface{}, error) {
+	c.cancelLock.Lock()
+	defer c.cancelLock.Unlock()
+
+	if c.reason != nil {
+		return nil, c.reason
+	}
+
 	return c.call, nil
 }
 
@@ -124,7 +125,7 @@ func (c *CallTracer) CallStart(depth int, from, to types.Address, callType int,
 func (c *CallTracer) CallEnd(depth int, output []byte, err error) {
 	c.activeCall.Output = hex.EncodeToHex(output)
 
-	var gasUsed uint64 = 0
+	gasUsed := uint64(0)
 
 	if c.activeCall.startGas > c.activeAvailableGas {
 		gasUsed = c.activeCall.startGas - c.activeAvailableGas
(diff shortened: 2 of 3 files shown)


state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,158 @@
...
+		c.Cancel(err)
+	}
+}
+
+func (c *CallTracer) CaptureState(memory []byte, stack []*big.Int, opCode int,
+	contractAddress types.Address, sp int, host tracer.RuntimeHost, state tracer.VMState) {
+	if c.cancelled() {
+		state.Halt()
+
+		return
```
igorcrevar: This return is not needed

jsonrpc/debug_endpoint.go
```diff
@@ -248,13 +250,19 @@ func newTracer(config *TraceConfig) (
...
-	tracer := structtracer.NewStructTracer(structtracer.Config{
-		EnableMemory:     config.EnableMemory && !config.DisableStructLogs,
-		EnableStack:      !config.DisableStack && !config.DisableStructLogs,
-		EnableStorage:    !config.DisableStorage && !config.DisableStructLogs,
-		EnableReturnData: config.EnableReturnData,
-		EnableStructLogs: !config.DisableStructLogs,
-	})
+	var tracer tracer.Tracer
+
+	if config.Tracer == "callTracer" {
```
igorcrevar: Maybe we should per-define all available tracers in some constants?

goran-ethernal: I'll move this `callTracer` string to a constant.

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,158 @@
...
+		Calls:    nil,
+		startGas: gas,
+	}
+
+	if depth == 1 {
+		c.call = call
+		c.activeCall = call
+	} else {
+		call.parent = c.activeCall
+		c.activeCall.Calls = append(c.activeCall.Calls, call)
```
igorcrevar, line 120: Do we need to append only to `.Calls` for a root call (`depth`==1)?

goran-ethernal: I don't think so. If depth is 1, that is the parent call, or the first call. If dept is 2, it will append the new call to the `Calls` list of that parent. If the depth is 3, it will append it to the calls of the child call in depth 2.

jsonrpc/debug_endpoint.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/7cb7be10c158113cc224e93b3c8123133040f980

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/7cb7be10c158113cc224e93b3c8123133040f980

state/runtime/tracer/calltracer/call_tracer.go
igorcrevar: why don't we model than this list as single `Child` field?

igorcrevar: But, it may not be that important; perhaps it's required according to some specifications. I will approve it.

igorcrevar (approved):

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,156 @@
...
+	startGas uint64
+}
+
+type CallTracer struct {
+	call               *Call
+	activeCall         *Call
+	activeGas          uint64
+	activeAvailableGas uint64
+
+	cancelLock *sync.Mutex
```
Stefan-Ethernal: No need to instantiate mutex, not sure are there any benefits if it is a pointer? Also probably it is better to use an `RWMutex`.
```suggestion
	cancelLock sync.RWMutex
```

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (c *CallTracer) Clear() {
+	c.call = nil
+	c.activeCall = nil
+}
+
+func (c *CallTracer) GetResult() (interface{}, error) {
+	return c.call, nil
+}
```
Stefan-Ethernal: Maybe we should propagate reason here, based GetResult implementation in the struct tracer implementation, WDYT?
Otherwise we should probably remove `reason` field altogether.

```suggestion
func (c *CallTracer) GetResult() (interface{}, error) {
    if c.reason != nil {
        return nil, c.reason
    }

	return c.call, nil
}
```

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,156 @@
...
+		call.parent = c.activeCall
+		c.activeCall.Calls = append(c.activeCall.Calls, call)
+		c.activeCall = call
+	}
+}
+
+func (c *CallTracer) CallEnd(depth int, output []byte, err error) {
+	c.activeCall.Output = hex.EncodeToHex(output)
+
+	var gasUsed uint64 = 0
```
Stefan-Ethernal: Nitpick, consistency-wise:

```suggestion
	gasUsed := uint64(0)
```

Stefan-Ethernal (approved):
Generally LGTM

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/786e10e03573e7a20dd6ec34eb45ff613da5909c

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/786e10e03573e7a20dd6ec34eb45ff613da5909c

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/786e10e03573e7a20dd6ec34eb45ff613da5909c

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,157 @@
...
+	startGas uint64
+}
+
+type CallTracer struct {
+	call               *Call
+	activeCall         *Call
+	activeGas          uint64
+	activeAvailableGas uint64
+
+	cancelLock sync.Mutex
```
Stefan-Ethernal: Optional nitpick:
```suggestion
	cancelLock sync.RWMutex
```

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (c *CallTracer) Clear() {
+	c.call = nil
+	c.activeCall = nil
+}
+
+func (c *CallTracer) GetResult() (interface{}, error) {
+	c.cancelLock.Lock()
+	defer c.cancelLock.Unlock()
```
Stefan-Ethernal: Optional nitpick:
```suggestion
	c.cancelLock.RLock()
	defer c.cancelLock.RUnlock()
```

state/runtime/tracer/calltracer/call_tracer.go
```diff
@@ -0,0 +1,157 @@
...
+
+	c.reason = err
+	c.stop = true
+}
+
+func (c *CallTracer) cancelled() bool {
+	c.cancelLock.Lock()
+	defer c.cancelLock.Unlock()
```
Stefan-Ethernal:
```suggestion
	c.cancelLock.RLock()
	defer c.cancelLock.RUnlock()
```

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/c10c000425d0d7047c6339bd81e6efe57c988599

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/c10c000425d0d7047c6339bd81e6efe57c988599

state/runtime/tracer/calltracer/call_tracer.go
goran-ethernal: Fix: https://github.com/0xPolygon/polygon-edge/pull/2001/commits/c10c000425d0d7047c6339bd81e6efe57c988599

PR #2026, opened 2023-11-02
Fix gas cost overflow for opCall opcode

Stefan-Ethernal:
# Description

This PR introduces a fix to EVM in terms that it checks for overflows in case of `opCall` gas cost cannot fit uint64. In that case, the `gas uint64 overflow` error is returned and execution is halted.

# Changes include

# Breaking changes

# Checklist

## Testing

### Manual tests

# Documentation update

# Additional comments

### DIFF
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -358,6 +358,16 @@ func BigIntDivCeil(a, b *big.Int) *big.Int {
 		Div(result, b)
 }
 
+// SafeAddUint64 sums two unsigned int64 numbers if there are no overflow.
+// In case there is an overflow, it would return 0 and true, otherwise sum and false.
+func SafeAddUint64(a, b uint64) (uint64, bool) {
+	if a > math.MaxUint64-b {
+		return 0, true
+	}
+
+	return a + b, false
+}
+
 // EncodeUint64ToBytes encodes provided uint64 to big endian byte slice
 func EncodeUint64ToBytes(value uint64) []byte {
 	result := make([]byte, 8)
--- a/helper/common/common.go
+++ b/helper/common/common.go
@@ -358,14 +358,15 @@ func BigIntDivCeil(a, b *big.Int) *big.Int {
 		Div(result, b)
 }
 
-// SafeAddUint64 sums two unsigned int64 numbers if there are no overflow.
+// SafeAddUint64 sums two unsigned int64 numbers if there is no overflow.
 // In case there is an overflow, it would return 0 and true, otherwise sum and false.
 func SafeAddUint64(a, b uint64) (uint64, bool) {
-	if a > math.MaxUint64-b {
+	sum := a + b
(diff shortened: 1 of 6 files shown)


helper/common/common.go
```diff
@@ -358,6 +358,16 @@ func BigIntDivCeil(a, b *big.Int) *big.Int {
 		Div(result, b)
 }
 
+// SafeAddUint64 sums two unsigned int64 numbers if there are no overflow.
```
begmaroman: "SafeAddUint64 sums two unsigned int64 numbers if there is no overflow"

helper/common/common.go
```diff
@@ -358,6 +358,16 @@ func BigIntDivCeil(a, b *big.Int) *big.Int {
...
+// In case there is an overflow, it would return 0 and true, otherwise sum and false.
+func SafeAddUint64(a, b uint64) (uint64, bool) {
```
begmaroman, line 363: Does it make sense to use atomic.AddUint64 here?

begmaroman (approved):

helper/common/common.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/polygon-edge/pull/2026/commits/76c28464ccc8971a6a08240d2c4fa3b78686d750

helper/common/common.go
Stefan-Ethernal: If you are suggesting it because of thread safety, I don't feel we need it, because parameters are passed into the function by value and also they are not modified, and instead, a new local variable is returned, so in that context I don't see there should be any race conditions. Introducing of the `atomic` would make implementation a bit less readable IMO.

goran-ethernal (approved):

rachit77 (approved):

state/runtime/evm/instructions.go
```diff
@@ -1352,7 +1358,7 @@ func (c *state) buildCreateContract(op OpCode) (*runtime.Contract, error) {
 
 	if hasTransfer {
 		if c.host.GetBalance(c.msg.Address).Cmp(value) < 0 {
-			return nil, fmt.Errorf("bad")
```
vcastellm, line 1355: bad xD

vcastellm (approved):
LGTM have you been able to reproduce?

Stefan-Ethernal:
> have you been able to reproduce?

Only through the unit test (https://github.com/0xPolygon/polygon-edge/pull/2026/files#diff-6f681b2f9f20cf2822bc04c36bd04e44b02b3048a32313ba05b8d90cd2dbacc3R708-R740), and compared that part with Geth implementation. IMO it is pretty much a corner use case (we would need to have a contract, that invokes other contract and providing nearly `math.MaxUint64` for initial gas when calling another contract and EIP-150 hardfork needs to be disabled).

